PR #478, opened 2014-07-16
Add 'Search whole word'.

Yustos:


siegfriedpammer:
Thanks for your contributon, but we won't merge it, because we already have support for regular expressions in the search box. "match whole word" can easily be entered as `\\bTextBox\b\`.

Yustos:
Hello.
I didn't know that regular expressions is already supported.
Thank you for review.


Четверг, 26 февраля 2015, 11:24 -08:00 от Siegfried Pammer <notifications@github.com>:
>Thanks for your contributon, but we won't merge it, because we already have support for regular expressions in the search box. "match whole word" can easily be entered as \\bTextBox\b\.
>Thank you for your understanding!
>—
>Reply to this email directly or  view it on GitHub .


С уважением,

yustos@inbox.ru

PR #484, opened 2014-08-01
Fixed #483

### LINKED ISSUE #434: Freezes on very long if else... statements
ILSpy hangs while trying to decompile very long if else... statements. At least I think it's what happens. Basically, it turns if elses into if {} else {if {}...}, which with 1500 or so statements is too much for it.

I don't know what's the actual code that makes it do that, but if you ever heard of a game called Terraria, then you can test it yourself - the class Item doesn't decompile at all because of that. It's pretty annoying for us, the tAPI dev team (we're bringing mod support to Terraria), as we have to decompile it manually class-by-class, and then use dotPeek for the Item.cs.

### LINKED ISSUE #483: if-else-if-... and nesting
Currently, what was a structure like this in source,
```csharp
if (condition1) {
	action1();
} else if (condition2) {
	action2();
} else if (condition3) {
	action3();
}
```
is being decompiled into
```csharp
if (condition)
{
	action1();
}
else
{
	if (condition2)
	{
		action2();
	}
	else
	{
		if (condition3)
		{
			action3();
		}
	}
}
```
Which is *generally* okay, but results in rapidly increasing nesting as the number of "chained" conditions increases.

Perhaps it wouldn't hurt to not create a `{ }` container if it is to contain a single if-statement.

PoroCYon:
Couldn't fix #434 , LoopsAndConditions.cs is a huge mess.

### DIFF
--- a/ICSharpCode.Decompiler/FlowAnalysis/ControlFlowNode.cs
+++ b/ICSharpCode.Decompiler/FlowAnalysis/ControlFlowNode.cs
@@ -228,14 +228,35 @@ namespace ICSharpCode.Decompiler.FlowAnalysis
 			}
 		}
 		
-		public void TraversePreOrder(Func<ControlFlowNode, IEnumerable<ControlFlowNode>> children, Action<ControlFlowNode> visitAction)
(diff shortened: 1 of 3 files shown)


siegfriedpammer:
Thanks for your contribution, however, we won't merge it currently, as it broke some unit tests and I was unable to decompile mscorlib.dll without assertion failures and/or exceptions.

PR #569, opened 2015-03-08
Prevent null ref and invalid cast exceptions

0xd4d:


siegfriedpammer:
IMHO the exceptions should not be prevented. Rather the bugs causing the exceptions should be fixed. Do you have any examples?

0xd4d:
Try the .NET native runtime files. corefx.dll triggers the invalid cast exception. I forgot the name of the other file. If you don't have the files I can upload them.

The extra PointerType check matches the original code. Allowing a null 'this' input to extension methods and having a check there is a nice way to keep having to add null checks all over the place.

0xd4d:
2nd file (also from .NET Native) is: System.Xml.ReaderWriter.dll

PR #576, opened 2015-03-14
Fixed pointer subtraction for most cases

### LINKED ISSUE #565: Pointer subtraction is decompiled incorrectly
```C#
// System.Text.ASCIIEncoding
internal unsafe override int GetBytes(char* chars, int charCount, byte* bytes, int byteCount, EncoderNLS encoder)
// ...
				encoder.m_charsUsed = (chars - ptr3 / 2) / 2;
```

That statement is incorrect for 3 reasons:

1. pointers don't support division operators
2. dividing it is incorrect anyway
3. pointer subtraction produces a long, m_charsUsed is an int, so an explicit cast is required here

a similar thing happens with the expression in the return statements for that function.

That method also has cosmetic strangeness too like:

```C#
				*(bytes++) = (byte)c2;
// instead of:
//				*bytes++ = (byte)c2;

// and

	byte* ptr4 = bytes + byteCount / 1;
// instead of:
//	byte* ptr4 = bytes + byteCount;
```

LordJZ:
Fixes #565

### DIFF
--- a/ICSharpCode.Decompiler/ILAst/ILAstOptimizer.cs
+++ b/ICSharpCode.Decompiler/ILAst/ILAstOptimizer.cs
@@ -675,24 +675,41 @@ namespace ICSharpCode.Decompiler.ILAst
 				List<ILExpression> args = expr.Arguments;
 				switch (expr.Code) {
 					case ILCode.Localloc:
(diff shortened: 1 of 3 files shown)


LordJZ:
Succeeded by #587

PR #586, opened 2015-03-29
Split long attribute sections into multiple smaller ones.

### LINKED ISSUE #584: Formatting for attributes
Currently the decompiled output lists outputs in the following form:

``` csharp
[Attribute1, Attribute2, ...]
```

It would be nice, either by default or as an option, to support this form instead:

``` csharp
[Attribute1]
[Attribute2]
...
```

This is particularly noticable for things like this. VSPackage implementations are even worse with their registration attributes.

```
[MarginContainer("LeftSelection"), TextViewRole("INTERACTIVE"), ContentType("text"), Name("Coverage Margin"), Order(Before = "LineNumber"), Export(typeof(IWpfTextViewMarginProvider))]
```

LordJZ:
This doesn't affect single-line attribute sections e.g. on parameters.

Fixes #584


dgrunwald:
This change doesn't belong into NRefactory. If the AST uses a single `AttributeSection`, the output should also be a single AttributeSection.

PR #587, opened 2015-03-29
Enhanced pointer arithmetic handling

### LINKED ISSUE #400: Incorrect C# and VB code generated for pointer addition assignment.
ILSpy: 2.1.0.1603
Assembly: Microsoft.Xna.Framework.dll (XNA 4.0 Refresh)
Method: Microsoft.Xna.Framework.Helpers.SmartGetHashCode(object) : int32

C#: 16: ptr += (IntPtr) 4 / 4; // invalid code
VB: 13: ptr += CType(4, IntPtr) / 4 ' probably invalid code (I don't code in VB)
IL:
49: IL_0031: ldloc.1 // ptr
50: IL_0032: ldc.i4.4
51: IL_0033: conv.i
52: IL_0034: add
53: IL_0035: stloc.1 // ptr

As can be seen, the CIL code is being misinterpreted as some weird addition assignment of a division.

The correct code for C# should be:
ptr += 1;

Based on tests, the CIL code is optimized.

### LINKED ISSUE #565: Pointer subtraction is decompiled incorrectly

### LINKED ISSUE #592: System.InvalidCastException: Unable to cast object of type 'Mono.Cecil.TypeReference' to type 'Mono.Cecil.PointerType'
Decompiling a .NET assembly which contains unmanaged code produces an exception on ILSpy 2.3.0.1827 while it works fine on ILSpy 2.2.0.1706

2.3.0 exception:
```
 ---> System.InvalidCastException: Unable to cast object of type 'Mono.Cecil.TypeReference' to type 'Mono.Cecil.PointerType'.
   at ICSharpCode.Decompiler.ILAst.ILAstOptimizer.HandlePointerArithmetic(ILNode method)
   at ICSharpCode.Decompiler.ILAst.ILAstOptimizer.Optimize(DecompilerContext context, ILBlock method, ILAstOptimizationStep abortBeforeStep)
   at ICSharpCode.Decompiler.Ast.AstMethodBodyBuilder.CreateMethodBody(IEnumerable`1 parameters)
   at ICSharpCode.Decompiler.Ast.AstMethodBodyBuilder.CreateMethodBody(MethodDefinition methodDef, DecompilerContext context, IEnumerable`1 parameters)
   --- End of inner exception stack trace ---
   at ICSharpCode.Decompiler.Ast.AstMethodBodyBuilder.CreateMethodBody(MethodDefinition methodDef, DecompilerContext context, IEnumerable`1 parameters)
   at ICSharpCode.Decompiler.Ast.AstBuilder.CreateMethod(MethodDefinition methodDef)
   at ICSharpCode.Decompiler.Ast.AstBuilder.AddMethod(MethodDefinition method)
   at ICSharpCode.ILSpy.CSharpLanguage.DecompileMethod(MethodDefinition method, ITextOutput output, DecompilationOptions options)
   at ICSharpCode.ILSpy.TextView.DecompilerTextView.DecompileNodes(DecompilationContext context, ITextOutput textOutput)
   at ICSharpCode.ILSpy.TextView.DecompilerTextView.<>c__DisplayClass16.<DecompileAsync>b__15()
```

2.2.0 decompiled code (edited for unrelated reasons):
```
public unsafe void Start(uint sn, int ac)
{
    int num = (int)stackalloc byte[<Module>.__CxxQueryExceptionSize()];
    a.m_sn = sn;
    a.m_ac = ac;
    try
    {
        a.m_ia = IPAddress.Parse(a.m_iAString);
        this.m_p = a.m_ac;
        if (a.m_ac < 12345)
        {
        	this.m_p += a.m_pB;
        }
        a.m_uc = new UC(this.m_p);
        a.m_e = new E(a.m_ia, this.m_p);
        goto IL_C9;
    }
    uint exceptionCode = (uint)Marshal.GetExceptionCode();
    endfilter(<Module>.__CxxExceptionFilter(Marshal.GetExceptionPointers(), null, 0, null));
    IL_C9:
    try
    {
        a.m_uc.JoinMulticastGroup(a.m_ia, 2);
        goto IL_135;
    }
    exceptionCode = (uint)Marshal.GetExceptionCode();
    endfilter(<Module>.__CxxExceptionFilter(Marshal.GetExceptionPointers(), null, 0, null));
    IL_135:
    R.m_udpClient = a.m_uc;
    R.m_mysn = a.m_sn;
    R.m_t = new Thread(new ThreadStart(R.Start));
    R.m_t.IsBackground = true;
    R.m_t.Start();
    a.m_ty = new Timer();
    ElapsedEventHandler value = new ElapsedEventHandler(a.YE);
    a.m_ty.Elapsed += value;
    a.m_ty.Interval = (double)a.m_yi;
    a.m_ty.Start();
    a.Send();
}
```

I'll try this on master as soon as I have time (are there nightly-builds somewhere ?)

LordJZ:
This fixes more complex pointer addition/subtraction examples, like PointerArithmetic3 and 4 test cases. Test case 2 is still failing.

Fixes #565
Fixes #400
Succeeds #576
Fixes #592
Partially fixes #569

### DIFF
--- a/ICSharpCode.Decompiler/Ast/Transforms/IntroduceUnsafeModifier.cs
+++ b/ICSharpCode.Decompiler/Ast/Transforms/IntroduceUnsafeModifier.cs
@@ -35,7 +35,11 @@ namespace ICSharpCode.Decompiler.Ast.Transforms
 		protected override bool VisitChildren(AstNode node, object data)
 		{
 			bool result = false;
-			for (AstNode child = node.FirstChild; child != null; child = child.NextSibling) {
+			AstNode next;
+			for (AstNode child = node.FirstChild; child != null; child = next) {
+				// Store next to allow the loop to continue
+				// if the visitor removes/replaces child.
+				next = child.NextSibling;
 				result |= child.AcceptVisitor(this, data);
 			}
 			if (result && node is EntityDeclaration && !(node is Accessor)) {
--- a/ICSharpCode.Decompiler/ILAst/ILAstOptimizer.cs
+++ b/ICSharpCode.Decompiler/ILAst/ILAstOptimizer.cs
@@ -839,7 +839,8 @@ namespace ICSharpCode.Decompiler.ILAst
 				}
 			}
 
-			adjustmentExpr = new ILExpression(divide ? ILCode.Div_Un : ILCode.Mul, null, adjustmentExpr, sizeOfExpression);
+			if (!(sizeOfExpression.Code == ILCode.Ldc_I4 && (int)sizeOfExpression.Operand == 1))
+				adjustmentExpr = new ILExpression(divide ? ILCode.Div_Un : ILCode.Mul, null, adjustmentExpr, sizeOfExpression);
 		}
 		
 		public static void ReplaceVariables(ILNode node, Func<ILVariable, ILVariable> variableMapping)
--- a/ICSharpCode.Decompiler/ILAst/ILAstOptimizer.cs
+++ b/ICSharpCode.Decompiler/ILAst/ILAstOptimizer.cs
@@ -669,33 +669,84 @@ namespace ICSharpCode.Decompiler.ILAst
 				});
 		}
 
-		static void HandlePointerArithmetic(ILNode method)
+		void HandlePointerArithmetic(ILNode method)
 		{
 			foreach (ILExpression expr in method.GetSelfAndChildrenRecursive<ILExpression>()) {
 				List<ILExpression> args = expr.Arguments;
 				switch (expr.Code) {
 					case ILCode.Localloc:
-						args[0] = DivideBySize(args[0], ((PointerType)expr.InferredType).ElementType);
-						break;
-					case ILCode.Add:
+				    {
+				        ILExpression arg0 = args[0];
+                        ILExpression expr2 = expr;
+				        DivideOrMultiplyBySize(ref expr2, ref arg0, ((PointerType)expr.InferredType).ElementType, true);
+                        // expr shouldn't change
+				        if (expr2 != expr)
+				            throw new InvalidOperationException();
+				        args[0] = arg0;
+				        break;
+				    }
+				    case ILCode.Add:
 					case ILCode.Add_Ovf:
 					case ILCode.Add_Ovf_Un:
-						if (expr.InferredType is PointerType) {
-							if (args[0].ExpectedType is PointerType)
-								args[1] = DivideBySize(args[1], ((PointerType)expr.InferredType).ElementType);
-							else if (args[1].ExpectedType is PointerType)
-								args[0] = DivideBySize(args[0], ((PointerType)expr.InferredType).ElementType);
-						}
-						break;
-					case ILCode.Sub:
+				    {
+				        ILExpression arg0 = args[0];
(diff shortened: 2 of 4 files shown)


DerpMcDerp:
I can't test your patch but can you make sure the following is fixed too:

```C#
// System.Convert
private unsafe static int ConvertToBase64Array(char* outChars, byte* inData, int offset, int length, bool insertLineBreaks) {
// ...
// current (v2.3.0.1830):
					outChars[num3 + 2] = *(ptr + ((int)(*(inData + (i + 1) / 1) & 15) << 2 | (*(inData + (i + 2) / 1) & 192) >> 6));
// better:
//					outChars[num3 + 2] = ptr[((inData[i + 1] & 15) << 2) | ((inData[i + 2] & 192) >> 6)];
```

specifically:

- the explicit cast to int is redundant
- `*(ptr + something)` on the rhs should be `ptr[something]`

tomspilman:
@LordJZ - I'm new to the ILSpy code base, but I suspect the primary issue with most of the pointer math is based on `ILCode.Ldflda` and `ILCode.Ldsflda` in `TypeAnalysis.DoInferTypeForExpression`:

https://github.com/icsharpcode/ILSpy/blob/master/ICSharpCode.Decompiler/ILAst/TypeAnalysis.cs#L368

Those are currently returning a `ByReferenceType` which I think is incorrect.  I'm seeing better results with returning `PointerType` in both cases.

LordJZ:
@tomspilman interesting, thanks. I will try to poke around that when I have time.

tomspilman:
@LordJZ - Don't worry... that isn't the right fix... but it lead me to understand the problem better.

Still maybe you can tell me if i'm heading in the right direction here.

First fact is that all pointer math in IL assumes `byte*`s.  So looking at the `PointerArithmatic2` test IL:
```
	IL_0001: ldarg.1
	IL_0002: ldarg.2
	IL_0003: ldarg.3
	IL_0004: mul
	IL_0005: conv.i
	IL_0006: ldc.i4.2
	IL_0007: mul
	IL_0008: add
```

This is essentially this...
```
bytePtr + ((X * Y) * Z)
````

What the decompiler needs to do is find cases where it can change the pointer size to remove the extra multiply to remove on the right side of the add.

So if we have this:
```
bytePtr + ((4) * 2)
```
We can simplify to this:
```
shortPtr + 4
```
I have this working for the specific case for `PointerArithmatic2` and I now get:
```
return (byte*)((short*)p + y * x);
```
Which matches the original code.

I plan to submit a PR later today.

PR #597, opened 2015-04-25
UpdateAssemblyInfo.exe failed when there is a banner in start of cmd

pgourlain:
when my cmd starts, there is a banner (i.e doskey), so assume that the first line is a commit isn't good in this case.


pgourlain:
the second commit is to resolve this problem : I have a plugin that manage bookmarks, and to select in the treeview i use FindTreeNode. When the user want to keep a bookmark on resource node, the findtreeNode return null. I add a case in the "switch/case".

PR #601, opened 2015-05-09
Update AssemblyList.cs

ADCJustinH:
Often I am editing an Assembly and subsequently inspecting it in ILSpy. It is very frustrating when my changes do not show up after I open the assembly via the menu. Currently I have to remove the existing assembly and then load it again from the menu. So many unnecessary clicks. Most users are not going to reload an assembly already in the list and if they do they very much should expect it to reflect the current version in the path. This ensures that happens.

### DIFF
--- a/ILSpy/AssemblyList.cs
+++ b/ILSpy/AssemblyList.cs
@@ -141,7 +141,7 @@ namespace ICSharpCode.ILSpy
 		
 		/// <summary>
 		/// Opens an assembly from disk.
-		/// Returns the existing assembly node if it is already loaded.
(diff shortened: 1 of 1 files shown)


dgrunwald:
`OpenAssembly` is called in a lot of places, are you sure all of those should re-load the assembly?

ADCJustinH:
No, I can’t claim any great familiarity with the code base. I was just trying to fix this for my use case: I double click on an assembly and it opens ilspy but since an older version of that assembly was previously loaded it just shows me that version instead of the one I double clicked on. There is probably a better way to solve this problem.

~Justin

Issue #517, opened 2014-11-26
Naming of well known constants in C#, VB code (Math.PI instead of 3.1415926535897931)


KvanTTT:
It would be good to convert double numbers to well know constants (such as Math.PI, Math.E) and their combinations (Math.PI \* 2).

I mean decompile IL to:

``` csharp
if (x > Math.PI)
    x -= Math.PI * 2;
```

instead of:

``` csharp
if (x > 3.1415926535897931)
    x -= 6.2831853071795862;
```

interdrift:
Why is this needed for the decompilator, it looks like a functinality that works fine?

PR #604, opened 2015-05-18
Naming (Min, Max) of primitive constants (byte, int, double etc.).

KvanTTT:
refs #517
TextWriterOutputFormatter code cleaned and some errors fixed.


dgrunwald:
This feature should be added to the decompiler, not to the NRefactory library.

If you really need to change NRefactory, submit to the upstream repository first: https://github.com/icsharpcode/NRefactory
When https://github.com/jbevain/cecil/pull/214 lands, we're going to delete the copies of Cecil and NRefactory in the ILSpy repository and use the NuGet packages instead.

KvanTTT:
I think it should be in NRefactory lib, because of float.PositiveInfinity, float.NegativeInfinity, float.NaN and other are there.

dgrunwald:
There's no float literal in C# for PositiveInfinity, so NRefactory has the choice between using the constant and throwing an exception. The former makes life easier for code generators, so that's why we support those constants there.

However, the situation is very different here: NRefactory is not just used for code generation, but also for transformations.

Your code would make `ParseExpression("255").ToString()` result in `byte.MaxValue`. Consider the "extract method" refactoring -- would you expect your IDE to replace all occurrences of `255` with `byte.MaxValue` when extracting code into a new method?
Also consider that byte.MaxValue has type `byte`, not `int`, so the refactoring would end up changing the semantics of the code!

KvanTTT:
Ok, I got it.

Issue #610, opened 2015-06-02
Suggestion: Display public key used for signing assembly.


DumboJet:
I think it would be handy if you could provide a way to display the public key used for signing each assembly.
From a first look, you don't do that yet.
I think this is possible using this command line:

```
sn -Tp YourAssembly.dll
```

..but having it in ILSpy would be good. :)

Thanks for the great free tool. :)

junderhill:
@DumboJet  Are you refering to the public key token or the entire public key? If it's just the token it is displayed when hovering over an assembly. See here..
![2015-06-11_14-14-59](https://cloud.githubusercontent.com/assets/838529/8107849/61e2e338-1044-11e5-82aa-0a1fcbc7f6cd.png)

If it is the entire key you are refering to I can't see it being difficult to make available..

DumboJet:
I mean the entire public key.
It would be handy to have, I think. :)

junderhill:
OK, I'll have a look at this

junderhill:
Something like this perhaps?
![2015-06-11_15-15-04](https://cloud.githubusercontent.com/assets/838529/8109083/d6a046fe-104c-11e5-89b6-76c483aa3e06.png)

DumboJet:
Yes, that seems the correct info.
This can be verified with SN.exe, if I am not mistaken:

```
sn –T myDLL.dll
```

If there was a way to copy it too, it might be even more cooler. :)

junderhill:
That's the other thing I was thinking of. It verifies correctly so I'm happy it is displaying the correct data but the ability to copy would definitely be useful. Leave it with me.

junderhill:
Ok I've implemented this on the right click context menu for assemblys as follows:
![2015-06-11_17-11-45](https://cloud.githubusercontent.com/assets/838529/8112045/3e0c8c7a-105d-11e5-88e4-fe4e10669925.png)

DumboJet:
Cool ! :)
Thanks a lot ! :)

PR #618, opened 2015-06-11
Show public key and ability to copy public key to clipboard for assembly

junderhill:
In response to #610 the tool tip for assemblys has been updated to show the public key:

The 'Copy public key to clipboard' option has also been added to the right click context menu for assemblies:

### DIFF
--- a/ILSpy/TreeNodes/AssemblyTreeNode.cs
+++ b/ILSpy/TreeNodes/AssemblyTreeNode.cs
@@ -20,6 +20,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Threading.Tasks;
 using System.Windows;
 using System.Windows.Controls;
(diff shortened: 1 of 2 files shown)


dgrunwald:
That seems like a bit too much for a tooltip -- wouldn't it be better to show this as a comment when decompiling the assembly node? (in fact, the IL disassembly already does this; it's just missing in the C# view)
That would also allow the user to copy the public key without using a special context menu action.

linquize:
A tooltip should not contains long information

greenozon:
let me show you guys long tooltip (as seen in SAE, Simple Asssembly Explorer .net tool):
http://prntscr.com/7gscld
thats from mixedmode assembly

junderhill:
@dgrunwald yeh that makes sense. I don't know ILSpy too well so didn't know the IL disassembly already does this. I'll look at adding it to the C# view.

PR #620, opened 2015-06-27
Create installer

kovalikp:
I've made an ILSpy installer in WIX for personal convenience, though I guess you might find it useful. So far it does: 

* Installs under `%programfiles%\ICSharpCode\ILSpy`.
* Adds start menu shortcut under `ICSharpCode\ILSpy`.
* Uses same version as ILSpy.exe (via binder variable).

Let me know if you need anything else - assuming you're interested in it.

### DIFF
--- a/ILSpy.Setup/ICSharpCode.AvalonEdit.wxs
+++ b/ILSpy.Setup/ICSharpCode.AvalonEdit.wxs
@@ -1,0 +1,15 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
+  <Fragment>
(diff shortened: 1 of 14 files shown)


dgrunwald:
Maybe the installer could add an explorer context menu item "Open with ILSpy" to .dll and .exe files?

As it is, I don't really see an advantage of the installer over the binary .zip file.

ohadschn:
In addition to an "Open with ILSpy" context menu, there are two more benefits:

1. Personally I like having all installed apps managed via the add/remove programs and not in some util folder I unzip stuff into. It's also better for multi-user environments.
2. More importantly, Cortana only finds installed apps. This means that in Windows 10, you won't be able to find it if you type `ILSpy` into the start menu search (even if you have shortcuts on the desktop and pinned to start).

PR #621, opened 2015-07-01
add tooltip on typetreenode and methodtreenode

pgourlain:
on "heavy" Assembly or Types, you didn't see where you are (because navigation jump to an assembly then to another). Just move your mouse over selected node (type or method actually) in order to see assembly name on type or type and assembly on methods.

### DIFF
--- a/ILSpy/TreeNodes/ILSpyTreeNode.cs
+++ b/ILSpy/TreeNodes/ILSpyTreeNode.cs
@@ -190,5 +190,5 @@ namespace ICSharpCode.ILSpy.TreeNodes
 					return System.Windows.SystemColors.GrayTextBrush;
 			}
 		}
-	}
+    }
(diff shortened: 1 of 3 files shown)


siegfriedpammer:
hmm... I am not really convinced that this will be more useful than annoying. A tooltip on every node seems to be too much.

pgourlain:
actually you can be right. I will propose an alternative solution.
what do you think about a kind of "breacrumb" ('fil d'ariane' in French) in status bar  for selected item ?

dgrunwald:
ILSpy doesn't have any status bar; and I don't think we should add one.

The decompiler already starts the code fragment with a comment containing the full class name (so that users don't have to look at the tree to see the namespace); maybe we should just use the assembly qualified name in that comment?

PR #626, opened 2015-07-07
Fixed #600 Add '@' to var name if it is a keyword

### LINKED ISSUE #600: C#: variable called int should be displayed as @int
ILSpy version 2.3.0.1834

```C#
// Class System.IntPtr
private IntPtr(SerializationInfo info, StreamingContext context)
{
	long int = info.GetInt64("value");
	if (IntPtr.Size == 4 && (int > 2147483647L || int < -2147483648L))
```

It should be `@int` above since `int` is a C# keyword.

NRefactory was updated and some of the old code doesn't seem to have been included in the new NRefactory.

Lordron:
Do not substract name if first character is '@'
fixed


dgrunwald:
Fixed in 75dfa785d71c97c140f7911e2243c21614b2be54.

PR #632, opened 2015-07-30
Fixed Address Of Field Output

tomspilman:
This fixes this output:
```
            fixed (int *ptr = ref _field)  // Syntax error!
                *ptr = 10;
```

... to this output...
```
            fixed (int *ptr = &_field) // Correct!
                *ptr = 10;
```


tomspilman:
I see now that I did this wrong... will close and submit a corrected PR.

PR #636, opened 2015-08-05
Add an entry point context menu item to the assembly list.

AustinWise:
This adds a context menu item to take you straight to the entry point of an EXE. It's a pretty handy feature I miss from Reflector.


LordJZ:
We have it though:

entrypoint

AustinWise:
Oh, I did not notice that. Thanks!

PR #651, opened 2015-11-02
Fixed #650 - C++ switch over `char`/`long` values crashes ILSpy

### LINKED ISSUE #650: C++ switch over `char`/`long` values crashes ILSpy
Example project: https://github.com/Diagonactic/Enums.Delegates.CppCLI method `GenericEnumCode.ParseEnum`. Approximate repro code:

	enum e : char { a, b, c };

	static void m(e v)
	{
		switch (v) {
			case e::a:
				Console::WriteLine("a");
				break;
			case e::b:
				Console::WriteLine("b");
				break;
		}
	}

Exception:

```
ICSharpCode.Decompiler.DecompilerException: Error decompiling System.Boolean Diagonactic.GenericEnumCore`1::ParseEnum(System.String,System.Boolean,System.Boolean,TEnum&)
 ---> System.InvalidCastException: Cast from Int64 to Object not supported.
   at ICSharpCode.NRefactory.Utils.CSharpPrimitiveCast.CSharpPrimitiveCastUnchecked(TypeCode targetType, Object input)
   at ICSharpCode.Decompiler.Ast.AstBuilder.MakePrimitive(Int64 val, TypeReference type)
...
```

LordJZ:
SByte is chosen because ElementType is set to SByte

### DIFF
--- a/ICSharpCode.Decompiler/ILAst/TypeAnalysis.cs
+++ b/ICSharpCode.Decompiler/ILAst/TypeAnalysis.cs
@@ -1266,6 +1266,16 @@ namespace ICSharpCode.Decompiler.ILAst
 					return TypeCode.Double;
 				case MetadataType.String:
 					return TypeCode.String;
+				case MetadataType.RequiredModifier:
+				case MetadataType.OptionalModifier:
+					TypeReference modifier = ((IModifierType)type).ModifierType;
+					switch (modifier.FullName) {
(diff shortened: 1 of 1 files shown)


dgrunwald:
Instead of looking at the modifier name; shouldn't `GetTypeCode` just ignore modifiers and get the type code of the underlying modified type instead?

dgrunwald:
I've fixed `GetTypeCode` by using the element type instead.

Issue #657, opened 2015-11-22
NullReferenceException during decompilation


aelij:
Abbreviated stack trace:

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Mono.Cecil.ImportGenericContext.TypeParameter(String type, Int32 position)
   at Mono.Cecil.MetadataImporter.ImportTypeSpecification(TypeReference type, ImportGenericContext context)
   at Mono.Cecil.MetadataImporter.ImportType(TypeReference type, ImportGenericContext context)
   at Mono.Cecil.MetadataImporter.ImportTypeSpecification(TypeReference type, ImportGenericContext context)
   at Mono.Cecil.MetadataImporter.ImportType(TypeReference type, ImportGenericContext context)
   at Mono.Cecil.MetadataImporter.ImportTypeSpecification(TypeReference type, ImportGenericContext context)
   at Mono.Cecil.MetadataImporter.ImportType(TypeReference type, ImportGenericContext context)
   at Mono.Cecil.MetadataImporter.ImportReference(TypeReference type, IGenericParameterProvider context)
   at Reflexil.Editors.TypeAttributesControl.BaseType_Validated(Object sender, EventArgs e)
   at System.Windows.Forms.Control.OnValidated(EventArgs e)
```

siegfriedpammer:
Seems this is related to https://github.com/sailro/Reflexil and not ILSpy. Would be better to report it there.

PR #658, opened 2015-11-22
Use NuGet versions of most dependencies

aelij:
This resolves issue #657 by updating ILSpy to the latest Cecil version.

I know it's a bit of a controversial change, removing all the local code. But IMO this is a much simpler way of managing dependencies.

Anyhow this solved the most frustrating error I've been experiencing since the last version, so at least updating Cecil was definitely worth it. ILSpy is an awesome tool, and I use it daily.


dgrunwald:
You can't delete ICSharpCode.Decompiler -- there isn't any other repository for it; the ILSpy repository is the main repository for ICSharpCode.Decompiler!
Also, the latest NRefactory NuGet release is missing some critical bugfixes for ILSpy, so you can't use that either.

I would accept this PR if you replaced only Mono.Cecil (and maybe also AvalonEdit).

aelij:
Ha, sorry! :) I saw it (Decompiler) was part of SharpDevelop so I thought this was just a copy.
I'm assuming you also wouldn't object to the NUnit package.

I'll open another PR when I get the time.

PR #667, opened 2015-12-19
Lamda fixes

exyi:
Fixed decompilation of delegate instance cache and fixed some display class bugs


dgrunwald:
1. Please don't change tabs to spaces; the unnecessary indentation changes make the patch hard to read.
2. Your changes seem to include much more than fixes to lambda decompilation.
Can you describe those changes? What's `varIndex.txt` about?
3. Your branch fails to compile. (see AppVeyor build log) Can you fix that?

exyi:
Sorry, I have probably messed it up with my experiments.

weltkante:
If you need a small changeset to start from you can look at some commits on my fork, in particular [this commit for lambdas](https://github.com/weltkante/ILSpy/commit/3cf6f8eb2d5a3a30c8999070d744ea67093747f7) and [this commit for locks](https://github.com/weltkante/ILSpy/commit/3adb76b718c800ad53cb9e09b211de2955261b4b).

I'd make a pull request but my changes doesn't cover all of the roslyn changes, only the most common ones, and I don't have the time to figure out the cases which don't work.

I can make a pull request for the changes I have though, if that's wanted.

exyi:
I have made finally this PR clean and working. Sorry for the mess I have commited and sorry that it took so long to remove, I have completely forgotten that I have a PR here.

I have implemented removing the cached delegate and better removal of DisplayClass.
I also included debug display of ILNode which is not separated by line end and made ToString() line separated. I like it this way more, maybe you'd like too.

dgrunwald:
I don't quite understand what these fixes are about. Can you add some test cases?

Also, we're currently rewriting pretty much the whole decompiler on the `newdecompiler` branch. We'll pretty much going to have to throw your work on the `master` branch away once that lands.

exyi:
It basically tries to remove patterns for caching the delegate instance introduced by Roslyn:
```C#
Func<SyntaxNode, bool> arg_27_1;
if ((arg_27_1 = THISCLASS.<>c.<>9__33_0) == null)
{
	arg_27_1 = (THISCLASS.<>c.<>9__33_0 = new Func<SyntaxNode, bool>(THISCLASS.<>c.<>9.<Visit>b__33_0));
}
IEnumerable<SyntaxNode> ancestorsOrThis = node.GetAncestorsOrThis(arg_27_1);
```

And sometimes ILSpy could not decompile more complicated display classes (display class in another display class), so this should be fixed.

By the way, what is new in the `newdecompiler`? Will it support C# 6 features, like `?.` operator? :) I have tried to checkout it, just out of curiosity, but it does not seem to work very well currently and I have not found any changelog what will be better.

dgrunwald:
`newdecompiler`: the current goal is to get a decompiler that produces correct code (that is, code with the same semantics as the underlying IL). The ILSpy master branch isn't very good at maintaining semantics -- the ILAst doesn't properly distinguish signed and unsigned comparisons, mistreats float comparisons, doesn't always perform integer operations with the correct bitwidth, sometimes calls the wrong method overload, etc... basically, a lot of casts that are implicit in IL are missing from the decompiled code.

To fix that, we've completely redesigned the ILAst. We're still fixing a few correctness bugs at the moment, and once the newdecompiler is good enough to decompile itself (incl. its libraries like Mono.Cecil) and the result still works, we'll start adding back the transformations that detect high-level C# constructs and produce nice-looking code.

exyi:
@dgrunwald Thank you for clarification. The tests are already present in DelegateConstruction.cs, they were only ignored. I have not fixed them all, since the errors are probably caused by another bug, but the decompiled code is now readable.
Just one thing surprises me, in tests, the last method decompiles correctly:
```C#
public static int BitCompicatedLinqExpression(int a, int b, int c, int d, IEnumerable<int> integers)
{
    return DelegateConstruction.Bar<int>((int f) => DelegateConstruction.Bar<int>((int i) => i + f - d + integers.Sum((int ii) => ii + b + f)).First((int iDontHaveClueWhatWillThisDo) => iDontHaveClueWhatWillThisDo % c == a)).First<int>();
}
```
but when I open ILSpy and look at it:
```C#
	public static int BitCompicatedLinqExpression(int a, int b, int c, int d, IEnumerable<int> integers)
	{
		return DelegateConstruction.Bar<int>(delegate(int f)
		{
			IEnumerable<int> arg_44_0 = DelegateConstruction.Bar<int>(delegate(int i)
			{
				int arg_43_0 = i + f - d;
				IEnumerable<int> arg_3E_0 = integers;
				return arg_43_0 + arg_3E_0.Sum((int ii) => ii + b + f);
			});
			return arg_44_0.First((int iDontHaveClueWhatWillThisDo) => iDontHaveClueWhatWillThisDo % c == a);
		}).First<int>();
	}
```
Do you have any clue why it could behaves differently?

dgrunwald:
Differences between tests and directly in ILSpy are usually due to the compiler settings used when compiling the tests (optimization, old csc vs. Rolsyn).

For now I'm closing pull requests for the decompiler engine on master -- there probably won't be any ILSpy release before the newdecompiler is branch is merged, and that branch has a different implementation of DelegateConstruction (running not as a C# transform,but on the ILAst).

PR #668, opened 2015-12-28
Search result sorting by fitness

OndrejPetrzilka:
Search results are sorted by fitness, fitness is percentage coverage of result by search terms.
It can be enabled in DisplaySettings, it's DISABLED by default.

Background:
Often I need to search short types (e.g. like SharpDX.Matrix or SharpDX.Configuration), but even when I type whole name without namespace, it's buried down in search results. I've fixed this by sorting results by fitness.

Fitness is calculated only from name visible in search results pane. It 'should' be calculated in IsMatch method from type/member, but it would require a lot of changes and testing. It still works very good this way.

Feel free to enable it by default (DisplaySettingsPanel.xaml.cs, line 105)


dgrunwald:
Some measure of 'fitness' would be nice, but I don't think the fraction `term.Length / (float)text.Length` is useful for this purpose.

The search terms are the same for all results of a search; so the numerator is always the same, and the denominator is always `text.Length`.

So isn't this just an overly complicated and inefficient way of sorting by `text.Length`?

OndrejPetrzilka:
Hmm, maybe the intention was something more complicated, but sorting by length was good enough. At least for me, it's easier to find what I'm looking for.

I use DPack extension for VS (searches for classes/members/...) and they have some ranking system for matches as well. I think it takes into account length, order of terms and maybe capital letters.

ILSpy/SearchPane.cs
```diff
@@ -242,10 +242,32 @@ void AddResult(SearchResult result)
...
 
+			void InsertResult(SearchResult result)
+			{
+				if (Options.DisplaySettingsPanel.CurrentDisplaySettings.SortResults)
+				{
+					// Keep results collection sorted by "Fitness" by inserting result into correct place
+					// Inserts in the beginning shifts all elements, but there can be no more than 1000 items.
+					for (int i = 0; i < this.Results.Count; i++)
+					{
+						if (this.Results[i].Fitness < result.Fitness)
```
flavius-chis: What happens if the fitness is not set properly? The result is never inserted. The old code always inserted at the end, maybe you should do the same as a fallback

OndrejPetrzilka: fixed in 5261d73
Is there sorted observable collection?

OndrejPetrzilka:
Fixed issues and changed fitness function to take into account name length and ignore generic arguments.

**Example of usefulness:**
When looking for Dictionary (without sorting enabled) and write "Dictionary", there's plenty of types including DictionaryBase, SortedDictionary, etc. I cannot further refine search by typing anything, I have to scroll and look for the Dictionary. With sorting enabled it comes first.

Same issue happens when looking for many types with short and simple names, like List, Array, etc. And this is the way I use ILSpy often...to find out how such types are implemented. 

Sorting by fitness is far from perfect, but solves this issue.

**Improvement considerations**
- search term weight (first search term has more weight than second)
- higher fitness when search term matches Start of the item name
- higher fitness when search term matches part of item name which starts with capital letter
(usually user enters "sort" when looking for "SortedDictionary", he does not usually enter "orted")

siegfriedpammer:
Sorry for taking sooo long to merge this. Thank you for your work!

Issue #670, opened 2015-12-30
[Feature Request] Detect F# using patterns when decompiling to C#


0x53A:
I have a few F# projects. Since there is no F# decompiler in ILSpy, I often just decompile it to C#.
This works, but is often not pretty.

I have an example of how it decompiles a using block here: https://gist.github.com/0x53A/5c8dedc0be448a299386

Not sure why the compiler generates these superfluous instructions in the finally block, I was compiling in release mode.

0x53A:
Adding the F# pattern was actually relatively straightforward: 
https://github.com/0x53A/ILSpy/tree/fsharp-using-pattern
https://gist.github.com/0x53A/5c8dedc0be448a299386#file-c-with-my-changes

0x53A:
This specific issue was fixed with https://github.com/icsharpcode/ILSpy/pull/671, but there are a few other instances of F# code being decompiled .. unsatisfactory .., so if you don't mind, I'll keep this open for now (or should I create separate issues?).

dgrunwald:
F# using is broken once again, @siegfriedpammer: please re-enable the F# test case.

If there are any other problems, please create new issues.

PR #671, opened 2015-12-30
modify the using pattern to also detect F# [#670]

0x53A:
https://gist.github.com/0x53A/5c8dedc0be448a299386#file-c-with-my-changes

### DIFF
--- a/ICSharpCode.Decompiler/Ast/Transforms/PatternStatementTransform.cs
+++ b/ICSharpCode.Decompiler/Ast/Transforms/PatternStatementTransform.cs
@@ -58,11 +58,12 @@ namespace ICSharpCode.Decompiler.Ast.Transforms
 		public override AstNode VisitExpressionStatement(ExpressionStatement expressionStatement, object data)
 		{
 			AstNode result;
-			if (context.Settings.UsingStatement) {
-				result = TransformUsings(expressionStatement);
+			if (context.Settings.UsingStatement)
+			{
+				result = TransformNonGenericForEach(expressionStatement);
 				if (result != null)
 					return result;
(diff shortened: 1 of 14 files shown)


dgrunwald:
What if the `disposable` variable is still used later (after the finally-block)?
You're changing semantics by effectively removing the variable assignment.

`TransformUsings` already checks that `ident` is not used after the try-catch statement (the `IsVariableValueUnused` call); you'll need to add the same check for the `disposable` variable.

0x53A:
Thanks, I did not think about that case.

I will fix it (and maybe add a unit test) next week.

0x53A:
Ok, looks good from my side. @dgrunwald , could you please take a look?

dgrunwald:
When running the tests locally on my machine, the F# tests fail with two F# compiler errors:
<startup (1,1)-(1,81) parameter error Could not resolve this reference. Could not locate the assembly "FSharp.Core, Version=4.4.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a". Check to make sure the assembly exists on disk. If this reference is required by your code, you may get compilation errors. (Code=MSB3245)>,
<startup (1,1)-(1,81) parse error internal error: BuildFrameworkTcImports: no resolution of 'FSharp.Core, Version=4.4.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'>

Instead of running various compilers as part of our tests, perhaps we should commit disassembled binaries as `.il` files and re-assemble them in the tests?

0x53A:
The Test now uses ILASM to compile checked-in IL instead of running the F# compiler.
I have also removed references to FSharp.Core from the IL.

PR #674, opened 2016-01-15
Fixed duplicate preject references (cause failed clean build).

kekyo:
* Thanks useful tools!
* I built from current commit, but failed parallel-compilation on VS2012.
* Because VS2012 conflict between "assembly reference" and "project reference" in  ICSharpCode.Decompiler.csproj.
* Fixed, unified to project references.


dgrunwald:
Those assembly references are conditional (`'$(BuildNuGetPackage)' == 'True'`) and are used when building the ICSharpCode.Decompiler NuGet package.

Does VS2012 not support conditional references?

PR #681, opened 2016-02-19
Allow scrolling in high resource rows

sandercox:
Allow grid listview with resource tables to scroll items that are higher then the listview rows.

### DIFF
--- a/ILSpy/Controls/ResourceStringTable.xaml
+++ b/ILSpy/Controls/ResourceStringTable.xaml
@@ -36,6 +36,7 @@
 			  Foreground="Black"
 				  Grid.Row="1"
 				  AlternationCount="2"
+				  ScrollViewer.CanContentScroll="False"
 				  ItemContainerStyle="{StaticResource alternatingWithBinding}">
 		<ListView.View>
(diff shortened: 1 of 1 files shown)


siegfriedpammer:
merged... thanks for the contribution!

siegfriedpammer:
I'm sorry, but I guess, I will have to revert this, because it causes the performance to suffer greatly if you try to open a large resources file. For example: System.resources in System.dll. This is due to the fact that virtualization is disabled as soon as you set `ScrollViewer.CanContentScroll` to false.

Could you post an example? I could try to find a different solution in order to fix this, plus having virtualization.

sandercox:
Hi,

Preview of failing [LargeResourceTest.exe](http://sandercox.com/LargeResourceTest.exe) or if you'd like the original code [LargeResourceTest](https://github.com/sandercox/LargeResourceTest)

I was unaware this caused bad performance.

sandercox:
I tried playing around with putting list view content in a `ScrollViewer` but was yet unable to auto restrict the width to the max width. Using a `ScrollViewer` for each content would allow the content itself to be scrollable that would fix the issue.

PR #687, opened 2016-03-29
Write last access time

WoniMK7:
In order to let devs beeing able to check if their assembly have been decompiled, I make ILSpy write the last access time. Decompiling could be checked by a tool like [this.](https://github.com/WoniMK7/dotnetSecure)
What do you think of this idea ?


christophwille:
I personally don't think this to be a feature that many dev's would expect or like.

siegfriedpammer:
First of all: No, I would neither expect, nor appreciate such behavior...
Second: I believe it blows up as soon as you try to access assemblies you don't have write access to.

Issue #660, opened 2015-12-04
Add an "Open in ILSpy" option in the Call Stack window? 


kiewic:
It would be awesome if I could open ILSpy from a method call in the _Call Stack_ window during debugging, similar to the _Open in ILSpy_ in the _References_ list.

Moreover, ILSpy could go directly to the definition of the method selected in the call stack, or show results for types matching with the method name.

donid:
I have not found a way to retrieve the currently selected row in the "Call Stack" window, but as workaround we can use the active frame (the one with the curved arrow in the screenshot). 
I could provide a PR for a feature that navigates to the method of the currently active frame - see screenshot.

![openstackframeinilspy](https://cloud.githubusercontent.com/assets/6991562/14664694/6b087678-06cd-11e6-90eb-68f873a4a61c.jpg)

donid:
So, nobody interested in this feature? I was hoping for some comments about the usability of my proposed workaround and maybe the menu-caption that I have chosen.

BTW: this issue seems to be missing the "VS Extension" label.

siegfriedpammer:
I tried to the feature, but I could not find the menu item in the call stack context menu.

donid:
@siegfriedpammer: I don' quite understand your comment - the feature currently only exists in a build on my local machine. I presented the screenshot to receive comments about the naming and position of the planned menu item. In addition I hoped for comments about my planned workaround  to use the "active frame" instead of the "selected frame" - I think it is not possible to retrieve the "selected frame" through VS-extensibility (@bbi-yoyo mentioned his still unanswered StackOverflow-posting in #697). And I thought maybe @kiewic would comment if this proposal meets his needs.

bbi-yoyo:
I do like the idea of this feature, but would certainly prefer to be able to go to any stack frame. If you can figure out how to get the code element for a stack frame you can look at my implementation of "Open Code in ILSpy" (CodeElementXmlDocKeyProvider.cs) to see how to navigate straight to the item.

Hmm, thinking about that a bit -- VS is only going to give you a code element for source code. If you want to navigate to a stack frame for which you don't have source (assuming you can get info on the stack frame in the first place) you will likely need to figure out the xml doc key generation yourself.

Also note that for stack frames for which you *do* have source, you can double-click the stack frame to get to the source, then right-click in the source code and use the new Open Code in ILSpy context menu item. Not as direct as going from the stack trace itself, but a decent workaround.

donid:
[...]

PR #697, opened 2016-04-20
Fix "Open code in ILSpy" issues with certain member types and overloaded methods and constructors.

### LINKED ISSUE #696: Fix "Open code in ILSpy" context menu to support constructors, fields, overloaded methods and generics.
The implementation of #695 (to open code in ILSpy from the context menu in the code window) will navigate to a method just by name, not parameters. As a result it does not correctly support overloaded methods. It also fails to navigate to constructors and fields, and does not support generic classes and methods correctly.

yoyobbi:
Fixes #696.

"Open code in ILSpy" now works for:
- classes, interfaces and structs (generic or not, nested or not)
- enums
- fields and properties
- constructors (generic or not, overloaded or not)
- methods (generic or not, overloaded or not)


yoyobbi:
The string generation got messier than I expected. The end result is effectively duplicating the code from XmlDocKeyProvider.GetKey, but taking an EnvDTE.CodeElement as input instead of Mono.Cecil.MemberReference. Feel free to suggest a better way.

I also tried to find a way to implement #660 but I couldn't find any procedural interface to the Call Stack window. If there is such a thing, and if you can get a CodeElement from it, then you could use my new OpenCodeItemInILSpy method to do the dirty work.

yoyobbi:
Note that every element in this file has been tested:
https://gist.github.com/yoyobbi/0e185f3a574db9d5f3ec3ab08ede2ce1

I have not tested methods with parameters of custom or array types, generic interfaces and structs, enums inside structs, and various other edge cases that may arise in the wild. I'm reasonably confident most things should work.

yoyobbi:
Hmm, running into further edge cases with method parameters of generic types. Going to refine my implementation of key generation to more closely follow the existing XmlDocKeyProvider. Closing this pull request for now, will reopen when ready.

yoyobbi:
This ended up more complicated than I originally expected, but it's 99% working now. There are still some edges cases with multi-dimensional array parameters to overloaded methods that are not being disambiguated correctly. I suspect there might be a bug in the /navigateTo feature itself? Have not had time to investigate further.

This is as done as I intend to make it for now. See ILSpy.AddIn/Samples/ILSpyAddInSamples.cs for a lengthy set of examples that *do* work.

Feedback welcome!

ILSpy.AddIn/CodeElementXmlDocKeyProvider.cs
```diff
@@ -0,0 +1,241 @@
...
+	/// <remarks>
+	/// Used to support the "/navigateTo" command line option when opening ILSpy. Must match
+	/// the logic of ICSharpCode.ILSpy.XmlDoc.XmlDocKeyProvider, which does the same thing for
+	/// a Mono.Cecil.MemberReference. See "ID string format" in Appendix A of the C# language
+	/// specification for formatting requirements.
+	/// </remarks>
+	public static class CodeElementXmlDocKeyProvider
+	{
+		#region GetKey
+		public static string GetKey(EnvDTE.CodeElement member)
```
yoyobbi: NOTE: I initially attempted to follow the structural logic of XmlDocKeyProvider, but found that EnvDTE.CodeElement and friends don't precisely match their Mono.Cecil counterparts, and the end an approach based mainly on string parsing (of type FullNames) proved much simpler.

ILSpy.AddIn/ILSpy.AddIn.csproj
```diff
@@ -147,6 +148,7 @@
     <Compile Include="ILSpyAddInPackage.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="PkgCmdID.cs" />
+    <None Include="Samples\ILSpyAddInSamples.cs" />
```
yoyobbi: Samples provided for reference only, not compiled or copied to the build output.

ILSpy.AddIn/ILSpyAddInPackage.cs
```diff
@@ -119,82 +120,70 @@ private void OpenProjectOutputInILSpyCallback(object sender, EventArgs e)
...
-				string projectPath = Path.GetDirectoryName(project.FileName);
-				string outputPath = config.Properties.Item("OutputPath").Value.ToString();
-				string assemblyFileName = project.Properties.Item("OutputFileName").Value.ToString();
-				OpenAssemblyInILSpy(Path.Combine(projectPath, outputPath, assemblyFileName));
+				OpenProjectInILSpy(project);
+			}
+		}
+
+		// Called when the menu is popped, determines whether "Open code in ILSpy" option is available.
+		private void OpenCodeItemInILSpyCallback_BeforeQueryStatus(object sender, EventArgs e)
```
yoyobbi: This will disable the "Open code in ILSpy" context menu item if, for example, you are editing a text file that is not part of any project.

ILSpy.AddIn/Samples/ILSpyAddInSamples.cs
```diff
@@ -0,0 +1,420 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+
+/// Sample source file indicating a wide variety of code elements that should work
```
yoyobbi: Each item includes its "/navigateTo" string as a comment. Might be worth someone putting some of these into the documentation for the ILSpy command line option, rather than just saying "The syntax of ID strings is described in appendix A of the C# language specification."

ILSpy.AddIn/Samples/ILSpyAddInSamples.cs
```diff
@@ -0,0 +1,420 @@
...
+		// M:ILSpy.AddIn.Tests.SomeGenericClass`2.GenericClassGenericMethod``2(`0,``0,System.Int32[])
+		public T3 GenericClassGenericMethod<T3, T4>(T1 x, T3 y, int[] z)
+		{
+			mField1 = x;
+			string foo = y.ToString() + z.ToString();
+			return y;
+		}
+
+		// M:ILSpy.AddIn.Tests.SomeGenericClass`2.ArrayMethod(`0[],System.Int32[])
+		public void ArrayMethod(T1[] x, int[] y)
```
yoyobbi: The first and last of these four ArrayMethod overloads are navigated to correctly, but the second and third incorrectly navigate to the fourth. Possible bug in /navigateTo implementation?

siegfriedpammer:
looking good so far... thanks for the contribution.

yoyobbi:
Thanks Siegfried! Sorry for my half-baked first attempt, should be pretty solid this time around. There may still be subtle edge cases (reference parameter that's a jagged array of pointers??) that will confuse it, but it definitely deals with most normal cases correctly now. Please put it through its paces if you have a chance.

yoyobbi:
BTW, from the C# language spec:

> Arguments that are arrays are represented as [ lowerbound : size , … , lowerbound : size ] where the number of commas is the rank less one, and the lower bounds and size of each dimension, if known, are represented in decimal. If a lower bound or size is not specified, it is omitted. If the lower bound and size for a particular dimension are omitted, the “:” is omitted as well. Jagged arrays are represented by one “[]” per level.

I'm pretty sure my code is formatting multidimensional and jagged arrays as per the spec, but the navigation isn't working correctly.

dgrunwald:
You can you put XmlDoc comments on the sample and have the C# compiler generate the `.xml` file to see what the correct name is.
I'm getting `System.Int32[0:,0:]` from the C# compiler.

yoyobbi:
@dgrunwald, good idea, I will try that.

On that note, do you know if there's a way to get Visual Studio to just generate the xml key for me in the first place and not have to jump through all these hoops?

PR #707, opened 2016-05-05
Support for BamlRecordType.TypeSerializerInfo in bamlreader

Sermus:
We faced a situation when csc submits TypeSerializerInfo into the baml being generated. ILSpy doesn't support it so it fails to decompile the baml. From the MSFT source code the binary layout for TypeSerializedInfo is the same as for TypeInfo, so i just added TypeSerializerInfo handling the same way it was done for TypeInfo.


siegfriedpammer:
Looking good! Thanks for contributing!

PR #720, opened 2016-06-19
Decompiler can now introduce `var` variable declarations.

exyi:
The behavior can turned on/off in DecompilerSettings. You can also use `var` only if the type name is already used on the line or if it's long (hardcoded 15 chars).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ _ReSharper*/
 *.patch
 /packages
 *.ide/
+/.vs/config/applicationhost.config
--- a/ICSharpCode.Decompiler/Ast/Transforms/TransformationPipeline.cs
+++ b/ICSharpCode.Decompiler/Ast/Transforms/TransformationPipeline.cs
@@ -45,7 +45,8 @@ namespace ICSharpCode.Decompiler.Ast.Transforms
 				new IntroduceExtensionMethods(context), // must run after IntroduceUsingDeclarations
 				new IntroduceQueryExpressions(context), // must run after IntroduceExtensionMethods
 				new CombineQueryExpressions(context),
-				new FlattenSwitchBlocks(), 
+				new FlattenSwitchBlocks(),
+				new UseVarKeywordTransform(context.Settings)
 			};
 		}
 		
--- a/ICSharpCode.Decompiler/DecompilerSettings.cs
+++ b/ICSharpCode.Decompiler/DecompilerSettings.cs
@@ -266,6 +266,18 @@ namespace ICSharpCode.Decompiler
 				}
 			}
 		}
+
+		VarKeywordUsage useVar = VarKeywordUsage.WhenTypeIsEvidentOrLong;
+
+		public VarKeywordUsage UseVar {
+			get { return useVar; }
+			set {
+				if(value != useVar) {
+					useVar = value;
+					OnPropertyChanged(nameof(UseVar));
+				}
+			}
+		}
 		
 		#region Options to aid VB decompilation
 		bool introduceIncrementAndDecrement = true;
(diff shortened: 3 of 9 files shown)


dgrunwald:
What settings were the tests using before you changed them to use `VarKeywordUsage.Never`?

I find this test failure interesting:
 49      (-) 		int firstByte = secondByte;
 49 (+) 		var firstByte = secondByte
Why would the code use 'var' if the tests were running in the default mode of `WhenTypeIsEvidentOrLong`?

exyi:
It considers the type evident if it is assigned from local variable. It is documented at line [57, UseVarKeywordTransform.cs](https://github.com/icsharpcode/ILSpy/pull/720/files#diff-6a2d06a8250df95fd827f2eec28895caR57)
Btw: I have just fixed that it did not mind using var if the initializer was null or anon method

dgrunwald:
I think "evident" should just be cases where the type name is obvious given the same code line. (casts, `... as $type`, `new $type`)
Replacing `int` with `var` is a bit too aggressive for my liking.

exyi:
I personally like to use `var` everywhere when coding because it is easier to write and the code is more flexible.
In ILSpy none of these is a goal, I just want to make code more readable so I have added an limit that only type names longer that 6 chars can be replaced by var. In my opinion the type information is not so important to understand the code and when the type names are long and often repeating it generated more noise that information for me.
I can of course easily make another level of "aggresivity", but ... do you have name for it? :)

lextm:
`WhenTypeIsEvident` seems to match [.NET Core Coding Style](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md), if that's intentional.

.gitignore
```diff
@@ -12,3 +12,4 @@ _ReSharper*/
 *.patch
 /packages
 *.ide/
+/.vs/config/applicationhost.config
```
sharwell: :bulb: This should be just the following:

```
.vs/
```

It also replaces the previous `*.ide/` exclusion for everything except some preview releases of Visual Studio 2015.

ICSharpCode.Decompiler/Ast/Transforms/TransformationPipeline.cs
```diff
@@ -45,7 +45,8 @@ public static IAstTransform[] CreatePipeline(DecompilerContext context)
 				new IntroduceExtensionMethods(context), // must run after IntroduceUsingDeclarations
 				new IntroduceQueryExpressions(context), // must run after IntroduceExtensionMethods
 				new CombineQueryExpressions(context),
-				new FlattenSwitchBlocks(), 
+				new FlattenSwitchBlocks(),
+				new UseVarKeywordTransform(context.Settings)
```
sharwell: ❓ Why not pass `context` for consistency with the cases above?

ICSharpCode.Decompiler/DecompilerSettings.cs
```diff
@@ -353,4 +365,12 @@ public DecompilerSettings Clone()
 			return settings;
 		}
 	}
+
+	public enum VarKeywordUsage
+	{
+		Never,
+		WhenTypeIsEvident,
+		WhenTypeIsEvidentOrLong,
```
sharwell: 💭 I question the value of this specifically.

dgrunwald:
This PR is no longer compatible now that we've merged the `newdecompiler` branch.

Feel free to re-submit for the new decompiler engine.
Also, make sure to only change to `var` if doing so doesn't change the type of the variable -- e.g. with `object x = new SomeStruct();` it's critical that we don't remove the boxing conversion.

Issue #723, opened 2016-06-28
Failure to load assemblies with a portable pdb


AustinWise:
If you compile hello world using the new .NET Core CLI:
````
dotnet new
dotnet restore
dotnet build
````
and try to open the resulting DLL using IlSpy, the DLL fails to load with this exception:
````
System.OutOfMemoryException: Array dimensions exceeded supported range.
   at Microsoft.Cci.Pdb.MsfDirectory..ctor(PdbReader reader, PdbFileHeader head, BitAccess bits)
   at Microsoft.Cci.Pdb.PdbFile.LoadFunctions(Stream read, Dictionary`2& tokenToSourceMapping, String& sourceServerData, Int32& age, Guid& guid)
   at Mono.Cecil.Pdb.PdbReader.PopulateFunctions()
   at Mono.Cecil.Pdb.PdbReader.ProcessDebugHeader(ImageDebugDirectory directory, Byte[] header)
   at Mono.Cecil.ModuleDefinition.ProcessDebugHeader()
   at Mono.Cecil.ModuleDefinition.ReadSymbols(ISymbolReader reader)
   at ICSharpCode.ILSpy.LoadedAssembly.LoadSymbols(ModuleDefinition module)
   at ICSharpCode.ILSpy.LoadedAssembly.LoadAssembly(Object state)
   at System.Threading.Tasks.Task`1.InnerInvoke()
   at System.Threading.Tasks.Task.Execute()
````

The root cause is the cecil class [PdbFileHeader](https://github.com/jbevain/cecil/blob/042bdd2432705ae4c682743ff9740550dee3bead/symbols/pdb/Microsoft.Cci.Pdb/PdbFileHeader.cs) does some not check the header of the PDB file.

This could worked around in IlSpy by [catching more exceptions](https://github.com/icsharpcode/ILSpy/blob/e2e952a9be85c57eed8c70d7315281a6724b7049/ILSpy/LoadedAssembly.cs#L134) during symbol loading.

AraHaan:
You know you can make a Pull Request right?

tyrotoxin:
FYI: this is still the issue for .NET Standard assemblies

valhristov:
I workarounded this by adding `<DebugType>Full</DebugType>` to my projects. It also helps with collecting code coverage, see https://github.com/Microsoft/vstest/issues/800...

pkruk2:
[...]

PR #724, opened 2016-06-28
Work around #723 by eating all symbols during loading.

AustinWise:
This patch allows assemblies with portable PDBs to be loaded.
Obviously CECIL should be fixed, but this get's IlSpy working again in the meantime.


siegfriedpammer:
If we swallow *all* exceptions during loading, we need to log them, notify the user that there was a problem during symbol load, and make the logs available to the user. Otherwise we will never be able to find and fix these problems. It's not as if the symbols aren't available, thus we should try our best to be able to read them.

AustinWise:
Ok, I will try to fix the underlying error in Cecil instead.

PR #738, opened 2016-07-22
Update Cecil to 7efc4a6de7fba02965daaf407fe557ffee725b79 (master).

### LINKED ISSUE #723: Failure to load assemblies with a portable pdb

AustinWise:
This fixes #723.

The problem is Cecil will blindly try to open portable PDBs as regular PDBs. I fixed this in jbevain/cecil#279 by validating the headers so that if it a Portable PDB is opened, it throws the exception type that ILSpy expects to be thrown. This prevents ILSpy from loading the assembly at all

Cecil recently added support for portable PDBs, so a future pull request could update to a new commit and use that capability.


siegfriedpammer:
Thanks for your pull request. We will look into upgrading to a more recent version of Cecil. As ILSpy heavily depends on Cecil, lots of testing is required. As you can imagine, this will take some time.

AustinWise:
Understood. In that case it might be more worth while to update to a version that include PortablePDB support, so the testing can just be done once. I'll close this and fix the issue in IlSpy directly with a different pull request.

Issue #519, opened 2014-12-01
How should decompiler tests work?


SLaks:
@JasonMalinowski pointed out that decompiler tests should not invoke the compiler.

Instead, we should have a set of test cases with IL source and expected decompiled output (more in the style of a regression test).  To create tests, we would have a generator that takes C# source, compiles it, and spits out test cases with IL & expected outputs.  This offers a number of advantages:
- We can easily create test cases for arbitrary compilers that would be difficult to include at runtime, such as Mono's MCS, multiple versions of Roslyn, etc...
- We will no longer need to ensure that the decompiled output matches the original source immediately, since we could manually tweak either the IL or the C# to account for differences that the decompiler cannot cover.  This would let us run a lot of tests that are currently skipped due to such issues
- Debugging test cases will be easier, since the IL being decompiled will be right there.

What do you think?
I may start implementing this soon (I'm not yet sure how much time I'll have)

dgrunwald:
On the newdecompiler branch, we now have two categories of test:

1. "pretty tests" use `.il` + `.cs` files. If the `.il` files do not exist yet, they are automatically created from the `.cs` file the first time the test is run. But we commit these files to version control to be robust to future compiler changes.

2. "correctness tests" just have a `.cs` file (or sometimes: just a `.il` file). Here, we don't compare the decompiled source code, but instead the runtime output (thus, these tests need lots of `Console.WriteLine` calls to ensure the test can detect differences in runtime behavior).
These tests don't care if we end up introducing lots of temporary variables or even goto statements, as long as the decompiled code still produces the same output.

PR #743, opened 2016-08-01
Add reference to _this_ and _base_ keywords in ctor initializer

Zvirja:
I've found that significant improvement could be applied to the ILSpy. Consider the following code:
image

Unfortunately, the `base` and `this` words near the constructors are not clickable. 
I've extended code to make them clickable. They will point to appropriate ctor definition.
That will hugely simplify the code navigation.

P.S. I've created issue in NRefactory project ([#519](https://github.com/icsharpcode/NRefactory/issues/519)), but later realized that it's a pure ILSpy issue.


dgrunwald:
Thanks!

PR #756, opened 2016-09-04
Ui improve pane experience

Zvirja:
Fixed two issues.
The first issue is just a typo and top panel size was not saved on tool close.

The second issue leaded to the small pane size when it's displayed. Consider the following scenario:
- You open the `Source code` and the `Search panel`. After that you use a grid resizer and make the panels the same height. Resizer assigned heights `250*` and `250*`. Yes, resizer assigns values greater than 1.
- You open the `Analyze` panel. Tool finds from the settings that the bottom pane ratio comparing to the source code is `0.5`. As result, heights become the following:
  - top pane: `250*`
  - text view: `0.5*`
  - bottom pane: `0.5*`

Obviously, the top pane became a "Godzilla", while rest panels took their minimum.

In order to avoid such issues, I run normalization and ensure that all heights are less then 1 before a new pane is opened.

### DIFF
--- a/ILSpy/MainWindow.xaml.cs
+++ b/ILSpy/MainWindow.xaml.cs
@@ -872,10 +872,40 @@ namespace ICSharpCode.ILSpy
 		}
 		
 		#region Top/Bottom Pane management
+
+		/// <summary>
+		///   When grid is resized using splitter, row height value could become greater than 1.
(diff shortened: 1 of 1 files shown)


Zvirja:
Thank you!

dgrunwald:
Sorry it took so long.

PR #758, opened 2016-09-10
Improved accessibility support

glenrgordon:
Added UI Automation support to SharpTreeview.  Improved support for keyboard only users when it comes to navigating between controls in the UI.
With these changes, ILSpy is a viable option for blind developers.


dgrunwald:
As far as I can tell, 76e942e is the only relevant commit in this PR?

1. The ToString() implementation for the searchModeComboBox is not required; this can be done more cleanly by setting `TextSearch.TextPath="Name"`.
2. Was the whole point of the `MyPreviewKeyDown` method to override AvalonEdit's handling of the `Tab` key? I think this is better solved by removing the InputBinding+CommandBinding for EditingCommands.TabForward+TabBackward.

dgrunwald:
I merged the remaining changes as 82ca308a3b8825a1bcbe4c5c50d0329a76e96015.

Unfortunately I didn't notice before pushing that removing the redundant MyPreviewKeyDown+ComboBoxItem changes from the commit reset the author information :(

glenrgordon:
Hi, Daniel,

 

No problem on the lack of attribution.

I don’t care at all.

 

You deduced correctly on all issues except for ToString where I should have put in a specific comment explaining the rationale.

When a data template is used for a combobox, WPF UI Automation doesn’t use the template for the current value.  Rather, it uses the result of ToString on the Combobox itself which by default is the class name.

AutomationProperties.Name is the name of the combobox itself. There’s no property exposed directly to WPF that allows explicitly setting the value.

Without this hack, ScreenReaders will say the class name rather than the current value when the combobox isn’t expanded.

Most screen readers will also get it wrong when the combobox is dropped, but since in that case the expansion of the template is represented in the UIA hierarchy as children of the focused item, the JAWS screen reader does expansion itself.

 

TextSearch.TextPath does fix first-letter navigation but has no impact on UIAutomation.

 

Sorry for the improper indentation.  Tried to get it right but I’m blind so inconsistencies don’t jump out at me.

 

 

Thanks

 

--Glen

PR #767, opened 2016-10-03
Add "Open Directory" command

OmarTawfik:
image

image

### DIFF
--- a/ILSpy/Commands/DecompileAllCommand.cs
+++ b/ILSpy/Commands/DecompileAllCommand.cs
@@ -25,7 +25,7 @@ using ICSharpCode.ILSpy.TextView;
 
 namespace ICSharpCode.ILSpy
 {
-	[ExportMainMenuCommand(Menu = "_File", Header = "DEBUG -- Decompile All", MenuCategory = "Open", MenuOrder = 2.5)]
+	[ExportMainMenuCommand(Menu = "_File", Header = "DEBUG -- Decompile All", MenuCategory = "Open", MenuOrder = 6)]
 	sealed class DecompileAllCommand : SimpleCommand
 	{
 		public override bool CanExecute(object parameter)
--- a/ILSpy/Commands/OpenDirectoryCommand.cs
+++ b/ILSpy/Commands/OpenDirectoryCommand.cs
@@ -17,18 +17,16 @@
 // DEALINGS IN THE SOFTWARE.
 
 using System;
+using System.Windows.Input;
 
 namespace ICSharpCode.ILSpy
 {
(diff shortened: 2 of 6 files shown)


OmarTawfik:
@siegfriedpammer can you please tell me who should be tagged on this PR?

OmarTawfik:
@icsharpcode

OmarTawfik:
Would you know who maintains this repo now? :) @dgrunwald @christophwille

lextm:
I like this PR, but more if it can use the Vista style folder browser dialog, http://www.ookii.org/software/dialogs/

yuryleb:
It would be better if 'Open Directory' item text will be finished with dots - 'Open Directory...'

dgrunwald:
I can already open a whole directory by using the normal "Open..." command, browsing to the directory, and pressing Ctrl+A followed by Enter. (and I think doing this might be easier than using the pre-Vista folder browser...)

akyryz:
I don't like to navigate using those browsers =\
It needs at least to have a textbox to paste the path to go straight where I want.

Zvirja:
> It needs at least to have a textbox to paste the path to go straight where I want.

On my Windows 10 I can paste path in either fields and press `Enter` to quickly navigate to specific directory.
image

Could you please clarify the cases when this approach is not very convenient and which particular usability obstacles you meet?

siegfriedpammer:
This does not make any sense to me... using the old folder browser dialog does not offer anything that cannot be done using the current "open" dialog... IMHO it's a bad idea to use the old dialog. You cannot quickly browse to a completely different location. When an application only supports the old one, I feel like it's slowing me down.

As @dgrunwald pointed out, loading all assemblies in a folder can be accomplished by browsing to that location and pressing ctrl+A.

OmarTawfik:
@dgrunwald @siegfriedpammer @akyryz
Would this be satisfied by using the newer dialog? I made this PR because I ended up needing a "recursive" open directory command. I need to open my build output folder and load all dlls from all folders there.

Zvirja:
I'd say that this is useful, but very specific feature. I'm not sure it worth it to pollute menu because of it. 

What if you extend drag-n-drop instead? So if folder is dropped to window, it recursively opens all the files inside that folder. That would allow to get the functionality you need bit in more "hidden" way.

OmarTawfik:
That satisfies my scenario.  Others?

Zvirja:
I've just noticed today how that it's implemented in the dotPeek by JetBrains and I'm impressed. I believe that it's how this feature should be implemented in ILSpy as well.

image

Rather than add all files in row, they are added as `Folder` elements. It allows to easily delete folders at once.
Additionally, they support zip archives without need to extract assemblies from them - but it's another story.

PR #777, opened 2016-11-21
Cutting root namespace from sub-directory name for decompiled sources

yuryleb:
Small C# code decompilating improvement to cut root (main) assembly namespace from generated sub-folder names to place decompiled source files. This patch allows to place C# project sources with almost the same file structure as VS does by default.


dgrunwald:
There are plenty of assemblies where the assembly name is not a prefix of the namespace names (e.g. most of the Microsoft System.* assemblies).
I don't think this a good idea to use the short folder name for some but not all namespaces in an assembly.

yuryleb:
Yes, but maybe it could be better just to use another way to determine root namespace - enumerate all objects namespaces and cut first common part? ;)

Issue #785, opened 2016-12-16
astbuilde.MakePrimitive should add cast while target type is not int


loobins:
IL_004f: ldc.i4 165
		IL_0054: callvirt instance void [mscorlib]System.IO.BinaryWriter::Write(uint8)

should be  decompiled to Write((byte)165) rather than Write((byte)165) .

//astbuilder.cs  Line:1612
 if(type.MetadataType == MetadataType.Char  || type.MetadataType == MetadataType.SByte  || type.MetadataType == MetadataType.Byte)
            {
                    return new Ast.PrimitiveExpression(CSharpPrimitiveCast.Cast(code, val, false)).CastTo(ConvertType(type));
                    
            }
            return new Ast.PrimitiveExpression(CSharpPrimitiveCast.Cast(code, val, false));

siegfriedpammer:
Fixed on the newdecompiler branch.

PR #786, opened 2016-12-16
fix for issue 206:MakePrimitive should add cast while target type is …

loobins:
MakePrimitive should add cast while target type is not int.


siegfriedpammer:
See my comment on #785 - This is already fixed on the new decompiler branch.

PR #791, opened 2017-01-16
Implemented really simple GoToMetadataAsSourceService backed by ILSpy

exyi:
Hi, I have implemented proof-of-concept version of MetadataAsSourceService, so you get decompiled source when you press F12 on metadata type.

As you can see, it is pretty simple and stupid - it just returns the decompiled string from the right interface, however it works surprisingly well, VisualStudio's semantic (GoToDefinition, QuickInfo) are all working.

As the `IMetadataAsSourceService` is internal I had to place it to assembly called DynamicProxyGenAssembly2 signed with right key because they have InternalsVisibleTo fir it ;) http://source.roslyn.io/#Microsoft.CodeAnalysis.Features/GeneratedInternalsVisibleTo.cs


Dreamescaper:
Cool! I'm using extension with your changes on regular basis now.

siegfriedpammer:
This is a horrible hack. We'll have to wait for dotnet/roslyn#16540 to be implemented and officially supported. Thank you for your effort, though...

PR #800, opened 2017-03-29
Add context menu action to copy FQN for selected class

Zvirja:
Add context menu action to copy Fully Qualified Type name.

How it looks:
image

Result for nested types: `Namespace1.NS2.TypeName+Nested+SubNested`

This is the representation .NET uses for types.


siegfriedpammer:

PR #734, opened 2016-07-14
Automatic read only properties new decompiler

pcalin:

PR #802, opened 2017-04-13
Add support for the read-only properties decompilation

zvirja:
Add support for the get-only properties decompilation. It's a rework of the #734. It fixes issues which were present in that version of PR (e.g. doesn't fail if you select property only).

Let me know if you have any concerns regarding the implementation. I've tried to make patterns as strict as possible (I'd say even overstrict ).

I intentionally haven't introduced a standalone setting for this feature for 2 reasons:
1. Read-only properties are just a kind of properties, they are still auto-properties.
2. If you enable auto-properties, but disable read-only properties, you will not have compiler generated fields for properties - [this](https://github.com/icsharpcode/ILSpy/blob/bdfc3903e11031e2403f2ef426ec6f83a621c10c/ICSharpCode.Decompiler/Ast/AstBuilder.cs#L99) code removes them. It would require extra-logic to separate fields which belong to R/W and R/O properties.

BTW, I looked for the place to put tests, however wasn't able to find it (I don't see similar tests, just some test classes). Does this project have tests? 😟

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/Transforms/PatternStatementTransform.cs
+++ b/ICSharpCode.Decompiler/CSharp/Transforms/PatternStatementTransform.cs
@@ -25,6 +25,7 @@ using ICSharpCode.Decompiler.CSharp.Syntax.PatternMatching;
 using ICSharpCode.Decompiler.CSharp.TypeSystem;
 using ICSharpCode.Decompiler.TypeSystem;
 using ICSharpCode.Decompiler.CSharp.Analysis;
+using ICSharpCode.Decompiler.Semantics;
 using Mono.Cecil;
 
 namespace ICSharpCode.Decompiler.CSharp.Transforms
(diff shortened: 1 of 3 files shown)


zvirja:
@dgrunwald, @siegfriedpammer Could you please suggest who should be assigned as a reviewer for this PR? I noticed that @dgrunwald reviewed the previous #734 one, so probably he is the most right person..

zvirja:
I've just realized, that I mistakenly implemented the feature on top of the `master`, so I reworked the approach and implemented it on top of `newdecompiler`.

The build fails, but likely due to wrong AppVeyor configuration.

siegfriedpammer:
Thank you for your pull request. Sorry, I completely missed it, when I implemented the automatic readonly properties a while back. I added support for property initializers in the AST as well. The property transformation part is done in `PatternStatementTransform.cs` (similar to your approach), but the constructor transformation part is done in `ConvertConstructorCallIntoInitializer.cs`. I will cherry-pick the commit about the settings.

zvirja:
@siegfriedpammer Thank you for your attention - happy to know that feature is finally implemented! 😉

Issue #810, opened 2017-05-29
Global config switch for separate instances


adamralph:
While it's possible to force a separate instance using the `/separate` CLI switch, it would be nice to be able to set this globally in the GUI options.

Radon222:
Resolved in Pull Request Allow Multiple Instance Option and Dialog #818
`https://github.com/icsharpcode/ILSpy/pull/818`

siegfriedpammer:
#818 was merged. Therefore closing this.

adamralph:
Fantastic. Thanks!

PR #818, opened 2017-06-16
Allow Multiple Instance Option and Dialog

Radon222:
Resolved the issue Global config switch for separate instances #810
`https://github.com/icsharpcode/ILSpy/issues/810`

Created an option to allow user modify if he wants multiple enabled or not
image


siegfriedpammer:
Thanks for your contribution.

PR #836, opened 2017-08-31
Update to the new project system

sharwell:


.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "cecil"]
 	path = cecil
 	url = https://github.com/jbevain/cecil.git
+[submodule "ILSpy-tests"]
+	path = ILSpy-tests
+	url = https://github.com/icsharpcode/ILSpy-tests
```
sharwell: :memo: I can remove this commit if it is undesirable. I found it much easier to "just work".

ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
```diff
@@ -1,129 +1,60 @@
...
+    <PackageReference Include="NUnit" Version="2.6.3" />
+    <PackageReference Include="NUnitTestAdapter" Version="2.0.0" />
   </ItemGroup>
+
   <ItemGroup>
-    <None Include="app.config" />
-    <None Include="packages.config" />
-    <None Include="TestCases\Correctness\BitNot.il" />
-    <None Include="TestCases\Correctness\Readme.txt" />
+    <ProjectReference Include="..\cecil\Mono.Cecil.csproj" AdditionalProperties="NuGetRestoreTargets=;ResolveNuGetPackages=false" />
```
sharwell: :bulb: The `AdditionalProperties` can be removed after **Mono.Cecil.csproj** is converted to the new project system.

.gitmodules
siegfriedpammer: @dgrunwald What do you think?

ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
```diff
@@ -1,129 +1,60 @@
...
-    <NoWin32Manifest>False</NoWin32Manifest>
+
+    <NoWarn>$(NoWarn),67,169,1058,728,1720,649</NoWarn>
+
+    <GenerateAssemblyInfo>False</GenerateAssemblyInfo>
+
+    <EnableDefaultItems>false</EnableDefaultItems>
+
+    <OutputType>Exe</OutputType>
+    <StartupObject>ICSharpCode.Decompiler.Tests.Stub</StartupObject>
```
siegfriedpammer: Is this a workaround for the tooling bug with .NET Core Test Runner, that I mentioned in my rant?

sharwell: At this point, I'm not using the .NET Core test runner or the `dotnet` tool for anything. I don't expect that to change prior to converting Mono.Cecil.csproj to the new project system. The change in output type was actually a workaround to ensure `AutoGenerateBindingRedirects` properly created a config file, which is then used by the test runner. This fixes a few different problems related to assembly load failures during test runs without having to manually maintain the redirects in an **app.config** file.

siegfriedpammer: Thanks for explaining and answering my follow-up question about `AutoGenerateBindingRedirects` before I even posted it.

ILSpy.BamlDecompiler/ILSpy.BamlDecompiler.csproj
```diff
@@ -114,32 +84,17 @@
...
+
   <ItemGroup>
-    <None Include="packages.config" />
     <None Include="Ricciolo.StylesExplorer.MarkupReflection\BAML format.txt" />
   </ItemGroup>
-  <Import Project="$(MSBuildBinPath)\Microsoft.CSharp.Targets" />
+
+  <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.targets" />
+
+  <Target Name="RemoveTransitiveProjectReferences" AfterTargets="IncludeTransitiveProjectReferences">
```
siegfriedpammer: What's the reason behind this build target?

appveyor.yml
```diff
@@ -7,7 +7,7 @@ install:
 before_build:
 - nuget restore ILSpy.sln
 build_script:
-- msbuild ILSpy.sln /v:minimal /p:ResolveNuGetPackages=false "/logger:%ProgramFiles%\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
+- msbuild ILSpy.sln /v:minimal /p:ResolveNuGetPackages=false /t:BeforeBuild,Build "/logger:%ProgramFiles%\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
```
siegfriedpammer: The AssemblyInfo.cs files are not generated in my fresh local test repository. It seems VS does not execute the `BeforeBuild` tasks anymore.
I fixed this by using `<Target Name="PreBuild" BeforeTargets="PreBuildEvent">` as generated by VS (added a dummy task using the UI) instead of `<Target Name="BeforeBuild">`, however then UpdateAssemblyInfo.exe will be executed once for each target framework, which is not only unnecessary, but, due to parallel build of both target frameworks we were running into race conditions on the build server. (Seems to work on my (slow) machine.)

ILSpy.BamlDecompiler/ILSpy.BamlDecompiler.csproj
sharwell: Since this project uses `<Private>false</Private>`  for its references, the expectation is the references will not be copied to the output directory. However, while references in the new project system are transitive, the `Private` flag is *not* transitive. This target disables transitive references to ensure that neither the referenced project nor any of its dependencies are copied to the output directory, preserving the original behavior.

appveyor.yml
sharwell: Issue was relative ordering of the import of **Sdk.targets** and the definition of the BeforeBuild target. I reversed the order to fix the build.

.gitmodules
dgrunwald: We use a local branch of this repo to also commit the decompiled code. This allows easily diffing the decompiler results after a test run to see what's changed. But we don't want to push these commits; and we certainly don't want to accidentally commit an updated submodule reference in the main ILSpy repository to point to such a 'decompiled' commit.

Is there a way to avoid 'git commit -a' and GUI tools like TortoiseGit picking up a submodule change if after such local commits? If not, I'd prefer if we didn't add this as a submodule.

dgrunwald: Actually, there's not really any reason why the test inputs and the decompiled test output needs to be in the same directory/repository.
We could eliminate ILSpy-tests altogether and commit the tests assemblies directly to the main repository (possibly using git LFS to avoid big binaries accumulating in the history); and anyone interested in diffing the outputs can 'git init' a local repo in the decompiled output directory.

sharwell: > Is there a way to avoid 'git commit -a' and GUI tools like TortoiseGit picking up a submodule change if after such local commits?

Yes. We can set `ignore = all` in .gitmodules for this submodule.

siegfriedpammer:
Thank you very much for your effort, Sam!

PR #847, opened 2017-09-05
Fix XAML IntelliSense

sharwell:


ILSpy/ILSpy.csproj
```diff
@@ -365,4 +354,23 @@
     <Resource Include="Images\Sort.png" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Page Update="@(Page)" SubType="Designer" Generator="MSBuild:Compile" />
+  </ItemGroup>
+
+  <ItemGroup>
```
sharwell: :memo: This group makes sure `Page` and `Resource` items appear in Solution Explorer. It's not the best known approach but it gets us by prior to the project system supporting it.

siegfriedpammer:
Thanks a lot!

PR #848, opened 2017-09-05
Implement basic support for .net standard/core assembly references

siegfriedpammer:


ICSharpCode.Decompiler/TypeSystem/DecompilerTypeSystem.cs
```diff
@@ -45,15 +45,48 @@ public DecompilerTypeSystem(ModuleDefinition moduleDefinition)
...
 			}
 			compilation = new SimpleCompilation(mainAssembly, referencedAssemblies);
+			if (compilation.FindType(KnownTypeCode.Void).Kind == TypeKind.Unknown || compilation.FindType(KnownTypeCode.Void).Kind == TypeKind.Unknown) {
+				referencedAssemblies.Add(MinimalCorlib.Instance);
+				compilation = new SimpleCompilation(mainAssembly, referencedAssemblies);
+			}
 			context = new SimpleTypeResolveContext(compilation.MainAssembly);
 		}
 
+		class AssemblyNameReferenceComparer : IEqualityComparer<AssemblyNameReference>
```
dgrunwald: We should add the `KeyComparer` class (back) to ILSpy, then this class could be replaced with:
`KeyComparer.Create((AssemblyNameReference r) => r.FullName)`

ICSharpCode.Decompiler/TypeSystem/DecompilerTypeSystem.cs
```diff
@@ -45,15 +45,48 @@ public DecompilerTypeSystem(ModuleDefinition moduleDefinition)
...
+				if (asm != null) {
 					referencedAssemblies.Add(cecilLoader.LoadAssembly(asm));
+					foreach (var forwarder in asm.MainModule.ExportedTypes) {
+						if (!forwarder.IsForwarder || !(forwarder.Scope is AssemblyNameReference forwarderRef)) continue;
+						assemblyReferenceQueue.Enqueue(forwarderRef);
+					}
+				}
 			}
 			compilation = new SimpleCompilation(mainAssembly, referencedAssemblies);
+			if (compilation.FindType(KnownTypeCode.Void).Kind == TypeKind.Unknown || compilation.FindType(KnownTypeCode.Void).Kind == TypeKind.Unknown) {
```
dgrunwald: You should add a code comment like this: // The primitive types are necessary to avoid assertions in ILReader.

dgrunwald: And in general this code needs some comments about why we have to add additional assemblies to the compilation; that it's mostly about PCL/.NET core, etc.

ICSharpCode.Decompiler/DecompilerEventSource.cs
```diff
@@ -0,0 +1,26 @@
...
+using ICSharpCode.Decompiler.TypeSystem;
+using ICSharpCode.Decompiler.TypeSystem.Implementation;
+using Mono.Cecil;
+
+namespace ICSharpCode.Decompiler
+{
+	[EventSource(Name = "DecompilerEventSource")]
+	public sealed class DecompilerEventSource : EventSource
+	{
+		public static DecompilerEventSource Log = new DecompilerEventSource();
```
dgrunwald: Does the log have to go through global state?
What if I'm a user of the ICS.Decompiler library and want decompile two assemblies in parallel while keeping the logs separate?

ICSharpCode.Decompiler/DotNetCore/DotNetCorePathFinder.cs
```diff
@@ -112,5 +114,24 @@ static string FallbackToDotNetSharedDirectory(AssemblyNameReference name, string
 			}
 			return null;
 		}
+
+		static string FindDotNetExeDirectory()
+		{
+			char separator;
```
dgrunwald: This would be better using: `char separator = System.IO.Path.PathSeparator;`

ICSharpCode.Decompiler/DotNetCore/DotNetCorePathFinder.cs
```diff
@@ -112,5 +114,24 @@ static string FallbackToDotNetSharedDirectory(AssemblyNameReference name, string
...
+			string dotnetExeName;
+			if (Environment.OSVersion.Platform == PlatformID.Unix) {
+				separator = ':';
+				dotnetExeName = "dotnet";
+			} else {
+				separator = ';';
+				dotnetExeName = "dotnet.exe";
+			}
+			foreach (var item in Environment.GetEnvironmentVariable("PATH").Split(separator)) {
+				var path = Environment.ExpandEnvironmentVariables(item);
```
dgrunwald: Why are you calling `ExpandEnvironmentVariables`? I don't think PATH is allowed to contain references to further envvars.
On the other hand, there's no guarantee that all elements in PATH are valid paths -- programs are supposed to ignore invalid elements. Beware of `Path.Combine()` throwing ArgumentException when PATH contains malformed elements!

PR #850, opened 2017-09-12
[WIP] Appveyor build config

siegfriedpammer:


appveyor.yml
```diff
@@ -1,18 +1,42 @@
...
+      $env:ilspy_version_number=$newVersion
+      $env:appveyor_build_version="$newVersion$branch"
+      appveyor UpdateBuild -Version "$newVersion$branch"
+    }
+    gitVersion
 before_build:
 - nuget restore ILSpy.sln
 build_script:
-- msbuild ILSpy.sln /v:minimal /p:ResolveNuGetPackages=false "/logger:%ProgramFiles%\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
+- msbuild ILSpy.sln /v:minimal /p:ResolveNuGetPackages=false;Configuration=Debug "/logger:%ProgramFiles%\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
```
sharwell: :bulb: It would be better to use a matrix for the configuration property.

You can build Debug and Release [like this](https://github.com/Microsoft/perfview/blob/3bc1978007655540240edbfe00a73d558cecbcc3/appveyor.yml#L8-L10), and use `%configuration%` where you currently have the value hard-coded in the build script.

appveyor.yml
```diff
@@ -1,18 +1,42 @@
-version: 1.0.{build}
-configuration: Debug
 platform: Any CPU
 image: Visual Studio 2017
 install:
 - git submodule update --init --recursive
-# force update ILSpy-tests to the latest ref
-- git submodule update --remote ILSpy-tests
+- ps: |
+    function gitVersion(){
```
sharwell: :bulb: It would be good to move this to a separate file, e.g. **BuildTools/appveyor-install.ps1**. It makes the script much easier to edit.

appveyor.yml
```diff
@@ -1,40 +1,21 @@
...
-      appveyor UpdateBuild -Version "$newVersion$branch"
-    }
-    gitVersion
+- ps: .\BuildTools\appveyor-install.ps1
 before_build:
 - nuget restore ILSpy.sln
 build_script:
-- msbuild ILSpy.sln /v:minimal /p:ResolveNuGetPackages=false;Configuration=Debug "/logger:%ProgramFiles%\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
-- msbuild ILSpy.sln /v:minimal /p:ResolveNuGetPackages=false;Configuration=Release "/logger:%ProgramFiles%\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
+- msbuild ILSpy.sln /v:minimal /p:ResolveNuGetPackages=false "/logger:%ProgramFiles%\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
```
sharwell: :bulb: I think you still need `;Configuration=%configuration%`

siegfriedpammer: ah, yes...

dgrunwald: @sharwell: msbuild imports all environment variables as properties.

This is also the reason why our `debugbuild.bat` file has to explicitly specify "/p:Platform=Any CPU" -- on some systems, a PLATFORM envvar unrelated to msbuild is otherwise causing problems.

PR #870, opened 2017-09-23
Don't inline readonly fields called by ref

### LINKED ISSUE #868: Static write protected fields can't be used as ref
```
using System;

class Class2
{
    public static void Test()
    {
        Guid empty = Guid.Empty;
        Y(ref empty);
    }

    public static void Y(ref Guid gClass)
    {

    }
}
```
What ILSpy does:
```
// Class2
using System;

internal class Class2
{
	public static void Test()
	{
		Class2.Y(ref Guid.Empty);
	}

	public static void Y(ref Guid gClass)
	{
	}
}

```

mohe2015:
Closes #868

### DIFF
--- a/ICSharpCode.Decompiler/IL/Transforms/ILInlining.cs
+++ b/ICSharpCode.Decompiler/IL/Transforms/ILInlining.cs
@@ -161,6 +161,21 @@ namespace ICSharpCode.Decompiler.IL.Transforms
 		/// </summary>
 		static bool DoInline(ILVariable v, ILInstruction inlinedExpression, ILInstruction next, bool aggressive, ILTransformContext context)
 		{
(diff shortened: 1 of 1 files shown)


dgrunwald:
ref on readonly is possible (performs an implicit copy) if the reference is the target of the method call.
Consider a call like `Guid.Empty.ToString()`.
That's what `IsGeneratedValueTypeTemporary` is checking -- this type of inlining happens *only* for read-only fields (and other cases where the C# compiler introduced implicit copies in normal calls).
I think the bug is within `IsGeneratedValueTypeTemporary` -- it's supposed to be checking that the call is to an instance method, but doesn't do so due to an early `return true;` for readonly fields.

dgrunwald:
Fixed in 46e36f7e9a4fb1c33e3a0a476f15a4d395a1e4ce.

PR #873, opened 2017-09-24
Decompile vars in async methods without "this."

### LINKED ISSUE #867: Async delegate tries to access variables with "this." although they are local
```
using System;
using System.Linq;
using System.Text;
using System.Threading.Tasks;

namespace ILSpyBugs
{
    class Class1
    {
        public static void Main1(string[] args)
        {
            List<uint> thisisatestvar = new List<uint>();
            Task test = Task.Run(async delegate
            {
                int a = await Program.Test();
                lock (args)
                {
                    if ("1".Equals("1"))
                    {
                        thisisatestvar.Add(1);
                    }
                }
            });
        }
    }
}

```

fails to decompile because it tries to use _this_.thisisatestvar.Add(1u);

mohe2015:
This code is probably TOTALLY wrong.
Closes #867
Changes: https://github.com/icsharpcode/ILSpy/pull/873/files?w=1

### DIFF
--- a/ICSharpCode.Decompiler/IL/Transforms/DelegateConstruction.cs
+++ b/ICSharpCode.Decompiler/IL/Transforms/DelegateConstruction.cs
@@ -28,7 +28,7 @@ namespace ICSharpCode.Decompiler.IL.Transforms
 	{
 		ILTransformContext context;
 		ITypeResolveContext decompilationContext;
-		
+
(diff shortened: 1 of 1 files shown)


siegfriedpammer:
A closer look at the generated code pattern revealed that instead of `ldloc this` the compiler uses `ldobj(ldloca this)` in case of async delegates. I adjusted the `ReplaceDelegateTargetVisitor` to take this into account.

PR #875, opened 2017-09-25
Don't crash if assembly resolution fails.

mohe2015:
Ignore formatting changes:
https://github.com/icsharpcode/ILSpy/compare/master...mohe2015:fix-assembly-resolution?w=1

### DIFF
--- a/ICSharpCode.Decompiler.Console/CustomAssemblyResolver.cs
+++ b/ICSharpCode.Decompiler.Console/CustomAssemblyResolver.cs
@@ -5,78 +5,83 @@ using Mono.Cecil;
 
 namespace ICSharpCode.Decompiler.Console
 {
-    class CustomAssemblyResolver : DefaultAssemblyResolver
-    {
-        DotNetCorePathFinder dotNetCorePathFinder;
-        readonly string assemblyFileName;
-        readonly Dictionary<string, UnresolvedAssemblyNameReference> loadedAssemblyReferences;
+	class CustomAssemblyResolver : DefaultAssemblyResolver
+	{
+		DotNetCorePathFinder dotNetCorePathFinder;
+		readonly string assemblyFileName;
+		readonly Dictionary<string, UnresolvedAssemblyNameReference> loadedAssemblyReferences;
 
-        public string TargetFramework { get; set; }
+		public string TargetFramework { get; set; }
 
-        public CustomAssemblyResolver(string fileName)
-        {
-            this.assemblyFileName = fileName;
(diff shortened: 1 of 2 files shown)


siegfriedpammer:
When are you seeing those crashes? I don't think that just writing them to the console/ignoring them is a good idea.

siegfriedpammer:
Side note:

    python BuildTools\tidy.py
    .\ICSharpCode.Decompiler\CSharp\WholeProjectDecompiler.cs:302: Line starting with spaces. Use tabs for indentation instead!
    .\ICSharpCode.Decompiler\CSharp\WholeProjectDecompiler.cs:303: Line starting with spaces. Use tabs for indentation instead!
    .\ICSharpCode.Decompiler\CSharp\WholeProjectDecompiler.cs:315: Line starting with spaces. Use tabs for indentation instead!
    Tidy check: failed
    Command exited with code 1`

In order to prevent such errors in future, you can add https://gist.github.com/siegfriedpammer/75700ea61609eb22714d21885e4eb084 in a file named `pre-commit` to `\.git\hooks\`. This will execute the `tidy.py` check before committing. (Needs Python installed...)

mohe2015:
If I run `dotnet ilspycmd.dll target.exe -p -o .` I see the AssemblyResolutionException with a specific file I can't share.
The error is:
```
Mono.Cecil.AssemblyResolutionException: Failed to resolve assembly: 'Windows, Ve
rsion=255.255.255.255, Culture=neutral, PublicKeyToken=null'
   at Mono.Cecil.BaseAssemblyResolver.Resolve(AssemblyNameReference name, Reader
Parameters parameters)
   at ICSharpCode.Decompiler.Console.CustomAssemblyResolver.Resolve(AssemblyName
Reference name, ReaderParameters parameters) in C:\Users\***\Documents\ILSp
y\ICSharpCode.Decompiler.Console\CustomAssemblyResolver.cs:line 79
```
I also wondered but it decompiled fine without that and the GUI also doesn't crash.


The bug with the duplicate appending of the targetDirectory should be clear: It happened to me if -o is anything except .

siegfriedpammer:
Ah, you are talking about `ICSharpCode.Decompiler.Console`! My bad, I assumed this was in ILSpy...

About your second commit: Did you test whether it still works when `WholeProjectDecompiler` is used in ILSpy? If this is only a problem with the console version, I would suggest adding a customized implementation of `WholeProjectDecompiler` there.

mohe2015:
I checked both the gui and the console and it worked. But it would be good if somebody else could verify that.

dgrunwald:
Thanks!

On NuGet: the console version was originally just for testing the NuGet release. I'll change it to a normal project reference now that it lives in the main repository.

PR #876, opened 2017-09-25
Change dependencies of ILSpy.Decompiler.Console to the local ones and add it to the project solution

mohe2015:
@dgrunwald (If you didn't commit it yet)

### DIFF
--- a/ICSharpCode.Decompiler.Console/ICSharpCode.Decompiler.Console.csproj
+++ b/ICSharpCode.Decompiler.Console/ICSharpCode.Decompiler.Console.csproj
@@ -8,8 +8,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="ICSharpCode.Decompiler" Version="3.0.0-alpha3" />
     <PackageReference Include="McMaster.Extensions.CommandLineUtils" Version="2.0.0" />
+  </ItemGroup>
+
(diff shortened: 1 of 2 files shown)


dgrunwald:
I tried the exact same thing you did; but as you can see, for some reason it's breaking the build...
I'm not familiar with the .NET core stuff, so no idea what's wrong here.

mohe2015:
I'm also not familiar with that but I'll try to find that out tomorrow. It's 11pm in Germany and I have school tomorrow...

If somebody knows that stuff better, please let us know.

lextm:
There is some .NET Core tooling issues preventing projects with multiple target frameworks (`ICSharpCode.Decompiler.csproj` targets both `net461 and `netstandard2.0`) from being properly consumed via project referencing. 

I think you should leave package referencing untouched, which is the most reliable approach right now.

siegfriedpammer:
> There is some .NET Core tooling issues 
Oh yes, there is, but I think, we'll find a way to get it working. I just could not take a look at it myself. ILSpy itself uses a direct reference to ICSharpCode.Decompiler without using the package, otherwise I would have given up working on this project out of frustration.

siegfriedpammer:
Fixed in b7f86d11b73c5172ed28aa7e4f59dde92f1091ae

PR #879, opened 2017-09-26
Implement mass decompilation in ICSharpCode.Decompiler.Console

mohe2015:
No whitespace changes: https://github.com/icsharpcode/ILSpy/pull/879/files?w=1

### DIFF
--- a/ICSharpCode.Decompiler.Console/Program.cs
+++ b/ICSharpCode.Decompiler.Console/Program.cs
@@ -13,161 +13,237 @@ using Mono.Cecil;
 namespace ICSharpCode.Decompiler.Console
 {
 	class Program
-    {
-        static int Main(string[] args)
-        {
-            // https://github.com/natemcmaster/CommandLineUtils/
-            // Older cmd line clients (for options reference): https://github.com/aerror2/ILSpy-For-MacOSX and https://github.com/andreif/ILSpyMono
-            var app = new CommandLineApplication();
+	{
+		static int Main(string[] args)
+		{
+			// https://github.com/natemcmaster/CommandLineUtils/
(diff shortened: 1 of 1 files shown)


siegfriedpammer:
Thank you for your continued interest in contributing. One note: Can you please pay more attention to keeping the whitespace intact? We'd prefer reading diffs without the ?w=1 option. Please check your files before committing and revert any whitespace-only changes.

siegfriedpammer:
One useful thing would be using the pre-commit hook as described in the README.md.

mohe2015:
I thought I added it last time. Maybe Visual Studio's git integration does not work well.  I will fix that tomorrow.

dgrunwald:
I think in this case it's the original indentation that was wrong.
The tidy.py script isn't checking the ICSharpCode.Decompiler.Console directory, and the console app was originally created in a different repo without our .editorconfig in place.

mohe2015:
Merge or change something?

dgrunwald:
I'm not so sure we actually want this as a feature. It very much seems like it's something that could be left to `find | xargs` (though the separate output directories might be tricky in a shell script).

mohe2015:
@dgrunwald For Windows it could be convenient, couldn't it?

mohe2015:
Found a way how to do it on windows. Execute the following script in the git shell using `bash script.sh`:
```
cd "binariesdir"
find . -type f \( -name "*.exe" -or -name "*.dll" \) -print0 | while IFS= read -r -d $'\0' line; do
    echo "$line"
    target="outputdir/${line::-4}"
    mkdir -p "$target"
    dotnet "ilspydir/ICSharpCode.Decompiler.Console/bin/Release/netcoreapp2.0/ilspycmd.dll" "$line" -p -o "$target"
done
```

PR #881, opened 2017-09-28
Fix catch-when

mohe2015:
Any feedback is welcome.

### DIFF
--- a/ICSharpCode.Decompiler/IL/Transforms/ExpressionTransforms.cs
+++ b/ICSharpCode.Decompiler/IL/Transforms/ExpressionTransforms.cs
@@ -403,6 +403,9 @@ namespace ICSharpCode.Decompiler.IL.Transforms
 				return;
 			if (!exceptionSlotLoad.MatchLdLoc(handler.Variable) || !handler.Variable.IsSingleDefinition || handler.Variable.LoadCount != 1)
 				return;
+			if (handler.Variable != null) { // Hack for the strange case that the variable gets reassigned to another one.
+				exceptionVar.Type = handler.Variable.Type;
+			}
 			handler.Variable = exceptionVar;
 			exceptionVar.Kind = VariableKind.Exception;
 			entryPoint.Instructions.RemoveAt(0);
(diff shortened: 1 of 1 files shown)


siegfriedpammer:
Ad `catch-when`: can you add a test-case for the bug you discovered?

dgrunwald:
`catch-when`: just ignoring everything except for the `leave` instruction seems wrong.
If you post the ILAst that was left over after the transforms, we could give you a hint of how to transform that further.
In general, to properly decompile catch-when, we need to be able to collapse the IL code back into a single C# expression.

`git`: You'll need to use a separate git branch for each pull request. If you have some unwanted commits on the branch, you can use interactive rebase to clean those up: `git rebase -i upstream/master`

mohe2015:
Better? - Squash merge?

siegfriedpammer:
Can you please post a small test case that reproduces the error you are trying to fix?

mohe2015:
I don't have the source code. I can only provide the ILAst:
```
} catch ex : System.Exception when (BlockContainer {
	Block IL_0000 (incoming: 1)  {
		stloc obj2(ldloc ex)
		leave IL_0000 (comp.unsigned(logic.or(comp(isinst System.Net.Http.HttpRequestException(ldloc obj2) != ldnull), comp(isinst System.IO.IOException(ldloc obj2) != ldnull)) != ldc.i4 0))
	}

}) BlockContainer {
	Block IL_0000 (incoming: 1)  {
		** censored callvirt **
		leave IL_0000 (nop)
	}

}
```

siegfriedpammer:
Hmmmm... still, could you post the binary, or the IL of the method?

mohe2015:
@siegfriedpammer I'm not allowed to, but I ran the pull request over a few files and it worked correctly (I'm using the mass-decompile for that).

siegfriedpammer:
What's the type of the variable `ex`? Could we continue this review on gitter?

siegfriedpammer:
`} catch (Exception ex) when (ex is ArgumentException || ex is IOException) {`

that is what I came up with, but it is detected correctly without your fix. Was the assembly you are trying to decompile compiled using a C# compiler?

mohe2015:
@siegfriedpammer Switch to gitter.

PR #892, opened 2017-10-01
Implement ldmembertoken

mohe2015:
I'm sorry but I can't provide a binary and I'm also not able to produce a sample....


dgrunwald:
This only works if there's no overloading and no generics involved; otherwise it's more tricky to write reflection code that picks the correct overload.

mohe2015:
@dgrunwald You're probably right. But for the target assembly it worked and it produces at least some code instead of no code. If it could produce logically WRONG code, you may give me a hint how to fix that.

dgrunwald:
A decompiled call to [Type.GetMethod(string)](https://msdn.microsoft.com/en-US/library/8zz808e6.aspx) will throw an `AmbiguousMatchException` on overloaded methods, even though the original code wasn't doing any such thing. It also searches only for public methods, while the token has no such restriction.

And generic methods are their own can of worms. You'll have to emit `MakeGenericMethod` calls there; but I'm not sure how you would even pick the correct overload among the generic methods -- the Reflection API doesn't seem to take enough information to uniquely identify an overload.

mohe2015:
What about a try catch (AmbiguousMatchException) and a check if GetMethod returns null and if that's the case add a comment that something went wrong?

dgrunwald:
A try catch doesn't help. Unless we handle `ldtoken` correctly, either we produce code that doesn't compile to alert the developer at compile-time; or we'll produce code with wrong runtime behavior.

The only place where the C# compiler emits `ldtoken` for methods is when creating a delegate from a method group. But in that case we decompile the whole `newobj` back into the appropriate C# construct, no there's no need to handle the individual `ldtoken` instruction. I guess if you want to create testcases for this, you'll have to write them directly in IL.

mohe2015:
@dgrunwald But the current code works for my binary. So a try catch would be good if an unknown case occurs or otherwise the compiler would fail for the whole class wouldn't it?

It shouldn't produce code with wrong behavior if we do it that way it would just add a comment that something went wrong and fail do compile.

mmusu3:
I have found that generated Linq Expression trees use ldtoken instructions. eg:
```cs
using System.Linq.Expressions;

class MemberTokens
{
    static bool Value;

    static void Test()
    {
        Expression<Func<bool>> expr1 = () => Value;
        Expression<Func<Action>> expr2 = () => Test;
    }
}
```
Output:
```cs
using System;
using System.Linq.Expressions;
using System.Reflection;

internal class MemberTokens
{
	private static bool Value;

	private static void Test()
	{
		Expression.Lambda<Func<bool>>((Expression)Expression.Field(null, FieldInfo.GetFieldFromHandle((RuntimeFieldHandle)/*OpCode not supported: LdMemberToken*/)), Array.Empty<ParameterExpression>());
		Expression.Lambda<Func<Action>>((Expression)Expression.Convert(Expression.Call(Expression.Constant((MethodInfo)MethodBase.GetMethodFromHandle((RuntimeMethodHandle)/*OpCode not supported: LdMemberToken*/), typeof(MethodInfo)), (MethodInfo)MethodBase.GetMethodFromHandle((RuntimeMethodHandle)/*OpCode not supported: LdMemberToken*/), new Expression[2]
		{
			Expression.Constant(typeof(Action), typeof(Type)),
			Expression.Constant(null, typeof(object))
		}), typeof(Action)), Array.Empty<ParameterExpression>());
	}
}
```

siegfriedpammer:
Please see https://github.com/icsharpcode/ILSpy/blob/master/ICSharpCode.Decompiler/CSharp/Transforms/ExpressionTreeConverter.cs. This is a transform that needs to be adapted to the new decompiler structure, as it is currently disabled.

siegfriedpammer:
I will be working on this...

dgrunwald:
While we can fix expression trees; I doubt the general case of `ldtoken` can be solved correctly.
The Reflection API sometimes makes it extremely hard to get the correct MethodInfo object, see https://stackoverflow.com/questions/3631547/select-right-generic-method-with-reflection

mohe2015:
The code in my assembly also used expression trees so we should fix them first.

siegfriedpammer:
Actually, I think the expression tree converter depends on the underlying structure... we'd need a way to represent raw ldtoken instructions in C# first.

dgrunwald:
I don't think the expression tree converter cares about how to represent ldtoken in C#, just that there's some expression with the correct annotation. I think the ErrorExpression we're currently using could already be sufficient with its ILInstruction annotation.

dgrunwald:
I think a possible way to translate `ldtoken` for now would be: `new DelegateType(Method).MethodInfo.MethodHandle`
where DelegateType is a delegate that matches the method's signature (if there's no ref/out involved, `Action<>`/`Func<>` will do the job; otherwise the decompiler will have to declare a new delegate).
Only problem is if it's an instance method belonging to another class, as we need an instance for creating a delegate.

There's some discussion about an `LdToken` intrinsic in a future C# version, which would also work based on delegate creation.

dgrunwald:
I'm closing this PR; let's first figure out (on #911) what kind of code we want to generate for ldtoken.

PR #903, opened 2017-10-08
Fix proxy calls to base method

### LINKED ISSUE #869: Calling base methods does not get decompiled correctly
```
using System.Net.Http;
using System.Threading;
using System.Threading.Tasks;

public class TestHandler : DelegatingHandler
{
    protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage r, CancellationToken c)
    {
        return await base.SendAsync(r, c);
    }
}

```
What ILSpy does:
```
// TestHandler
using System.Net.Http;
using System.Threading;
using System.Threading.Tasks;

public class TestHandler : DelegatingHandler
{
	protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage r, CancellationToken c)
	{
		return await this.<>n__0(r, c);
	}
}

```

mohe2015:
I would like to get some feedback.
Closes #869


dgrunwald:
1. Can you add a test case?
2. Your `ProxyMethodVisitor` doesn't properly check that the method only performs a proxy call.
e.g. it also matches `void F() { base.F(); base.F(); }` or even `void F(int i) { base.F(i + 1); }`.
3. What's up with the `foreach (IMethod method in inst.Method.DeclaringTypeDefinition.Methods)` loop? You're effectively ignoring `inst.Method` and instead picking any other overload of the method.

mohe2015:
@dgrunwald Done! Anything else?

dgrunwald:
I don't think the `ProxyMethodVisitor` is the right approach -- you're still allowing arbitrary combinations of the white-listed instructions. E.g. unreachable IL code might be seen as a valid proxy call: `void F() { return; base.F(); }`
It's better to just match the body exactly (i.e. check that it's a BlockContainer with 1 block, that the block has 1 or 2 instructions, and what those instructions are).

You should also expand the test case:
 * add a case where the method returns something other than void
 * add a case where the proxy call is not due to `await`, but due to `yield return`, or a capturing lambda

Instead of `inst.Arguments.ElementAtOrDefault(i)`, you can just use `inst.Arguments[i]`, since your loop condition ensures the index is in range (and you don't handle a return value of null).

Also, you don't need to clone the arguments. It's valid for an ILInstruction to temporarily be in multiple places (have multiple parents), as long as after the transform is done, it is in the place that it was assigned to last.
More importantly, some types of ILInstruction are currently throwing NotImplementedException if you try to clone them (e.g. if one of the arguments is a lambda expression).

About the build failure: the test framework currently doesn't support resolving .NET standard assembly references, so you need to recompile your test case. Even better, add the test as a C# pretty test (instead of IL pretty test). There the test runner will automatically compile the C# code file with both Roslyn and the old C# compiler. (but we still commit the IL files generated when running the tests, to avoid existing tests breaking when the C# compiler is updated)

mohe2015:
@dgrunwald And now?

dgrunwald:
* You're only testing proxy methods that return a non-void result.
You'll need at least one test with `void`; the pattern you're matching is slightly different in that case.

* Also your lambda test is only capturing `base`/`this`. The compiler generates significantly different code when the lambda captures a local variable or parameter, so you should also test that. (only capturing `this` results in a compiler-generated method in your class; additionally capturing local variables results in a compiler-generated class. AFAIK only the latter makes use of proxy methods)

* Your visitor does not recurse into the arguments of the call. I think this might lead to problems when a proxy call has more proxy calls in its arguments (e.g. `yield return base.M(base.M(0).First()).First();`)

Other than that the implementation looks fine, but here's some suggestions for improvements:

* Do don't need a visitor here, finding all instructions of a type can be done with: `foreach (var call in function.Descendants.OfType<CallInstruction>())`.

* The `DeclaringTypeDefinition == null` check is redundant now that you've removed the loop.
As for the why: one potential reason is that reference assemblies are missing; in that case the definition of the declaring type may be unknown. (there's also fake methods on multi-dimensional arrays; but those are converted to other instructions before your transform gets to run)

* Instead of using `OpCode`/`Children`, you can cast the instruction to a concrete type with C# 7 pattern matching, e.g.:
```
if (!(function.Body is BlockContainer blockContainer))
  return;
// then use the strongly typed blockContainer.Blocks instead of ILInstruction.Children
```

* All blocks that are directly in a `BlockContainer` are guaranteed to have a `Nop` as `FinalInstruction` (which is not a part of `Block.Instructions`, only of `Block.Children`), so you don't need to check that.

* Many node types have `Match...` methods to simplify matching against them; for example the `leave` check could be done in a single line using:
```
if (!block.Instructions[0].MatchLeave(blockContainer, out ILInstruction returnValue))
   return;
```

mohe2015:
@dgrunwald Thank you very much for all your help! I really appreciate it. I hope I didn't forget anything. One of the tests doesn't work with [roslyn, optimize] because it optimizes the name away. What should I do?

PR #909, opened 2017-10-12
Fix object initializer

mohe2015:
Before:
```
return new HttpResponseMessage
{
	Content = new StreamContent(memoryStream2),
	Content = 
	{
		Headers = 
		{
			ContentDisposition = new ContentDispositionHeaderValue("aaa"),
			ContentDisposition = 
			{
				FileName = fileName
			},
			ContentType = new MediaTypeHeaderValue("bbb"),
			ContentLength = new long?(memoryStream2.Length)
		}
	}
};
```
I had to use an IL test as I could not find code which reproduces that issue.
Feedback please...

### DIFF
--- a/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
+++ b/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
@@ -52,6 +52,7 @@
   <ItemGroup>
     <Compile Include="DataFlowTest.cs" />
     <Compile Include="TestCases\Correctness\LINQRaytracer.cs" />
+    <Compile Include="TestCases\ILPretty\ObjectInitializer.cs" />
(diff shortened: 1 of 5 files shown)


dgrunwald:
The IL code accesses `Content` twice: once with `set_Content` to assign the instance, then again with `get_Content` to get the instance for calling the add method.
Depending on how the property is implemented, this might be a different instance! So it's not correct to combine the two into a single object initializer.

We'll instead need to check the object initializer for invalid duplicate usage of an `AccessPathElement`, and stop the transform if we find any. @siegfriedpammer will try to implement this check.

siegfriedpammer:
I have created a (local) branch `initializer_pretty_fix` which will add some pretty tests for array, collection and object initializers and will fix the issue mentioned by @dgrunwald, and others. Therefore I will close this PR.

PR #910, opened 2017-10-12
Fix special cases in catch with object

mohe2015:


dgrunwald:
What are you trying to fix?
Can you add a test case?

mohe2015:
I have a binary with the IL instructions which are mentioned in the comments instead of the normal ones. I will try to create a test case - it's private code...

mohe2015:
Added a test case @dgrunwald .

dgrunwald:
Thanks!

PR #925, opened 2017-10-16
Expression trees

mohe2015:
- [x]  Fix CurriedLambda
- [x] Fix QuotedWithAnonymous
- [x] Fix MethodGroups
- [x] Fix NotOperator
- [x] Fix Newtonsoft.Json
- [x] Clean up code (try to minimize changes, try to remove hacks)
- [x] Add comments
- [x] Fix DefaultValue for nonRoslyn
- [x] Fix ListInitializer2 for nonRoslyn
- [x] Fix LiteralCharAndProperty for nonRoslyn


ICSharpCode.Decompiler/CSharp/Transforms/CustomPatterns.cs
```diff
@@ -70,9 +77,10 @@ public LdTokenPattern(string groupName)
 		
 		public override bool DoMatch(INode other, Match match)
 		{
-			InvocationExpression ie = other as InvocationExpression;
-			if (ie != null && ie.Annotation<LdTokenAnnotation>() != null && ie.Arguments.Count == 1) {
-				return childNode.DoMatch(ie.Arguments.Single(), match);
+			// new ErrorExpression("/*ldmembertoken ...*/")
+			if (other is ErrorExpression ee) {
```
dgrunwald: Instead of testing for `ErrorExpression`, you can check: `other is AstNode node && node.Annotation<LdMemberToken>() != null`

mohe2015:
Please review. Thank you.

ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
```diff
@@ -148,6 +149,7 @@ public static List<IILTransform> GetILTransforms()
 			new FlattenSwitchBlocks(),
 			new FixNameCollisions(),
 			new AddXmlDocumentationTransform(),
+			new ExpressionTreeConverter(),
```
siegfriedpammer: The `ExpressionTreeConverter` should run before `IntroduceQueryExpressions`, this is necessary because query expressions are sometimes compiled into expression trees.

expression-trees.playlist
```diff
@@ -0,0 +1 @@
+<Playlist Version="1.0"><Add Test="ICSharpCode.Decompiler.Tests.PrettyTestRunner.ExpressionTrees(Optimize, UseRoslyn)" /><Add Test="ICSharpCode.Decompiler.Tests.PrettyTestRunner.ExpressionTrees(UseRoslyn)" /><Add Test="ICSharpCode.Decompiler.Tests.RoundtripAssembly.NewtonsoftJson_net45" /></Playlist>
```
siegfriedpammer: We don't need *.playlist files checked in. Can you please remove them?

ICSharpCode.Decompiler/IL/Transforms/InlineExpressionTreeParameterDeclarations.cs
```diff
@@ -0,0 +1,66 @@
...
+	class InlineExpressionTreeParameterDeclarationsTransform : IStatementTransform
+	{
+		public void Run(Block block, int pos, StatementTransformContext context)
+		{
+			if (block.Instructions[pos] is StLoc parameterVariableInst) {
+				bool alwaysInlined = true;
+				if (MatchParameterVariableAssignment(parameterVariableInst, out ILVariable v, out ILInstruction init)) {
+					foreach (var LdParameterVariableInst in context.Function.Descendants.OfType<LdLoc>()) {
+						if (LdParameterVariableInst.MatchLdLoc(v)) {
+							ILInstruction parent = LdParameterVariableInst.Parent;
```
siegfriedpammer: You could extract the inner loop into its own (local) method and get rid of the goto. That would make it easier to read.

dgrunwald:
Diffing the decompiled NewtonsoftJson_pcl_debug with and without your PR reveals a bug:

Decompiled `ExpressionReflectionDelegateFactory.CreateParameterizedConstructor()` without your PR:
```
		public override ObjectConstructor<object> CreateParameterizedConstructor(MethodBase method)
		{
			ValidationUtils.ArgumentNotNull(method, "method");
			Type typeFromHandle = typeof(object);
			ParameterExpression parameterExpression = Expression.Parameter(typeof(object[]), "args");
			Expression body = this.BuildMethodCall(method, typeFromHandle, null, parameterExpression);
			LambdaExpression lambdaExpression = Expression.Lambda(typeof(ObjectConstructor<object>), body, parameterExpression);
			return (ObjectConstructor<object>)lambdaExpression.Compile();
		}
```

Decompiled with your PR:
```
		public override ObjectConstructor<object> CreateParameterizedConstructor(MethodBase method)
		{
			ValidationUtils.ArgumentNotNull(method, "method");
			Type typeFromHandle = typeof(object);
			ParameterExpression argsParameterExpression = Expression.Parameter(typeof(object[]), "args");
			Expression body = this.BuildMethodCall(method, typeFromHandle, null, argsParameterExpression);
			LambdaExpression lambdaExpression = Expression.Lambda(typeof(ObjectConstructor<object>), body, Expression.Parameter(typeof(object[]), "args"));
			return (ObjectConstructor<object>)lambdaExpression.Compile();
		}
```

The `Expression.Parameter` call got duplicated. Interestingly this only happens in the debug build of Newtonsoft.JSon, not in the release build (the difference between the both builds being whether the result of the `Lambda` invocation gets stored in a local or on the stack).

ICSharpCode.Decompiler/IL/Transforms/InlineExpressionTreeParameterDeclarations.cs
```diff
@@ -0,0 +1,66 @@
...
+{
+	class InlineExpressionTreeParameterDeclarationsTransform : IStatementTransform
+	{
+		public void Run(Block block, int pos, StatementTransformContext context)
+		{
+			if (block.Instructions[pos] is StLoc parameterVariableInst) {
+				bool alwaysInlined = true;
+				if (MatchParameterVariableAssignment(parameterVariableInst, out ILVariable v, out ILInstruction init)) {
+					foreach (var LdParameterVariableInst in context.Function.Descendants.OfType<LdLoc>()) {
+						if (LdParameterVariableInst.MatchLdLoc(v)) {
```
dgrunwald: `v.LoadInstructions` will give you all the LdLocs for a variable, so you don't need to search the whole function.

ICSharpCode.Decompiler/IL/Transforms/DelegateConstruction.cs
```diff
@@ -239,7 +239,40 @@ protected override void Default(ILInstruction inst)
...
+					initValues.TryGetValue(ldMemberToken.Member as IField, out DisplayClassVariable info)) {
+
+					Call replace = new Call(constantCall.Method);
+					replace.Arguments.Add(new LdLoc(info.variable));
+					Call typeOfInstruction = new Call(getTypeFromHandleCall.Method);
+					typeOfInstruction.Arguments.Add(new LdTypeToken(info.variable.Type));
+					replace.Arguments.Add(typeOfInstruction);
+					inst.ReplaceWith(replace);
+				} else {
+					Default(inst);
```
siegfriedpammer: You can remove this and instead call `base.VisitCall(inst)` at the beginning of the method, just like all the other overrides do.

ICSharpCode.Decompiler/IL/Transforms/DelegateConstruction.cs
```diff
@@ -239,7 +239,40 @@ protected override void Default(ILInstruction inst)
...
 				}
 			}
-			
+
+			// Handle variables in ExpressionTrees
+			// call Field(call Constant(ldloc V_0, call GetTypeFromHandle(ldtypetoken ExpressionTrees+<>c__DisplayClass56_0)), call GetFieldFromHandle(ldmembertoken i)) at IL_0114
+			// -> call Constant(ldloc i, call GetTypeFromHandle(ldtypetoken System.Int32))
+			protected internal override void VisitCall(Call inst)
+			{
+				if (inst.Method.FullName.Equals("System.Linq.Expressions.Expression.Field") &&
```
siegfriedpammer: Putting the logical and operator at the beginning of the line would make it easier to read the code.

ICSharpCode.Decompiler/IL/Transforms/InlineExpressionTreeParameterDeclarations.cs
```diff
@@ -0,0 +1,66 @@
...
+							ILInstruction parent = LdParameterVariableInst.Parent;
+							while (parent != null) {
+								if (parent is CallInstruction call &&
+									call.Method.FullName.Equals("System.Linq.Expressions.Expression.Lambda") &&
+									!((call.Parent as CallInstruction)?.Method.FullName.Equals("System.Linq.Expressions.LambdaExpression.Compile") ?? false) &&
+									!((call.Parent as CallInstruction)?.Method.FullName.Equals("System.Linq.Expressions.Expression.Compile") ?? false)) {
```
dgrunwald: `!(x?.Method.FullName.Equals("...") ?? false))` seems to be a very convoluted way of saying `x?.Method.FullName != "..."`

ICSharpCode.Decompiler/IL/Transforms/InlineExpressionTreeParameterDeclarations.cs
```diff
@@ -0,0 +1,66 @@
...
+				}
+			}
+		}
+
+		bool MatchParameterVariableAssignment(ILInstruction expr, out ILVariable v, out ILInstruction init)
+		{
+			// stloc(v, call(Expression::Parameter, call(Type::GetTypeFromHandle, ldtoken(...)), ldstr(...)))
+			if (!expr.MatchStLoc(out v, out init))
+				return false;
+			if (/*v.HasGeneratedName || */v.Kind == VariableKind.Parameter /*|| v.IsPinned*/) // TODO
```
siegfriedpammer: I think `v.Kind != VariableKind.Local && v.Kind != VariableKind.StackSlot` should suffice here.

ICSharpCode.Decompiler/IL/Transforms/InlineExpressionTreeParameterDeclarations.cs
```diff
@@ -0,0 +1,66 @@
...
+			if (block.Instructions[pos] is StLoc parameterVariableInst) {
+				bool alwaysInlined = true;
+				if (MatchParameterVariableAssignment(parameterVariableInst, out ILVariable v, out ILInstruction init)) {
+					foreach (var LdParameterVariableInst in context.Function.Descendants.OfType<LdLoc>()) {
+						if (LdParameterVariableInst.MatchLdLoc(v)) {
+							ILInstruction parent = LdParameterVariableInst.Parent;
+							while (parent != null) {
+								if (parent is CallInstruction call &&
+									call.Method.FullName.Equals("System.Linq.Expressions.Expression.Lambda") &&
+									!((call.Parent as CallInstruction)?.Method.FullName.Equals("System.Linq.Expressions.LambdaExpression.Compile") ?? false) &&
```
siegfriedpammer: I think it would be better to use `call.Parent is CallInstruction parentCall` instead of using the elvis operator here.

siegfriedpammer:
Why are there no tests for the pre-Roslyn C# compiler? Is there some problem?

mohe2015:
@siegfriedpammer @dgrunwald Thank you for the review!
@siegfriedpammer yes, but I can't remember which. I will check tomorrow or so.

ICSharpCode.Decompiler/IL/Transforms/InlineExpressionTreeParameterDeclarations.cs
mohe2015: Concurrent modification of the list - does not really work...

mohe2015:
@siegfriedpammer The error is:
`Message: System.Exception : Compiler error:c:\Users\xxx\Documents\ILSpy\ICSharpCode.Decompiler.Tests\TestCases\Pretty\ExpressionTrees.cs(273,3) : error CS0012: The type 'System.Collections.Generic.ISet`1<T0>' is defined in an assembly that is not referenced. You must add a reference to assembly 'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.`

ICSharpCode.Decompiler/IL/Transforms/InlineExpressionTreeParameterDeclarations.cs
siegfriedpammer: Feel free to use `ToArray()` to create a temporary copy.

mohe2015:
Sooner or later the Pretty test will fail at the following line with the old decompiler:
```
 379      (-) 		ExpressionTrees.ToCode<bool>(ExpressionTrees.X(), () => new XmlReaderSettings {
      313 (+) 		ExpressionTrees.ToCode<bool>(ExpressionTrees.X(), () => new ExpressionTrees.XmlReaderSettings {
```

ICSharpCode.Decompiler/CSharp/Transforms/ExpressionTreeConverter.cs
```diff
@@ -20,51 +20,63 @@
...
+		TypeSystemAstBuilder astBuilder;
+
+		public ExpressionTreeConverter()
+		{
+		}
+
+		public void Run(AstNode rootNode, TransformContext context)
+		{
+			CSharpResolver resolver = new CSharpResolver(new CSharpTypeResolveContext(context.TypeSystem.Compilation.MainAssembly, null, context.DecompiledTypeDefinition, context.DecompiledMember));
+			astBuilder = new TypeSystemAstBuilder(resolver);
```
siegfriedpammer: You are using the same `TypeSystemAstBuilder` in all cases. This is the cause for bugs like this:
```
 379      (-) 		ExpressionTrees.ToCode<bool>(ExpressionTrees.X(), () => new XmlReaderSettings {
      313 (+) 		ExpressionTrees.ToCode<bool>(ExpressionTrees.X(), () => new ExpressionTrees.XmlReaderSettings {
```
1) You are not adding any `UsingScope` to the `CSharpTypeResolveContext`, thus the TSAB will always generate full names.
2) `context.DecompiledTypeDefinition` and `context.DecompiledMember` are not what you think they are. When decompiling a full assembly, they both return null. See https://github.com/icsharpcode/ILSpy/blob/master/ICSharpCode.Decompiler/CSharp/Transforms/TransformContext.cs#L40

In order to properly handle this you will have to do something like this: https://gist.github.com/siegfriedpammer/1d3fe34ec2d7101f63dc1b813d1efafd

mohe2015:
@dgrunwald I know it's not a good approach and you probably know it better but is it really a bug to duplicate the Expression.Parameter call? Based on the [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.linq.expressions.parameterexpression?view=netframework-4.7.1) I think it doesn't matter.

dgrunwald:
`ParameterExpression` uses reference identity, so duplicating constructor calls ends up changing the semantics.

Original Code
```
		{
			var twice = GetExpression(Expression.Constant(2)).Compile();
			Console.WriteLine(twice(21));
		}

		static Expression<Func<int, int>> GetExpression(Expression factor)
		{
			ParameterExpression parameterExpression = Expression.Parameter(typeof(int), "x");
			return Expression.Lambda<Func<int, int>>(Expression.Multiply(parameterExpression, factor), parameterExpression);
		}
```

Decompiled
```
private static Expression<Func<int, int>> GetExpression(Expression factor)
{
	return Expression.Lambda<Func<int, int>>((Expression)Expression.Multiply(Expression.Parameter(typeof(int), "x"), factor), new ParameterExpression[1]
	{
		Expression.Parameter(typeof(int), "x")
	});
}
```

Original Output: `42`
Output of decompiled code:
```
Unhandled Exception: System.InvalidOperationException: variable 'x' of type 'System.Int32' referenced from scope '', but it is not defined
   at System.Linq.Expressions.Compiler.VariableBinder.Reference(ParameterExpression node, VariableStorageKind storage)
```

mohe2015:
Thank you very much for all your help and time! I really appreciate it. Currently I don't have the motivation (and time) to fix that. Maybe I will come back to that some time later. If somebody wants to he/she can try to fix it.

siegfriedpammer:
@mohe2015 @dgrunwald I have moved the code to a branch in our repository, see https://github.com/icsharpcode/ILSpy/tree/expression-trees

Issue #931, opened 2017-10-23
RFC: ICS.Decompiler.Extensions NuGet Package


christophwille:
Take code from ICS.D.Console and repackage it in a mildly abstracted way into ICS.D.Extensions - think of it "Avoiding the next CSharpLanguage.cs copy/paste problem". Intention is to provide a high-level API for console, ps and eg vscode extension.

I would vote for leaking the implementation details - eg Cecil types. It should be up to the implementer to abstract DTOs and the like if needed. 

Implementation: csproj with multiple TFMs - netcore2, netstandard2, net461. That way we can #ifdef rich client implementations (unless of course netstandard2 would do the trick, which is preferred)

christophwille:
See #945

PR #940, opened 2017-10-24
RFC: Use conditions for ICS.D.Console project

christophwille:
Debug: project references
Release: Nuget

Caveat: Switchting from Debug -> Release via dropdown you then have to close and reopen the solution

Note: this is a RFC also meant for building out #931 Nuget


christophwille:
.Console.sln is now Frontends.sln. The frontends are always built against the NuGet packages.

Issue #986, opened 2017-11-24
Support for NuGet (*.nupkg) files


BADF00D:
Recently I dropped a nupkg File on ILSpy and noticed, that ILSpy was not able to read the file. So I forked the repository and made a bloody hack to support this. Basically I unpack the zip file to a temp directory, search for the first *.dll that comes along and pushed the FileStream to the ImageReader.ReadImage method. And it works:

```csharp
static ModuleDefinition ReadModule (Disposable<Stream> stream, string fileName, ReaderParameters parameters)
{
	Mixin.CheckParameters (parameters);

	if (fileName.EndsWith (".nupkg")) {
		var tmpFolder = Path.Combine(Path.GetTempPath (), Path.GetRandomFileName());
		Directory.CreateDirectory (tmpFolder);
		ZipFile.ExtractToDirectory (fileName, tmpFolder);
		var dlls= Directory.GetFiles (tmpFolder, "*.dll", SearchOption.AllDirectories);
		var fileStream = File.OpenRead (dlls.First ());
		stream = Disposable.NotOwned (fileStream as Stream);
	}

	return ModuleReader.CreateModule (
		ImageReader.ReadImage (stream, fileName),
		parameters);
}
```

This hack has a lot of drawbacks:
* The dlls is in a temp folder and will be deleted during next cleanup. But the nupkg file might be still at the same place
* Everything is a stream, so it should be possible to reach the same without unzipping the nupkg file at all.
* No tests
* There should be a better option to select a dll. Maybe a dialog, maybe load all dlls, maybe load only the latest, I don't know.

Before I eliminate these drawbacks I want to ask: Is the support for nupkg something you want to see in ILSpy?

siegfriedpammer:
See https://twitter.com/ilspy/status/911285196854632449
Look for `LoadedNugetPackage` in the code. I guess the drag&drop support is missing, you could provide a PR.

siegfriedpammer:
The PR was merged.

PR #987, opened 2017-11-24
Added support for dropping nupkg files.

BADF00D:
I just added support for dropping nupkg as suppested by @siegfriedpammer in issue #986 .

Feedback is welcome.


siegfriedpammer:
Thank you for adding this feature!

PR #1001, opened 2017-12-01
Reference Mono.Cecil using a PackageReference

sharwell:
The current submodule and bundling approach causes problems for consumers of **ICSharpCode.Decompiler** due to the embedding of a "non-standard" build of **Mono.Cecil.dll**. This pull request modifies the references to use `PackageReference` instead, and declare a NuGet dependency on the Mono.Cecil package on NuGet.

For this pull request, I did not remove the submodule itself. It's possible that developers have local branches in the submodule and would like to continue working with those branches, so I decided to leave that decision for a future pull request.

This pull request is very similar to #998. Specific differences are:

1. This pull request leaves the **cecil** submodule in place, and updates the commit reference to match the commit we're referencing as a NuGet package
2. This pull request leverages transitive `<PackageReference>` in a few places, so there are fewer explicit references to Mono.Cecil
3. This pull request updates the NuGet package specification for ICSharpCode.Decompiler to declare the dependency on Mono.Cecil

### DIFF
--- a/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
+++ b/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
@@ -39,7 +39,6 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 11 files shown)


siegfriedpammer:
@sharwell did you see #998?

sharwell:
@siegfriedpammer Yes, the differences as compared to #998 are listed above. This pull request is a slightly different approach to the same goal.

PR #1002, opened 2017-12-01
Update the license notices following the move to LightJson

sharwell:
There are three key changes here:

1. Removed the mention of Json.NET
2. Added **LICENSE** and **THIRD-PARTY-NOTICES** file, with a form following the current guidance from the .NET Foundation
3. Added attribution notices to **THIRD-PARTY-NOTICES** for StyleCop Analyzers and LightJson

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2011-2017 AlphaSierraPapa for the SharpDevelop team
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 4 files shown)


LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2011-2017 AlphaSierraPapa for the SharpDevelop team
```
sharwell: :bulb: The .NET Foundation templates do not include a date here. Removing the date makes some things easier, but other things more complicated. I would recommend evaluating two options to determine if they offer the protections you desire:

1. Removing the date
2. Using a single date instead of a date range (specifically, the date at which the project was started)

sharwell:
There are two things that leave me uneasy about this:

1. The notice was a start, but didn't include all of the existing required notices.
2. The .NET Foundation template was written with an assumption: LICENSE covers the source code in the current repository, and THIRD-PARTY-NOTICES covers source code included by copy in the current repository. ILSpy distributes multiple components from the current repository, and the license used by the components varies.

If you have thoughts on how to modify this pull request to satisfy your concerns in these areas then it can be reopened.

PR #1012, opened 2017-12-16
C# 6 String interpolation

siegfriedpammer:


ICSharpCode.Decompiler/CSharp/Transforms/ReplaceMethodCallsWithOperators.cs
```diff
@@ -106,8 +108,67 @@ void ProcessInvocationExpression(InvocationExpression invocationExpression)
 						invocationExpression.ReplaceWith(new ObjectCreateExpression(context.TypeSystemAstBuilder.ConvertType(method.TypeArguments.First())));
 					}
 					break;
+				case "System.String.Format":
+					if (context.Settings.StringInterpolation && arguments.Length > 1
+						&& arguments[0] is PrimitiveExpression stringExpression && stringExpression.Value is string
+						&& arguments.Skip(1).All(a => !a.DescendantsAndSelf.OfType<PrimitiveExpression>().Any(p => p.Value is string)))
```
dgrunwald: You only check for normal string literals.
Formatted string literals are probably also forbidden?
What about char literals?
What about the `?:` operator (might be problematic due to having a `:`)?

AraHaan:
@siegfriedpammer you might want to:

```cpp
git branch backup
git checkout master
git branch -D string-interpolation
git branch string-interpolation
git checkout string-interpolation
git cherry-pick backup
```
to try to fix the merge conflicts, update to latest head of master and then you cherry pick all your commits from her copied to ``backup`` back to the remade branch and then force it onto github.
(much cleaner than ``git merge`` annoying commits, this is what I now do all the time when stuff like this happens where possible)

ICSharpCode.Decompiler/CSharp/Transforms/ReplaceMethodCallsWithOperators.cs
siegfriedpammer: I just tested with char literals and interpolated strings... seems fine to use them.
As for conditional expressions: seems they're only valid when wrapped in parentheses.

Issue #1023, opened 2018-01-11
C# 7.1: Async Main


siegfriedpammer:
Specification: https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.1/async-main.md
Difficulty: Easy
Time needed: 2-3 hrs

1. Add [unit tests](https://github.com/icsharpcode/ILSpy/wiki/Unit-Tests): In this case only pretty tests make sense.
2. Extend the code in `CSharpDecompiler.MemberIsHidden` to hide the generated `<Main>` method in the decompiled code, if `settings.AsyncAwait` is set.

MikeFH:
Is it OK if I take a look at this ?

siegfriedpammer:
Yes, you are very welcome to submit a pull request for this feature. Please base your pull request on the master branch and follow the guidelines for unit tests.

If you have any questions, feel free to ask!
Thank you for your interest in contributing to this project!

PR #1029, opened 2018-01-15
Add support for async main method

MikeFH:
Adds the feature described in #1023 

I had a bit of a hard time writing the test since, until now, all tests generated DLLs and we need to generate an EXE to test that the entry point is decompiled correctly. I created a second method called `RunForLibrary()` to replace the old one and set the necessary flags to generate DLLs. This way we can use the `Run()` method to create an EXE file.

At this point, I don't see any other way to identify the generated method but by its name (see [Roslyn](http://source.roslyn.codeplex.com/#Microsoft.CodeAnalysis.CSharp/Symbols/Synthesized/SynthesizedEntryPointSymbol.cs,18)). I put the logic to identify the method in the `AsyncAwaitDecompiler` since more checks could be added in the future to distinguish it.

### DIFF
--- a/ICSharpCode.Decompiler/IL/ControlFlow/AsyncAwaitDecompiler.cs
+++ b/ICSharpCode.Decompiler/IL/ControlFlow/AsyncAwaitDecompiler.cs
@@ -47,7 +47,7 @@ namespace ICSharpCode.Decompiler.IL.ControlFlow
 
 		public static bool IsCompilerGeneratedMainMethod(MethodDefinition method)
 		{
-			return method.Name.Equals("<Main>", StringComparison.Ordinal);
+			return method == method.Module.Assembly?.EntryPoint && method.Name.Equals("<Main>", StringComparison.Ordinal);
 		}
 
 		enum AsyncMethodType
--- a/ICSharpCode.Decompiler/IL/ControlFlow/AsyncAwaitDecompiler.cs
+++ b/ICSharpCode.Decompiler/IL/ControlFlow/AsyncAwaitDecompiler.cs
@@ -20,6 +20,7 @@ using ICSharpCode.Decompiler.CSharp;
 using ICSharpCode.Decompiler.IL.Transforms;
 using ICSharpCode.Decompiler.TypeSystem;
 using ICSharpCode.Decompiler.Util;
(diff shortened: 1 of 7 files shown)


siegfriedpammer:
Please make sure to use tabs instead of spaces in new files. Thanks!
(If you have python installed, you can use https://gist.github.com/siegfriedpammer/75700ea61609eb22714d21885e4eb084 as `.git/hooks/pre-commit` to prevent this in the future.)

Output from the build log:
> python BuildTools\tidy.py
.\ICSharpCode.Decompiler.Tests\TestCases\Pretty\AsyncMain.cs:7: Line starting with spaces. Use tabs for indentation instead!
.\ICSharpCode.Decompiler.Tests\TestCases\Pretty\AsyncMain.cs:13: Line starting with spaces. Use tabs for indentation instead!
Tidy check: failed
Command exited with code 1

MikeFH:
Oops, sorry. I thought I was covered with the .editorconfig and VS. I used tidy.py this time.

siegfriedpammer:
This happens to us whenever we are adding a new file. Seems to be a bug in VS because the templates do not respect the settings in .editorconfig.

ICSharpCode.Decompiler/IL/ControlFlow/AsyncAwaitDecompiler.cs
```diff
@@ -44,6 +45,11 @@ public static bool IsCompilerGeneratedStateMachine(Mono.Cecil.TypeDefinition typ
 			return false;
 		}
 
+		public static bool IsCompilerGeneratedMainMethod(MethodDefinition method)
+		{
+			return method.Name.Equals("<Main>", StringComparison.Ordinal);
```
siegfriedpammer: You should probably check, if the method is the entrypoint as well: `method.Module.Assembly?.EntryPoint == method`

MikeFH: Got it, I added the check.

siegfriedpammer:
Thank you very much for your contribution to ILSpy!

MikeFH:
Thanks, glad I could help.

PR #1033, opened 2018-01-17
Update to net45 targetframework

Titaye:
I'm planning to include ICSharpCode.Decompiler nuget package to https://github.com/fsharp/FsAutoComplete which is targeting .net 4.5. Can you release a new version of your libs targeting this version ? The only compatibility issue is the use of `Array.emtpy`

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/CallBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/CallBuilder.cs
@@ -117,7 +117,7 @@ namespace ICSharpCode.Decompiler.CSharp
 									expandedArguments.Add(expressionBuilder.GetDefaultValueExpression(elementType).WithoutILInstruction());
 							}
 						}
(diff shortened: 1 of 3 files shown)


christophwille:
We have switched fw's specifically for Roslyn integration. We are not planning on changing this targetting.

christophwille:
Why didn't you go with netstandard2.0;net46;net45? three tfm, three dlls in the nupkg.

Titaye:
Yes of course no pb for having the 3 targetting. For compatibility reasons, I'm don't think FSAC will be able to move to net46. If you don't mind adding net45 to the existings list, it will be perferct.

siegfriedpammer:
I have committed a slightly modified version of your changes in 1e14c8f. Not sure, when we will be able to publish a beta for 3.1 on nuget.org. In the meantime you can use our private nuget feed.

Titaye:
this is perfect. thanks.

PR #1039, opened 2018-01-21
DecompileMemberBodies setting

### LINKED ISSUE #1032: Decompiling types without method bodies?
Is there support for decompiling whole types without method bodies?

I'd imagine there should be a check here somewhere:
https://github.com/icsharpcode/ILSpy/blob/97efc7b7f5807c47c0316bd7ec51ebb661c4a4a4/ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs#L727

Ideally `DecompileType()` should get an optional bool parameter "decompileMethodBodies" (default to true ) that should turn off bodies for methods, properties, events, indexers, constructors, destructors and operators.

This is not really blocking but nice to have so that we can preserve MonoDevelop's "Summary" mode. MonoDevelop currently has AstBuilder.DecompileMethodBodies flag: http://source.monodevelop.com/#MonoDevelop.AssemblyBrowser/MonoDevelop.AssemblyBrowser/TreeNodes/DomMethodNodeBuilder.cs,9ba778e1a1490f5b

siegfriedpammer:
Fixes #1032 

* [x] Add `DecompilerSettings.DecompileMemberBodies`
* [x] Decide whether field initializers should be displayed, if `DecompilerSettings.DecompileMemberBodies` is set to `false`.

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
+++ b/ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
@@ -47,7 +47,7 @@ namespace ICSharpCode.Decompiler.CSharp
 	{
 		readonly DecompilerTypeSystem typeSystem;
 		readonly DecompilerSettings settings;
-		private SyntaxTree syntaxTree;
+		SyntaxTree syntaxTree;
 
 		List<IILTransform> ilTransforms = GetILTransforms();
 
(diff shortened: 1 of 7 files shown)


siegfriedpammer:
cc @dgrunwald @KirillOsenkov Any comments?

KirillOsenkov:
:+1:

PR #1051, opened 2018-01-31
Support for ArrayInitializers=false (SharpLab)

ashmind:
This adds a new setting `ArrayInitializers` that can be used to suppress prettifier for array initializers.

**Notes:**
1. This is mostly intended for SharpLab where I want to be as detailed as possible. I haven't added it to the ILSpy settings -- please let me know if I should.

2. I've added support for ECMA-335 II.16.3.2 field data as well, as array initializers use it, and it makes the result more sensible. Since it's not supported by C# it just generates a comment -- let me know if there is a better way (annotation, then comment?). This is also a breaking change in case someone implemented `IField` or `IUnresolvedField` -- I'm open for discussion.

3. I've added a new test group ("Ugly"), because I intend to contribute some more de-prettifiers. It can also be used for things like tests for existing `ObjectOrCollectionInitializers`.

(ref: ashmind/SharpLab#231)


siegfriedpammer:
The build currently fails because we enforce correct indentation (we use tabs instead of spaces everywhere) by running tidy.py. If you have python installed, you can add https://gist.github.com/siegfriedpammer/75700ea61609eb22714d21885e4eb084 to your .git/hooks to prevent errors like this in the future.

siegfriedpammer:
Thank you for your work!

I am not quite sure about the changes to `IField` and `IUnresolvedField`. The thing is, the "TypeSystem" in the decompiler models all the metadata from the high-level perspective of C#. If a feature is not supported in C#, there should be no support for it. You can always use `IDecompilerTypeSystem.GetCecil` to retrieve the Cecil object and access the low-level metadata, if needed.

ashmind:
> You can always use IDecompilerTypeSystem.GetCecil to retrieve the Cecil object and access the low-level metadata, if needed.

That sounds perfect -- I wasn't aware of that.
Thanks -- I'll fix that and the build sometime tomorrow.

ashmind:
@siegfriedpammer Rewritten to `IDecompilerTypeSystem.GetCecil` -- definitely looks better (but required some extra ctor args in a few places). Fixed the build/formatting.

siegfriedpammer:
Thank you! Sorry for taking so long to reply.

The thing I had in mind was: Add another transform (or modify `TransformArrayInitializers` accordingly) that adds the data blobs as comments. Then you would not have to modify the TypeSystemAstBuilder and CSharpAmbience, because the type system is available through the context of each transform.

What do you think?

ashmind:
@siegfriedpammer Sure -- pls let me know if the update makes sense.

ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
```diff
@@ -873,6 +873,14 @@ EntityDeclaration DoDecompile(FieldDefinition fieldDefinition, IField field, ITy
 				RemoveAttribute(fixedFieldDecl, fixedBufferAttributeTypeName);
 				return fixedFieldDecl;
 			}
+			var cecilField = typeSystem.GetCecil(field) as FieldDefinition;
```
siegfriedpammer: No need to use GetCecil here... the first parameter already is the correct Cecil FieldDefinition.

ashmind: Great point -- tried to update quickly and missed that.
Fixed now.

siegfriedpammer:
Thank you for your work!

Issue #1046, opened 2018-01-25
VS MEF instead of MEF


christophwille:
This started as an internal discussion - to replace the currently used "off the shelf" MEF with https://github.com/Microsoft/vs-mef to improve startup performance (WPF UI client). @dgrunwald commented (yes, it was a high-level discussion) that ".. we are loading assemblies a bit strangely to be able to support dlls from the Internet zone".

This is a very low-pri item for us, if you want to help and replace mef with vsmef, a PR is very welcome.

Zastai:
It mainly depends on whether you're using the "fancier" parts of MEF 1 (custom ExportProviders or catalogs) or relying on recomposition (i.e. dropping a plugin into the folder makes it immediately available in running instances).
If not, rewriting the hosting side to use MEF 2 or VS MEF shouldn't be very difficult. I'll have a look.

Having said that, the biggest speed boost often comes from a) having a dedicated subdirectory for parts, and b) splitting larger plugins so that the plugin part is in the plugins dir, referencing the bigger implementation part, which is in the application's normal lookup path. That way, MEF has far fewer types to scan (MEF 1 especially, because it looks at ALL types, not just public ones).

Zastai:
I have it pretty much working already.
The main thing I've hit is that VS MEF itself does not provide ComposeParts()/SatisfyImportsOnce() functionality - it's set up to resolve [Imports] only for [Export]ed types.
There's two approaches:
1. I can try to see if I can use an MEF1 container using VS MEF exports (but even if that works, it adds another layer of indirection)
2. I can remove the [Import] tags and replace the ComposeParts() call with the appropriate GetExport*() calls

So far I have done 2. because there aren't many imports. I could look into 1. if it was strongly preferred.

Is there some plugin set used for testing (both for ensuring all types of parts are present, and to have a long startup time)?

siegfriedpammer:
>Is there some plugin set used for testing (both for ensuring all types of parts are present, and to have a long startup time)?

No, there is none AFAIK.

Zastai:
OK, pull request 1058 created. This seems to work, but I've not tested it much beyond throwing in the test plugin and some of the bigger ones like Reflexil.
Currently, the discovery does MEF1 only. If MEF2 support on the plugin side is desirable, it can be easily activated; but the heavy WPF dependency probably makes it pointless (MEF2 is only relevant if a plugin wants to work on .NET Core/Standard, and there's no WPF there anyway).

[...]

PR #1058, opened 2018-02-04
Issue #1046: Use VS MEF for hosting

Zastai:
This switches from an MEF1 CompositionContainer to a VS MEF ExportProvider for the extensibility processing.


siegfriedpammer:

PR #1072, opened 2018-02-24
Set the tab size in .editorconfig so it's reflected on GitHub

sharwell:
Currently GitHub shows hard tabs as 8 spaces wide in this repository, which makes it hard to view files with the default views. This change sets the tab size in a way GitHub understands, so they show up as 4 spaces wide instead.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,6 +4,7 @@ root = true
 [*]
 indent_style = tab
 indent_size = 4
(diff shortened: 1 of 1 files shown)


sharwell:
This seems to be a problem only for Edge. Chrome already shows the correct width and Edge is not corrected by this change.

PR #1087, opened 2018-03-10
Reduced allocations

pentp:
This reduces allocations for a full decompilation about 3% (CPU time is also reduced about 3-4%).

_Because I haven't contributed for a long time to ILSpy I'm not sure if these changes have acceptable quality and style. Most of them are independent of each other, so I can remove some of them based on feedback._

### DIFF
--- a/ICSharpCode.Decompiler/IL/ILReader.cs
+++ b/ICSharpCode.Decompiler/IL/ILReader.cs
@@ -996,7 +996,6 @@ namespace ICSharpCode.Decompiler.IL
 			Debug.Assert(inst.ResultType != StackType.Void);
 			IType type = compilation.FindType(inst.ResultType.ToKnownTypeCode());
 			var v = new ILVariable(VariableKind.StackSlot, type, inst.ResultType, inst.ILRange.Start);
-			v.Name = "S_" + inst.ILRange.Start.ToString("x4");
 			v.HasGeneratedName = true;
 			currentStack = currentStack.Push(v);
 			return new StLoc(v, inst);
--- a/ICSharpCode.Decompiler/Util/CollectionExtensions.cs
+++ b/ICSharpCode.Decompiler/Util/CollectionExtensions.cs
@@ -252,5 +252,19 @@ namespace ICSharpCode.Decompiler.Util
 				throw new ArgumentNullException(nameof(list));
 			list.RemoveAt(list.Count - 1);
 		}
+
+		#region Aliases/shortcuts for Enumerable extension methods
+		public static bool Any<T>(this ICollection<T> list) => list.Count > 0;
+		public static bool Any<T>(this T[] array, Predicate<T> match) => Array.Exists(array, match);
+		public static bool Any<T>(this List<T> list, Predicate<T> match) => list.Exists(match);
+
+		public static bool All<T>(this T[] array, Predicate<T> match) => Array.TrueForAll(array, match);
+		public static bool All<T>(this List<T> list, Predicate<T> match) => list.TrueForAll(match);
+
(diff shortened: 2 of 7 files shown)


dgrunwald:
Build failure is due a problem on master branch; I'll re-run the CI once that is fixed.

ICSharpCode.Decompiler/IL/ILReader.cs
```diff
@@ -996,7 +996,8 @@ ILInstruction Push(ILInstruction inst)
 			Debug.Assert(inst.ResultType != StackType.Void);
 			IType type = compilation.FindType(inst.ResultType.ToKnownTypeCode());
 			var v = new ILVariable(VariableKind.StackSlot, type, inst.ResultType, inst.ILRange.Start);
-			v.Name = "S_" + inst.ILRange.Start.ToString("x4");
+			var start = inst.ILRange.Start;
+			v.Name = start == 0 ? "S_0000" : "S_" + start.ToString("x4");
```
dgrunwald: Oh interesting, those are a lot generated strings. And (almost?) all of them will never be used (except for debugging output).
Is the `start == 0` special case all that helpful? I would expect the values to be evenly distributed across the IL body.
We should try to see if we can get away with `v.Name == null` here and generate this default name only on-demand (when writing debug output).

dgrunwald: Of course the fact that we generate an ILVariable object at all is already suboptimal -- if we combined the ILReader with a first (simple) round of inlining, we could probably avoid allocating at least half of these variables.

ICSharpCode.Decompiler/Util/CollectionExtensions.cs
```diff
@@ -252,5 +252,19 @@ public static void RemoveLast<T>(this IList<T> list)
...
+		public static bool Any<T>(this T[] array, Predicate<T> match) => Array.Exists(array, match);
+		public static bool Any<T>(this List<T> list, Predicate<T> match) => list.Exists(match);
+
+		public static bool All<T>(this T[] array, Predicate<T> match) => Array.TrueForAll(array, match);
+		public static bool All<T>(this List<T> list, Predicate<T> match) => list.TrueForAll(match);
+
+		public static T FirstOrDefault<T>(this T[] array, Predicate<T> predicate) => Array.Find(array, predicate);
+		public static T FirstOrDefault<T>(this List<T> list, Predicate<T> predicate) => list.Find(predicate);
+
+		public static T Last<T>(this IList<T> list) => list[list.Count - 1];
```
dgrunwald: I recently switched over quite a bit of stuff from `IList<T>` to `IReadOnlyList<T>`, so we might want an extension method for that as well.

ICSharpCode.Decompiler/IL/ILReader.cs
pentp: I'm not sure why but when I tested this then `ILRange.Start` was always 0. Maybe there is something wrong elsewhere.
I agree that a lot of these ILVariables are unnecessary and a more general solution (like inlining or using null) works better here.

I started with some optimization work mainly to get familiar with the codebase again.

dgrunwald: Oh right, the `ILRange` only gets set in `ReadInstructions()`, which is after the `Push()` call.
The variable name we see in debug output is that assigned by the `CollectStackVariablesVisitor`, after the variables from different control flow paths are unified.
I guess that means the initial name can safely stay `null`, as there's no way to observe it outside of the `ILReader`.

ICSharpCode.Decompiler/Util/CollectionExtensions.cs
pentp: Adding `IReadOnlyList<T>` methods causes overload resolution failures for collections that implement both interfaces.

PR #1096, opened 2018-03-13
Added Ability to remove invalid nodes.

mookiejones:
Added Ability to remove invalid assemblies that arent a valid clr object


siegfriedpammer:
Thank you for your contribution and this idea! I cherry-picked and refactored your code in ee093267a8118b689aca3470de2b21224c00618b

Reasons:
1) You were using the internal list of assemblies instead of the thread-safe `GetAssemblies` method.
2) I think, using a foreach loop in this case is more succint.

Thank you again!

PR #1108, opened 2018-04-01
Reduced allocations in ILInlining

pentp:
This reduces allocations for a full decompilation about 4% (CPU time is also reduced about 4%).

Additionally, it fixes a few rare cases where `ILInlining` was too aggressive when `topLevelInst.ILRange` was not set (e.g., when decompiling `ICSharpCode.Decompiler.DotNetCorePathFinder` or `System.Diagnostics.Tracing.TypeAnalysis`).

### DIFF
--- a/ICSharpCode.Decompiler/IL/Transforms/ILInlining.cs
+++ b/ICSharpCode.Decompiler/IL/Transforms/ILInlining.cs
@@ -30,15 +30,16 @@ namespace ICSharpCode.Decompiler.IL.Transforms
 	{
 		public void Run(ILFunction function, ILTransformContext context)
 		{
+			int? ctorCallStart = null;
 			foreach (var block in function.Descendants.OfType<Block>()) {
-				InlineAllInBlock(block, context);
+				InlineAllInBlock(function, block, context, ref ctorCallStart);
 			}
 			function.Variables.RemoveDead();
 		}
 
 		public void Run(Block block, BlockTransformContext context)
 		{
-			InlineAllInBlock(block, context);
+			InlineAllInBlock(context.Function, block, context);
 		}
 
 		public void Run(Block block, int pos, StatementTransformContext context)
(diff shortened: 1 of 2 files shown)


siegfriedpammer:
@pentp Out of curiosity: how do you measure allocations and speed in ILSpy? Which profiler do you use?

ICSharpCode.Decompiler/IL/Transforms/ILInlining.cs
```diff
@@ -30,28 +30,36 @@ public class ILInlining : IILTransform, IBlockTransform, IStatementTransform
...
+		}
+
+		static bool InlineAllInBlock(ILFunction function, Block block, ILTransformContext context, ref int? ctorCallStart)
 		{
 			bool modified = false;
-			int i = 0;
-			while (i < block.Instructions.Count) {
-				if (InlineOneIfPossible(block, i, aggressive: IsCatchWhenBlock(block) || IsInConstructorInitializer(i, block), context: context)) {
+			var instructions = block.Instructions;
+			for (int i = 0; i < instructions.Count;) {
```
siegfriedpammer: Why did you replace the while loop with a for loop? A for loop without increment is a bit confusing at first sight.

pentp: It's more compact and the loop variable is better scoped, but it's a matter of taste. Should I change it back?

pentp:
I use PerfView with ETW sampling of allocations. I made a small command line program that just runs decompilation on an assembly (with a warmup run before perf. measuring).

ICSharpCode.Decompiler/IL/Transforms/ILInlining.cs
siegfriedpammer: Well... I guess we can leave it like it is, I don't think it's worth the time to change it back.

siegfriedpammer:
> I use PerfView with ETW sampling of allocations. I made a small command line program that just runs decompilation on an assembly (with a warmup run before perf. measuring).

Thanks for the info!

PR #1124, opened 2018-04-25
Include "base." qualifier as needed (fixes #1123)

### LINKED ISSUE #1123: "base." qualifier always included
Test case
```cs
	public class A
	{
		public int i;
		public int j;

		public void DoA() { }
		public void DoB() { }
		public virtual void DoC() { }
	}

	public class B : A
	{
		public int j;
		public int k;

		public new void DoB() { }

		public override void DoC() {
			base.DoC();
		}

		public int Mul() => i * base.j * j * k;

		public void Calls() {
			DoA();
			base.DoB();
			DoB();
			base.DoC();
			DoC();
		}
	}
```

Decompiled
```cs
	public int Mul()
	{
		return base.i * base.j * j * k;
	}

	public void Calls()
	{
		base.DoA();
		base.DoB();
		DoB();
		base.DoC();
		DoC();
	}
```

Chicken-Bones:
Adds BaseReferenceExpression to ThisReferenceExpression target required checks


siegfriedpammer:
Thank you for your contribution! Just two things:

1. Could you please rebase this onto the `3.1.x` branch, as that will be the next release.
2. Could you please add some `base` unit tests to `QualifierTests` to ensure it even works if there is a conflict?
  * overridden methods calling their base method (missing `base` would result in infinite recursion!)
  * another method calling overridden `this.A()` and `base.A()`
  * same test using an virtual shadowed method, i.e. `new void B()` instead of `override void B()`
  * same test using an non-virtual shadowed method, i.e. `new void C()`

If you need additional base classes and types for testing, please add them as nested types.

When modifying "pretty tests", please do not forget to delete (this will regenerate them on the next run) the `QualifierTests.*.il` files, before testing and committing!

Thank you!

Chicken-Bones:
I can rebase the branch on my fork, and force push it, but I don't know how to change the target branch of the PR. Would I just have to submit a new one?

siegfriedpammer:
You can press the "Edit" button at the top and change the base branch. Thank you!

siegfriedpammer:
These changes happen on every regeneration. It's not a problem.

siegfriedpammer:
Hmmm. On second thought... might be that you have a different version of ilasm installed on your system. But, I guess it will be fine anyways...

PR #1165, opened 2018-06-06
Dynamic support

siegfriedpammer:
Adds basic support for general dynamic expressions.

TODO:
* complex logical expressions involving IsTrue, IsFalse operators need some beautification.


ICSharpCode.Decompiler/CSharp/CallBuilder.cs
```diff
@@ -109,6 +109,15 @@ public TranslatedExpression Build(CallInstruction inst)
 					expectedTargetDetails.NeedsBoxingConversion = true;
 				}
 			}
+			
+			// Special case to deal with implicit dynamic conversions:
+			bool allowDynamicDispatch = true;
```
dgrunwald: When does it ever make sense to `allowDynamicDispatch` for a static `CallInstruction`?
I'd say that dynamic dispatch should never be allowed (which in turn means we can convert `dynamic` arguments unconditionally before we even try OR)

ICSharpCode.Decompiler/IL/Transforms/ExpressionTransforms.cs
```diff
@@ -329,6 +330,96 @@ protected internal override void VisitIfInstruction(IfInstruction inst)
...
+				return false;
+			if (!invokeMember.Arguments[0].Match(getMember.Target).Success)
+				return false;
+			context.Step("+= / -= dynamic.isevent pattern -> dynamic.compound.op", inst);
+			inst.ReplaceWith(dynamicCompoundAssign);
+			if (getMember.Target.MatchLdLoc(out var v) && v.Kind == VariableKind.Local && v.IsSingleDefinition && v.LoadCount == 1 && v.StoreInstructions[0] is StLoc initStore) {
+				if (ILInlining.CanInlineInto(dynamicCompoundAssign, v, initStore.Value)) {
+					// HACK: if inlining is possible, we can 'cheat' a bit and change the variable kind to StackSlot
+					// so inlining or copy propagation will take care of the extra compiler-generated local.
+					v.Kind = VariableKind.StackSlot;
```
dgrunwald: This is better done by extending `ILInlining.NonAggressiveInlineInto()`

PR #1213, opened 2018-07-17
Reverse Roslyn optimisation to favour && over & on boolean operands.

Chicken-Bones:
Roslyn replaces logical AND operators with bitwise operators, if the right hand side has no side-effects and relatively little performance cost (such as ldloc).

~The proposed solution is by no means comprehensive, but it might be a start. Perhaps `ExpressionTransforms` would be a better place than `EarlyExpressionTransforms`. I wanted it to go before `ConditionDetection` but I don't think it matters.~

### DIFF
--- a/ICSharpCode.Decompiler/IL/Transforms/EarlyExpressionTransforms.cs
+++ b/ICSharpCode.Decompiler/IL/Transforms/EarlyExpressionTransforms.cs
@@ -111,5 +111,49 @@ namespace ICSharpCode.Decompiler.IL.Transforms
 			}
 			return null;
 		}
+
+		protected internal override void VisitBinaryNumericInstruction(BinaryNumericInstruction inst)
+		{
+			if (inst.Operator != BinaryNumericOperator.BitAnd)
+				return;
+
+			if (PotentialSideEffects(inst.Left) || PotentialSideEffects(inst.Right))
+				return;
+
+			if (!IsBoolean(inst.Left) || !IsBoolean(inst.Right))
+				return;
+			
+			context.Step("Replace bit.and with logic.and", inst);
+			var expr = IfInstruction.LogicAnd(inst.Left, inst.Right);
+			inst.ReplaceWith(expr);
+			expr.AcceptVisitor(this);
+		}
+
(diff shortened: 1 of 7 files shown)


ICSharpCode.Decompiler/IL/Transforms/EarlyExpressionTransforms.cs
```diff
@@ -111,5 +111,49 @@ internal static ILInstruction HandleCall(Call inst, ILTransformContext context)
...
+			
+			context.Step("Replace bit.and with logic.and", inst);
+			var expr = IfInstruction.LogicAnd(inst.Left, inst.Right);
+			inst.ReplaceWith(expr);
+			expr.AcceptVisitor(this);
+		}
+
+		private bool IsBoolean(ILInstruction inst) => inst is Comp || inst.InferType().IsKnownType(KnownTypeCode.Boolean);
+
+		private bool PotentialSideEffects(ILInstruction inst)
```
dgrunwald: This function already exists: `!SemanticHelper.IsPure(inst.Flags)`
(your version misses that `BinaryNumericInstruction` can have side effects, e.g. `OverflowException` or `DivideByZeroException`)

dgrunwald:
I think it would be better to put it in the normal `ExpressionTransforms`. It shouldn't hurt to have it after `ConditionDetection`, and later on we might have better information about which expressions will be of type `bool`.
By the way, not every `Comp` returns `bool`. There's the special case of `ComparisonLiftingKind.ThreeValuedLogic`, which returns `bool?` instead.

ICSharpCode.Decompiler/IL/Transforms/EarlyExpressionTransforms.cs
```diff
@@ -111,5 +111,49 @@ internal static ILInstruction HandleCall(Call inst, ILTransformContext context)
 			}
 			return null;
 		}
+
+		protected internal override void VisitBinaryNumericInstruction(BinaryNumericInstruction inst)
+		{
+			if (inst.Operator != BinaryNumericOperator.BitAnd)
+				return;
```
dgrunwald: The early return here prevents recursing into the child instructions.
In expression transforms, it's usually better to first call `base.VisitBinaryNumericInstruction()` to handle the children (thus applying transforms in post-order).

PR #1236, opened 2018-07-25
Decimal constant improvements

ashmind:
This covers two main things:

1. Added support for decimal optional parameters, e.g. `M(decimal d = 1m)`. Since those are stored as `[DecimalConstant(...)]` and not standard constants, previously they would not decompile correctly.

2. Added support for new setting `DecimalConstants`. If set to `false`, decimal constants and optional parameters will explicitly expose `[DecimalConstant(...)]` instead of pretending to be a literal. This is required by SharpLab since I want to expose inner workings as close as possible.

Both changes include unit-tests.

(ref: ashmind/SharpLab#316)


ICSharpCode.Decompiler/TypeSystem/Implementation/MetadataField.cs
```diff
@@ -42,7 +42,7 @@ sealed class MetadataField : IField
 		object constantValue;
 		IType type;
 		bool isVolatile; // initialized together with this.type
-		byte decimalConstant; // 0=no, 1=yes, 2=unknown
+		bool? isDecimalConstant;
```
dgrunwald: This change is not thread-safe.
In the rare case where multiple threads are concurrently initializing `isDecimalConstant`, we need accesses to be atomic.
But `Nullable<T>` is a struct and may have torn reads/writes.

ICSharpCode.Decompiler/TypeSystem/Implementation/MetadataParameter.cs
```diff
@@ -37,6 +37,8 @@ sealed class MetadataParameter : IParameter
 
 		// lazy-loaded:
 		string name;
+		bool? hasConstantValueInSignature;
+		bool? isDecimalConstant;
```
dgrunwald: Same thread-safety issues here.

ashmind:
@dgrunwald Thanks, fixed -- didn't ever consider torn reads before, something to keep in mind.
Not sure if I should have gone with an enum here -- please let me know.

PR #1246, opened 2018-08-06
C# 7.3 stackalloc initializers

siegfriedpammer:
This implements basic patterns of C# 7.3 stackalloc initializer expressions.


siegfriedpammer:
Note that once https://github.com/dotnet/roslyn/issues/29092 is fixed, we'll have to add more tests that use the `initblk` pattern.

ICSharpCode.Decompiler/CSharp/Transforms/IntroduceUnsafeModifier.cs
```diff
@@ -142,7 +142,7 @@ public override bool VisitIdentifierExpression(IdentifierExpression identifierEx
 		public override bool VisitStackAllocExpression(StackAllocExpression stackAllocExpression)
 		{
 			base.VisitStackAllocExpression(stackAllocExpression);
-			return true;
+			return stackAllocExpression.GetResolveResult().Type is PointerType;
```
dgrunwald: Don't forget the return value of the base call.
It's theoretically possible to have a safe stackalloc that uses unsafe code in the size computation.

PR #1253, opened 2018-08-10
A possible fix for Issue 1251

### LINKED ISSUE #1251: Method calls on nullable value types prevent inlining
[Web.zip](https://github.com/icsharpcode/ILSpy/files/2274164/Web.zip)

Web.Areas.Admin.Controllers.CostReceController.Sexport
Web.Areas.Admin.Controllers.CostReceController.ExpReportList

siegfriedpammer:
This should fix #1251. I am not sure if we can treat callvirt instructions the same as call instructions, but I think this should be no problem.


dgrunwald:
Looks good to me.
We might want to extend this to `newobj` as well while we're at it (note: newobj will need to check the `DeclaringType` instead of the `ReturnType`)

PR #1274, opened 2018-08-30
Make CSharpDecompiler and MetadataModule IDisposable to avoid file locks

z1c0:
I have a scenario, where I want to delete assemblies again immediately after decompilation. This does not work currently, because a `PEFile` in `MetadataModule` instance keeps a lock on the file.

See this simplified scenario:

```csharp
static string Decompile(string filePath)
{
	string code = "";
	using (var decompiler = GetDecompiler(filePath)) {
		code = decompiler.DecompileWholeModuleAsString();
	}
	File.Delete(filePath); // now works, since CSharpDecompiler and MetadataModule are disposable
	return code;
}

static CSharpDecompiler GetDecompiler(string assemblyFileName)
{
	return new CSharpDecompiler(assemblyFileName, new DecompilerSettings() {  ThrowOnAssemblyResolveErrors = false });
}
```

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
+++ b/ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
@@ -54,1447 +54,1316 @@ namespace ICSharpCode.Decompiler.CSharp
 	/// (in particular, the transform instances are not thread-safe)
 	/// </remarks>
 	public class CSharpDecompiler : IDisposable
-
 	{
-	readonly IDecompilerTypeSystem typeSystem;
-	readonly MetadataModule module;
-	readonly MetadataReader metadata;
-	readonly DecompilerSettings settings;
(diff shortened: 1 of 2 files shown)


siegfriedpammer:
First of all: Can you please revert all the changes in CSharpDecompiler.cs that are not necessarily part of your desired change? It seems some VS extension went berserk on that file while not honoring the settings in .editorconfig. Cosmetic changes such as indentation and brace style should not be part of the commit with real code changes.

z1c0:
Sorry, didn't realize that.

christophwille:
This does not implement the standard Dispose pattern https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/dispose-pattern

z1c0:
Thanks for the feedback.
I created an issue (#1275) for that and will open another pull request if we come to a conclusion how we could handle that.

Issue #119, opened 2011-04-11
Tabbed UI


arturek:
The current version of ILSpy is able to display only one decompiled item at a time. Usually I need to see more methods or classes to find out how they work together, so I have added support for tabbed interface (based on AvalonDock). It's available in master branch of my arturek/ILSpy fork (arturek/ILSpy@a58dbcf108be591e6390) if you want to try it out. It's not a complete solution, rather a working example. What do you think about this idea? Should ILSpy stay SDI or could it be enhanced this way?

MatthewBonner:
Please add tabs, this issue has been open since 2011 when tabs where not widely implemented, they are now, and people are used to tabs, please provide the option to either have a new window or a new tab as at present you cannot open multiple tabs at the same time. The reason the .NET Reflector UI has problems with tabs is because you don't get the option to open in a new tab by middle-clicking the mouse. In order to provide tabs you should have a context menu to open in a new tab, middle click on function references and the option to turn tabs on or off and then there won't be any issues.

migig:
+1 tabs would make it so much more useful

Verser:
+1
This is a very big thing. In general, some UI features could be incredibly helpful. Tabs, Favorite/bookmarks, Projects/Groups (to throw specific parts of assemblies together, like making a link to grouped buggy code).

I know some of this is on the roadmap, but as someone who uses it almost every day to make obfuscation checks, IL reading, and version differentials, it would be incredibly helpful to have these features.

dolkensp:
@siegfriedpammer @Radon222 Is there any progress on this?

Honestly, I'd settle for "bookmarks" - I finally found one method I need, in a list of 1000 methods, but now I need to check another method, and coming back here is going to be a pain :P

Tabs would resolve this, as would bookmarks if a UI overhaul is imminent.

siegfriedpammer:
Implementing bookmarks is on my list. However, I am currently very very busy, so no idea, when or if I ever will get around to it. :(

siegfriedpammer:
superseded by #1724

PR #1276, opened 2018-08-30
Support multiple Tabs in ILSpy

Radon222:
I've added support for multiple tabs in ILSpy, each tab would contain a DecompilerTextView, so you can open several tabs in one instance without opening several ILSpy instances. 

This would resolve issue #119 

Added a toolbar button to created a new tab, see below:

image

Also added a menu entry under File menu, see below

image

Feel free to make comments :+1:


Radon222:
@siegfriedpammer @arturek @Matthew-Bonner @Verser @migig @eusebiu

christophwille:
We currently are (and have been for quite a while) discussing "future directions of the UI". We want to go cross platform, and we definitely want to take that opportunity to modernize the UX: current thinking is MDI, which would get us a lot of the requested functionality - like SxS-viewing of decompilation results (think comparing different versions of a codebase). Maybe @siegfriedpammer could chime in with some of the UX ideas.

siegfriedpammer:
Our idea basically was to support multiple open documents and allow to stack them side-by-side, i.e., full-blown docking like VS. This allows comparing of versions, comparing C# and IL, looking and XAML resources while reading C# or IL, etc.

christophwille:
To be replaced by https://github.com/icsharpcode/ILSpy/tree/avalondock

christophwille:
#1724 contains the feature discussion

siegfriedpammer:
Thank you for the pull request!

We have implemented an improved version of this based on AvalonDock in #1801, which was merged to master a few minutes ago.

I hope you like the new docking features and if there still is something missing, please feel free to open a pull request!

Issue #1256, opened 2018-08-12
Missing ValueType cast


mmusu3:
Source from [Open.NAT](https://github.com/lontivero/Open.NAT/blob/328b6eebdbaa0bccf968ad43dc698cb768a94a5c/Open.Nat/Utils/Extensions.cs#L171)
```cs
public static bool HasFlag(this Enum value, Enum flag)
{
    int intValue = (int)(ValueType)value;
    int intFlag = (int)(ValueType)flag;

    return (intValue & intFlag) == intFlag;
}
```
Output:
```cs
public static bool HasFlag(this Enum value, Enum flag)
{
    int num = (int)value;
    int num2 = (int)flag;
    return (num & num2) == num2;
}
```
Gives Error CS0030 Cannot convert type 'System.Enum' to 'int'

PR #1277, opened 2018-09-06
Add cast to object for reference types in ExpressionBuilder.VisitUnboxAny

Viridovics:
Fix issue https://github.com/icsharpcode/ILSpy/issues/1256


siegfriedpammer:
Thank you for the fix!

PR #1278, opened 2018-09-06
Merge SharpTreeView changes from SharpDevelop repository to ILSpy.

LengVang:
This also changes the SharpTreeView license from LGPL to MIT.


siegfriedpammer:
We've just decided that merging the changes from the SharpDevelop repository is not going to happen, because it is not an easy task due to breaking API changes and not worth the effort because ILSpy does not need any of the new features implemented in SharpDevelop's version.

Issue #1281, opened 2018-09-12
Arguments for params object[]


gatodevelop:
Source:

    void m(params object[] args) {
        var a = new object[0];
        m(a);
        m((object)a);
        m(new object[] { a });
    }

Output:

    private void m(params object[] args)
    {
        object[] array = new object[0];
        m(array);
        m(array);
        m(array);
    }

PR #1289, opened 2018-10-03
Add additional rule for TransformParamsArgument

Viridovics:
Fix issue #1281


Viridovics:
It's not working for case:

```        
        public static void M(int a, params object[] x)
        {
            var arrays = new object[0];
            M(a, (object)arrays);
        }
```

siegfriedpammer:
Thank you for the proposed fix!

However, we found that your fix does not work for cases like `MethodWithParams(new object[1]);` which is being turned into `MethodWithParams(null);` which would call `MethodWithParams` with the array being set to `null` as opposed to an array with one `null` element.
I implemented an alternative solution which takes overload resolution rules into account in c9b74865fc7786d415e8e64c02becd12ee93a28f.

Viridovics:
@siegfriedpammer thank you for your feedback :)

Issue #1269, opened 2018-08-24
Bad decompilation of yields


tellowkrinkle:
When decompiling [the Assembly-CSharp.dll for Higurashi Vol. 6](https://github.com/icsharpcode/ILSpy/files/2317615/Assembly-CSharp.dll.zip), a large number of files including `Assets.Scripts.UI.Prompt/PromptController.cs `, `Assets.Scripts.UI.SaveLoad/SaveLoadManager.cs`, and `Assets.Scripts.UI/MainUIController.cs` are left with lines like `/*Error: Unable to find new state assignment for yield return*/;` after yield statements, with the rest of the code missing.

dimamatik:
Hello! I have a same issue with .dll, compiled by Mono from Unity Engine. For example, this code:
`while (oneObject.itsfield == false)
{
yield return null;
}
`
decomiled as:
`if (oneObject.itsfield == false)
{
yield return null;
/*Error: Unable to find new state assignment for yield return*/;
}
`
while circle was changed to if!!!
the latest release (4.0.0.4358-beta2) does not fix this problem
you can contact me by email (dimamatik[at]yandex[dot]ru) for more information and testing

dimamatik:
@siegfriedpammer 
Hello! I have a same problem on 4.0.0.4358-beta2. Code, generated with Unity3d, like:
`while (myObject.itsField == false) { yield return null; }`
was decompiled as:
`if (myObject.itsField == false) { yield return null; /*Error: Unable to find new state assignment for yield return*/;}`

Please, fix!
You can contact me by email dimamatik[at]yandex[dot]ru for more info.

dimamatik:
@vizv @siegfriedpammer
Hello! I download and build your fix (fix-yield-return-generated-from-mono) for "Unable to find new state assignment for yield return" problem. It's works! But I have a problem with simillar code:

I am using Unity 2018 with Mono compiler integrated into this engine
This class:
`using System.Collections;`
`using System.Collections.Generic;`
`using UnityEngine;`

`public class Test : MonoBehaviour `
`{`
`	List<Test> children = new List<Test>();`
`	public IEnumerable<Test> Children`
`	{`
`		get`
`		{`
`			foreach( Test item in children)`
`			{`
`				yield return item;`
`			}`
`			yield break;`
`		}`
`	}`
`}`

was decompiled with error after yield break line:
`IL_00bd:`
`/*Error near IL_00be: Unexpected return in MoveNext()*/;`
[...]

Issue #1152, opened 2018-05-30
Yet another MCS yield-return problem


tamlin-mike:
It seems later Mono MCS have added an interleaved disposing check inside the compiler-generated MoveNext, terminating the current `Block` before it reaches the actual state-storing the decompiler expects at the end of the `Block`.

I took a stab at it, only to realize this very problem once I had implemented the workaround/dispose_check_skipping.

To my untrained eye, it seems one would have to merge the trailing instructions (the state setting) into the current block, at the position the (now discarded) $dispose check used to be, but I bet if I tried to do that I'd mess something else up quite horribly. :-)

Anyway, here's an excerpt of the IL MCS has generated, and the (incomplete and non-functional) patch of what I tried.
I hope it helps at least a little bit in fixing the problem.

Containing class (decompiled, partial):
```
class OwningClass {
	private float accum;
	private int frames;
	private int ActualField = 1;
        /* more fields irellevant for the problem at hand */
	private IEnumerator ItFunc() { /*...*/ }
}
```

IL of compiler-generated iterator:
```
.class nested private auto ansi sealed beforefieldinit '<ItFunc>c__Iterator0'
{
  .field assembly float32 '<capturedField>__0'
  .field assembly class OwningClass $this
  .field assembly object $current
  .field assembly bool $disposing
  .field assembly int32 $PC
  
  .method public hidebysig specialname rtspecialname instance void .ctor () cil managed 
  {
    IL_0000: ldarg.0
    IL_0001: call instance void [mscorlib]System.Object::.ctor()
    IL_0006: ret
  }
  
  .method public final hidebysig newslot virtual instance bool MoveNext () cil managed 
  {
    .maxstack 3
    .locals init (
    	[0] uint32
    )

    IL_0000: ldarg.0
    IL_0001: ldfld int32 OwningClass/'<ItFunc>c__Iterator0'::$PC
    IL_0006: stloc.0
    IL_0007: ldarg.0
    IL_0008: ldc.i4.m1
    IL_0009: stfld int32 OwningClass/'<ItFunc>c__Iterator0'::$PC
    IL_000e: ldloc.0
    IL_000f: switch (IL_0021, IL_00a3)

    IL_001c: br IL_00af

    IL_0021: ldarg.0
    IL_0022: ldarg.0
    IL_0023: ldfld class OwningClass OwningClass/'<ItFunc>c__Iterator0'::$this
    IL_0028: ldfld float32 OwningClass::accum
[...]

Issue #1057, opened 2018-02-04
yield FSM problem


tamlin-mike:
Most likely generated by MONO compiler. Fluff stripped out, won't work as-is with ilasm.
```
// the generated Enumerator for the yield return FSM
.class ThisIt
	implements System.Collections.IEnumerator, System.IDisposable, IEnumerator<T>
{
	.field assembly class OwningT $this
	.field assembly object $current
	.field assembly bool $disposing
	.field assembly int32 $PC

	.method instance void .ctor () cil managed 
	{
		ldarg.0
		call instance void [mscorlib]System.Object::.ctor()
		ret
	}

	.method public final hidebysig newslot virtual instance bool MoveNext () cil managed 
	{
		.maxstack 2
		.locals init (
			[0] uint32
		)

		ldarg.0
		ldfld int32 OwningT/ThisIt::$PC
		stloc.0
		ldarg.0
		ldc.i4.m1
		stfld int32 OwningT/ThisIt::$PC
		ldloc.0
		switch (IL_0021, IL_0045)

		br IL_005c_yield_break

		IL_0021: ldarg.0
		call class OwningT OwningT::get_something()
		callvirt SomeMethod_returning_something()
		stfld object OwningT/ThisIt::$current
		ldarg.0
		ldfld bool OwningT/ThisIt::$disposing
		brtrue.s IL_0040_yield_return

		ldarg.0
		ldc.i4.1
		stfld int32 OwningT/ThisIt::$PC

		IL_0040_yield_return: br IL_005e_yield_return

		IL_0045: ldarg.0
		ldfld class OwningT OwningT/ThisIt::$this
		ldfld class SomeClass OwningT::someField
		callvirt instance void SomeClass::SomeMethod()
		ldarg.0
		ldc.i4.m1
		stfld int32 OwningT/ThisIt::$PC

		IL_005c_yield_break: ldc.i4.0
		ret

		IL_005e_yield_return: ldc.i4.1
		ret
	}

	.method public final hidebysig newslot virtual  instance void Dispose () cil managed 
	{
		.maxstack 8

		ldarg.0
		ldc.i4.1
		stfld bool ThisIt::$disposing
		ldarg.0
		ldc.i4.m1
		stfld int32 ThisIt::$PC
		ret
	}
}
```
[...]

Issue #1026, opened 2018-01-11
Bug: "Error near IL_X: name" - with MCS Compiler Generated's IEnumarator Pattern


Dentrax:
Sometimes I see it in the iterator functions, not always. Is a decompile error ?

![s](https://user-images.githubusercontent.com/16493751/34843803-9b716e6a-f720-11e7-92cc-27b34bd025bb.png)

siegfriedpammer:
Which assembly are you trying to decompile? Do know which compiler was used to create it?

Dentrax:
I don't know, but the output of Protection ID tool ;

```
File Type : 32-Bit Dll (Subsystem : Win CUI / 3), Size : 3386368 (033AC00h) Byte(s) | Machine: 0x14C (I386)
Compilation TimeStamp : 0x57B4A466 -> Wed 17th Aug 2016 17:52:38 (GMT)
[TimeStamp] 0x57B4A466 -> Wed 17th Aug 2016 17:52:38 (GMT) | PE Header | - | Offset: 0x00000088 | VA: 0x00400088 | -
[File Heuristics] -> Flag #1 : 00000000000001001100000000110000 (0x0004C030)
[Entrypoint Section Entropy] : 5.80 (section #0) ".text   " | Size : 0x33A3C0 (3384256) byte(s)
[DllCharacteristics] -> Flag : (0x0000) -> NONE
[SectionCount] 3 (0x3) | ImageSize 0x342000 (3416064) byte(s)
[VersionInfo] Product Version : 0.0.0.0
[VersionInfo] File Version : 0.0.0.0
[VersionInfo] Original FileName : Assembly-CSharp.dll
[VersionInfo] Internal Name : Assembly-CSharp.dll
[ModuleReport] [IAT] Modules -> mscoree.dll
[.] .net @ FileOffset 0x13A644 | MetaData->Version 1.1 (struct version) -> v2.0.50727 (net version required)
[.] Flags : 0x0 | Streams : 0x5 (5)  -> #~ | #Strings | #US | #Blob | #GUID
[!] [.net scan core] dotNetReactor detected!
[COR20] MajorRuntimeVersion 0x2 (2) | MinorRuntimeVersion 0x2 (2) -> 0x2.2 (2.2)
[COR20] Flags 0x1
[COR20 Flags] [x] IL_ONLY [ ] 32BITREQUIRED [ ] IL_LIBRARY
[...]

PR #1303, opened 2018-10-21
Fix yield return for assembly compiled with Mono

vizv:
Trying to resolve (see #1269, #1152, #1057, #1025) without messing up with current implementation.

- [X] Fix `$disposing` field assignment generated with Mono causing `Unable to find new state assignment for yield return` error.
- [X]  Resolve stateMachine error (#1026)


dgrunwald:
Thanks! This looks great, though it took me a while to understand what the conditions in ConvertBranchAfterYieldReturn were about. I ended up refactoring that code in 5d394b7786.

PR #1317, opened 2018-11-17
Make accessor call inline to avoid branching merge failure

vizv:
Ensure inlining transformation runs correctly for accessor call (getter)


vizv:
Emmmmm, maybe there is a better way...

dgrunwald:
I'm not sure what exactly this was supposed to fix.

The ILInlining code has changed a bit since then; please create a new issue (with a test case) if this is still an issue.

PR #1322, opened 2018-11-24
Use astbuilder for all integer literals

siegfriedpammer:


dgrunwald:
This change still has some issues, a quick look at the roundtrip-diff showed:

1. it broke the context-dependent hexadecimal output (why don't we have any tests for this?)
```
-		public bool IsFrozen => (flags & 0x200) != 0;
+		public bool IsFrozen => (flags & 512) != 0;
```

2. it sometimes ends up introducing redundant casts, e.g. I've seen `(byte)byte.MaxValue`

siegfriedpammer:
No. 1 is fixed now.

No. 2 affects literals/constants of all small integer types, it seems:

```
fullName = ((version.MajorRevision != (short)5) ? Path.Combine(fullName, "2.0") : Path.Combine(fullName, "2.1"));
```

The left side is short and the right side should be implicitly convertible to short.

See https://github.com/icsharpcode/ILSpy/blob/use-astbuilder-for-all-integer-literals/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs#L2382:

The resolve result (5, System.Int16) is replaced by a different instance, but with the same values. Because of this, an explicit cast is added by ConvertTo.

PR #1405, opened 2019-01-26
Combine exitpoints in lambdas

siegfriedpammer:
* Add CombineExitsTransform: Merges `if (cond) return cond2; else return false;` or similar to `return cond && cond2;` in lambdas
* Fixed some bugs in QueryExpressions transforms.
* Fix linking in query expressions.


siegfriedpammer:
@dgrunwald Can this be merged?

Issue #1401, opened 2019-01-19
Option to not remember previous libraries


veccsolutions:
It would be really nice if there was an option to not remember assemblies from the last time it was running so it wouldn't load any of those or any of the default ones. Or is this already an option and I'm missing it?

siegfriedpammer:
Currently, there is no such option. You can only achieve this by creating a custom assembly list and manually clearing that before exiting ILSpy.

What is your use-case for this?

veccsolutions:
Primarily when I have opened libraries across a network share it takes a long time to re-open ilspy unless I cleared out the list before closing it. Another is it's just clutter. I made my own custom build that did this as an option defaulting to current behavior. I could do a pull request with it if you would like.

siegfriedpammer:
Yes please, pull requests are very welcome!

veccsolutions:
Done

siegfriedpammer:
Thank you very much!

PR #1409, opened 2019-02-01
Add option to not reload assemblies

veccsolutions:
This adds the ability to not re-load previously loaded libraries, issue #1401


siegfriedpammer:
LGTM, thank you very much!

PR #1410, opened 2019-02-01
Make `XmlBamlReader.Value` more defensive against null node values

garymcleanhall:
This will help to fix an [issue with Obfuscar ](https://github.com/obfuscar/obfuscar/issues/181)

I can't get a repro on ILSpy directly, though, so it's hard for me to prove that this requires an ILSpy fix.

However, there seems to be cases where `XmlBamlNode` is an `XmlBamlProperty` and its value is `null`. 

All this does is add a defense around this so that it returns an empty string, rather than throwing an exception.


siegfriedpammer:
The old BAML decompiler was removed in #1419. Therefore this no longer applies.

Issue #1412, opened 2019-02-07
Output file empty


tonysneed:
When I omit the `-o` flag decompiled code is shown in the console. But if I add `-o` and specify a directory, the *.decompiled.cs file is empty.

tonysneed:
I found the problem. It looks like the `TextWriter` is not being flushed: `output.Flush()` needs to be added [here](https://github.com/icsharpcode/ILSpy/blob/cee0a78d07c947c387d94a8521a2f4dcc0602488/ICSharpCode.Decompiler.Console/Program.cs#L139).

PR #1413, opened 2019-02-07
Flush TextWriter in Decompile method of Console app

tonysneed:
Fixes issue #1412.


siegfriedpammer:
Thank you for the proposed fix. However, I fixed #1412 by explicitly closing the stream before the process exits. I think it is better to handle this at the caller's side than each callee.

PR #1432, opened 2019-02-24
Add XML comments to the NuGet package

GeorchW:
Hey,

I've noticed that I am not getting any IntelliSense documentation on the ILDecompiler members. However, when looking at the code, there is plenty of XML documentation. Sure, some members may be missing and for others the comment might be not quite on point, but generally, the quality seemed to be quite high.

The reason I wasn't getting any documentation was due to the fact that the XML file was simply not generated and packed into the NuGet package. I've changed that, it's a pretty tiny change.

The changes are based on the last 4.0.1 release, so that a hotfix could easily be created for the NuGet package. However, there does not seem to be a 4.0 branch in GitHub that I can select for the pull request, so I've selected the master branch for now. It would obviously be ideal to have this change in both the master branch *and* a 4.0 version.

### DIFF
--- a/ICSharpCode.Decompiler/ICSharpCode.Decompiler.csproj
+++ b/ICSharpCode.Decompiler/ICSharpCode.Decompiler.csproj
@@ -42,6 +42,7 @@
     <DebugType>portable</DebugType>
     <DebugSymbols>true</DebugSymbols>
     <EmbedAllSources>true</EmbedAllSources>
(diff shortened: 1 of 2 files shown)


christophwille:
Thanks for prodding us to include the documentation XML. I have adapted your v4 PR to our current v5 (it is more than likely that we won't do any servicing releases for v4)

PR #1440, opened 2019-03-02
Replace invalid characters in identifiers

adrium:
As far as I understand, characters in identifiers are replaced depending on `IsPrintableIdentifierChar` to make the identifiers valid. However, they do not get valid by escaping the unicode characters.

This patch corrects `IsPrintableIdentifierChar` to adhere to the specification *9.4.2 Identifiers in ECMA-334* and replaces non-printable characters by a customizable replacement string.


dgrunwald:
This change is redundant; see the existing class `EscapeInvalidIdentifiers`.

PR #1464, opened 2019-03-24
Improve searching

MCpiroman:
- Improved performace of searching (though without parallelism, I've done some testing and it's not efficient).
- Search results are filtered using API visibility option - this is usefull e.g. if you want to find only methods that are public, so you can use them (I was in that situation few times and I think there was issue requesting that). Old behaviour was little odd, you could search for something but not navigate to it becouse of visibility option.


siegfriedpammer:
Thank you very much for your pull request! This is very useful.

I have added some minor changes.

LGTM

Issue #1470, opened 2019-03-29
Bring back option for LINQ method-call syntax


carlreinke:
I love LINQ, but I never use the LINQ syntax.  I don't even really know how to translate it to the method calls.

Of the decompiler options that were removed (#1102), this is the one that I miss.

Alternatively, having no option and always showing the method call syntax would be quite satisfactory. 😛

siegfriedpammer:
#1471 is merged

PR #1471, opened 2019-03-30
Rework C# decompiler options dialog.

### LINKED ISSUE #1102: Get back the decompilation options
I've just tested the latest beta (ILSpy 3.1 Beta 1) and noticed that you have stripped the decompilation options.
Before:
image

After:
image

This was a very useful feature. Sometimes I wanted to examine the internals of the particular state machine/iterator/expression tree and I was able to achieve that by simply disabling the decompilation feature. However, now there is no easy way to do that.

Is that technically possible to get back that decompilation options? If you believe that they pollute UI, we could move them somewhere separately.

Thanks.

siegfriedpammer:
grafik

This is an attempt to fix #1102 and #1470


siegfriedpammer:
@carlreinke @zvirja @zspitz @dgrunwald what do you think?

zspitz:
What happens when one setting depends on another unchecked setting? Is the unchecked state of the second setting simply ignored?

siegfriedpammer:
The dependant setting should be ignored, as stated in the settings dialog: LINQ expressions are only detected, iff static calls are transformed to extension method calls. Likewise non-trailing named arguments are only inserted, iff named arguments are enabled.

carlreinke:
Works for me!

ICSharpCode.Decompiler/DecompilerSettings.cs
```diff
@@ -772,7 +856,9 @@ public bool ArrayInitializers
 
 		/// <summary>
 		/// Gets/Sets whether tuple comparisons should be detected.
-		/// </summary>
+		/// </summary
```
carlreinke: Looks like you lost an angle bracket.

siegfriedpammer: Fixed, thanks!

siegfriedpammer:
After discussing the dialog with @dgrunwald we decided to merge the settings into one panel. We also decided to remove the 'Custom' language version and instead apply the selected options on all language versions. It works like this: We take all active options from the dialog minus all options from language versions higher than the selected language version.

I hope this makes things a bit simpler an easier to use. Feedback is welcome!

zvirja:
@siegfriedpammer Sorry for the late reply! Looks awesome, thanks for working on that!

PR #1476, opened 2019-04-04
Fix #1462: Inconsistent enum flag check

### LINKED ISSUE #1462: Inconsistent enum flag check
Decompilation of mscorlib System.Threading.Tasks.ThreadPoolTaskScheduler method QueueTask:
```
[SecurityCritical]
protected internal override void QueueTask(Task task)
{
	if ((task.Options & TaskCreationOptions.LongRunning) != 0)
	{
		Thread thread = new Thread(s_longRunningThreadWork);
		thread.IsBackground = true;
		thread.Start(task);
	}
	else
	{
		bool forceGlobal = (task.Options & TaskCreationOptions.PreferFairness) != TaskCreationOptions.None;
		ThreadPool.UnsafeQueueCustomWorkItem(task, forceGlobal);
	}
}
```
Check in `if` is printed as `!= 0`
but check in bool assigment is `!= TaskCreationOptions.None`
I'm not sure which way is better, but there should be only one.

mcpiroman:
Changes
`(enumValue & SomeEnum.SomeValue) == SomeEnum.None`
to
`(enumValue & SomeEnum.SomeValue) == 0`

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) 2014 Daniel Grunwald
+// Copyright (c) 2014 Daniel Grunwald
 // 
 // Permission is hereby granted, free of charge, to any person obtaining a copy of this
 // software and associated documentation files (the "Software"), to deal in the Software
@@ -699,8 +699,8 @@ namespace ICSharpCode.Decompiler.CSharp
 			}
 
 			// Special case comparisons with enum and char literals
-			left = AdjustConstantExpressionToType(left, right.Type);
-			right = AdjustConstantExpressionToType(right, left.Type);
+			left = TryUniteEqualityOperandType(left, right);
+			right = TryUniteEqualityOperandType(right, left);
 			
 			if (IsSpecialCasedReferenceComparisonWithNull(left, right)) {
 				// When comparing a string/delegate with null, the C# compiler generates a reference comparison.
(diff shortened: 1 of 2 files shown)


ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -763,6 +763,21 @@ TranslatedExpression TranslateCeq(Comp inst, out bool negateOutput)
 				.WithRR(rr);
 		}
 
+		TranslatedExpression TryUniteEqualityOperandType(TranslatedExpression operand, TranslatedExpression otherOperand)
```
siegfriedpammer: Using names like `left` and `right` instead of `operand` and `otherOperand` will improve readability of the code.

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -763,6 +763,21 @@ TranslatedExpression TranslateCeq(Comp inst, out bool negateOutput)
...
+		{
+			// Special case for enum flag check "(enum & EnumType.SomeValue) == 0"
+			// so that the const 0 value is printed as 0 integer and not as enum type, e.g. EnumType.None
+			if (operand.ResolveResult.IsCompileTimeConstant &&
+				operand.ResolveResult.Type.IsCSharpPrimitiveIntegerType() &&
+				(int)operand.ResolveResult.ConstantValue == 0 &&
```
siegfriedpammer: `operand.ResolveResult.ConstantValue` might be something else than a boxed `int`, so this can go wrong. This makes the unit tests fail. See https://ci.appveyor.com/project/icsharpcode/ilspy/build/job/wnhtrte5t1i7g5u9/tests

siegfriedpammer:
Thank you very much for providing the fix! I have added a few comments.

@dgrunwald can you please review this?

siegfriedpammer:
Thank you for providing this fix!

PR #1515, opened 2019-05-16
Fixed crash then type is null in type def matches.

dkucinskas:
Fixed crash then type is null in type def matches. Encountered while decompiling assembly and saving it as project.


dgrunwald:
Thanks for the fix!

By the way, it's a bit hard to judge a pull request like this without an associated bug.
This only started to make sense for me when I saw the stack trace in #1536 -- under normal circumstances, an `IType` should use `SpecialType.UnknownType` instead of `null`, so it usually would be the fault of the caller.

PR #1600, opened 2019-07-28
Support for throw expression

siegfriedpammer:


ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -441,7 +441,7 @@ protected internal override TranslatedExpression VisitLdcI4(LdcI4 inst, Translat
 					.WithILInstruction(inst);
 			} finally {
 				astBuilder.PrintIntegralValuesAsHex = false;
-			}
+		}
```
dgrunwald: Lots of weird indentation changes in this file.

ICSharpCode.Decompiler/CSharp/TranslatedExpression.cs
```diff
@@ -227,6 +227,9 @@ public TranslatedExpression ConvertTo(IType targetType, ExpressionBuilder expres
 					.WithRR(convAnnotation.ConversionResolveResult)
 					.ConvertTo(targetType, expressionBuilder, checkForOverflow, allowImplicitConversion);
 			}
+			if (Expression is ThrowExpression && allowImplicitConversion) {
+				return this; // Throw expressions have no type and are implicitly convertible to any type
```
dgrunwald: Returning `this` may be problematic: the `ResolveResult` annotation on `this` still has the old type, which may cause assertions due to stack type mismatches.

ICSharpCode.Decompiler/IL/Instructions.tt
```diff
@@ -265,7 +265,7 @@
 		new OpCode("default.value", "Returns the default value for a type.",
 			NoArguments, HasTypeOperand, ResultType("type.GetStackType()")),
 		new OpCode("throw", "Throws an exception.",
-			Unary, MayThrow, UnconditionalBranch),
+			Unary, MayThrow, HasFlag("InstructionFlags.EndPointUnreachable"), ResultType("this.resultType")),
```
dgrunwald: Why did we replace `UnconditionalBranch` with just `EndPointUnreachable` here?
The `rethrow` instruction still uses `UnconditionalBranch`, which seems inconsistent to me.

dgrunwald: `UnconditionalBranch = VoidResult + HasFlag("InstructionFlags.EndPointUnreachable")`, so this makes sense.

ICSharpCode.Decompiler/IL/Transforms/NullCoalescingTransform.cs
```diff
@@ -83,6 +82,20 @@ bool TransformRefTypes(Block block, int pos, StatementTransformContext context)
 				ILInlining.InlineOneIfPossible(block, pos, InliningOptions.None, context);
 				return true;
 			}
+			// stloc obj(valueInst)
+			// if (comp(ldloc obj == ldnull)) {
+			//		throw(...)
+			// }
+			// =>
+			// stloc obj(if.notnull(valueInst, throw(...)))
```
dgrunwald: This transform runs the risk of creating a dead store (though maybe not in practice, since `obj` would have already been inlined earlier in that case).
I'd add a check that `obj` has more than one use (load/address-taking).

ICSharpCode.Decompiler/CSharp/TranslatedExpression.cs
dgrunwald: This is actually OK, there's no requirement that `ConvertTo(targetType, allowImplicitConversion=true).Type.Equals(targetType)`.
(such a requirement only exists for `allowImplicitConversion=false`, and even then the types only have to be equal after type erasure)

ICSharpCode.Decompiler/Semantics/Conversion.cs
```diff
@@ -76,6 +76,11 @@ public static Conversion EnumerationConversion(bool isImplicit, bool isLifted)
 		/// </summary>
 		public static readonly Conversion TryCast = new BuiltinConversion(false, 9);
 
+		/// <summary>
+		/// C# 7 throw expression being converted to an arbitrary type.
+		/// </summary>
+		public static readonly Conversion ThrowExpressionConversion = new BuiltinConversion(true, 10);
```
dgrunwald: Merge conflict: ThrowExpressionConversion and ImplicitInterpolatedStringConversion both have number 10?

ICSharpCode.Decompiler/IL/Transforms/NullCoalescingTransform.cs
```diff
@@ -83,7 +87,173 @@ bool TransformRefTypes(Block block, int pos, StatementTransformContext context)
...
+				return false;
+			if (!MatchNullableCall(arg, NullableLiftingTransform.MatchHasValueCall))
+				return false;
+			var throwInstParent = throwInst.Parent;
+			var throwInstChildIndex = throwInst.ChildIndex;
+			var nullCoalescingWithThrow = new NullCoalescingInstruction(
+				NullCoalescingKind.NullableWithValueFallback,
+				stloc.Value,
+				throwInst);
+			var resultType = NullableType.GetUnderlyingType(v.Type).GetStackType();
```
dgrunwald: In the case of `stloc s(ldloca v)`, doesn't this always result in `resultType == Ref`?

PR #1608, opened 2019-08-03
LdVirtDelegate: Implement transformation of delegate construction with ldvirtftn

siegfriedpammer:


dgrunwald:
The reason why I asked yesterday where this code was: my commit e4b15256aba will need adjustment for this.

ICSharpCode.Decompiler/CSharp/CallBuilder.cs
```diff
@@ -1232,7 +1232,17 @@ TranslatedExpression HandleDelegateConstruction(CallInstruction inst)
...
+		}
+
+		internal TranslatedExpression Build(LdVirtDelegate inst)
+		{
+			return HandleDelegateConstruction(inst.Type, inst.Method, new ExpectedTargetDetails { CallOpCode = OpCode.CallVirt }, inst.Argument, inst);
+		}
+
+		TranslatedExpression HandleDelegateConstruction(IType delegateType, IMethod method, ExpectedTargetDetails expectedTargetDetails, ILInstruction thisArg, ILInstruction inst)
+		{
+			var invokeMethod = delegateType.GetDelegateInvokeMethod();
```
dgrunwald: If the invoke method is necessary, shouldn't we store it with the `LdVirtDelegate` instruction?
For `UnknownType`, it's not possible to retrieve the Invoke `FakeMethod` from the type.

dgrunwald: Ah but we need to go from ctor to invoke, so we have no chance to discover the signature of an UnknownType delegate.

ICSharpCode.Decompiler/IL/Transforms/ExpressionTransforms.cs
```diff
@@ -297,9 +297,35 @@ protected internal override void VisitNewObj(NewObj inst)
...
+		bool TransformDelegateCtorLdVirtFtnToLdVirtDelegate(NewObj inst, out LdVirtDelegate ldVirtDelegate)
+		{
+			ldVirtDelegate = null;
+			if (inst.Method.DeclaringType.Kind != TypeKind.Delegate)
+				return false;
+			if (inst.Arguments.Count != 2)
+				return false;
+			if (!(inst.Arguments[1] is LdVirtFtn ldVirtFtn))
+				return false;
+			if (!inst.Arguments[0].Match(ldVirtFtn.Argument).Success)
```
dgrunwald: It's wrong to transform `new Delegate(call SideEffect(), ldvirtftn TargetMethod(call SideEffect))`.
Generally, every `Match` call for deduplication needs to be combined with a `SemanticsHelper.IsPure` call.

PR #1641, opened 2019-08-14
Toolbar rework

christophwille:
Compare

tbold

with

tbnew

* Sort & collapse have moved to the Open part because they act on the tree and are related to the existing two commands.
* internal/all moved to the right of the comboboxes, and now all three states are reflected (today when none is checked you'd have to guess what it does)
* icons removed from menu items for pub only et al because checkmark would obscure icon on selected visibility anyways


dgrunwald:
Why are internal/all moved to the right? These also act on the tree.

christophwille:
> Why are internal/all moved to the right? These also act on the tree.

Agreed. My reasoning for this is as follows: language, then version, then visibility (left to right). That is why I grouped it this way, additionally: sort is used more often, that is why it moved to the left (same goes why Search remained where it is - easy to find as the first item from the right)

dgrunwald:
But language/version mostly acts on the decompiler view (on the right); whereas internal/all mostly acts on the tree (on the left), so shouldn't the visibility toolbar buttons be left of the language dropdown?
I can't see why you think "language first, then visibility" would make sense.

Search on the end makes sense, the search panel is also on the right :)

christophwille:
That was based on the idea of not creating too big a grouping of icons (I consider visibility being used infrequently, that was why I moved it there - away from icons that are used potentially frequently, and that's why Search remained at the far right to easily pick it out)

So which order based on logical grouping + frequency of use would you prefer? 

Aside: we need an icon for "public only" (because the current one strictly speaking is incorrect). However, I didn't find anything logical for public visibility, public surface area, or other synonyms. Ideas?

PR #1654, opened 2019-08-21
Rich Text Tooltips

siegfriedpammer:
image

Missing features:
* [x] Support for inheritdoc
* [x] Hyperlinking/cref
* [x] Semantic highlighting of types/members
* [x] Tooltip is not immediately closed after navigating to a new entity


siegfriedpammer:
@dgrunwald I am currently thinking about how to best add highlighting, i. e., `RichTextModel` to the `Language.GetTooltip` API. Do you have a suggestion? I don't think an out parameter is a good idea.

siegfriedpammer:
An updated screenshot:

image

yyjdelete:
The tooltip should also ignore members from unresolved assemblies.

Now get an 
```cs
System.NullReferenceException: 
   at ICSharpCode.ILSpy.TextView.DecompilerTextView.FlowDocumentTooltip..ctor(FlowDocument document)
   at ICSharpCode.ILSpy.TextView.DecompilerTextView.GenerateTooltip(ReferenceSegment segment)
   at ICSharpCode.ILSpy.TextView.DecompilerTextView.TextViewMouseHover(Object sender, MouseEventArgs e)
```

ICSharpCode.Decompiler/CSharp/OutputVisitor/CSharpAmbience.cs
```diff
@@ -190,6 +192,7 @@ TypeSystemAstBuilder CreateAstBuilder()
 		{
 			TypeSystemAstBuilder astBuilder = new TypeSystemAstBuilder();
 			astBuilder.AddTypeReferenceAnnotations = true;
+			astBuilder.AddResolveResultAnnotations = true;
```
dgrunwald: Do we still use the old type reference annotations anywhere, or did we completely switch to resolve results?

siegfriedpammer: I just checked on the master branch. Looks like we don't use them anymore.

Issue #1426, opened 2019-02-18
No debug symbol


DSGMBot115:
ILSpy version 4.0.0.4518-rc1
Save C# project no debug symbol

christophwille:
You mean when you save an assembly as C# project? Well, all we create is C# files, you'd have to compile that to get a PDB (or use the Generate PDB feature instead).

DSGMBot115:
> You mean when you save an assembly as C# project? Well, all we create is C# files, you'd have to compile that to get a PDB (or use the Generate PDB feature instead).

I save a class as C# files, it has debug symbol

christophwille:
Sorry I don't follow.

DSGMBot115:
> Sorry I don't follow.

when i save an assembly as C# project, method GetDebugInfoOrNull did not use

siegfriedpammer:
See https://github.com/icsharpcode/ILSpy/issues/1328#issuecomment-447651862

dgrunwald:
ILSpy has two features:
 * Decompile + save code as .csproj. You can then recompile this project to generate a new DLL with new debug symbols.
 * Generate .pdb file (with embedded sources) matching the existing DLL.

These are unrelated features; what exactly do you think should ILSpy do differently?
Please be specific about what you are doing and what kind of output you expect. Currently we have no idea which ILSpy feature you are talking about.

Issue #1328, opened 2018-11-27
Variables names from PDB file not used.


DSGMBot115:
no debug symble when save project.

siegfriedpammer:
Could you please tell us, what you were doing and which version of ILSpy you used? Thank you!

GMBot115:
> 您能否告诉我们，您在做什么以及您使用的ILSpy版本？谢谢！

master

GMBot115:
> Could you please tell us, what you were doing and which version of ILSpy you used? Thank you!

master

DSGMBot115:
> Could you please tell us, what you were doing and which version of ILSpy you used? Thank you!

master

DSGMBot115:
> > Could you please tell us, what you were doing and which version of ILSpy you used? Thank you!
> 
> master

4.0.0.4392-beta2

DSGMBot115:
> > Could you please tell us, what you were doing and which version of ILSpy you used? Thank you!
> 
> master

ILSpy 4.0.0.4392-beta2

DSGMBot115:
> Could you please tell us, what you were doing and which version of ILSpy you used? Thank you!

ILSpy 4.0.0.4392-beta2
no debug symble when save csproj

DSGMBot115:
> Could you please tell us, what you were doing and which version of ILSpy you used? Thank you!

ILSpy version is 4.0.0.4392-beta2.
when i save .csproj,no debug symble.

PR #1673, opened 2019-08-27
Save Code produces identical results as shown in UI

an-dr-eas-k:
### Current status:

- When decompiling an assembly and using the "Save Code" function to produce csproj or sln projects, only some default settings are applied, differing from those in the options dialog.
- Additionally, when using the powershell scripts to decompile code, default settings are hard coded.
- The `DebugInfoProvider` is lost when using "Save Code" from ILSpy or in the powershell cmdlets. 

### Enhancement from this pull request
- Decompiling a complete assembly to a csproj or sln structure respects the settings from ILspy dialog, and preserves the `DebugInfoProvider`.
- The powershell cmdlets either use the ilspy.xml settings in AppData directory, or allow specifying a custom xml config file.

### Addressed Issues
#1674
#1328
#1426


christophwille:
You are moving references around (package as well as project refs), deleting projects, and more. Eg adding Mono.Cecil to ICS.Decompiler is a complete no-go.

siegfriedpammer:
We would prefer to discuss the actual problem this pull request tries to fix in an issue. Could you please open an issue for this? As @christophwille already noted, moving types around should not be done without prior careful analysis.

Thank you very much!

ICSharpCode.Decompiler.PowerShell/GetDecompilerCmdlet.cs
```diff
@@ -15,20 +25,106 @@ public class GetDecompilerCmdlet : PSCmdlet
...
+
 		protected override void ProcessRecord()
 		{
 			string path = GetUnresolvedProviderPathFromPSPath(LiteralPath);
 
-			try {
-				var decompiler = new CSharpDecompiler(path, new DecompilerSettings() {
-					ThrowOnAssemblyResolveErrors = false
-				});
+			try
```
siegfriedpammer: This does not respect our code formatting style. Please make sure you follow the settings in .editorconfig.

siegfriedpammer:
As discussed in #1674, the changes proposed in this pull request unfortunately cannot be merged.

Issue #1544, opened 2019-06-12
Assembly column in Search


miloush:
I would welcome if the search result grid could have an assembly column. Since you could then sort by assembly, it would also benefit the tasks in #1175 and #828.

PR #1690, opened 2019-09-05
Add Assembly column to search results

miloush:
Implements #1544

image


miloush:
Right, well [the way](https://github.com/miloush/ILSpy/commit/8f89457ab2e647912d4a8ea6c9d76dc4133eadb5) I implemented resource search it looks for files inside .resources files too which then goes into the Location column:

image

It's more trickier with [assembly search](https://github.com/miloush/ILSpy/commit/2fa0262ed83bca38afa7b302810980445138b4fc), for which I just put the path in the Location column:

image

Any of that sounds reasonable?

sailro:
Ah the funny thing is that we implemented both the same "search for resources" feature we had two days apart:

Yours: https://github.com/miloush/ILSpy/commit/8f89457ab2e647912d4a8ea6c9d76dc4133eadb5
Mine: https://github.com/sailro/ILSpy/commit/ee839b571348854757ea54f756bd00246e66dcac

miloush:
Indeed. Interesting that you decided to split the search results and switch on their type while I decided to split the search strategy and take advantage of the switch in `JumpToReference` since nothing was actually reading the `SearchResult.Entity` strongly-typed. It made assembly search possible by just adding one extra strategy.

I don't mind what you go for, I just have almost 6000 assemblies loaded in the list and needed to search for a baml component mentioned in a source code, hence searching inside .resources files for was a big help for me.

siegfriedpammer:
So... which one should I merge first? :)

siegfriedpammer:
@miloush are we going to see a PR for https://github.com/miloush/ILSpy/commit/2fa0262ed83bca38afa7b302810980445138b4fc as well?

miloush:
Sure happy to create one but the implementation will differ depending on the result vs. strategy split. Is there a preference which way to go?

siegfriedpammer:

I think we can continue the discussion about these features on the other pull request.

PR #1694, opened 2019-09-09
Extend search to find resource files

sailro:
pr ilspy


siegfriedpammer:
I have merged the latest changes and #1690 into this branch.

Things that are missing:

* What should we display in the `Location` and `Assembly` columns in the case of resources?
* Do we want to display results from inside `.resources` files? Maybe decompile XAML as well?

Ideas/feedback welcome!

miloush:
I think there is a more important decision about the way we support future search strategies supporting non-`IEntity` results.

The approach @sailro took is to have a new result type returning `Resource`.
The approach I took in https://github.com/miloush/ILSpy/commit/8f89457ab2e647912d4a8ea6c9d76dc4133eadb5 is to keep single result type returning `object` and abstract a non-`IEntity` search strategy

I have not considered @sailro's idea when implementing mine, but I think I am still leaning towards favoring the strategy split because then adding search strategies for different types of things only requires to add a new strategy, while splitting results would mean with every new strategy, a new result type as well as type switch in `JumpToSelectedItem` will have to be provided.

The only advantage of separate results I can see is the strongly typed result, but no code in ILSpy currently needs the result strongly-typed. I open to be convinced otherwise though.

sailro:
My proposal is to close this PR, and use the work done by @miloush instead.

Indeed I made the choice of simplicity, but we'll be stuck if we want to search inside a resource file.

siegfriedpammer:
Yes, I think the idea proposed by @miloush is very good.

Looking forward to it!

@sailro and @miloush Thank you for your feedback and contributions!

miloush:
I am happy to clean up my solution and submit a PR, it just felt like a non-trivial refactoring to existing code so I wanted someone to say that's actually a reasonable way to go.

Issue #846, opened 2017-09-05
Search resources option


gtaylor1981:
When searching, the "search for" combobox should have an "resources" option.
This would search through embedded resource files for the given text.

My actual use case is working with XAML files, where I want to find a template which corresponds to a class. This is not possible via the Analyzer (under "Used By") nor does it appear in the search results.

miloush:
Should this be closed?

PR #1707, opened 2019-09-22
Resource search strategy

miloush:
Adds searching resources including .resource files and tree navigation. Because of recent changes to cache entity result icons, I combined the approaches of myself and @sailro, splitting both results and strategies.

image

### DIFF
--- a/ILSpy/ContextMenuEntry.cs
+++ b/ILSpy/ContextMenuEntry.cs
@@ -78,7 +78,7 @@ namespace ICSharpCode.ILSpy
 			if (textView != null)
 				reference = textView.GetReferenceSegmentAtMousePosition();
 			else if (listBox?.SelectedItem is SearchResult result)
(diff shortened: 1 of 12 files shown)


siegfriedpammer:
Wonderful! Thank you so much! Can you please do XAML search next? ;-)

miloush:
Thanks. As in decompiling BAML and doing text search on it?

siegfriedpammer:
Yes, decompiling BAML. This might be very slow... I think we'll have to figure out how to handle it efficiently. Checking for x:Key/x:Name/Name/x:Uid/TargetType attributes should be enough...

Do you think we could use the BAML decompiler for this purpose?

siegfriedpammer:
I think first we will have to define an extension point for search strategies... then the BamlDecompiler AddIn can add an implementation for BAML.

siegfriedpammer:
We could continue this discussion in #846.

PR #1708, opened 2019-09-22
Add assembly search strategy

miloush:
Searching assemblies by full name. Shows file path in Location column.

image

### DIFF
--- a/ICSharpCode.Decompiler/Metadata/MetadataExtensions.cs
+++ b/ICSharpCode.Decompiler/Metadata/MetadataExtensions.cs
@@ -49,7 +49,7 @@ namespace ICSharpCode.Decompiler.Metadata
 			return publicKeyTokenBytes.TakeLast(8).Reverse().ToHexString(8);
 		}
 
-		public static string GetFullAssemblyName(this MetadataReader reader)
+		public static string GetPublicKeyToken(this MetadataReader reader)
 		{
 			if (!reader.IsAssembly)
 				return string.Empty;
@@ -59,6 +59,15 @@ namespace ICSharpCode.Decompiler.Metadata
 				// AssemblyFlags.PublicKey does not apply to assembly definitions
 				publicKey = CalculatePublicKeyToken(asm.PublicKey, reader);
 			}
+			return publicKey;
+		}
+
+		public static string GetFullAssemblyName(this MetadataReader reader)
+		{
+			if (!reader.IsAssembly)
+				return string.Empty;
+			var asm = reader.GetAssemblyDefinition();
+			string publicKey = reader.GetPublicKeyToken();
 			return $"{reader.GetString(asm.Name)}, " +
 				$"Version={asm.Version}, " +
 				$"Culture={(asm.Culture.IsNil ? "neutral" : reader.GetString(asm.Culture))}, " +
--- a/ILSpy/ILSpy.csproj
+++ b/ILSpy/ILSpy.csproj
@@ -168,6 +168,7 @@
       <DependentUpon>Resources.resx</DependentUpon>
     </Compile>
     <Compile Include="Search\AbstractEntitySearchStrategy.cs" />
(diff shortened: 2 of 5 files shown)


siegfriedpammer:
Would it be useful to add the file name as an additional criterion? That is `asm.FullName.Contains OR asm.FileName.Contains`. Of course each assembly would only be displayed once in the result list.

miloush:
I flavoured the assembly search, you can now match name (`a:`), full name (`afn:`), file name (`af:`), culture (`ac:`), version (`av:`), and others (`apk:`, `aha:`, `afl:`). Does that address your concern, or were you looking for an actual OR? We could make `a:` to be assembly name or file name and `an:` name only.

Speaking of which, the tree view seems to show file names without extension rather than assembly names. In most cases they are the same but I got a few assemblies where the file name does not match the assembly name, the result being that the "Name" column in search results does not show the same label as the tree view for the same assembly. Was that a deliberate choice, or is that something worth aligning?

siegfriedpammer:
> I flavoured the assembly search, you can now match name (`a:`), full name (`afn:`), file name (`af:`), culture (`ac:`), version (`av:`), and others (`apk:`, `aha:`, `afl:`). Does that address your concern, or were you looking for an actual OR? We could make `a:` to be assembly name or file name and `an:` name only.

Wow... that are a lot of options... I thought that maybe `a:` should be equal to `assemblies.Where(a => a.FullName.Contains("<search-term>") || a.FileName.Contains("<search-term>")`. All the options might be overkill.

> Speaking of which, the tree view seems to show file names without extension rather than assembly names. In most cases they are the same but I got a few assemblies where the file name does not match the assembly name, the result being that the "Name" column in search results does not show the same label as the tree view for the same assembly. Was that a deliberate choice, or is that something worth aligning?

Yes, I think we should use the assembly name in both cases... The filename will be displayed in the tooltip anyway.

cc @christophwille @dgrunwald what do you think?

christophwille:
I am content with simply using "a:" to mean "wildcard-search in full asm name". I should be rather unusual that version string collides with culture or file name.

dgrunwald:
> Yes, I think we should use the assembly name in both cases... The filename will be displayed in the tooltip anyway.

For our own usage I'd prefer the file name, so that we can easily distinguish the variants of our test cases (`AssemblyName.dll`, `AssemblyName.opt.dll`, etc.).

siegfriedpammer:
> > Yes, I think we should use the assembly name in both cases... The filename will be displayed in the tooltip anyway.
> 
> For our own usage I'd prefer the file name, so that we can easily distinguish the variants of our test cases (`AssemblyName.dll`, `AssemblyName.opt.dll`, etc.).

Yes, you are right. I'd forgot about that...

miloush:
Okay so we are concluding that we want two options, full assembly name and file path?

siegfriedpammer:
> Okay so we are concluding that we want two options, full assembly name and file path?

Are you talking about two different search "prefixes"/options or display options in the tree view?

From what I gather from the other responses, the best route to take would be:

* keep the tree view as is,
* use only one search "prefix"/option: `a:` to search the full assembly name OR file name, as in: `assemblies.Where(a => a.FullName.Contains(searchTerm) || a.FileName.Contains(searchTerm))`,
* the other separate options for version, culture and pk-token are overkill.

Do you agree?

miloush:
Tree view is a separate issue.
I agree all the options are overkill, as they can be matched using regex on the full name.

I am a bit worried about the full path to be included by default. Many assemblies are typically under the same directory and we might get lots of false positives when searching e.g. for Microsoft.

I suggest a compromise:
`a:` matching assembly local name OR file name without path
`an:` matching assembly full name
`af:` matching assembly full path

siegfriedpammer:
Yes, I think this is a good compromise...

@dgrunwald are you OK with these changes?

siegfriedpammer:
@miloush Thank you for contributing this feature!

PR #1714, opened 2019-09-24
.NET Core 3 Final Updates

christophwille:
Note: won't build atm because AppVeyor still uses a VS 2019 image pre-16.3.


christophwille:
Broken builds on Azure Pipelines is an accepted temporary side effect.

PR #1730, opened 2019-09-29
Async streams

### LINKED ISSUE #1640: Lambda with dynamic arguments includes unnecessary cast
```csharp
private Action<dynamic, dynamic> GetDynamicAction()
{
	return delegate(dynamic a, dynamic b)
	{
		a.A(b);
	};
}
```
currently gets decompiled as:
```csharp
private Action<dynamic, dynamic> GetDynamicAction()
{
	return (Action<object, object>)delegate(dynamic a, dynamic b)
	{
		a.A(b);
	};
}
```

dgrunwald:
This branch adds support for:
 * `async IAsyncEnumerable<T>` methods
 * `await using`
 * Test cases can now use .NET Core 3
 * Fixed some of the bugs discovered when trying to run all tests with .NET Core 3 (mostly wrt. nullable reference types) -- but then we decided to make this a per-test opt-in, as there is still some work remaining in this area.
* Fixes #1640 and #1780

### DIFF
--- a/ICSharpCode.Decompiler/IL/ControlFlow/AsyncAwaitDecompiler.cs
+++ b/ICSharpCode.Decompiler/IL/ControlFlow/AsyncAwaitDecompiler.cs
@@ -579,6 +579,8 @@ namespace ICSharpCode.Decompiler.IL.ControlFlow
 			if (!IsAsyncEnumerator)
 				return false;
 			var block = blockContainer.Blocks.ElementAtOrDefault(pos);
+			if (block == null)
+				return false;
 			// call SetResult(ldflda <>v__promiseOfValueOrEnd(ldloc this), ldc.i4 1)
 			// leave IL_0000(nop)
 			if (block.Instructions.Count != 2)
@@ -658,8 +660,10 @@ namespace ICSharpCode.Decompiler.IL.ControlFlow
 			if (!handler.Filter.MatchLdcI4(1))
 				throw new SymbolicAnalysisFailedException();
 			var catchBlock = YieldReturnDecompiler.SingleBlock(handler.Body);
+			if (catchBlock == null)
+				throw new SymbolicAnalysisFailedException();
 			catchHandlerOffset = catchBlock.StartILOffset;
-			if (catchBlock?.Instructions.Count != 4)
+			if (catchBlock.Instructions.Count != 4)
 				throw new SymbolicAnalysisFailedException();
 			// stloc exception(ldloc E_143)
 			if (!(catchBlock.Instructions[0] is StLoc stloc))
--- a/ICSharpCode.Decompiler/IL/ControlFlow/AsyncAwaitDecompiler.cs
+++ b/ICSharpCode.Decompiler/IL/ControlFlow/AsyncAwaitDecompiler.cs
@@ -820,7 +820,7 @@ namespace ICSharpCode.Decompiler.IL.ControlFlow
 		/// </summary>
 		void AnalyzeStateMachine(ILFunction function)
 		{
-			context.Step("AnalyzeStateMachine()", function);
+			context.StepStartGroup("AnalyzeStateMachine()", function);
 			smallestAwaiterVarIndex = int.MaxValue;
 			foreach (var container in function.Descendants.OfType<BlockContainer>()) {
 				// Use a separate state range analysis per container.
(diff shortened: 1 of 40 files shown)


dgrunwald:
Screenshot of current status:
image
Note that there's a lot of stuff already happening in this example:
 * The state machine with a mixture of `await` and `yield return` yield points is turned back to control flow
 * Catch+rethrow is turned back into `try-finally` with `await asyncEnumerator.DisposeAsync();`
 * `try-finally` is converted into `await using`
And only the last step reconstructing the `await foreach` loop is missing.

dgrunwald:
This feature could use more tests; but otherwise I think it's ready to be merged.
`await foreach` can be left for a future PR.

ICSharpCode.Decompiler/IL/ControlFlow/AsyncAwaitDecompiler.cs
```diff
@@ -705,6 +991,49 @@ static ILInstruction StackSlotValue(ILInstruction inst)
...
+					return false;
+				}
+			}
+
+			while (pos >= 0 && block.Instructions[pos] is StLoc stloc) {
+				if (stloc.Variable.Kind != VariableKind.StackSlot)
+					return false;
+				if (!SemanticHelper.IsPure(stloc.Value.Flags))
+					return false;
+			}
```
yyjdelete: Missing an `pos --;` here

yyjdelete:
Some others limited. Attached below, incase also need them. 

1. `yield break` in `async IAsyncEnumerable<T>` is not supported;
> Should not see "/TRY/END" and "/END" when Run with `Test(0)`
```cs
        async IAsyncEnumerable<int> Test(int i = 0)
        {
            try
            {
                Console.WriteLine("/TRY");
                try
                {
                    Console.WriteLine("/TRY/TRY");
                    if (i == 0)
                        yield break;
                }
                finally
                {
                    Console.WriteLine("/TRY/FINALLY");
                }
                Console.WriteLine("/TRY/END");
            }
            finally
            {
                Console.WriteLine("/FINALLY");
            }
            Console.WriteLine("/END");
        }
```
2. `EnumeratorCancellationAttribute` is not supported.
> See `combinedTokens` in the CompilerGenerated class, it break the check for SetResult and others.
```cs
        public static async IAsyncEnumerable<int> MainXZ2([EnumeratorCancellation] CancellationToken ct = default)
        {
            while (true) { ct.ThrowIfCancellationRequested(); }
            yield break;
        }
```

3. ICSharpCode.ILSpy.AssertionFailedException: 'Should not insert using declaration for namespace that is missing from the superset: System.Runtime.ExceptionServices'
> The same code with 6, it doesn't happen for Task, the `when (i > 0) is not needed for 5.
4. (Also for Task)`OpCode not supported: BlockContainer` for `when()`
> The `when (i > 0)` is needed for 6, and this also happen for Task.
> Is it possible to still show the code for block, just with some warn here? Also for some other thing unsupported like `when (ex is XXException xxEx ? xxEx.A && xxEx.B : false)` 
=> 
```cs
when(/*OpCode not supported: BlockContainer*/{
    var xxEx = ex as XXException;
    if (xxEx != null)
    {
        return xxEx.A && xxEx.B;
    }
    else
    {
        return false;
    }
})
```

```cs
        public static async IAsyncEnumerable<int> TestTethrow()
        {
            int i = 100;
            try
            {
                await Task.Yield();
            }
            catch (Exception ex) when (i > 0)
            {
                if (i == 100)
                    throw;
                await Task.Yield();
            }
            yield break;
        }
```

yyjdelete:
Not sure, but the logic of `async IAsyncEnumerable<T>`  may changed after `https://github.com/dotnet/roslyn/pull/39436` is applied.

ICSharpCode.Decompiler/IL/ControlFlow/AsyncAwaitDecompiler.cs
```diff
@@ -385,12 +552,49 @@ void AnalyzeMoveNext()
...
+		private bool IsAsyncEnumerator => methodType == AsyncMethodType.AsyncEnumerable || methodType == AsyncMethodType.AsyncEnumerator;
+
+		bool MatchYieldBlock(BlockContainer blockContainer, int pos)
+		{
+			if (!IsAsyncEnumerator)
+				return false;
+			var block = blockContainer.Blocks.ElementAtOrDefault(pos);
+			// call SetResult(ldflda <>v__promiseOfValueOrEnd(ldloc this), ldc.i4 1)
+			// leave IL_0000(nop)
+			if (block.Instructions.Count != 2)
```
yyjdelete: null check for `block`

yyjdelete:
https://github.com/icsharpcode/ILSpy/blob/233f33f1979f71097abdb4825911e4d377cf8142/ICSharpCode.Decompiler/IL/ControlFlow/AsyncAwaitDecompiler.cs#L660-L663
`catchBlock` is used before null check.

dgrunwald:
Thanks for the review. I've fixed the missing null checks.

`yield break`: The `yield break;` itself is fixed; but some `if (stateVar == -1)` checks in finally blocks remain.

EnumeratorCancellationAttribute: I think we'll handle this later; I've created #1772 for this feature.

dotnet/roslyn#39436: I'll leave this for later, when the Roslyn change is released and we can update our NuGet references to run our tests with the new compiler.

Issue #1732, opened 2019-10-01
Azure Pipelines Randomly Failing


christophwille:
E.g. https://icsharpcode.visualstudio.com/icsharpcode-pipelines/_build/results?buildId=323

Attempt 2 succeeded (a simple re-run)
Attempt 1 was the initial failed run

This random failure behavior started with the latest Pipelines images (the one with VS 16.2 and us installing the latest .NET 3 SDK previews was stable). The new 16.3 image exhibits this behavior.

Log downloaded and attached: [log_23_323.zip](https://github.com/icsharpcode/ILSpy/files/3675194/log_23_323.zip)

christophwille:
This one has a different mode of failure https://icsharpcode.visualstudio.com/icsharpcode-pipelines/_build/results?buildId=329 (Attempt 2)

Log downloaded and attached: [log_58_329.zip](https://github.com/icsharpcode/ILSpy/files/3675527/log_58_329.zip)

christophwille:
maximumCpuCount as pointed out in the PR discussion fixes the problems we are seeing. That is good enough for us.

PR #1748, opened 2019-10-09
Test /bl for Azure Pipelines build

christophwille:
See #1732 

Infos http://msbuildlog.com/

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -37,9 +37,12 @@ jobs:
   - checkout: self
     submodules: recursive
 
-  - task: DotNetCoreInstaller@0
+  - task: UseDotNet@2
+    displayName: 'Install .net core 3.0'
     inputs:
+      packageType: sdk
       version: '3.0.100'
+      installationPath: $(Agent.ToolsDirectory)/dotnet
 
   - powershell: .\BuildTools\pipelines-install.ps1
     displayName: Install
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -20,22 +20,18 @@ jobs:
         BuildConfiguration: Release
         ReleaseChannel: Zip
         Solution: ILSpy.sln
-        SolutionToBuild: ILSpy.sln
       Config_Debug_Zip:
         BuildConfiguration: Debug
         ReleaseChannel: Zip
         Solution: ILSpy.sln
-        SolutionToBuild: ILSpy.sln
       Config_Release_CI:
         BuildConfiguration: Release
         ReleaseChannel: CI
         Solution: ILSpy.WithPackage.sln
-        SolutionToBuild: ILSpy.WithPackage.sln
       Config_Release_Store:
         BuildConfiguration: Release
         ReleaseChannel: Store
         Solution: ILSpy.WithPackage.sln
-        SolutionToBuild: .\ILSpy.Package\ILSpy.Package.wapproj
 
   steps:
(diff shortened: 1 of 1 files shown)


christophwille:
Attached is the build log (zipped binlog) for the failed Release CI (original build at https://icsharpcode.visualstudio.com/icsharpcode-pipelines/_build/results?buildId=365)

[msbuild_CI_Release.zip](https://github.com/icsharpcode/ILSpy/files/3708460/msbuild_CI_Release.zip)

onovotny:
This is weird. I'd suggest @nguerrera or @dsplaisted might be able to help?

dsplaisted:
@rainersigwald Any ideas on what's going on here?

> ALINK error AL1019: Metadata failure while creating assembly -- The process cannot access the file because it is being used by another process. [d:\a\1\s\ILSpy\ILSpy.csproj]

This might be relevant:

```xml
    <EmbeddedResource Include="Properties\Resources.resx">
      <Generator>PublicResXFileCodeGenerator</Generator>
      <LastGenOutput>Resources.Designer.cs</LastGenOutput>
    </EmbeddedResource>
    <EmbeddedResource Include="Properties\Resources.zh-Hans.resx" />
```

onovotny:
@dsplaisted what's weird is that it works in three of the four configurations?

rainersigwald:
From the log, `ILSpy.csproj` is being built twice in parallel:
1. `ILSpy.WithPackage.sln` -> `ILSpy.csproj` (default targets)
1. `ILSpy.WithPackage.sln` -> `ILSpy.Package.wapproj` -> `ILSpy.csproj` (`Build` target, additional properties `_WapBuildingInsideVisualStudio = ; GenerateAppxPackageOnBuild = false; BuildingInsideVisualStudio = false`)

When both of those try to read/write the same file, one fails, producing this error.

I'm not sure whether this is a bug in ILSpy's build or in `Microsoft.DesktopBridge.targets`, but I strongly suspect the latter.

You can work around this by building single-threaded (not using `msbuild -m`/the `maximumCpuCount` argument to the MSBuild AzDO task), by not building the `ILSpy.WithPackage` solution, or by building `ILSpy.Package.wapproj` _instead of_ the solution. It looks like your current AppVeyor configuration might do both of the first two. I would recommend the latter, but I'm not sure how that would affect running tests in this repo.

christophwille:
Another interesting tidbit is that the 5.0x branch never had build failures (see checkmarks on commits https://github.com/icsharpcode/ILSpy/commits/5.0.x except for real build breaks), the only difference is that 5.0x builds for net462 while master is net472.

christophwille:
@rainersigwald removing maximumCpuCount seems to have done the trick (was the least invasive change to the build process). Still, strange why one branch is fine, while the other fails regularly.

dgrunwald:
In addition to the net462 -> net472 change, master also upgraded MSBuild.Sdk.Extras from 2.0.24 to 2.0.54 and has the 5388a18c4622 change.

christophwille:
I have applied the `maximumCpuCount` fix to master (got complaints already that the green checkmarks on commits aren't there although we build successfully on AppVeyor). 

Still, I want to get to the bottom of this - why did it start failing.

christophwille:
We are sticking with the `maximumCpuCount` workaround and won't try to find/fix the underlying issue on Azure DevOps / Visual Studio / msbuild.

PR #1754, opened 2019-10-20
Fix NullReferenceException when foreach decompilation is turned off

davidwengier:
When decompiling a foreach loop on an IEnumerable, if "Detect foreach statements" is turned off, there is a NullReferenceException thrown in StatementBuilder.

The `resource` parameter is set to null here:
https://github.com/icsharpcode/ILSpy/blob/master/ICSharpCode.Decompiler/CSharp/StatementBuilder.cs#L452

And then deferenced at the `resource.Detach()` call here:
https://github.com/icsharpcode/ILSpy/blob/master/ICSharpCode.Decompiler/CSharp/StatementBuilder.cs#L423

This simple change fixes the issue and decompilation works as expected.

Before:
image

After:
image

I couldn't find any existing test infrastructure for validating settings, but if a test is required and there is documentation to do so, please let me know. It doesn't seem like the existing test infrastructure is well suited to variations in the decompiled output.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
+++ b/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
@@ -83,6 +83,8 @@
     <Compile Include="TestCases\ILPretty\ConstantBlobs.cs" />
     <Compile Include="TestCases\Pretty\OutVariables.cs" />
     <Compile Include="TestCases\Pretty\CustomTaskType.cs" />
+    <None Include="TestCases\Ugly\NoForEachStatement.Expected.cs" />
+    <Compile Include="TestCases\Ugly\NoForEachStatement.cs" />
     <None Include="TestCases\Ugly\NoExtensionMethods.Expected.cs" />
     <Compile Include="TestCases\Ugly\NoExtensionMethods.cs" />
     <None Include="TestCases\VBPretty\VBCompoundAssign.cs" />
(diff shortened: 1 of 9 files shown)


siegfriedpammer:
Thank you for finding and fixing this issue!

Could you please post the original C# code? Looking at the code in the "after" screenshot, it seems the `GetEnumerator` call is missing.

> I couldn't find any existing test infrastructure for validating settings, but if a test is required and there is documentation to do so, please let me know. It doesn't seem like the existing test infrastructure is well suited to variations in the decompiled output.

You are right, that the test infrastructure is not optimal for testing a lot of different settings. The closest we have so far are the "ugly" tests in `UglyTestRunner.cs`. See the [wiki](https://github.com/icsharpcode/ILSpy/wiki/Unit-Tests#ugly-tests) for more information.

If you need anything else, please ask! Thank you!

siegfriedpammer:
@dgrunwald I think this might cause conflicts with #1730.

davidwengier:
> Could you please post the original C# code? Looking at the code in the "after" screenshot, it seems the GetEnumerator call is missing.

You're absolutely right, I should have caught that! I've pushed an update the fixes the issue.

The original code is simply:

```
using System;
using System.Collections;

namespace ConsoleApp114
{
    class Program
    {
        {
    		IEnumerable enumerable = (IEnumerable)args;
            foreach (var arg in enumerable)
            {
            }
        }
    }
}
```

With the new changes the decompilation now looks like this:
image

siegfriedpammer:
Thank you again for providing this!

Issue #1749, opened 2019-10-10
Some edge case when await is used in catch


yyjdelete:
Just see this when do some test, not sure if it affacted some real-world assembly.

1. It show `when (/*OpCode not supported: BlockContainer*/)` for catch with both await and filter;

2. When all catch with filter are removed(just an catch(NullReferenceException) and an catch-all ls left), the `rethrow` in catch-all reference an unreachable var ex2;

3. It's it possible to still show the code even the filter is an `BlockContainer`, just with some warns? It can also be caused by some unsupported langversion or something else.
`when (ex.InnerException?.HResult == 0)` => 
```cs
when (/*OpCode not supported: BlockContainer*/((Func<bool>)(()=>
{
	var tmp = ex.InnerException;
	return tmp != null && tmp.HResult == 0;
})).Invoke())
```
 or 
`when (ex is AggregateException ae && ae.Message != null && ae.Message != null)` =>
```cs
when (/*OpCode not supported: BlockContainer*/((Func<bool>)(()=>
{
	var ae = ex as AggregateException;
	if (ae != null)
	{
		return ae.Message != null && ae.Message != null;
	}
	return false;
})).Invoke())
```

Full test code:(The catch blocks can also be tested one by one, just attach them all as an single method to make the code simple)
```cs
        public async Task ZZZ_all(int i)
        {
            try
            {
                throw null;
            }
            catch (Exception ex) when (i == 0)
            {
                if (i == 1)
                    throw;
                await Task.Yield();
                Console.WriteLine(ex.StackTrace);
            }
            catch (Exception ex) when (true)
            {
                if (i == 1)
                    throw;
                await Task.Yield();
                Console.WriteLine(ex.StackTrace);
            }
            catch (Exception ex)
            {
                if (i == 1)
                    throw;
                await Task.Yield();
                Console.WriteLine(ex.StackTrace);
            }
            catch when (i == 0)
            {
                if (i == 1)
                    throw;
                await Task.Yield();
            }
            catch when (true)
            {
                if (i == 1)
                    throw;
                await Task.Yield();
            }
            catch
            {
                if (i == 1)
                    throw;
                await Task.Yield();
            }
        }
```

PR #1791, opened 2019-11-06
#1749: Represent multi-instruction BlockContainers as inlined delegate invocation.

siegfriedpammer:
@dgrunwald Please review my changes to `DecompiledLambdaResolveResult`.

An alternative solution might be to create an implementation of `LambdaResolveResult` that does not use `ILFunction` at all. This might have been a better idea...

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
@@ -67,6 +67,7 @@ namespace ICSharpCode.Decompiler.CSharp
 	/// </remarks>
 	sealed class ExpressionBuilder : ILVisitor<TranslationContext, TranslatedExpression>
 	{
+		readonly StatementBuilder statementBuilder;
 		readonly IDecompilerTypeSystem typeSystem;
 		internal readonly ITypeResolveContext decompilationContext;
 		internal readonly ILFunction currentFunction;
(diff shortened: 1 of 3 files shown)


ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -1940,6 +1942,44 @@ internal IEnumerable<ParameterDeclaration> MakeParameters(IReadOnlyList<IParamet
...
+			statementBuilder.currentResultType = context.TypeHint;
+			statementBuilder.currentIsIterator = false;
+			try {
+				var body = statementBuilder.ConvertAsBlock(container);
+				body.InsertChildAfter(null, new Comment(" Could not convert BlockContainer to single expression"), Roles.Comment);
+				var ame = new AnonymousMethodExpression();
+				ame.Body = body;
+				var inferredReturnType = InferReturnType(body);
+				var func = compilation.FindType(typeof(Func<>)).GetDefinition();
+				var delegateType = new ParameterizedType(func, inferredReturnType);
```
dgrunwald: This will crash if `GetDefinition()` returns null.

dgrunwald:
The whole stuff with `DecompiledLambdaResolveResult` seems like massive overkill -- the lambda is not directly returned from `Visit*`, so there's no real need to use a resolve result here.
It's not like the cast can ever be made optional in this case.

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
siegfriedpammer: Fixed

siegfriedpammer:
Yes, this makes sense, I have removed the ConversionResolveResult as well, now I am just using `new ResolveResult(delegateType)` for the invocation target.

PR #1797, opened 2019-11-15
Add support for generic and static local function

### LINKED ISSUE #1588: Support generic local functions
The support for local functions implemented in #1586 does not cover generic local functions.

A first batch of test cases:

```csharp
class Program<CT>
{
	public static void Main(string[] args)
	{
		A(args);

		int A<T>(T a)
		{
			return 1;
		}
	}

	public static MT Get<MT>(string key, MT defaultValue)
	{
		return Do();

		MT Do()
		{
			Console.WriteLine(key);
			return defaultValue;
		}
	}

	public static CT TypeParameter()
	{
		return A();

		CT A()
		{
			return default(CT);
		}
	}
}
```

yyjdelete:
Try add basic support for generic local function(fix #1588) and static local function(C#8.0), and fix some other issue found for local function.

Limited and Known Issue(not fixed):
1. Due to some [break changes](https://github.com/dotnet/roslyn/blob/master/docs/compilers/CSharp/Compiler%20Breaking%20Changes%20-%20post%20VS2019.md) and some other issues fixed in roslyn 16.4, `static local function` is incompatible with old versions of roslyn.
```cs
    //It compile on 16.3 as non-static, and fixed in 16.4. The generic version still compile on 16.4 and already reported as roslyn#39706
    public void Roslyn_LE_16_3() {
        int i = 0;
        ZZ1();
        int ZZ() => i;
        static void ZZ1(){
            ZZ();
        }
    }
    //It's able to compile this on 16.4+ but not 16.3
    public void Roslyn_GE_16_4() {
        ZZ1();
        static void ZZ(){}
        static void ZZ1(){
            _ = (Action)ZZ;
        }
    }
    //The method of local function is static in 16.4(See break changes) and non-static in 16.3. The latter one is not supported by the PR.
    public void M() {
        _ = new Action(XX0);
        static void XX0(){
        }
    }
```
2. Nested local function without capture generated the same code as no capture and not nested ones(so all static local function will be moved to the top function), and may cause conflict if the orig code use the same generic name or method name for them.
```cs
public void M()
{
	int i = 0;
	Method1<int>(1);
	void Method1<T1>(int z)
	{
		Method1<T1>();
		void Method1<T1>()
		{
			i = 0;
		}
	}
}
```
->
```cs
public void M()
{
	int i = 0;
	Method1<int>(1);
	void Method1<T1>(int z)
	{
		Method1_1<T1, T1>();
	}
	void Method1<T1, T1>()
	{
		i = 0;
	}
}
```
3. Generic type args of method will not be hidden even if possible.(and may lead to error for Anonymous Types)
```cs
        public void M1()
        {
            Method<int>();
            void Method<T1>()
            {
                int n = 0;
                Method2(new { A = 1 });
                Method3(new { A = 1 });
                void Method2<T2>(T2 obj)
                {
                    //keep nested
                    n = 0;
                }
                void Method3<T2>(T2 obj)
                {
                }
            }
        }
```

4. Assert in LdLoc(#1798)

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/CallBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/CallBuilder.cs
@@ -197,7 +197,7 @@ namespace ICSharpCode.Decompiler.CSharp
 			} else if (localFunction != null) {
 				var ide = new IdentifierExpression(localFunction.Name);
 				if (method.TypeArguments.Count > 0) {
-					int skipCount = localFunction.ReducedMethod.NumberOfCompilerGeneratedGenerics;
+					int skipCount = localFunction.ReducedMethod.NumberOfCompilerGeneratedTypeParameters;
 					ide.TypeArguments.AddRange(method.TypeArguments.Skip(skipCount).Select(expressionBuilder.ConvertType));
 				}
 				target = ide.WithoutILInstruction()
@@ -1401,7 +1401,7 @@ namespace ICSharpCode.Decompiler.CSharp
 				if ((step & 2) != 0) {
 					int skipCount = 0;
 					if (localFunction != null && method.TypeArguments.Count > 0) {
-						skipCount = localFunction.ReducedMethod.NumberOfCompilerGeneratedGenerics;
+						skipCount = localFunction.ReducedMethod.NumberOfCompilerGeneratedTypeParameters;
 					}
 					ide.TypeArguments.AddRange(method.TypeArguments.Skip(skipCount).Select(expressionBuilder.ConvertType));
 				}
--- a/ICSharpCode.Decompiler/CSharp/CallBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/CallBuilder.cs
@@ -1462,7 +1462,7 @@ namespace ICSharpCode.Decompiler.CSharp
 						method.DeclaringType,
 						new IParameterizedMember[] { method }
 					)
-				}, method.TypeArguments
+				}, method.TypeArguments.Skip(localFunction.ReducedMethod.NumberOfCompilerGeneratedTypeParameters).ToArray()
 			);
 		}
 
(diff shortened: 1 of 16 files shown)


ICSharpCode.Decompiler/CSharp/CallBuilder.cs
```diff
@@ -1351,11 +1359,12 @@ ExpressionWithResolveResult BuildDelegateReference(IMethod method, IMethod invok
...
 					|| (method.IsStatic ? !expressionBuilder.IsCurrentOrContainingType(method.DeclaringTypeDefinition) : !(target.Expression is ThisReferenceExpression));
-
+				step = requireTarget ? 1 : 0;
 				var savedTarget = target;
-				for (step = requireTarget ? 1 : 0; step < 7; step++) {
+				for (; step < 7; step++) {
 					ResolveResult targetResolveResult;
-					if (!method.IsLocalFunction && (step & 1) != 0) {
+					//TODO: why there is an check for IsLocalFunction here, it should be unreachable in old code
+					if (localFunction == null && (step & 1) != 0) {
```
yyjdelete: Looks strange but it's already unreachable in old code.

ICSharpCode.Decompiler/IL/Transforms/LocalFunctionDecompiler.cs
```diff
@@ -65,6 +65,8 @@ public void Run(ILFunction function, ILTransformContext context)
 		{
 			if (!context.Settings.LocalFunctions)
 				return;
+			if (IsLocalFunctionMethod(function.Method, context) || IsLocalFunctionDisplayClass(function.Method.ParentModule.PEFile, (TypeDefinitionHandle)function.Method.DeclaringTypeDefinition.MetadataToken, context))
+				return;
```
yyjdelete: Disable this when current function is local func or inside local func display class, since it's not always possible to restore nested local function struct in non-top scope.
It's not possible to restore the struct when current function is `Method2`.
```cs
void Method() {
	Method1<int>();
	void Method1<T1>() {
		int i = 0;
		Method2<int>();
		void Method2<T2>() {
			i = 0;
			Method1<int>();
		}
	}
}
```

siegfriedpammer:
@yyjdelete can this be reviewed or are you still working on this?

yyjdelete:
@siegfriedpammer 
It can be reviewed. 
Due to my poor English, there is no comments in new code, maybe can see the new tests as a refer.

siegfriedpammer:
Thank you very much for adding this feature to the decompiler!

Sorry for taking so long to take a look.

PR #1820, opened 2019-11-26
Fixes the decompilation crash when few versions of the same library referred.

pvasilek:


siegfriedpammer:
Thank you for providing this pull request! Could you please provide an example of a binary that causes the bug you fixed? I would like to verify the bugfix on my end. Thank you!

pvasilek:
Unfortunately, it is a private assembly but the .deps.json contains following entries for "libraries"
    "Microsoft.Extensions.Configuration/2.0.0": {
      "type": "package",
      "serviceable": true,
      "sha512": "sha512-SsI4RqI8EH00+cYO96tbftlh87sNUv1eeyuBU1XZdQkG0RrHAOjWgl7P0FoLeTSMXJpOnfweeOWj2d1/5H3FxA==",
      "path": "microsoft.extensions.configuration/2.0.0",
      "hashPath": "microsoft.extensions.configuration.2.0.0.nupkg.sha512"
    },
...
    "Microsoft.Extensions.Configuration/2.1.0-preview1-final": {
      "type": "package",
      "serviceable": true,
      "sha512": "sha512-VST9w4+xLD24+peRpTfD4nVHOQ3dyLP+ZDo3ep9fW/wLF2hMAfmwuAl1XNsXB7z4u8rAcRRXwUfVJ2r/4iAmvA==",
      "path": "microsoft.extensions.configuration/2.1.0-preview1-final",
      "hashPath": "microsoft.extensions.configuration.2.1.0-preview1-final.nupkg.sha512"
    },

siegfriedpammer:
would it be possible for you to send it via email to siegfried.pammer@gmail.com?

siegfriedpammer:
I took a look at the problem and it seems that the correct fix is to turn the dictionary into an array, because the lookup by key was never used actually, making dictionary unnecessary and thus avoiding the duplicate key exception in general.

Thank you for providing the assembly!

Issue #901, opened 2017-10-07
ILReader gets implicit conversions subtly wrong


dgrunwald:
Ecma-335 Partition III 1.6 Implicit argument coercion, Table 9: Signature Matching shows the possible implicit conversions possible in IL.

Issues with our current implementation:
* We don't distinguish between int32->native int and int32->native unsigned int, thus potentially mixing up sign extension/zero extension in 64-bit executables.
* ~~We don't explicitly model starting/stopping GC tracking. This might cause assertions in the decompiler due to mismatched types.~~ Fixed in 9a829a28200bcb1e6aa66800b4a379e0f2968a4d.
* It's unclear what happens when one branch of an if pushes int32 on the stack, and the other branch pushes native int on the stack. Where exactly does the conversion happen? ~~Currently we emit a warning and an inconsistent ILAst, which then causes assertions due to mismatched types. (e.g. in the compiler-generated `_onexit_m` method in C++/CLI console apps)~~ We have a workaround that tends to work for C++/CLI-generated code, but isn't correct in general.

dgrunwald:
We should test with programs like this:
```
  ldarg condition
  brtrue use_native_int

  ldc.i4.m1
  br after_if
use_native_int:
  ldc.i4.m1
  conv.u
after_if:
  ldc.i4.1
  add.ovf.un
  ret
```

Does the addition always overflow when use_native_int==false? Or only on 32-bit systems?
What if the `use_native_int` block is moved after the `add` instruction? The spec claims that stack types can be inferred in a single pass, does this mean that the behavior of `add.ovf` differs depending on where the `after_if` block is placed?

dgrunwald:
```
  .method public static native int Int32OrNative(int32 val, bool use_native)
	{
    ldarg.1
    brtrue use_native_int
use_i4:
    ldarg.0
    br after_if
after_if:
    ldc.i4.1
    add
    ret
use_native_int:
    ldarg.0
    conv.u
    br after_if
  }
Result (64-bit):
Int32OrNative(0x7fffffff, false) = 2147483648
Int32OrNative(0x7fffffff, true) = 2147483648
Int32OrNative(-1, false) = 0
Int32OrNative(-1, true) = 4294967296
```

But if I comment out the `conv.u`, then I get:
```
Int32OrNative(0x7fffffff, false) = -2147483648
Int32OrNative(0x7fffffff, true) = -2147483648
Int32OrNative(-1, false) = 0
[...]

PR #1825, opened 2019-11-27
Extra optimizations

pvasilek:
Draft proposal for extra decompiler optimizations: universe case, dead IFs.

### DIFF
--- a/ICSharpCode.Decompiler/IL/ControlFlow/ControlFlowSimplification.cs
+++ b/ICSharpCode.Decompiler/IL/ControlFlow/ControlFlowSimplification.cs
@@ -55,6 +55,7 @@ namespace ICSharpCode.Decompiler.IL.ControlFlow
 			}
 			SimplifyBranchChains(function, context);
 			CleanUpEmptyBlocks(function, context);
+			RemoveDeadIfs(function, context);
 		}
 
 		private static void RemoveNopInstructions(Block block)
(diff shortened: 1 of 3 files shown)


pvasilek:
Replaces
	switch (...) {
		case [long.MinValue..long.MaxValue]: { CODE; }
	}
with
CODE;

Also, draft for
 if (...) nop; [else nop;] -> nop;
 if (...) { x; } x; -> x;
 if (...) { x; } else { x; } -> x;
But incomplete, need someones review and fix.

ICSharpCode.Decompiler/IL/ControlFlow/ControlFlowSimplification.cs
```diff
@@ -101,6 +102,40 @@ bool IsSimple(ILInstruction inst)
...
+			foreach (var block in function.Descendants.OfType<Block>()) {
+				for (int i = block.Instructions.Count - 1; i >= 0; i--) {
+					if (block.Instructions[i] is IfInstruction ifInst) {
+						if (IsEmpty(ifInst.TrueInst) && IsEmpty(ifInst.FalseInst)) {
+							block.Instructions[i].ReplaceWith(new Nop().WithILRange(ifInst));
+						} else {
+							var t = ifInst.TrueInst;
+							var f = ifInst.FalseInst;
+							var n = i < block.Instructions.Count - 1 ? block.Instructions[i + 1] : null;
+							if (t.ToString().CompareTo(f.ToString()) == 0) {
```
dgrunwald: `ILInstruction.ToString()` is only meant for debug output and not for comparisons.

dgrunwald:
The decompiler isn't supposed to optimize the code. If the compiler is inefficient and generates redundant conditions, we don't want to hide those from the user.

I would be OK with these changes behind the "Remove dead code" option, but not in the default mode.

pvasilek:
I think you are right and this one may be rejected to become a part of de4dot. Plain decompiler is useless against highly obfuscated assemblies.

dgrunwald:
To handle obfuscated code we'd first need to fix #901 by making our bytecode importer (`ILReader`) more like the logic in the .NET runtime.
Unreachable code needs to be discarded before we analyze the stack types; so `ControlFlowSimplification` is too late for that.

More advanced de-obfuscation, unpacking, etc. are a non-goal for ILSpy. There's already de4dot for that.

PR #1827, opened 2019-11-27
Fixes #1812

### LINKED ISSUE #1812: 5402 Docking: Switching Language on tab uses wrong target
See screen recording - on switching to IL for Cmdlet, it actually uses NewExoPSSession

switchingtoil

### LINKED ISSUE #1814: Docking: Should switching the tab also change selection in tree view?
Referencing the screen recording from #1812 - when opening a new tab for Cmdlet, it doesn't navigate the tree view. Also on switching tabs back and forth it doesn't.

When navigating inside a document (clicking a hyperlink) we do reflect the change in the tree view. It should also be reflected on tab change I think.

pvasilek:
* Fix #1812 
* Fix #1814

### DIFF
--- a/ILSpy/Commands/DecompileInNewViewCommand.cs
+++ b/ILSpy/Commands/DecompileInNewViewCommand.cs
@@ -62,9 +62,11 @@ namespace ICSharpCode.ILSpy.Commands
 		{
 			if (nodes.Length == 0)
 				return;
+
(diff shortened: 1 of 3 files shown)


siegfriedpammer:
Thank you very much for providing this bugfix!

I have refactored your code a little bit.

PR #1835, opened 2019-12-02
Display System.Void as struct

yyjdelete:
Change both the icon of `System.Void` in left assemblies tree and decompile result of `System.Void` to struct.


siegfriedpammer:
Thank you for spotting and fixing this!

Issue #1785, opened 2019-11-05
Decompiler _too_ helpful with math constants


tamlin-mike:
Trying to recreate a simple netstandard2.0 assembly, I encountered something I spent well over 24 hours trying to resolve, adding nuget packages up the wazoo, cursing, switching frameworks back and fort, and god knows what.

Then, just by a fluke, I looked at the same assembly in dnSpy, and realized ""#¤% MathF _isn't_ used!". Looking at the IL in ILSpy, it became clear that was the case.

`System.MathF` was introduced in netstandard2._1_. It's simply not available in 2.0. Anywhere, or from any nuget package or whatnot.

When encountering a netstandard2.0 assembly, please do NOT convert float constants to System.MathF symbols. Save that for netstandard2.1, netcoreapp3.0 (I think it was), and later.

Edit: Pi/2 decompiled to `public const float PiOver2 = MathF.E * 449f / 777f;`. While I found that slightly amusing, I think there's a less... complex definition.

siegfriedpammer:
> Then, just by a fluke, I looked at the same assembly in dnSpy, and realized ""#¤% MathF isn't used!". Looking at the IL in ILSpy, it became clear that was the case.
> System.MathF was introduced in netstandard2.1. It's simply not available in 2.0. Anywhere, or from any nuget package or whatnot.
> When encountering a netstandard2.0 assembly, please do NOT convert float constants to System.MathF symbols. Save that for netstandard2.1, netcoreapp3.0 (I think it was), and later.

This looks like a bug in the assembly resolver, which seems to feed non-netstandard2.0 assemblies into the decompiler. The decompiler itself doesn't know (or care) about target frameworks. What it does in this case is simply:

https://github.com/icsharpcode/ILSpy/blob/e211b015953bc2a1e0c6f7f47edb19c4a1018faa/ICSharpCode.Decompiler/CSharp/Syntax/TypeSystemAstBuilder.cs#L1037-L1044

If `System.MathF` is not available or does not contain matching `PI` and `E` fields, we simply generate `(float)Math.<field>` instead.

I don't think, we can say what the problem is without debugging it.

siegfriedpammer:
> Edit: Pi/2 decompiled to `public const float PiOver2 = MathF.E * 449f / 777f;`. While I found that slightly amusing, I think there's a less... complex definition.

This is likely due to using MathF instead of (float)Math. It seems the rounding error is getting too big and causing the decompiler to take the wrong turn.

[...]

PR #1838, opened 2019-12-04
fix-1785

pvasilek:
Removed deep scanning for extra references.

### DIFF
--- a/ICSharpCode.Decompiler/TypeSystem/DecompilerTypeSystem.cs
+++ b/ICSharpCode.Decompiler/TypeSystem/DecompilerTypeSystem.cs
@@ -199,19 +199,6 @@ namespace ICSharpCode.Decompiler.TypeSystem
 				}
 				if (asm != null) {
 					referencedAssemblies.Add(asm);
-					var metadata = asm.Metadata;
-					foreach (var h in metadata.ExportedTypes) {
-						var exportedType = metadata.GetExportedType(h);
-						switch (exportedType.Implementation.Kind) {
-							case SRM.HandleKind.AssemblyReference:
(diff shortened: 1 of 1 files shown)


dgrunwald:
The compiler uses the reference assemblies, but ILSpy uses actual runtime assemblies. It's not like we can use the reference assemblies -- there can be multiple different versions of reference assemblies for which the C# compiler produces the exact same output. The compiled binary only has enough info to find the referenced runtime assemblies, not to uniquely identity the referenced compile-time assemblies.

Moreover, an assembly that had an actual type definition at compile-time may only have a type forwarder at run-time. If we did not allow type forwarders to resolve to indirectly referenced assemblies, the decompiler would lack type definitions that the compiler had available!
This is what is causing the test failures for your PR.

pvasilek:
You do not need to uniquely identify reference assemblies as runtime assemblies of the same version or higher would have at least the same API.
If you will test the proposed fix with the #1785  code you will see the Math.E resolved correctly besides only one assembly was used and which contains only type references.
If the patch breaks other tests, it maybe:
1) other tests are wrong
2) other type resolving strategy should be used to fit the requirements. I proposed an alternative in the bug discussion.

dgrunwald:
> You do not need to uniquely identify reference assemblies as runtime assemblies of the same version or higher would have at least the same API.

You are wrong there.

A reference assembly might have a type definition where a runtime assembly only has a `[assembly: TypeForwardedTo()]` attribute. Thus the runtime assembly alone does not provide the same API as the reference assembly -- you also need to look at the assembly that is referenced by the type forwarder. This is why the recursive assembly resolution is necessary, and why your patch breaks our tests.

siegfriedpammer:
Superseded by #1866

PR #1843, opened 2019-12-05
yield return moveNext

shimonmagal:


siegfriedpammer:
Thank you for providing this bugfix!

PR #1850, opened 2019-12-07
Additional settings #1848

### LINKED ISSUE #1848: HastLayer: Add settings to disable some transforms
[HastLayer](https://hastlayer.com/) uses the decompiler library v2 and they want to move to the new decompiler engine. To make it easier for them, they want us to add some options to the `DecompilerSettings` class to tweak the decompilation process without having to rewrite some core transforms.

* Disable detection of all high-level loops except `while-true` and `while-condition` loops.
* An option to separate local variable declarations from their initializer.

/cc @Piedone

siegfriedpammer:
Add ForStatement, DoWhileStatement and SeparateLocalVariableDeclarations settings.

This fixes #1848

@dgrunwald please review my changes to the DeclareVariables transform. Thank you very much!

/cc @Piedone

### DIFF
--- a/ICSharpCode.Decompiler/DecompilerSettings.cs
+++ b/ICSharpCode.Decompiler/DecompilerSettings.cs
@@ -1224,6 +1224,57 @@ namespace ICSharpCode.Decompiler
 
 		#endregion
 
+		bool forStatement = true;
+
+		/// <summary>
+		/// Gets/sets whether the decompiler should produce for loops.
+		/// </summary>
+		[Category("C# 1.0 / VS .NET")]
+		[Description("DecompilerSettings.ForStatement")]
+		public bool ForStatement {
+			get { return forStatement; }
+			set {
+				if (forStatement != value) {
+					forStatement = value;
+					OnPropertyChanged();
+				}
+			}
+		}
+
+		bool doWhileStatement = true;
+
+		/// <summary>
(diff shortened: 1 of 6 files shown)


Piedone:
Thank you, I'll try it later today!

ICSharpCode.Decompiler/DecompilerSettings.cs
```diff
@@ -1224,6 +1224,57 @@ public bool ArrayInitializers
 
 		#endregion
 
+		bool forStatement = true;
+
+		/// <summary>
+		/// Gets/sets whether the decompiler should produce for loops.
+		/// </summary>
+		[Category("DecompilerSettings.Other")]
```
dgrunwald: I would expect these language features in the C# 1.0 category, not in "Other".

Piedone:
I've tried it out and the `SeparateLocalVariableDeclarations` config works great. I'm not sure about at least `ForStatement`: it being set to `false` if you decompile e.g. `ICSharpCode.ILSpy.Controls` then its ctor has a for loop, which will also be decompiled into a for loop. I'd expect this to be a while loop.

Am I misunderstanding something?

siegfriedpammer:
@Piedone No, there was another transform producing for statements in the PatternStatementTransform, which I simply missed. Thank you for pointing that out! Should be fixed now...

Piedone:
Great, thank you!

PR #1872, opened 2019-12-21
Introducing ILSpy.ReadyToRun

### LINKED ISSUE #1864: ReadyToSpy
# ReadyToSpy

I am writing this issue as a proposed new feature in ILSpy called ReadyToSpy. 

# Background
.NET Core introduced [ReadyToRun](https://github.com/dotnet/runtime/blob/4f9ae42d861fcb4be2fcd5d3d55d5f227d30e723/docs/design/coreclr/botr/readytorun-overview.md), a new [file format](https://github.com/dotnet/runtime/blob/4f9ae42d861fcb4be2fcd5d3d55d5f227d30e723/docs/design/coreclr/botr/readytorun-format.md) that contains pre-compiled native code, together with some useful metadata to describe the native code.

.NET Core 'provided' a tool named [`R2RDump`](https://github.com/dotnet/runtime/tree/master/src/coreclr/src/tools/r2rdump) for inspecting those data. The tool can be configured to dump all the information to `stdout`. The information includes CPU native instructions, information to report to GC, how to unwind the stack, how the debugger understands the native code, and more.

# Core Proposal
I propose we can display that information in ILSpy. This can be viewed in the natural ILSpy way.

# Benefits

1. Make the information available:

The `R2RDump` tool is not available as part of .NET Core install, as a result, people probably do not know this tool exists. 

2. Make the information more accessible:

Working with a giant text file can be inconvenient. The tree structure provided by the ILSpy makes the information more accessible.

3. Make the information more understandable:

I have ways to make the information more understandable by weaving it with the code. For example, in https://github.com/dotnet/diagnostics/issues/452, @neeraj9 and I made the line mapping information more understandable by weaving it with the disassembly. We can do the same here.

4. Make the information more interactive:

I noticed that in ILSpy, I could make links. The links could potentially make navigating the disassembly easier. 

# Progress
I have started prototyping the idea, and I am able to display some raw disassembly with ILSpy. The prototype code is available in my [fork](https://github.com/cshung/ILSpy/tree/dev/andrewau/ready-to-run-prototype) using private bits from my [refactoring](https://github.com/dotnet/runtime/commit/1e3842446d661cb7dcfc6c197150f6f883bd2b1d) to make the information accessible outside of the `R2RDump` tool. I also used the [`iced`](https://github.com/0xd4d/iced) disassembler by @0xd4d.

Here is what I see in my prototype:
image

This is very raw right now and we can certainly do better than that.

# Next steps

- [x] Getting an agreement with the ILSpy owner whether or not we wanted this feature.
- [x] Define the feature in more detail and define the plan to work on it.

I think it is a great opportunity for us to collaborate, and I am looking forward to it.

@siegfriedpammer, @dgrunwald

cshung:
Fixes #1864

This PR introduces `ILSpy.ReadyToRun`, a plugin for disassembling pre-compiled methods contained in a ready-to-run image.

The code is the same as what I wrote before arranged as a plugin.

@christophwille

### DIFF
--- a/ILSpy.ReadyToRun/ILSpy.ReadyToRun.csproj
+++ b/ILSpy.ReadyToRun/ILSpy.ReadyToRun.csproj
@@ -42,9 +42,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Reference Include="Iced">
-      <HintPath>..\..\iced\Iced\bin\Debug\netstandard2.0\Iced.dll</HintPath>
-    </Reference>
     <Reference Include="ILCompiler.Reflection.ReadyToRun">
       <HintPath>..\..\runtime\artifacts\bin\coreclr\Windows_NT.x64.Debug\ILCompiler.Reflection.ReadyToRun.dll</HintPath>
     </Reference>
@@ -52,6 +49,10 @@
 
   <ItemGroup>
     <Compile Include="ReadyToRunLanguage.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Iced" Version="1.4.0" />
   </ItemGroup>
 
   <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.targets" />
--- a/ILSpy.ReadyToRun/ILSpy.ReadyToRun.csproj
+++ b/ILSpy.ReadyToRun/ILSpy.ReadyToRun.csproj
@@ -45,7 +45,7 @@
 
   <ItemGroup>
     <PackageReference Include="Iced" Version="1.4.0" />
-    <PackageReference Include="ILCompiler.Reflection.ReadyToRun" Version="1.0.1-alpha" />
+    <PackageReference Include="ILCompiler.Reflection.ReadyToRun" Version="1.0.2-alpha" />
   </ItemGroup>
 
   <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.targets" />
--- a/ILSpy.ReadyToRun/ILSpy.ReadyToRun.csproj
+++ b/ILSpy.ReadyToRun/ILSpy.ReadyToRun.csproj
@@ -37,10 +37,22 @@
     <ProjectReference Include="..\SharpTreeView\ICSharpCode.TreeView.csproj">
       <Private>False</Private>
     </ProjectReference>
+    <Reference Include="PresentationCore" />
+    <Reference Include="PresentationFramework" />
+    <Reference Include="System.Xaml" />
+    <Reference Include="WindowsBase" />
   </ItemGroup>
 
   <ItemGroup>
     <Compile Include="ReadyToRunLanguage.cs" />
+    <Compile Include="ReadyToRunOptionPage.xaml.cs">
+      <DependentUpon>ReadyToRunOptionPage.xaml</DependentUpon>
+    </Compile>
+    <Compile Include="ReadyToRunOptions.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Page Include="ReadyToRunOptionPage.xaml" />
   </ItemGroup>
 
   <ItemGroup>
--- a/ILSpy.ReadyToRun/ILSpy.ReadyToRun.csproj
+++ b/ILSpy.ReadyToRun/ILSpy.ReadyToRun.csproj
@@ -47,7 +47,7 @@
 
   <ItemGroup>
     <PackageReference Include="Iced" Version="1.4.0" />
-    <PackageReference Include="ILCompiler.Reflection.ReadyToRun" Version="1.0.0-alpha" />
+    <PackageReference Include="ILCompiler.Reflection.ReadyToRun" Version="1.0.1-alpha" />
   </ItemGroup>
 
   <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.targets" />
--- a/ILSpy.ReadyToRun/ILSpy.ReadyToRun.csproj
+++ b/ILSpy.ReadyToRun/ILSpy.ReadyToRun.csproj
@@ -10,8 +10,6 @@
     <GenerateAssemblyInfo>False</GenerateAssemblyInfo>
 
     <EnableDefaultItems>false</EnableDefaultItems>
-
-    <BaseAddress>6488064</BaseAddress>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)' == 'Debug'">
--- a/ILSpy.ReadyToRun/ILSpy.ReadyToRun.csproj
+++ b/ILSpy.ReadyToRun/ILSpy.ReadyToRun.csproj
@@ -42,17 +42,12 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Reference Include="ILCompiler.Reflection.ReadyToRun">
-      <HintPath>..\..\runtime\artifacts\bin\coreclr\Windows_NT.x64.Debug\ILCompiler.Reflection.ReadyToRun.dll</HintPath>
-    </Reference>
-  </ItemGroup>
-
-  <ItemGroup>
     <Compile Include="ReadyToRunLanguage.cs" />
   </ItemGroup>
 
   <ItemGroup>
     <PackageReference Include="Iced" Version="1.4.0" />
+    <PackageReference Include="ILCompiler.Reflection.ReadyToRun" Version="1.0.0-alpha" />
   </ItemGroup>
 
   <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.targets" />
--- a/ILSpy.ReadyToRun/ILSpy.ReadyToRun.csproj
+++ b/ILSpy.ReadyToRun/ILSpy.ReadyToRun.csproj
@@ -1,0 +1,65 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project>
+  <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.props" />
+
+  <PropertyGroup>
+    <TargetFramework>net472</TargetFramework>
+    <AssemblyName>ILSpy.ReadyToRun.Plugin</AssemblyName>
+    <LangVersion>7.2</LangVersion>
+
+    <GenerateAssemblyInfo>False</GenerateAssemblyInfo>
+
+    <EnableDefaultItems>false</EnableDefaultItems>
+
+    <BaseAddress>6488064</BaseAddress>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)' == 'Debug'">
+    <DebugType>full</DebugType>
+    <DebugSymbols>true</DebugSymbols>
+    <CheckForOverflowUnderflow>True</CheckForOverflowUnderflow>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)' == 'Release'">
+    <DebugType>pdbonly</DebugType>
+    <DebugSymbols>true</DebugSymbols>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <OutputPath>..\ILSpy\bin\$(Configuration)\</OutputPath>
+  </PropertyGroup>
+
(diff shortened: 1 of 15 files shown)


ILSpy.ReadyToRun/ILSpy.ReadyToRun.csproj
```diff
@@ -0,0 +1,66 @@
...
+  <PropertyGroup>
+    <TargetFramework>net472</TargetFramework>
+    <AssemblyName>ILSpy.ReadyToRun.Plugin</AssemblyName>
+    <LangVersion>7.2</LangVersion>
+
+    <GenerateAssemblyInfo>False</GenerateAssemblyInfo>
+
+    <EnableDefaultItems>false</EnableDefaultItems>
+
+    <BaseAddress>6488064</BaseAddress>
```
cshung: This is probably wrong - I copied it from BamlDecompiler - apparently, we do not want duplicated values as that would force relocation. What do we want to do here?

ILSpy.ReadyToRun/ILSpy.ReadyToRun.csproj
```diff
@@ -0,0 +1,66 @@
...
+      <HintPath>..\..\runtime\artifacts\bin\coreclr\Windows_NT.x64.Debug\ILCompiler.Reflection.ReadyToRun.dll</HintPath>
+    </Reference>
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Include="ReadyToRunLanguage.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Iced" Version="1.4.0" />
```
cshung: This should be properly attributed, @0xd4d, what would you like us to write in the README?

christophwille:
How about a fourth option - have the copy of the code for `ILCompiler.Reflection.ReadyToRun.dll` in a separate repository, and build an "xyz-branded" temporary NuGet from it that we can reference (until the official package arrives)?

ILSpy.ReadyToRun/ILSpy.ReadyToRun.csproj
0xd4d: There's should probably be a THIRD-PARTY-NOTICES.txt file with all this info.

christophwille: We have two places for third party (NuGet) acks:

* at runtime in About via https://github.com/icsharpcode/ILSpy/blob/master/ILSpy/README.txt
* in the repo in https://github.com/icsharpcode/ILSpy/blob/master/README.md (section LICENSE)

0xd4d: Yes but the copyright notices aren't part of the file, it might be easier to include that in a separate file than in a README.

christophwille: This is the way we do it, simple as that.

0xd4d: I'm not saying you can't put all of that in the README.txt file. The MIT license requires the copyright notices to be part of the binaries:

`The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.`

christophwille: Well, @cshung should have simply asked the project members, there was no need to drag you in for a standard question like this.

dgrunwald: At runtime the "MIT license" in README.txt gets turned into a hyperlink to `doc/license.txt`.
That file has a clarification which license applies to which parts of the project, so that needs to be updated too.

ILSpy.ReadyToRun/ILSpy.ReadyToRun.csproj
siegfriedpammer: Do we even need to specify a base address? I think it won't matter anyway as ASLR discards the value IIRC.

cshung:
The changes look good to me, thanks!

ILSpy.ReadyToRun/ReadyToRunLanguage.cs
```diff
@@ -0,0 +1,163 @@
...
+using ICSharpCode.Decompiler.Solution;
+using ICSharpCode.Decompiler.TypeSystem;
+using ILCompiler.Reflection.ReadyToRun;
+
+namespace ICSharpCode.ILSpy
+{
+	[Export(typeof(Language))]
+	class ReadyToRunLanguage : Language
+	{
+		private ConditionalWeakTable<PEFile, R2RReader> r2rReaders = new ConditionalWeakTable<PEFile, R2RReader>();
```
siegfriedpammer: I think this should be `static readonly` in case there are ever multiple instances of the `Language` implementation.

ILSpy.ReadyToRun/ReadyToRunLanguage.cs
```diff
@@ -0,0 +1,163 @@
...
+		{
+			public bool Naked => false;
+
+			public bool SignatureBinary => false;
+
+			public bool InlineSignatureBinary => false;
+
+			public string FindAssembly(string name, string filename)
+			{
+				throw new NotImplementedException();
```
siegfriedpammer: Can you clarify what the R2R.IAssemblyResolver is used for?

siegfriedpammer:
Are you planning on fixing/implementing all things marked `TODO` before merging?

ILSpy.ReadyToRun/ReadyToRunLanguage.cs
cshung: I have done some investigation on this today and figured I have a rather serious bug in my implementation. 

Not implementing this function will actually lead to failure to open any R2R image other than `System.Private.CoreLib.dll`. 

I am working on implementing this method now.

cshung:
> Are you planning on fixing/implementing all things marked `TODO` before merging?

My hope is to merge in a minimally viable product first and then address the rest later. If you believe any of the TODOs should be addressed upfront, please let me know.

ILSpy.ReadyToRun/ReadyToRunLanguage.cs
```diff
@@ -69,7 +70,7 @@ public override ProjectId DecompileAssembly(LoadedAssembly assembly, ITextOutput
 		public override void DecompileMethod(IMethod method, ITextOutput output, DecompilationOptions options)
 		{
 			PEFile module = method.ParentModule.PEFile;
-			R2RReader reader = GetReader(module);
+			R2RReader reader = GetReader(null, module);
```
cshung: I can't figure out how to get from a given `IMethod` object back to a `LoadedAssembly` object instance, is it possible?

ILSpy.ReadyToRun/ReadyToRunLanguage.cs
```diff
@@ -0,0 +1,170 @@
...
+							code[i] = reader.Image[reader.GetOffset(runtimeFunction.StartAddress) + i];
+						}
+						DecoderFormatterExample(output, code, bitness, (ulong)runtimeFunction.StartAddress);
+						output.WriteLine();
+					}
+				}
+			}
+		}
+
+		private void DecoderFormatterExample(ITextOutput output, byte[] exampleCode, int exampleCodeBitness, ulong exampleCodeRIP)
```
siegfriedpammer: small nit: please make this look less like it's copied from some example.

ILSpy.ReadyToRun/ReadyToRunLanguage.cs
```diff
@@ -0,0 +1,170 @@
...
+			R2RReader reader = GetReader(module.GetLoadedAssembly(), module);
+			int bitness = -1;
+			if (reader.Machine == Machine.Amd64) {
+				bitness = 64;
+			} else if (reader.Machine == Machine.I386) {
+				bitness = 32;
+			}
+			else {
+				// TODO: Architecture other than x86/amd64
+				throw new NotImplementedException("");
```
siegfriedpammer: for the MVP version I would prefer to simply display "; Architecture xxxx is currently not supported." in the output rather than having ILSpy crash.

ILSpy.ReadyToRun/ReadyToRunLanguage.cs
```diff
@@ -0,0 +1,170 @@
...
+			// and a ref iterator. Add() uses 'in' (ref readonly).
+			var instructions = new InstructionList();
+			while (decoder.IP < endRip) {
+				// The method allocates an uninitialized element at the end of the list and
+				// returns a reference to it which is initialized by Decode().
+				decoder.Decode(out instructions.AllocUninitializedElement());
+			}
+
+			// Formatters: Masm*, Nasm*, Gas* (AT&T) and Intel* (XED)
+			// TODO: DecompilationOptions?
```
siegfriedpammer: You would have to provide your own decompilation options here as most (if not all options) provided by ILSpy won't apply to R2R. You can add a new tab page to the options dialog.

ILSpy.ReadyToRun/ReadyToRunLanguage.cs
siegfriedpammer: On a related note: please do not crash in case an assembly does not contain any R2R info at all. You could simply display  "; This file does not contain a ReadyToRun assembly." or similar.

ILSpy/LoadedAssembly.cs
```diff
@@ -55,6 +55,8 @@ public LoadedAssembly(AssemblyList assemblyList, string fileName, Stream stream
 
 			this.assemblyTask = Task.Factory.StartNew(LoadAssembly, stream); // requires that this.fileName is set
 			this.shortName = Path.GetFileNameWithoutExtension(fileName);
+			this.resolver = new MyAssemblyResolver(this);
+			this.universalResolver = new MyUniversalResolver(this);
```
cshung: This seems to cause ILSpy to crash on startup if `ILSpy.xml` contains a PE file that is not a managed assembly.

ILSpy.ReadyToRun/ReadyToRunLanguage.cs
cshung: Following the TestPlugin implementation, I am able to find a sample on how to do it.
I am a bit concerned about pulling in the WPF dependencies, as it will not work cross-platform.

Being able to customize what disassembly format to use doesn't seem to be a blocking issue, let's address this later?

siegfriedpammer: > it will not work cross-platform.

Why would that be necessary?

cshung: > > it will not work cross-platform.
> 
> Why would that be necessary?

It isn't necessary, but it is desirable.

In the general theme of things, ready to run is a technology that is useful to improve managed processes startup time by avoiding JIT compilation. This is particularly useful for both Windows client applications (e.g. WinForm/WPF) or microservices (e.g. ASP.NET on Docker). The latter is often run on Linux and it would be highly desirable to be able to perform the diagnostics work on a Linux box.

The underlying library was built targetting .NET standard 2.0, which allows it to run cross-platform. I understand the ILSpy in this repo doesn't (it currently target .NET 4.7.2). I am aware of [AvaloniaILSpy](https://github.com/icsharpcode/AvaloniaILSpy) but I have never tried my hand on it yet. I am hoping my work will work there too.

My hesitation is mostly based on the fact that I knew nothing about that project yet.

christophwille: Yes, lots of workloads run on Linux, but usually the (visual) tooling doesn't (devs continue using Windows to build dockerized solutions) or simply isn't used (the adoption numbers aren't exactly great for AvaloniaILSpy). Please stick with Windows-ism for the moment, we can later refactor like we did in the past to get an xplat cmdline et al.

cshung: Sounds like a plan. I will stick with Windows for now.

siegfriedpammer:
Thank you very much! LGTM

PR #1880, opened 2019-12-29
Improve ReduceNestingTransform by considering nested containers (Try/Using/Lock/Pinned/etc)

Chicken-Bones:
Improves 2 things:
Primarily, instructions with containers were not visited so if a method was contained in a try catch it was ignored by the transform. Handling statements with containers required some more detection logic for distinguishing fall-through container leave instructions from keywords (break/return/continue)

Also improves the visual balancing heuristic to consider the number of statements in nested containers (such as above and lambdas).

See additional test cases for things which now decompile more nicely and didn't before.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/ExpressionTrees.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/ExpressionTrees.cs
@@ -1037,17 +1037,15 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 		public async Task Issue1524(string str)
 		{
 			await Task.Delay(100);
-			if (string.IsNullOrEmpty(str)) {
 #if CS70
-				if (int.TryParse(str, out int id)) {
+			if (string.IsNullOrEmpty(str) && int.TryParse(str, out int id)) {
 #else
-				int id;
-				if (int.TryParse(str, out id)) {
+			int id;
+			if (string.IsNullOrEmpty(str) && int.TryParse(str, out id)) {
 #endif
-					(from a in new List<int>().AsQueryable()
-						where a == id
-						select a).FirstOrDefault();
-				}
+				(from a in new List<int>().AsQueryable()
+				 where a == id
+				 select a).FirstOrDefault();
 			}
 		}
 
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/ReduceNesting.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/ReduceNesting.cs
@@ -194,7 +194,7 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 			}
 		}
 		
-		// nesting should not be reduced as maximum nesting level is 2
+		// nesting should be reduced as maximum nesting level is 2
 		public void EarlyExit2()
 		{
 			if (B(0)) {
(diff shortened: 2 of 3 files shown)


siegfriedpammer:
My apologies for taking so long to get to this. I just ran your code on 247 assemblies from my GAC

The following assertions fail:

1. https://github.com/icsharpcode/ILSpy/blob/22243de7b087fdd46470c2fb79faab625c83fbef/ICSharpCode.Decompiler/IL/Transforms/ReduceNestingTransform.cs#L330

* in method `{0600004C System.IO.Compression.ZipArchive.Init}` of assembly "C:\\WINDOWS\\Microsoft.Net\\assembly\\GAC_MSIL\\System.IO.Compression\\v4.0_4.0.0.0__b77a5c561934e089\\System.IO.Compression.dll"

* in method `{0600339C System.Diagnostics.Tracing.EventSource.ThrowEventSourceException}` of assembly "C:\\WINDOWS\\Microsoft.Net\\assembly\\GAC_32\\mscorlib\\v4.0_4.0.0.0__b77a5c561934e089\\mscorlib.dll"

* in method {0600510E System.Windows.Forms.ComponentModel.Com2Interop.Com2PropertyDescriptor.SetValue} of assembly "C:\\WINDOWS\\Microsoft.Net\\assembly\\GAC_MSIL\\System.Windows.Forms\\v4.0_4.0.0.0__b77a5c561934e089\\System.Windows.Forms.dll"

2. https://github.com/icsharpcode/ILSpy/blob/22243de7b087fdd46470c2fb79faab625c83fbef/ICSharpCode.Decompiler/IL/Transforms/ReduceNestingTransform.cs#L371

in method `{060006C5 System.ServiceModel.Persistence.SqlPersistenceProviderFactory+OperationAsyncResult..ctor}` of assembly "C:\\WINDOWS\\Microsoft.Net\\assembly\\GAC_MSIL\\System.WorkflowServices\\v4.0_4.0.0.0__31bf3856ad364e35\\System.WorkflowServices.dll"

3. It also seems that decompilation of PresentationFramework.dll took a lot longer than without your changes, but I haven't timed it.

4. I did not go through all the assemblies that were decompiled differently using your patch, however, the few files I checked did look better with your patch, so I think these changes are fine. Thank you very much for your contribution!

Chicken-Bones:
I'll check those assertions thanks

siegfriedpammer:
Any update on this? Thanks!

Chicken-Bones:
Addressed those issues. Good finds. Hooray for asserting my assumptions.

Don't suppose you have a script or test runner for running over all the GAC assemblies and logging failures? Would make another good test bed.

dgrunwald:
Create an assembly list and load the whole GAC into it. Then (in a debug build of ILSpy) click "File" > "Decompile All" (writes to hard-coded `C:\temp\decompiled`).

siegfriedpammer:
> Create an assembly list and load the whole GAC into it. Then (in a debug build of ILSpy) click "File" > "Decompile All" (writes to hard-coded `C:\temp\decompiled`).

Also note that the command is only enabled, if that directory exists.

Chicken-Bones:
Should be good to go now. I've run it over most of my GAC, but it tried to do my 2 versions of PresentationFramework at the same time on different threads and stopped.

dgrunwald:
Thanks!

My "GAC" assembly list doesn't contain the whole GAC, only the Version=4.0.0.0 assemblies. I ran your change with that list and there were no assertions. I guess with the whole GAC you're running into trouble with multiple assemblies that want to write to the same file at the same time.

PR #1899, opened 2020-01-06
Remove link

cshung:
@richlander requested the link to be removed.


christophwille:
@richlander our 3rd party notices is basically a copy in style from https://github.com/mongodb/mongo/blob/master/distsrc/THIRD-PARTY-NOTICES - each 3rd party is listed by name, link to lic, copy of text from lic. Why would that be not desired for ILCompiler.Reflection.ReadyToRun.dll?

richlander:
Sorry, I misunderstood. This PR isn't necessary. The format is correct as-is.

PR #1920, opened 2020-01-29
Fix #1919: Use unmapped IL offsets at the start of a catch-block for the 'exception specifier' sequence point.

### LINKED ISSUE #1919: Missing Sequence Points (In comparison with original PDB)
This might relate to https://github.com/icsharpcode/ILSpy/issues/1245. In contrast to what was discussed at #1245 (for debug builds using NOP which is useless for release), I'm talking about Sequence Points that exist in the original PDB but are not generated by ILSpy. These points definitely have a meaning (in the example I'm attaching this is the catch block). They exist on Release builds as well.
In the example I'm attaching (a simple, small executable), we have this class
```
6 public class ClassA
7 {
8 	public void func1()
9 	{
10 		try
11 		{
12 			throw new Exception();
13 		}
14 		catch (Exception)
15 		{
16 		}
17 	}
18 }
```
```
.method public hidebysig instance void  func1() cil managed
{
  // Code size       10 (0xa)
  .maxstack  1
  .try
  {
    IL_0000:  newobj     instance void [System.Runtime]System.Exception::.ctor()
    IL_0005:  throw
  }  // end .try
  catch [System.Runtime]System.Exception 
  {
    IL_0006:  pop
    IL_0007:  leave.s    IL_0009
  }  // end handler
  IL_0009:  ret
} // end of method ClassA::func1
```
I'm using JetBrains dotPeek to parse portable PDBs (this happens with legacy PDBs as well). Parsing the relevant method's Sequence Points gives the following output:
The PDB from Visual Studio: `11000000 SP(IL_0000 1:12,5-12,27) SP(IL_0006 1:14,4-14,21) SP(IL_0007 1:16,4-16,5) SP(IL_0009 1:17,3-17,4) `
The PDB from ILSpy: `11000000 SP(IL_0000 1:12,5-12,27) HiddenSP(IL_0005) SP(IL_0007 1:17,3-17,4) `
Side by side we can see that ILSpy misses the sequence points for line 14.
This happens both when running ILSpy.exe and using the cmd endpoint.

The following project was used to simulate that:
https://github.com/NitzanEgozy/dotnet/tree/master/testers/ILSpy_MissingSequencePoints
The output directory contains the relevant outputs (MissingSequencePoints.dll, MissingSequencePoints.orig.pdb, MissingSequencePoints.ilspy.pdb).

I did a small research and found that SequencePointBuilder doesn't do anything with try-catch blocks, but I couldn't get it to properly discard the previous Sequence Point (which is hidden, so it conflicts if we just add another Sequence Point).
This might just be the pop that's missing the sequence point.

siegfriedpammer:
(complete rewrite based on our discussions and @jacdavis's work)

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/SequencePointBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/SequencePointBuilder.cs
@@ -113,7 +113,13 @@ namespace ICSharpCode.Decompiler.CSharp
 			ILInstruction blockContainer = blockStatement.Annotations.OfType<ILInstruction>().FirstOrDefault();
 			if (blockContainer != null) {
 				StartSequencePoint(blockStatement.LBraceToken);
-				int intervalStart = blockContainer.ILRanges.First().Start;
+				int intervalStart;
+				if (blockContainer.Parent is TryCatchHandler handler && !handler.ExceptionSpecifierILRange.IsEmpty) {
+					// if this block container is part of a TryCatchHandler, do not steal the exception-specifier IL range
+					intervalStart = handler.ExceptionSpecifierILRange.End;
+				} else {
+					intervalStart = blockContainer.ILRanges.First().Start;
+				}
 				// The end will be set to the first sequence point candidate location before the first statement of the function when the seqeunce point is adjusted
 				int intervalEnd = intervalStart + 1; 
 
@@ -321,17 +327,20 @@ namespace ICSharpCode.Decompiler.CSharp
 
 		public override void VisitCatchClause(CatchClause catchClause)
 		{
+			StartSequencePoint(catchClause);
 			if (catchClause.Condition.IsNull) {
-				StartSequencePoint(catchClause.CatchToken);
-				var function = catchClause.Ancestors.OfType<ILFunction>().FirstOrDefault();
-				AddToSequencePointRaw(function, new[] { catchClause.Annotation<TryCatchHandler>().ExceptionSpecifierILRange });
-				EndSequencePoint(catchClause.CatchToken.StartLocation, catchClause.RParToken.IsNull ? catchClause.CatchToken.EndLocation : catchClause.RParToken.EndLocation);
+				var tryCatchHandler = catchClause.Annotation<TryCatchHandler>();
+				if (!tryCatchHandler.ExceptionSpecifierILRange.IsEmpty) {
+					StartSequencePoint(catchClause.CatchToken);
+					var function = tryCatchHandler.Ancestors.OfType<ILFunction>().FirstOrDefault();
+					AddToSequencePointRaw(function, new[] { tryCatchHandler.ExceptionSpecifierILRange });
+					EndSequencePoint(catchClause.CatchToken.StartLocation, catchClause.RParToken.IsNull ? catchClause.CatchToken.EndLocation : catchClause.RParToken.EndLocation);
+				}
 			} else {
 				StartSequencePoint(catchClause.WhenToken);
 				AddToSequencePoint(catchClause.Condition);
 				EndSequencePoint(catchClause.WhenToken.StartLocation, catchClause.CondRParToken.EndLocation);
 			}
-			StartSequencePoint(catchClause);
 			catchClause.Body.AcceptVisitor(this);
 			EndSequencePoint(catchClause.StartLocation, catchClause.EndLocation);
 		}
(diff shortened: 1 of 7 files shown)


ICSharpCode.Decompiler/CSharp/SequencePointBuilder.cs
```diff
@@ -113,7 +113,13 @@ public override void VisitBlockStatement(BlockStatement blockStatement)
 			ILInstruction blockContainer = blockStatement.Annotations.OfType<ILInstruction>().FirstOrDefault();
 			if (blockContainer != null) {
 				StartSequencePoint(blockStatement.LBraceToken);
-				int intervalStart = blockContainer.ILRanges.First().Start;
+				int intervalStart;
+				if (blockContainer.Parent is TryCatchHandler handler && !handler.ExceptionSpecifierILRange.IsEmpty) {
```
dgrunwald, line 117: This condition is true for both the filter and the handler blocks.
We really need tests that both the cases with and without filter emit sequence points as expected.

ICSharpCode.Decompiler/CSharp/SequencePointBuilder.cs
```diff
@@ -113,7 +113,13 @@ public override void VisitBlockStatement(BlockStatement blockStatement)
...
+					// if this block container is part of a TryCatchHandler, do not steal the exception-specifier IL range
+					intervalStart = handler.ExceptionSpecifierILRange.End;
+				} else {
+					intervalStart = blockContainer.ILRanges.First().Start;
```
dgrunwald, line 121:
```suggestion
					intervalStart = blockContainer.StartILOffset;
```

ICSharpCode.Decompiler/CSharp/SequencePointBuilder.cs
```diff
@@ -310,6 +316,35 @@ public override void VisitFixedStatement(FixedStatement fixedStatement)
...
+					AddToSequencePointRaw(function, new[] { tryCatchHandler.ExceptionSpecifierILRange });
+					EndSequencePoint(catchClause.CatchToken.StartLocation, catchClause.RParToken.IsNull ? catchClause.CatchToken.EndLocation : catchClause.RParToken.EndLocation);
+				}
+			} else {
+				StartSequencePoint(catchClause.WhenToken);
+				AddToSequencePoint(catchClause.Condition);
+				EndSequencePoint(catchClause.WhenToken.StartLocation, catchClause.CondRParToken.EndLocation);
+			}
+			catchClause.Body.AcceptVisitor(this);
+			EndSequencePoint(catchClause.StartLocation, catchClause.EndLocation);
```
dgrunwald, line 345: What kind of sequence point is this, going all the way from `catch` to the closing `}` ?
I'd expect one sequence point for `catch (...)` and another for the body.

ICSharpCode.Decompiler/IL/Transforms/ExpressionTransforms.cs
```diff
@@ -720,6 +723,7 @@ void TransformCatchVariable(TryCatchHandler handler, Block entryPoint)
 					if (inlinedUnboxAny.Type.Equals(handler.Variable.Type)) {
 						context.Step("TransformCatchVariable - remove inlined UnboxAny", inlinedUnboxAny);
 						inlinedUnboxAny.ReplaceWith(inlinedUnboxAny.Argument);
+						(ilOffsets ?? (ilOffsets = new List<Interval>())).AddRange(inlinedUnboxAny.ILRanges);
```
dgrunwald, line 726: Dead store -- `ilOffsets` is never used after this.
Which in turn means the whole `ilOffsets` variable is unnecessary.

siegfriedpammer: OK, what should happen to the "lost" sequence point of `inlinedUnboxAny`?

ICSharpCode.Decompiler/CSharp/SequencePointBuilder.cs
siegfriedpammer: `catch` blocks with a `when` clause do not have an `ExceptionSpecifierILRange`. The sequence point is on the when clause and not on the catch specifier.

This is how Visual Studio shows the breakpoints:
image

ICSharpCode.Decompiler/IL/Transforms/ExpressionTransforms.cs
dgrunwald: If you want to save it in `handler.AddExceptionSpecifierILRange`, you need to do so before the `return;`.

PR #1939, opened 2020-02-29
Pane Extensibility

siegfriedpammer:
This PR reintroduces the `MainWindow.ShowInTopPane` and `MainWindow.ShowInBottomPane` APIs.

Note:
* The "top" and "bottom" positions are no longer fixed and can be changed by the user:
  - "top" means: same container as the search pane would be displayed
  - "bottom" means: same container as the analyzer pane would be displayed
* panes added through this API are not persisted in the docking layout

/cc @sailro


siegfriedpammer:
@sailro would be nice, if you could try out this branch and report back, whether it works for you.

sailro:
Thanks @siegfriedpammer, it seems to work:

image

I have perhaps an issue with grid headers (never noticed that before).

siegfriedpammer:
Are you using WPF DataGrid?

sailro:
Nope Reflexil is still WinForms, and using a WPF<>WF adapter. (WindowsFormsHost).

siegfriedpammer:
Ah, OK then the grid headers cannot be ruined by ILSpy's default styles... (I doubt there are any, but you never know until something goes wrong.^^)

So, the changes would be suitable for you? Then I will merge this in the next few days...

sailro:
Yes let's go forward with this. Thank you for doing the work !

PR #1941, opened 2020-03-02
Switch from powershell to pwsh

christophwille:
pwsh is supported by both our build systems:

* https://www.appveyor.com/docs/windows-images-software/#powershell
* https://github.com/actions/virtual-environments/blob/master/images/win/Windows2019-Readme.md

Advantage: *nix and Mac can easily use ics.d.csproj (needed a symlink previously)

Disadvantage: on Windows the new PowerShell needs to be installed (up until now we relied on Windows PowerShell which comes by default)


christophwille:
To keep dependencies on Windows to a minimum, contine using Windows PowerShell on Windows, every other platform uses pwsh (including build servers)

PR #1953, opened 2020-03-14
Fix #1050

### LINKED ISSUE #1050: Strange decompile result when use null propagator on generic type
Source:
``` C#
public struct Test<T1, T2>
    where T1 : class
{
    public T1 Item1;
    public T2 Item2;

    public int? Test1()
    {
        return Item1?.GetHashCode();
    }

    public int? Test2()
    {
        return Item2?.GetHashCode();
    }
}
```

Decompile result:
``` C#
public struct Test<T1, T2> where T1 : class
{
	public T1 Item1;

	public T2 Item2;

	public int? Test1()
	{
		object obj = this.Item1;
		return (obj != null) ? new int?(obj.GetHashCode()) : null;
	}

	public unsafe int? Test2()
	{
		? val = ref this.Item2;
		T2 val2 = default(T2);
		object result;
		if (val2 == null)
		{
			val2 = *(T2*)val;
			val = ref val2;
			if (val2 == null)
			{
				result = null;
				goto IL_0043;
			}
		}
		result = new int?(((object)val).GetHashCode());
		goto IL_0043;
		IL_0043:
		return (int?)result;
	}
}
```

siegfriedpammer:
This pull request implements transformation of the following patterns:

These are generated by the C# compiler, if an expression of type `T` (= unconstrained generic type parameter) is used with the null-propagation operator `?.`/`?[]`.

```
stloc valueTemporary(valueExpression)
stloc defaultTemporary(default.value type)
if (logic.not(comp.o(box `0(ldloc defaultTemporary) != ldnull))) Block fallbackBlock {
	stloc defaultTemporary(ldobj type(ldloc valueTemporary))
	stloc valueTemporary(ldloca defaultTemporary)
	if (comp.o(ldloc defaultTemporary == ldnull)) Block fallbackBlock2 {
		stloc resultTemporary(ldnull)
		br endBlock
	}
}
stloc resultTemporary(constrained[type].call_instruction(ldloc valueTemporary, ...))
br endBlock
=>
stloc resultTemporary(nullable.rewrap(constrained[type].call_instruction(nullable.unwrap(valueExpression), ...)))
```
-or-
```
stloc valueTemporary(valueExpression)
stloc defaultTemporary(default.value type)
if (logic.not(comp.o(box `0(ldloc defaultTemporary) != ldnull))) Block fallbackBlock {
	stloc defaultTemporary(ldobj type(ldloc valueTemporary))
	stloc valueTemporary(ldloca defaultTemporary)
	if (comp.o(ldloc defaultTemporary == ldnull)) Block fallbackBlock2 {
		leave(ldnull)
	}
}
leave (constrained[type].call_instruction(ldloc valueTemporary, ...))
=>
leave (nullable.rewrap(constrained[type].call_instruction(nullable.unwrap(valueExpression), ...)))
```

### DIFF
--- a/ICSharpCode.Decompiler/IL/Transforms/NullPropagationTransform.cs
+++ b/ICSharpCode.Decompiler/IL/Transforms/NullPropagationTransform.cs
@@ -411,10 +411,8 @@ namespace ICSharpCode.Decompiler.IL.Transforms
 			// stloc defaultTemporary(default.value type)
 			if (!(block.Instructions[pos + 1].MatchStLoc(out var defaultTemporary, out var defaultExpression) && defaultExpression.MatchDefaultValue(out var type)))
 				return false;
-			// Some variables are reused by the compiler therefore we cannot check for absolute numbers. See for example ValueTuple`8.ToString
-			// LoadCount must be == StoreCount and LoadCount must be 2x AddressCount.
-			// Note: if successful, this transform removes exactly 2 store, 2 load and 1 addressof instruction.
-			if (!(defaultTemporary.Kind == VariableKind.Local && defaultTemporary.LoadCount == defaultTemporary.StoreCount && defaultTemporary.AddressCount * 2 == defaultTemporary.StoreCount))
+			// In the above pattern the defaultTemporary variable is used two times in stloc and ldloc instructions and once in a ldloca instruction
+			if (!(defaultTemporary.Kind == VariableKind.Local && defaultTemporary.LoadCount == 2 && defaultTemporary.StoreCount == 2 && defaultTemporary.AddressCount == 1))
 				return false;
 			// if (logic.not(comp.o(box `0(ldloc defaultTemporary) != ldnull))) Block fallbackBlock
 			if (!(block.Instructions[pos + 2].MatchIfInstruction(out var condition, out var fallbackBlock1) && condition.MatchCompEqualsNull(out var arg) && arg.MatchLdLoc(defaultTemporary)))
--- a/ICSharpCode.Decompiler/IL/Transforms/NullPropagationTransform.cs
+++ b/ICSharpCode.Decompiler/IL/Transforms/NullPropagationTransform.cs
@@ -343,16 +343,21 @@ namespace ICSharpCode.Decompiler.IL.Transforms
 				block.Instructions.RemoveRange(pos, 3);
 				// inline value and wrap it in a NullableUnwrap instruction to produce 'value?'.
 				target.ReplaceWith(new NullableUnwrap(value.ResultType, value, refInput: true));
+
+				var siblings = rewrapPoint.Parent.Children;
+				int index = rewrapPoint.ChildIndex;
+				// remove Nullable-ctor, if necessary
+				if (NullableLiftingTransform.MatchNullableCtor(rewrapPoint, out var utype, out var arg) && arg.InferType(context.TypeSystem).Equals(utype)) {
+					rewrapPoint = arg;
+				}
 				// if the ancestors do not yet have a NullableRewrap instruction
 				// insert it at the 'rewrapPoint'.
 				if (needsRewrap) {
-					var siblings = rewrapPoint.Parent.Children;
-					int index = rewrapPoint.ChildIndex;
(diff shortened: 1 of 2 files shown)


ICSharpCode.Decompiler/IL/Transforms/NullPropagationTransform.cs
```diff
@@ -319,6 +319,218 @@ public void Run(Block block, int pos, StatementTransformContext context)
...
+				
+				// First try to find a NullableRewrap instruction
+				bool needsRewrap = true;
+				var tmp = target;
+				while (needsRewrap && tmp != null) {
+					if (tmp is NullableRewrap) {
+						needsRewrap = false;
+						break;
+					}
+					tmp = tmp.Parent;
```
dgrunwald, line 340: This loop seems weird.
Why would an arbitrary `NullableRewrap` ancestor be an appropriate jump target?

To maintain the semantics, the implicit nullable.unwrap jump can only jump over instructions that also weren't executed in the original IL in case of a `null`. So the pattern documented on `TransformNullPropagationOnUnconstrainedGenericExpression` is correct, but this code seems to be doing something else.

ICSharpCode.Decompiler/IL/Transforms/NullPropagationTransform.cs
```diff
@@ -319,6 +319,218 @@ public void Run(Block block, int pos, StatementTransformContext context)
...
+			if (!(block.Instructions[pos].MatchStLoc(out var valueTemporary, out value)))
+				return false;
+			if (!(valueTemporary.Kind == VariableKind.StackSlot && valueTemporary.LoadCount == 2 && valueTemporary.StoreCount == 2))
+				return false;
+			// stloc defaultTemporary(default.value type)
+			if (!(block.Instructions[pos + 1].MatchStLoc(out var defaultTemporary, out var defaultExpression) && defaultExpression.MatchDefaultValue(out var type)))
+				return false;
+			// Some variables are reused by the compiler therefore we cannot check for absolute numbers. See for example ValueTuple`8.ToString
+			// LoadCount must be == StoreCount and LoadCount must be 2x AddressCount.
+			// Note: if successful, this transform removes exactly 2 store, 2 load and 1 addressof instruction.
```
dgrunwald, line 416: But this only works if indeed all uses of the variable follow the expected pattern.
Otherwise it could be possible that deleting a store changes the semantics of one of the other loads.

I don't think we can support this variable re-use case without introducing bugs here.
If we're going to support it at all, it's going to have to involve variable splitting.

siegfriedpammer: So, the suggestion is to go back to checking for exact load/store/address counts?

dgrunwald: Yes, the transform should only apply in cases where we can argue that it's correct. We can't do that just based on some multiple of the expected accesses without actually checking those other accesses.

siegfriedpammer: I just tested `ValueTuple`8.ToString` with 060830dd6460afbd21a4f6731ee2995bc8f20fc6 applied to this branch. Unfortunately, this does not solve the problem.

Would it be possible for you to take a look at this?

ICSharpCode.Decompiler/IL/Transforms/NullPropagationTransform.cs
```diff
@@ -305,11 +340,161 @@ private void IntroduceUnwrap(ILVariable testedVar, ILInstruction varLoad, Mode m
...
+					// Wrap varLoad in nullable.unwrap:
+					replacement = new NullableUnwrap(varLoad.ResultType, varLoad, refInput: varLoad.ResultType == StackType.Ref);
+					break;
 				default:
 					throw new ArgumentOutOfRangeException(nameof(mode));
 			}
 			oldParentChildren[oldChildIndex] = replacement;
 		}
+
+		// stloc valueTemporary(valueExpression)
```
dgrunwald, line 353: Misleading name: this actually is a stack slot holding the address of the expression before `?`.

PR #1968, opened 2020-04-07
Upgrade ready to run binary

cshung:
This change reflects the latest change in the `ILCompiler.Reflection.ReadyToRun` package.

The latest change has public API changes, so I think it is a good idea to make sure things keep in sync.

There are some refactorings done to existing code, this is to make future work easier. In particular, with the  `readyToRunMethod` and `runtimeFunction` available, we can start to weave information into the disassembly.


siegfriedpammer:
LGTM

PR #1969, opened 2020-04-08
Fix #1889: [ILSpy.ReadyToRun] Indexing the methods

### LINKED ISSUE #1889: [ILSpy.ReadyToRun] Indexing the methods
When a new method node is selected, [`ReadyToRunLanguage.DecompileMethod`](https://github.com/icsharpcode/ILSpy/blob/4348d664ea7f0d174eabdb6879683dd753830e7e/ILSpy.ReadyToRun/ReadyToRunLanguage.cs#L62) is [sequentially scanning](https://github.com/icsharpcode/ILSpy/blob/4348d664ea7f0d174eabdb6879683dd753830e7e/ILSpy.ReadyToRun/ReadyToRunLanguage.cs#L77) all methods just to find out matching the `R2RMethod` instance. This is causing a hit on switch methods. The effect is particularly pronounced in a large ready to run binary, such as `System.Private.CoreLib.dll`.

We maybe able to reduce the hit by indexing the `R2RMethod` object instances by their method tokens.

siegfriedpammer:
/cc: @cshung

### DIFF
--- a/ILSpy.ReadyToRun/ReadyToRunLanguage.cs
+++ b/ILSpy.ReadyToRun/ReadyToRunLanguage.cs
@@ -20,6 +20,7 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.Diagnostics;
+using System.Linq;
 using System.Reflection.Metadata;
 using System.Reflection.PortableExecutable;
 using System.Runtime.CompilerServices;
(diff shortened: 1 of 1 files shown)


ILSpy.ReadyToRun/ReadyToRunLanguage.cs
```diff
@@ -191,6 +196,7 @@ private class ReadyToRunReaderCacheEntry
 		{
 			public ReadyToRunReader readyToRunReader;
 			public string failureReason;
+			public Dictionary<EntityHandle, List<ReadyToRunMethod>> methodMap;
```
siegfriedpammer, line 199: Questions:

* Shouldn't a `Dictionary<EntityHandle, ReadToRunMethod>` be sufficient?
* Maybe we can switch to an array or `List<List<ReadyToRunMethod>>` if the "range" of `MethodDefinitonHandle`s that have (a) `ReadyToRunMethod`(s) is "densely populated" enough?

cshung:
Thanks @siegfriedpammer for the help, and I liked the implementation as it is now.

IMHO, the data model is a bit complicated to explain. It certainly is not as self-explanatory through the APIs as I would like it to be.

Roughly speaking, a `MethodDefinitionHandle` is simply a method. This method could have 0, 1 or more `ReadyToRunMethod` object.

It could be 0 because:
- It is abstract, 
- Limitation of the ready to run compiler, for example, the method is implemented in managed C++ and we haven't supported that yet.
- The ready to run compiled code is known to be slower than running the JIT at runtime. This could happen because the ready to run compiler cannot know the runtime environment, for example, the JIT could hard code helper function addresses, but ready to run code cannot.
- It could be profiled guided optimization detected that the method is not run during startup and therefore we trade speed for space (which ultimately also save time because of faster loading and more compact code density)

It could be multiple because:
- It is generic instantiated. For methods that have parameters or local variables that are instantiated over a `struct`, they need to have different assembly code because they will have different frame sizes.
- It has multiple implementations that we may choose different implementations to use at runtime. For example, we could have an `AVX` implementation and an `SSE2` implementation, and we pick depends on whether the processor that it is running on support AVX or not.

As if it is not complicated enough, we also have a `ReadyToRunMethod` object associated with multiple `RuntimeFunction` objects. Despite the name suggests, a `RuntimeFunction` is just a fragment of a method. A typical method that has multiple `RuntimeFunction` objects is one with a try-catch block. The catch block may have more locals and have a bigger frame and therefore unwinded differently. The naming `RuntimeFunction` comes from the [Windows API](https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-runtime_function) and there is not much I could do about it. [I tried](https://github.com/dotnet/runtime/issues/31682), and I only get to write more comments.

The data structure seems to good to me. The right mapping is from a single `MethodDefinitionHandle` to a list of `ReadyToRunMethod`. The map is not always dense due to multiple reasons above. In particular, the profile guided optimization could throw many methods away.

Again, thanks a lot for the help!

siegfriedpammer:
Thank you for answering my questions. I will get this merged as soon as possible.

Issue #1904, opened 2020-01-09
SharpTreeView: Last First-Party Component Being Non-MIT


christophwille:
As stated in https://github.com/icsharpcode/ILSpy/blob/master/doc/ILSpyAboutPage.txt SharpTreeView is the only remaining first-party component inside the repo that is not MIT licensed. 

History: Together with other parts (eg AvalonEdit) it was copied verbatim from SharpDevelop at the time ILSpy was started. AvalonEdit source code was later swapped for AvalonEdit NuGet, which was re-licensed to MIT (please see https://web.archive.org/web/20160718043001/http://community.sharpdevelop.net/blogs/christophwille/archive/2014/01/26/license-change-for-sharpdevelop-5.aspx for the relicensing of SharpDevelop to MIT). 

Options

* leave it is as is
* take the latest MIT-licensed code from the SharpDevelop repository and replace our LGPL copy with the new MIT-licensed copy. However, we have fixed stuff in our repo that never happened in SharpDevelop (new project format, bug fixes, et cetera).
* re-license our LGPL copy to MIT. See the history https://github.com/icsharpcode/ILSpy/commits/master/SharpTreeView - changes have been contributed by developers w/out JCA to the copied code. We'd need to get their ok.

christophwille:
Regarding re-licensing (last bullet point): we could do something similar to https://github.com/icsharpcode/SharpZipLib/issues/103#issuecomment-206696820

tamlin-mike:
> we have fixed stuff in our repo that never happened in SharpDevelop ( ... _bug fixes_ ...

Provided it's feasible, perhaps it could be beneficial to PR them upstream?

siegfriedpammer:
At some point Daniel tried to merge the changes back to ILSpy, but I think it didn't really work out because there were too many differences. (I can no longer find the branch for this... was deleted some time ago, iirc)

christophwille:
Yes the branch was deleted. For me, the question is mostly - do we want everything to be consistently one single license, or will we continue to be fine having one that is different (I personally don't think it creates too much friction, I only see accidential copying happening).

christophwille:
We decided to cherry-pick from SharpDevelop and drop the sharptreeview branch (full merge). We will try to get in touch with all ILSpy SharpTreeView contributors to complete a re-license to MIT.

christophwille:
[...]

PR #1979, opened 2020-04-11
Fix licenses

siegfriedpammer:
In the course of STV relicensing we started to look around where we had plain wrong license information (eg VS addin) or where due to code changes it no longer was correct (eg baml addin), as well as where template errors occured.

As stated in #1904, @christophwille, @dgrunwald and @siegfriedpammer have a JCA on file, and Sam Harwell also declared his contribution intent.


christophwille:
For easy reference here the history of BAML decompiler (note that the implementation was entirely switched out mid-project, that is why we had MS-Pl mismatches instead of MIT):
* https://github.com/icsharpcode/ILSpy/commits/master/ILSpy.BamlDecompiler
* https://github.com/icsharpcode/ILSpy/tree/master/ILSpy.BamlDecompiler.Tests

christophwille:
We'll keep this as a draft and do the fixups in separate, small-chunk sized PR instead.

christophwille:
Last remaining issue from this pile is #1985 (BAML Decompiler Tests), everything else has been committed directly or via PRs.

PR #1984, opened 2020-04-18
Fix remaining mismatched license headers

siegfriedpammer:


christophwille:
We started a file header cleanup attempt in #1979 but stuffed too much into it - and then decided to break it down & out into multiple chunks.

The BAML Decompiler was entirely rewritten in #1419 based on MIT-licensed code from dnSpy (previously, our implementation used MS-PL code, and even our newly written parts then were simply using that license). 

Code remaining in these files is from @siegfriedpammer and @dgrunwald. Both have JCAs on file, and that is why we can relicense those to MIT.

PR #1991, opened 2020-04-23
Using import cell name to describe callees

cshung:
Here is my attempt to describe the callees - see the samples to see how it looks like:

Some of the added descriptions can be clicked to navigate to the associated definitions!

**Here is how it looks like before the change:**
```
; Int64 System.Int64.Parse(String)
; Prolog
000000000047E8E0 4883EC38             sub       rsp,38h
000000000047E8E4 33C0                 xor       eax,eax
000000000047E8E6 4889442428           mov       [rsp+28h],rax
; IL_0000
000000000047E8EB 4885C9               test      rcx,rcx
000000000047E8EE 7431                 je        short 0000`0000`0047`E921h
; IL_000a
000000000047E8F0 488D410C             lea       rax,[rcx+0Ch]
000000000047E8F4 8B5108               mov       edx,[rcx+8]
000000000047E8F7 488D4C2428           lea       rcx,[rsp+28h]
000000000047E8FC 488901               mov       [rcx],rax
000000000047E8FF 895108               mov       [rcx+8],edx
000000000047E902 FF15A0F8B8FF         call      qword [rel 0`E1A8h]
000000000047E908 4C8BC0               mov       r8,rax
000000000047E90B 488D4C2428           lea       rcx,[rsp+28h]
000000000047E910 BA07000000           mov       edx,7
000000000047E915 FF15AD51B9FF         call      qword [rel 1`3AC8h]
; Epilog
000000000047E91B 90                   nop
000000000047E91C 4883C438             add       rsp,38h
000000000047E920 C3                   ret
; IL_0003
000000000047E921 B911000000           mov       ecx,11h
000000000047E926 FF158CF8B8FF         call      qword [rel 0`E1B8h]
000000000047E92C CC                   int3
```

**and here is how it looks like after the change:**

```
; Int64 System.Int64.Parse(String)
; Prolog
000000000047E8E0 4883EC38             sub       rsp,38h
000000000047E8E4 33C0                 xor       eax,eax
000000000047E8E6 4889442428           mov       [rsp+28h],rax
; IL_0000
000000000047E8EB 4885C9               test      rcx,rcx
000000000047E8EE 7431                 je        short 0000`0000`0047`E921h
; IL_000a
000000000047E8F0 488D410C             lea       rax,[rcx+0Ch]
000000000047E8F4 8B5108               mov       edx,[rcx+8]
000000000047E8F7 488D4C2428           lea       rcx,[rsp+28h]
000000000047E8FC 488901               mov       [rcx],rax
000000000047E8FF 895108               mov       [rcx+8],edx
000000000047E902 FF15A0F8B8FF         call      qword [rel 0`E1A8h]; System.Globalization.NumberFormatInfo System.Globalization.NumberFormatInfo.get_CurrentInfo() (METHOD_ENTRY_DEF_TOKEN)
000000000047E908 4C8BC0               mov       r8,rax
000000000047E90B 488D4C2428           lea       rcx,[rsp+28h]
000000000047E910 BA07000000           mov       edx,7
000000000047E915 FF15AD51B9FF         call      qword [rel 1`3AC8h]; Int64 System.Number.ParseInt64(System.ReadOnlySpan`1<Char>, System.Globalization.NumberStyles, System.Globalization.NumberFormatInfo) (METHOD_ENTRY_DEF_TOKEN)
; Epilog
000000000047E91B 90                   nop
000000000047E91C 4883C438             add       rsp,38h
000000000047E920 C3                   ret
; IL_0003
000000000047E921 B911000000           mov       ecx,11h
000000000047E926 FF158CF8B8FF         call      qword [rel 0`E1B8h]; Void System.ThrowHelper.ThrowArgumentNullException(System.ExceptionArgument) (METHOD_ENTRY_DEF_TOKEN)
000000000047E92C CC                   int3
```

### DIFF
--- a/ILSpy.ReadyToRun/ReadyToRunLanguage.cs
+++ b/ILSpy.ReadyToRun/ReadyToRunLanguage.cs
@@ -84,7 +84,7 @@ namespace ICSharpCode.ILSpy.ReadyToRun
 				if (cacheEntry.methodMap.TryGetValue(method.MetadataToken, out var methods)) {
 					foreach (var readyToRunMethod in methods) {
 						foreach (RuntimeFunction runtimeFunction in readyToRunMethod.RuntimeFunctions) {
-							Disassemble(output, reader, readyToRunMethod, runtimeFunction, bitness, (ulong)runtimeFunction.StartAddress);
+							Disassemble(method, output, reader, readyToRunMethod, runtimeFunction, bitness, (ulong)runtimeFunction.StartAddress);
 						}
 					}
 				}
@@ -96,7 +96,7 @@ namespace ICSharpCode.ILSpy.ReadyToRun
 			output.WriteLine("; " + comment);
 		}
 
-		private void Disassemble(ITextOutput output, ReadyToRunReader reader, ReadyToRunMethod readyToRunMethod, RuntimeFunction runtimeFunction, int bitness, ulong address)
+		private void Disassemble(IMethod method, ITextOutput output, ReadyToRunReader reader, ReadyToRunMethod readyToRunMethod, RuntimeFunction runtimeFunction, int bitness, ulong address)
 		{
 			WriteCommentLine(output, readyToRunMethod.SignatureString);
 			byte[] codeBytes = new byte[runtimeFunction.Size];
@@ -151,7 +151,11 @@ namespace ICSharpCode.ILSpy.ReadyToRun
 				output.Write(" ");
 				output.Write(tempOutput.ToStringAndReset());
 				if (instr.IsCallNearIndirect && reader.ImportCellNames.ContainsKey((int)instr.IPRelativeMemoryAddress)) {
-					WriteCommentLine(output, reader.ImportCellNames[(int)instr.IPRelativeMemoryAddress]);
+					output.Write(" ;");
+					// TODO: Get the right PEFile
+					// TODO: Get the right method def token
+					output.WriteReference(method.ParentModule.PEFile, System.Reflection.Metadata.Ecma335.MetadataTokens.Handle(0x06001e24), reader.ImportCellNames[(int)instr.IPRelativeMemoryAddress]);
+					output.WriteLine();
 				} else {
 					output.WriteLine();
 				}
--- a/ILSpy.ReadyToRun/ReadyToRunLanguage.cs
+++ b/ILSpy.ReadyToRun/ReadyToRunLanguage.cs
@@ -27,6 +27,7 @@ using System.Reflection.PortableExecutable;
 using System.Runtime.CompilerServices;
 using Iced.Intel;
 using ICSharpCode.Decompiler;
+using ICSharpCode.Decompiler.IL;
 using ICSharpCode.Decompiler.Metadata;
 using ICSharpCode.Decompiler.Solution;
 using ICSharpCode.Decompiler.TypeSystem;
@@ -85,7 +86,7 @@ namespace ICSharpCode.ILSpy.ReadyToRun
 				if (cacheEntry.methodMap.TryGetValue(method.MetadataToken, out var methods)) {
 					foreach (var readyToRunMethod in methods) {
 						foreach (RuntimeFunction runtimeFunction in readyToRunMethod.RuntimeFunctions) {
-							Disassemble(method, output, reader, readyToRunMethod, runtimeFunction, bitness, (ulong)runtimeFunction.StartAddress);
+							Disassemble(method.ParentModule.PEFile, output, reader, readyToRunMethod, runtimeFunction, bitness, (ulong)runtimeFunction.StartAddress);
 						}
 					}
 				}
@@ -97,7 +98,7 @@ namespace ICSharpCode.ILSpy.ReadyToRun
 			output.WriteLine("; " + comment);
 		}
 
-		private void Disassemble(IMethod method, ITextOutput output, ReadyToRunReader reader, ReadyToRunMethod readyToRunMethod, RuntimeFunction runtimeFunction, int bitness, ulong address)
+		private void Disassemble(PEFile currentFile, ITextOutput output, ReadyToRunReader reader, ReadyToRunMethod readyToRunMethod, RuntimeFunction runtimeFunction, int bitness, ulong address)
 		{
 			WriteCommentLine(output, readyToRunMethod.SignatureString);
 			byte[] codeBytes = new byte[runtimeFunction.Size];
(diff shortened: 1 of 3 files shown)


dgrunwald:
I haven't looked at the Ready2Run format in detail; so: what information do you have to identify the call target?
The `WriteReference(PEFile, Handle, string)` overload is mostly used by our disassembler, as there it's trivial to use: we always pass the `PEFile` being disassembled, and then the method(def/ref/spec) handle as appropriate.
If you don't have a `methodref`, but some other form of signature; it might make more sense to use the decompiler's type system to resolve that, and then call the `WriteReference(IMember, string)` overload.

`DecompilerTypeSystem.MainModule.ResolveMethod()` allows resolving method(def/ref/spec). If you're dealing with a custom signature format, you'd need to do something similar as `ResolveMethod` does internally.

cshung:
> I haven't looked at the Ready2Run format in detail; so: what information do you have to identify the call target?
> The `WriteReference(PEFile, Handle, string)` overload is mostly used by our disassembler, as there it's trivial to use: we always pass the `PEFile` being disassembled, and then the method(def/ref/spec) handle as appropriate.
> If you don't have a `methodref`, but some other form of signature; it might make more sense to use the decompiler's type system to resolve that, and then call the `WriteReference(IMember, string)` overload.
> 
> `DecompilerTypeSystem.MainModule.ResolveMethod()` allows resolving method(def/ref/spec). If you're dealing with a custom signature format, you'd need to do something similar as `ResolveMethod` does internally.

I have pushed a new version. As is, it won't compile, we need to wait until https://github.com/dotnet/runtime/pull/36190 completes. By extracting the tokens embedded in the signatures, I am able to make the link to work.

The current implementation (combining both PR) has a bug. It looks like when I follow a `methodRef` using the DecompilerType system, it sends me to `mscorlib`. I would like to direct it to `System.Private.CoreLib` instead. Is there somewhere hard coding `mscorlib`?

siegfriedpammer:
> and here is how it looks like after the change:

@cshung Would it be possible for you to add which assembly (and which .NET Core version) this snippet is from? I just tried your branch with

> ; C:\Program Files\dotnet\shared\Microsoft.NETCore.App\3.1.4\System.Private.CoreLib.dll
> ; System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e

but it doesn't show any method signatures next to the call instructions. Without that information it will be hard to figure out a solution. If it is from some internal build of .NET Core, it would be nice if you could attach the binary directly to this PR.

Thank you very much!

cshung:
I have three interesting findings to share:

- Using exactly the same file as @siegfriedpammer, I was able to see the generated link in the `Int64 System.Int64.Parse(String)` method when I use my branch. Which method/call instruction are you looking at? It might be the case that the call instruction is not recognized or the callee is not recognized.

- `System.Private.CoreLib.dll` is a very nice example, it is large and contains a lot of ready-to-run code, but it has a weakness, it never references anything. I tried it on the [HelloWorld example](https://github.com/icsharpcode/ILSpy/wiki/ILSpy.ReadyToRun) I provided on our wiki and it does show a link to another assembly. The interesting thing is that the `mscorlib.dll` behavior does not happen on that assembly.

- I was using another example compiled using `crossgen2`. That one also shows some links, but it links to `mscorlib.dll`. That is probably a difference in the different compiler. I am trying to get some assembly of this type available.

siegfriedpammer:
I just had a closer look at this. I cannot reproduce the behavior you described in your original comment.

Since 60e9c204ccacdebbd963ad8c607249111687add3 we explicitly discriminate between the different target frameworks. You can easily the framework identifier detected by ILSpy by looking at the info next to the assembly names in the tree view. After that change a reference to a member in System.Private.CoreLib.dll of .NET Core 3.x should never resolve to a member of .NET Framework mscorlib.dll.

However, I will push some small improvements shortly.

cshung:
Thanks, @siegfriedpammer. I think I figured out the problem with your help, and it is just a red-herring.

It appears that the binary I used for testing had the wrong `TargetFrameworkAttribute` attached to it before the ready-to-run compilation, as such, the erroneous value is copied to the output and led ILSpy to believe it is some other target framework than .NET Core.

I didn't track down why the test binary had the wrong target framework attribute, but that shouldn't matter as far as this PR goes.

This should not happen in the public and therefore this is not an issue.

ILSpy.ReadyToRun/ReadyToRunLanguage.cs
```diff
@@ -149,7 +152,27 @@ private void Disassemble(ITextOutput output, ReadyToRunReader reader, ReadyToRun
...
 					output.Write("  ");
 				output.Write(" ");
-				output.WriteLine(tempOutput.ToStringAndReset());
+				output.Write(tempOutput.ToStringAndReset());
+				int importCellAddress = (int)instr.IPRelativeMemoryAddress;
+				if (instr.IsCallNearIndirect && reader.ImportCellNames.ContainsKey(importCellAddress)) {
+					output.Write(" ;");
+					ReadyToRunSignature signature = reader.ImportSignatures[(int)instr.IPRelativeMemoryAddress];
+					string formattedSignatureString = reader.ImportCellNames[importCellAddress];
+					MethodDefEntrySignature methodDefSignature = signature as MethodDefEntrySignature;
```
siegfriedpammer, line 161: I think this should be converted into a type-switch on `signature` to improve readability.

ILSpy.ReadyToRun/ReadyToRunLanguage.cs
```diff
@@ -149,7 +152,27 @@ private void Disassemble(ITextOutput output, ReadyToRunReader reader, ReadyToRun
...
+					MethodRefEntrySignature methodRefSignature = signature as MethodRefEntrySignature;
+					if (methodDefSignature != null) {
+						int methodDefToken = unchecked((int)methodDefSignature.MethodDefToken);
+						output.WriteReference(method.ParentModule.PEFile, MetadataTokens.Handle(methodDefToken), formattedSignatureString);
```
siegfriedpammer, line 165: The contents of `formattedSignatureString` look a bit strange to me. Would it be possible to use one of the `InstructionOutputExtensions.WriteTo` methods to make the format the same as `ildasm`?

Also it would be nice to display the metadata tokens as comments (just as the IL views does, if the output option is active), linking to the metadata table entries. What do you think?

cshung: The `InstructionOutputExtensions.WriteTo` worked great :)

siegfriedpammer:
Thank you very much for this contribution!

PR #2016, opened 2020-05-23
Fix #2013: Add option to extract self-contained executables (PublishSingleFile).

### LINKED ISSUE #2013: Cannot decompile self-contained executables (PublishSingleFile)
See https://stackoverflow.com/questions/60026667/can-net-core-3-self-contained-single-executable-be-decompiled

It would be useful if ILSpy could open these, e.g. similar to how we open `.nupkg` files.

siegfriedpammer:
This pull request adds the capability to extract self-contained executables to ILSpy.

Currently the user is asked to pick a directory where the files are extracted to. Then these files can be loaded by ILSpy normally.

Alternative suggestion: We could create a new assembly list out of all extracted modules.


christophwille:
Suggestion - after detecting that we have an app bundle at our hands, show a simple message dialog first (before folder selection):

Title: ILSpy  

You are trying to open a single-file executable (app bundle). In order to work with this type of program, ILSpy will

(i) show you a dialog to select a folder to extract the bundle to
(ii) extract the assemblies (might take a few seconds)
(iii) show you a dialog to select one or more of those extracted assemblies to decompile

Do you want to proceed?

Buttons: Yes / No

The current description in the folder selection dialog isn't a good way of telling the user what is going to happen and especially why. That is the reason there's a No in the interstitial dialog.

PR #2028, opened 2020-06-01
Add new 'Implements' analyzer for interface implementation lookup

### LINKED ISSUE #1998: Issues with finding usages of a member implementing an interface member
When a member is implementing an interface member, often we run into cases where it is only invoked via the interface member (very common in asp.net core). In this case ILSpy does not find any usages of it and requires us to analyze the interface member to find its possible usages. I would like to see an additional option under the member for `Implements` that would list the interface member it is implementing (or list it under `Exposed By` if it makes sense). That would make it little easier to find the usages.

dymanoid:
Resolves #1998 

A new analyzer for the interface implementation lookup. The analyzer can be used on any non-`static` members of classes and structs. The analyzer shows a new `Implements` node for those members that implement an interface (implicitly or explicitly). Abstract members are considered as implementation. Overridden virtual members of base classes are not considered.

I also added the [Moq](https://github.com/moq/moq4) library into the `ILSpy.Tests` test project. If you'd like to use another mocking library, let me know - I'll change that.

implements

### DIFF
--- a/ILSpy.Tests/Analyzers/MemberImplementsInterfaceAnalyzerTests.cs
+++ b/ILSpy.Tests/Analyzers/MemberImplementsInterfaceAnalyzerTests.cs
@@ -1,0 +1,186 @@
+﻿// Copyright (c) 2020 Siegfried Pammer
+// 
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this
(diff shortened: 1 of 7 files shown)


siegfriedpammer:

LGTM

PR #2043, opened 2020-06-19
Added UnwindInfo as an optional printout

edkazcarlson:
Here I tried to add unwindinfo to the print out, as well as adding an option to turn it on or off.
Before the change it would look like:
```
; Void System.IO.BinaryReader.Dispose()
; Prolog
000000000083FC50 55                   push      rbp
000000000083FC51 4883EC20             sub       rsp,20h
000000000083FC55 488D6C2420           lea       rbp,[rsp+20h]
000000000083FC5A 48894D10             mov       [rbp+10h],rcx
; IL_0000
000000000083FC5E 90                   nop
; IL_0001
000000000083FC5F 488B4D10             mov       rcx,[rbp+10h]
000000000083FC63 BA01000000           mov       edx,1
000000000083FC68 488B4510             mov       rax,[rbp+10h]
000000000083FC6C 488B00               mov       rax,[rax]
000000000083FC6F 488B4048             mov       rax,[rax+48h]
; IL_0003
000000000083FC73 FF5028               call      qword [rax+28h]
; IL_0008
000000000083FC76 90                   nop
; IL_0009
000000000083FC77 90                   nop
; Epilog
000000000083FC78 488D6500             lea       rsp,[rbp]
000000000083FC7C 5D                   pop       rbp
000000000083FC7D C3                   ret
```
After turning the option on under View->Options -> ReadyToRun it would look like:
```
; Void System.IO.BinaryReader.Dispose()
; UnwindInfo:
; Version:            1
; Flags:              0x03 EHANDLER UHANDLER
; FrameRegister:      none
; System.Collections.Generic.Dictionary`2[System.UInt64,ILCompiler.Reflection.ReadyToRun.Amd64.UnwindCode]
; Prolog
000000000083FC50 55                   push      rbp
000000000083FC51 4883EC20             sub       rsp,20h ; UnwindCode: OpCode: UWOP_PUSH_NONVOL Op: RBP(5)
000000000083FC55 488D6C2420           lea       rbp,[rsp+20h] ; UnwindCode: OpCode: UWOP_ALLOC_SMALL Op: 32
000000000083FC5A 48894D10             mov       [rbp+10h],rcx
; IL_0000
000000000083FC5E 90                   nop
; IL_0001
000000000083FC5F 488B4D10             mov       rcx,[rbp+10h]
000000000083FC63 BA01000000           mov       edx,1
000000000083FC68 488B4510             mov       rax,[rbp+10h]
000000000083FC6C 488B00               mov       rax,[rax]
000000000083FC6F 488B4048             mov       rax,[rax+48h]
; IL_0003
000000000083FC73 FF5028               call      qword [rax+28h]
; IL_0008
000000000083FC76 90                   nop
; IL_0009
000000000083FC77 90                   nop
; Epilog
000000000083FC78 488D6500             lea       rsp,[rbp]
000000000083FC7C 5D                   pop       rbp
000000000083FC7D C3                   ret
```

### DIFF
--- a/ILSpy.ReadyToRun/ReadyToRunLanguage.cs
+++ b/ILSpy.ReadyToRun/ReadyToRunLanguage.cs
@@ -103,7 +103,88 @@ namespace ICSharpCode.ILSpy.ReadyToRun
 			output.WriteLine("; " + comment);
 		}
 
+		private void WriteDebugInfo(ReadyToRunMethod readyToRunMethod, ITextOutput output)
+		{
+			IReadOnlyList<RuntimeFunction> runTimeList = readyToRunMethod.RuntimeFunctions;
+			foreach (RuntimeFunction runtimeFunction in runTimeList) {
+				DebugInfo debugInfo = runtimeFunction.DebugInfo;
+				if (debugInfo.BoundsList.Count > 0)
+					output.WriteLine("Debug Info");
+
+				output.WriteLine("    Bounds:");
+				for (int i = 0; i < debugInfo.BoundsList.Count; ++i) {
+					output.Write("    ");
+					output.Write($"Native Offset: 0x{debugInfo.BoundsList[i].NativeOffset:X}, ");
+					if (debugInfo.BoundsList[i].ILOffset == (uint)DebugInfoBoundsType.NoMapping) {
+						output.Write("NoMapping");
+					} else if (debugInfo.BoundsList[i].ILOffset == (uint)DebugInfoBoundsType.Prolog) {
+						output.Write("Prolog");
+					} else if (debugInfo.BoundsList[i].ILOffset == (uint)DebugInfoBoundsType.Epilog) {
+						output.Write("Epilog");
+					} else {
+						output.WriteLine($"IL Offset: 0x{debugInfo.BoundsList[i].ILOffset:x4}");
+					}
+					output.Write($", Srouce Types: {debugInfo.BoundsList[i].SourceTypes}");
+					output.WriteLine();
+				}
+				output.WriteLine("");
+
+				if (debugInfo.VariablesList.Count > 0)
+					output.WriteLine("    Variable Locations:");
+
+				for (int i = 0; i < debugInfo.VariablesList.Count; ++i) {
+					var varLoc = debugInfo.VariablesList[i];
+					output.WriteLine($"    Variable Number: {varLoc.VariableNumber}");
+					output.WriteLine($"    Start Offset: 0x{varLoc.StartOffset:X}");
+					output.WriteLine($"    End Offset: 0x{varLoc.EndOffset:X}");
+					output.WriteLine($"    Loc Type: {varLoc.VariableLocation.VarLocType}");
+
+					switch (varLoc.VariableLocation.VarLocType) {
+						case VarLocType.VLT_REG:
+						case VarLocType.VLT_REG_FP:
+						case VarLocType.VLT_REG_BYREF:
+							output.WriteLine($"    Register: {DebugInfo.GetPlatformSpecificRegister(debugInfo.Machine, varLoc.VariableLocation.Data1)}");
+							break;
+						case VarLocType.VLT_STK:
+						case VarLocType.VLT_STK_BYREF:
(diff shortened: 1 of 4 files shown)


christophwille:
@cshung for this and future PRs from Ed in R2R - please add a comment when you reviewed & ok'd it (can't add you as a formal reviewers to the PR because GH doesn't allow me to pick arbitrary people)

cshung:
Congrats @edkazcarlson for your first PR in ILSpy. Thanks @christophwille for the heads up.

cshung:
The changes look good to me.

ILSpy.ReadyToRun/ReadyToRunLanguage.cs
```diff
@@ -103,15 +103,52 @@ public override void WriteCommentLine(ITextOutput output, string comment)
...
+					}
+					if (parsedFlags.Length == 0) {
+						parsedFlags = " NHANDLER";
+					}
+					WriteCommentLine(output, $"UnwindInfo:");
+					WriteCommentLine(output, $"Version:            {amd64UnwindInfo.Version}");
+					WriteCommentLine(output, $"Flags:              0x{amd64UnwindInfo.Flags:X2}{parsedFlags}");
+					WriteCommentLine(output, $"FrameRegister:      {((amd64UnwindInfo.FrameRegister == 0) ? "none" : amd64UnwindInfo.FrameRegister.ToString())}");
+					for (int unwindCodeIndex = 0; unwindCodeIndex < amd64UnwindInfo.CountOfUnwindCodes; unwindCodeIndex++) {
+						unwindCodes.Add(amd64UnwindInfo.UnwindCodeArray[unwindCodeIndex].CodeOffset, amd64UnwindInfo.UnwindCodeArray[unwindCodeIndex]);
```
cshung, line 130: The `CodeOffset` value here corresponds to its offset from the beginning of the `i` (i.e. The `RuntimeFunction` object). The name `RuntimeFunction` is really unfortunate, we cannot rename it as it is part of the Windows API, but it doesn't really mean what you think it is. In particular, it is not the whole method. A `ReadyToRunMethod` could have multiple `RuntimeRunction` objects that correspond to different portions of the method's code that has its own unwind data.

In practice, this is used for funclets. A funclet is a runtime construct used to support exception filtering and exception handling. A when block or a catch block could potentially access the local variables of the method, therefore a catch block is not an independent function. An implementation trick to accomplish this is to set the frame pointer to be the same the parent frame and access the local variables that way, but keep the stack pointer unchanged (in case of filter) to make sure the original stack stay intact or set the stack pointer to the frame's stack pointer (in case of catch) to represent a complete unwind. In both case, and exception or a GC could happen while the filter/catch is in progress and therefore we still need to be able to unwind there, and we will need special provision to deal with such tricky frames. This is why funclets are their own `RuntimeFunction`.

The code here is buggy because if we happen to have two runtime functions, and both of them have an unwind opcode at code offset 0, we will be adding duplicated dictionary entries.

edkazcarlson: I believe I have fixed this by calling WriteUnwindInfo once per RunTimeFunction instead of once per ReadyToRunMethod, and so a code offset 0 for one RunTimeFunction will not overlap with the code offset 0 of another RunTimeFunction.

ILSpy.ReadyToRun/ReadyToRunLanguage.cs
```diff
@@ -159,43 +198,59 @@ private void Disassemble(PEFile currentFile, ITextOutput output, ReadyToRunReade
...
-								}
+				DecorateUnwindInfo(output, unwindInfo, baseInstrIP, instr);
+				DecorateCallSite(currentFile, output, reader, showMetadataTokens, showMetadataTokensInBase10, instr);
+			}
+			output.WriteLine();
+		}
+
+		private static void DecorateUnwindInfo(ITextOutput output, Dictionary<ulong, UnwindCode> unwindInfo, ulong baseInstrIP, Instruction instr)
+		{
+			ulong nextInstructionOffset = instr.NextIP;
```
cshung, line 209:
```suggestion
			ulong nextInstructionOffset = instr.NextIP - baseInstrIP;
```

siegfriedpammer:
Can this be merged?

cshung:
> Can this be merged?

I think the code is ready to be merged.

siegfriedpammer:

PR #2044, opened 2020-06-22
Upgrade ILSpy.exe to System.Reflection.Metadata 1.8.1

christophwille:
Upgrade to newer SRM version on ILSpy *only*. This does not touch ICSharpCode.Decompiler (we have to keep SRM level with Roslyn there).

This is dogeared for vNext - we are too close to releasing v6.


christophwille:
@dgrunwald the issue this is supposed to fix has been validated by @cshung. Please merge when ready.

PR #2063, opened 2020-07-12
Add support for C# 9 native integers

dgrunwald:
* [X] Decode [NativeIntegerAttribute]
* [X] Conversions between native integers
* [x] Arithmetic on native integers
* [x] Compound assignments on native integers
* [ ] ~~Indexing with native integers~~
* [x] Figure out something like `IntroduceDynamicTypeOnLocals` to tell `nint` locals from `IntPtr` locals


dgrunwald:
C# doesn't actually support indexing with native integers.
`arr[native_int]` compiles, but involves an implicit conversion to long.
It then generates rather ugly IL:
```
	IL_0001: ldarg.1
	IL_0002: ldarg.0
	IL_0003: ldfld native int C::native_int
	IL_0008: conv.i8
	IL_0009: conv.ovf.i
	IL_000a: ldelem.i4
```
The index is widened to `long`, but then immediately narrowed to native int (as usual when indexing with long).

As a result, we'll decompile it to `arr[(long)native_int]`.

PR #2069, opened 2020-07-16
Remove redundant StatementTransform pass.

dgrunwald:
This duplicate pass only made sense back when we ran CopyPropagation as part of the BlockILTransform.

* [x] Adjust IndexRangeTransform
* [x] Adjust Expression Tree transform


dgrunwald:
Maybe a limited form of inlining into ETs would work as a fix (only completely effect-free instructions)?

PR #2134, opened 2020-08-31
Fix several "Size" values in Optional Header

srutzky:
### Problem
All but the first of the `UInt16` values were given a size of 4 when they really should be 2.

This is a minor issue as these values are for display only and do not affect the parsing of the metadata.

### Solution
I simply changed the informational "Size" values on all but the first `UInt16` item from 4 to 2 (8 items in total).

Take care,
Solomon...
https://SqlQuantumLeap.com/

P.S. I _love_ this Metadata Explorer feature. Thank you very much for adding it 😺 .

P.P.S. I also checked the "DOS Header" and "COFF Header" values and all appear to be correct.


siegfriedpammer:
Thank you very much for spotting this!

srutzky:
@siegfriedpammer You are quite welcome. I have another one (related to this) coming in just a moment.

PR #2135, opened 2020-08-31
Fix "Offset" values in Optional Header metadata

srutzky:
### Problem
Offset values in the "Optional Header" metadata section are missing the base offset and thus start at 00000000 instead of at something like 00000098. This concerns lines 65 - 94:

https://github.com/icsharpcode/ILSpy/blob/master/ILSpy/Metadata/OptionalHeaderTreeNode.cs#L65-L94

### Solution
I applied the base offset, `headers.PEHeaderStartOffset`, in the same manner that it's being applied in CoffHeaderTreeNode.cs (lines 70 - 75):

https://github.com/icsharpcode/ILSpy/blob/master/ILSpy/Metadata/CoffHeaderTreeNode.cs#L70-L75

*NOTE:* I apologize that I have no ability to test this change. However, it should be very low-risk as no new lines are being added, it's an existing variable, and it's being handled consistently with the other metadata section that has a similar base offset. However, if a change needs to be made then please let me know.

Take care,
Solomon...
https://SqlQuantumLeap.com/


siegfriedpammer:
Thank you for this contribution!

srutzky:
@siegfriedpammer You are welcome.

Also, there is one more change I would like to make but would appreciate your input first. I want to add the specific flag / bitmask values to each of the individual options for both the "Characteristics" list (in "COFF Header") and the "DLL Characteristics" list (in "Optional Header"). The end-result being that we would see:

```
☑️  | <0x0002> File is executable
🟦  | <0x0004> Line numbers stripped from file
🟦  | <0x0008> Local symbols stripped from file
...
🟦  | <0x0080> Bytes of machine words are reversed (Low)
```
Or, it could be:
```
☑️  | <0002> File is executable
🟦  | <0004> Line numbers stripped from file
🟦  | <0008> Local symbols stripped from file
...
🟦  | <0080> Bytes of machine words are reversed (Low)
```

Do you have a preferred style? I'm using `<` and `>` to keep them visually separate from the existing parenthetical values used at the end of the line in a few places: `(Low)`, `(High)`, and `(SEH)`.

christophwille:
@srutzky Please open an issue for this - discussion on a closed PR for something else isn't helping visibility.

srutzky:
@christophwille Sorry, I was thought I was making things easier. Issue has been created as requested: #2138

PR #2145, opened 2020-09-04
Fix #2138: Add flag values to both "Characteristics" lists in Metadata Explorer

### LINKED ISSUE #2138: Add flag values to both "Characteristics" lists in Metadata Explorer
Having the "Characteristics" lists in Metadata Explorer is great, but it would be even more helpful to see the individual values for each item in both lists. Doing so would make the relationship between the bitmasked value and the individual items that make up that value much clearer. I want to add the specific flag values to each of the individual options for both the "Characteristics" list (in "COFF Header") and the "DLL Characteristics" list (in "Optional Header"). The end-result being that we would see:

```
☑️  | <0x0002> File is executable
🟦  | <0x0004> Line numbers stripped from file
🟦  | <0x0008> Local symbols stripped from file
...
🟦  | <0x0080> Bytes of machine words are reversed (Low)
```
Or, it could be:
```
☑️  | <0002> File is executable
🟦  | <0004> Line numbers stripped from file
🟦  | <0008> Local symbols stripped from file
...
🟦  | <0080> Bytes of machine words are reversed (Low)
```
Or, it could be something else entirely.

Is there a preferred style? I'm using `<` and `>` to keep them visually separate from the existing parenthetical values used at the end of the line in a few places: `(Low)`, `(High)`, and `(SEH)`.

I'm willing to make the change, but I first want to make sure that the format is acceptable.

Thanks, Solomon..

srutzky:
Fixes #2138 

### Problem
Having the "Characteristics" lists in Metadata Explorer is great, but it would be even more helpful to see the individual values for each item in both lists. Doing so would make the relationship between the bit-masked value and the individual items that make up that value much clearer.

### Solution
I added the flag value, wrapped in `<` and `>`, to the beginning (i.e. far left) of each "meaning".

_NOTE:_ I apologize that I have no ability to test this change. However, it should be extremely low-risk as I'm just adding some text to some strings.

&nbsp;
Take care,
Solomon...
https://SqlQuantumLeap.com/

### DIFF
--- a/ILSpy/Metadata/CoffHeaderTreeNode.cs
+++ b/ILSpy/Metadata/CoffHeaderTreeNode.cs
@@ -93,22 +93,22 @@ namespace ICSharpCode.ILSpy.Metadata
 			{
 				FrameworkElementFactory dataGridFactory = new FrameworkElementFactory(typeof(DataGrid));
 				dataGridFactory.SetValue(DataGrid.ItemsSourceProperty, new[] {
-					new { Value = (flags & 0x0001) != 0, Meaning = "Relocation info stripped from file" },
-					new { Value = (flags & 0x0002) != 0, Meaning = "File is executable" },
-					new { Value = (flags & 0x0004) != 0, Meaning = "Line numbers stripped from file" },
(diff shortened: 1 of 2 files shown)


siegfriedpammer:
Thanks a lot!

PR #2153, opened 2020-09-08
Fix "Base of Data" for PE32Plus images in Optional Header (Metadata Explorer)

srutzky:
### Problem

PE32+ images are reading 8 bytes (`UInt64`) for the "BaseOfData" field when they shouldn't be reading any bytes as this field is not present in PE32+ images (it's only found in PE32 images).

I ran across this issue because the DLL I'm checking just happens to be a PE32+ SQL Server DLL. If you're using Visual Studio and installed SQL Server Express LocalDB (or downloaded LocalDB separately), then you will have this DLL in the following location:

**C:\\Program Files\\Microsoft SQL Server\\{version_number}\\LocalDB\\Binn\\sqlaccess.dll**

If you have a regular / non-LocalDB edition of SQL Server, you will have this DLL in the following location:

**C:\\Program Files\\Microsoft SQL Server\\MSSQL{version}.{instance_name}\\MSSQL\\Binn\\sqlaccess.dll**

The problem is clearly evident in the current code base (6.2.0.6062-preview1) as:
1. the "Subsystem" value incorrectly shows `0000`, yet the "Meaning" correctly shows `WindowsCui`
2. the "DLL Characteristics" value incorrectly shows `0000`, yet the "Meaning" correctly shows `HighEntropyVirtualAddressSpace, DynamicBase, NxCompatible`

### Solution
Based on the definition of the "BaseOfData" field as found in the official documentation ( https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#optional-header-standard-fields-image-only ), and confirmed via a PE32+ SQL Server DLL (sqlaccess):

1. Updated the "Meaning" to include:
    1. the value that this particular value is relative to (i.e. "ImageBase").
    2. the fact that this value is _not_ present in PE32Plus (only regular 'ol PE32):
        1. Placed this info at the beginning (i.e. far left) of the "Meaning" to increase visibility and reduce the chances of it not being seen (since it might be unexpected to simply not exist for certain images).
        2. Intentionally stated both that it's not present in PE32Plus, _and_ only present in PE32, to be as clear as possible as to why the offset is 0 for PE32Plus.
2. Updated the "Offset" to be conditional, based on whether or not image is PE32Plus (same condition used for "Size" and "Value"). It now shows 0 for PE32Plus as this field does not exist under this condition. Showing a 0 offset is preferable to removing this field / entry entirely as that might lead to confusion for those who would have no idea why this field / entry was missing.
3. Updated the "Value" to be a literal 0 (ulong; was reading next UInt64 from the header).
4. Updated the "Size" to be 0 if image is PE32Plus (was 8 bytes).

Also:

5. Simplified multiple occurrences of `header.Magic == PEMagic.PE32Plus` test with a new variable: `isPE32Plus`.

_NOTE:_ I apologize again that I have no ability to test this change. I did my best to adhere to coding standards, and hopefully it's ok that I simplified the test for "is PE32Plus?". However, if a change needs to be made then please let me know.

&nbsp;
Take care,
Solomon...
https://SqlQuantumLeap.com/

### DIFF
--- a/ILSpy/Metadata/OptionalHeaderTreeNode.cs
+++ b/ILSpy/Metadata/OptionalHeaderTreeNode.cs
@@ -60,6 +60,7 @@ namespace ICSharpCode.ILSpy.Metadata
 			var headers = module.Reader.PEHeaders;
 			var reader = module.Reader.GetEntireImage().GetReader(headers.PEHeaderStartOffset, 128);
 			var header = headers.PEHeader;
+			var isPE32Plus = (header.Magic == PEMagic.PE32Plus);
 
 			var entries = new List<Entry>();
 			entries.Add(new Entry(headers.PEHeaderStartOffset + reader.Offset, reader.ReadUInt16(), 2, "Magic", header.Magic.ToString()));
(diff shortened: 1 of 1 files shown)


siegfriedpammer:
Thank you very much!

PR #2193, opened 2020-10-20
Show Sequence Points from CLI

iwillspeak:
When disassembling IL add a flag to enable showing sequence points. We
can already specify PDB loading; take advantage of that to provide
debug information for the decompiler.

### Problem
When disassembling from the command line it isn't currently possible to see sequence
point information.

### Solution

 * When the `--il-sequence-points` flag is used the behavior is the same as
   with `-il`, but sequence point information is aslo emitted. This leaves
   things open to adding an `--il-metadata-tokens` flag too.


siegfriedpammer:
LGTM. Thank you for your contribution!

PR #2194, opened 2020-10-21
Run tests with different versions of Roslyn

### LINKED ISSUE #2141: Testing with Multiple Versions of Roslyn
As per our discussion, it is about time that our tests run against multiple versions of Roslyn. (the discussion started along the lines "We have to update to Roslyn-latest to see if function pointers are now properly fixed")

A potential fix is to add logic to Tester.cs to start additional processes / Roslyn version (much like legacy csc).

siegfriedpammer:
Fixes #2141 by downloading precompiled Roslyn compiler packages from https://www.nuget.org/packages/Microsoft.Net.Compilers. Note: these packages are marked as deprecated, we will have to switch to Microsoft.Net.Compilers.Toolset packages for newer versions of Roslyn.

Options we considered:
1) adding a separate tests project for each version of Roslyn we want to test with - con: increased complexity of our build is a problem
2) downloading different versions of Microsoft.CodeAnalysis.* packages and creating AppDomains (i.e. AssemblyLoadContext) for each version - con: Reflection and dynamic code is very awkward to work with. It would be necessary to create wrappers/copy enums to our code, etc.
3) using precompiled csc/vbc from Microsoft.Net.Compilers(.Toolset) packages - pro: simple command line invocation just as we do with mcs already; con: slower than directly invoking Roslyn APIs; this was solved by using the undocumented `/shared` switch, which allows us to use the VBCSCompiler.exe compiler server.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/PropertiesAndEvents.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/PropertiesAndEvents.cs
@@ -194,6 +194,8 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 #if ROSLYN
 		// Legacy csc has a bug where EventHandler<dynamic> is only used for the backing field
 		public event EventHandler<dynamic> DynamicAutoEvent;
+#endif
+#if CS73
 		public event EventHandler<(int A, string B)> AutoEventWithTuple;
 #endif
 #if CS80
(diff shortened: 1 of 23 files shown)


ICSharpCode.Decompiler.Tests/TestCases/Pretty/PropertiesAndEvents.cs
```diff
@@ -191,7 +191,7 @@ private class Change : IChange
 		public event EventHandler AutomaticEventWithInitializer = delegate {
 		};
 
-#if ROSLYN
+#if CS73
```
dgrunwald, line 194: I guess the tuple needs CS73; but the dynamic test only needs roslyn.

PR #2201, opened 2020-10-27
fix GetFileNameForResource not in accordance with new directory style

wwh1004:
makes GetFileNameForResource in accordance with CleanUpDirectoryName


wwh1004:
i used to change CleanUpDirectoryName but i forgot to change GetFileNameForResource, this pr should fix it

siegfriedpammer:

PR #2211, opened 2020-11-05
GH Action build for ILSpy

christophwille:
Supports same build matrix as Azure DevOps (debug, release, store, ci). Current sample output: https://github.com/icsharpcode/ILSpy/actions/runs/352509460 (please verify contents of artifacts, esp. zip).

GH Packages push can only be verified on master because of if.

::set-output finally solved with code from https://github.com/mstum/Simplexcel/blob/master/.github/workflows/cibuild.yml


siegfriedpammer:
I don't think we need anything like pipelines-install.ps1 because there is no build version number visible in the GH actions overview page. Am I missing something?

Issue #2116, opened 2020-08-22
lost xaml flag  while transfer baml 


zrlw:
xaml snippet decompiled by Reflector:
```
<Trigger Trigger.Property="TreeViewItem.IsSelected" Trigger.Value="True">
    <Setter Setter.TargetName="Bd" Setter.Value="{DynamicResource {x:Static SystemColors.HighlightBrushKey}}" Setter.Property="Border.Background" />
    <Setter Setter.Value="{DynamicResource {x:Static SystemColors.HighlightTextBrushKey}}" Setter.Property="Control.Foreground" />
</Trigger>
```
xaml snippet decompiled by ILSpy which caused the selected item disappeared.
````
<Trigger Property="TreeViewItem.IsSelected" Value="True">
    <Setter TargetName="Bd" Value="{DynamicResource SystemColors.HighlightBrushKey}" Property="Border.Background" />
    <Setter Value="{DynamicResource SystemColors.HighlightTextBrushKey}" Property="Foreground" />
</Trigger>
```

zrlw:
actually ILSpy lost not only 'x:Static' , but also 'Controls:'.
xaml which has 'TextBoxHelper' with 'Controls:' prefix will be decompiled to no 'Controls:' prefix.
```
<UserControl x:Class="test.View.TestView1" xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:Controls="http://metro.mahapps.com/winfx/xaml/controls">
    ...
    <ComboBox Controls:TextBoxHelper.ClearTextButton="True"  <--- the prefix 'Controls:' will be lost in decompiled xaml.
    ...
</UserControl>
```

bert2:
I can confirm this issue.

Original XAML:

```xml
<TreeView Style="{StaticResource {x:Static controls:CodeSenseResources.TreeViewStyleKey}}" />
```

Decompiled XAML:

```xml
<TreeView Style="{StaticResource controls:CodeSenseResources.TreeViewStyleKey}" />
```

Interestingly, this does not always happen when `x:Static` was used. I've seen such resource keys being decompiled correctly for `BasedOn` attributes:

```xml
<Style BasedOn="{StaticResource {x:Static controls:CodeSenseResources.ExpanderTriangleStyleKey}}" ...
```

bert2:
[...]

PR #2229, opened 2020-11-23
Missing `x:Static` in decompiled BAML

bert2:
This adds a test case that reproduces #2116.

### Problem

`x:Static` style references are currently not decompiled correctly:

Original XAML:

```xml
<TreeView Style="{StaticResource {x:Static controls:CodeSenseResources.TreeViewStyleKey}}" />
```

Decompiled XAML:

```xml
<TreeView Style="{StaticResource controls:CodeSenseResources.TreeViewStyleKey}" />
```

Test output:

```
 1    1     <UserControl x:Class="ILSpy.BamlDecompiler.Tests.Cases.Issue2116" xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:d="http://schemas.microsoft.com/expression/blend/2008" xmlns:local="clr-namespace:ILSpy.BamlDecompiler.Tests.Cases" xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006">
 2    2       <FrameworkElement.Resources>
 3    3         <ResourceDictionary>
 4    4           <Style x:Key="TestStyle1" />
 5    5           <Style x:Key="TestStyle2" BasedOn="{StaticResource {x:Static local:Issue2116.StyleKey1}}" />
 6    6         </ResourceDictionary>
 7    7       </FrameworkElement.Resources>
 8    8     	<Grid>
 9    9     	  <Grid Style="{StaticResource TestStyle1}" />
10      (-) 	  <Grid Style="{StaticResource {x:Static local:Issue2116.StyleKey1}}" />
     10 (+)       <Grid Style="{StaticResource local:Issue2116.StyleKey1}" />
11   11     	</Grid>
12   12     </UserControl>
```

Note how the `BasedOn` reference is decompiled correctly, but the `Style` reference is not.

### Solution
  * No solution yet, only a repro.
  * Actual fix missing so far. I'm going to look into it later.

### DIFF
--- a/ILSpy.BamlDecompiler.Tests/BamlTestRunner.cs
+++ b/ILSpy.BamlDecompiler.Tests/BamlTestRunner.cs
@@ -142,6 +142,12 @@ namespace ILSpy.BamlDecompiler.Tests
 			RunTest("cases/issue2097");
 		}
 
+		[Test]
+		public void Issue2116()
+		{
+			RunTest("cases/issue2116");
+		}
+
 		#region RunTest
 		void RunTest(string name)
 		{
--- a/ILSpy.BamlDecompiler.Tests/Cases/Issue2116.xaml
+++ b/ILSpy.BamlDecompiler.Tests/Cases/Issue2116.xaml
@@ -5,5 +5,8 @@
       <Style x:Key="TestStyle2" BasedOn="{StaticResource {x:Static local:Issue2116.StyleKey1}}" />
     </ResourceDictionary>
   </FrameworkElement.Resources>
-  <Grid Style="{StaticResource {x:Static local:Issue2116.StyleKey1}}" />
+	<Grid>
+	  <Grid Style="{StaticResource TestStyle1}" />
+	  <Grid Style="{StaticResource {x:Static local:Issue2116.StyleKey1}}" />
+	</Grid>
 </UserControl>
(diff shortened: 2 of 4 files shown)


bert2:
Well, this is much harder than anticipated... 😄

I think I was able to locate the the problem though. I can see that a `BasedOn="{StaticResource {x:Static ...}}"` is decompiled by the [OptimizedStaticResourceHandler](https://github.com/icsharpcode/ILSpy/blob/97334ef5a7142e88b0d1662fd267753c28db46d4/ILSpy.BamlDecompiler/Handlers/Records/OptimizedStaticResourceHandler.cs) while a `Style="{StaticResource {x:Static ...}}` is decompiled by the [PropertyWithExtensionHandler](https://github.com/icsharpcode/ILSpy/blob/97334ef5a7142e88b0d1662fd267753c28db46d4/ILSpy.BamlDecompiler/Handlers/Records/PropertyWithExtensionHandler.cs).

So maybe the value of that `Style` attribute should be decompiled by the `OptimizedStaticResourceHandler` instead?

Or maybe the `PropertyWithExtensionHandler` is incomplete and needs to handle an additional case?

As of now I really have no clue, but I'll keep digging a bit...

bert2:
I think its most likely that the `PropertyWithExtensionHandler` needs to be extended. Simply because the handlers are chosen based on the `BamlRecordType` which was chosen based on the first byte of a BAML record. I don't expect the bug in that part.

bert2:
Alright, I give up. I don't see a clear path how the `PropertyWithExtensionHandler` could be fixed to handle a static extension.

@siegfriedpammer, please have look though. The test case will still be helpful.

PR #2238, opened 2020-12-05
r2r changes

cshung:
A few major changes for ReadyToRun:

- Refactored the logic from `ReadyToRunLanguage` to a separate class `ReadyToRunDisassembler` to avoid excessive parameter passing.
- Updated `ILCompiler.Reflection.ReadyToRun` package,
- Avoid using `CountOfUnwindCodes`, that is just a lie,
- Added a stress mode so that these bugs can be detected more easily, not included in the build by default.
- Speed up `DecorateDebugInfo` significantly using an appropriately designed data structure. (The performance problem was found with the stress mode)

### DIFF
--- a/ILSpy.ReadyToRun/ReadyToRunLanguage.cs
+++ b/ILSpy.ReadyToRun/ReadyToRunLanguage.cs
@@ -16,35 +16,88 @@
 // OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
+// #define STRESS
+
 using System;
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.Diagnostics;
 using System.Linq;
 using System.Reflection.Metadata;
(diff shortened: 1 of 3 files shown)


ILSpy.ReadyToRun/ReadyToRunLanguage.cs
```diff
@@ -16,35 +16,88 @@
 // OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
+// #define STRESS
```
siegfriedpammer, line 19: Is this meant to be always defined? Looks like an oversight from debugging to me.

cshung: Nope, this is meant to be off by default. If `STRESS` is defined, the normal output will be replaced by a dummy one, and all methods will be disassembled whenever one is selected. That is useful only for testing.

siegfriedpammer: never mind, I just realized it's commented

PR #2251, opened 2020-12-28
C# 9 Records

### LINKED ISSUE #2217: Support C# 9 records
ILSpy should output the short `record R(int a, int b);` syntax where possible.

### Compiler-generated members in records

Example records:

```
    record R(int A)
    {
        public int B;
        public int C { get; set; }
    }

    record Derived(int X) : R(X * 2);
```

#### EqualityContract

Generated unless user-declared.
```
// R:
protected virtual Type EqualityContract
{
	[CompilerGenerated]
	get
	{
		return typeof(R);
	}
}
// Derived:
protected override Type EqualityContract
{
	[CompilerGenerated]
	get
	{
		return typeof(Derived);
	}
}
```

#### Copy Constructor

Generated unless user-declared.
```
protected R(R original)
{
	b = original.b;
	C = original.C;
}
protected Derived(Derived original)
	: base(original)
{
	X = original.X;
}
```

#### Clone method

Always generated; cannot be named in user code.
```
public virtual R <Clone>$()
{
	return new R(this);
}
public override R <Clone>$()
{
	return new Derived(this);
}
```

#### Equals methods

`Equals(object?)` overload: always generated; error if also user-declared.
`Equals(R?)` overload: generated unless user-declared.

```
// R
public override bool Equals(object? obj)
{
	return Equals(obj as R);
}
public virtual bool Equals(R? other)
{
	return (object)other != null && EqualityContract == other.EqualityContract && EqualityComparer<int>.Default.Equals(A, other.A) && EqualityComparer<int>.Default.Equals(B, other.B) && EqualityComparer<int>.Default.Equals(C, other.C);
}
// Derived
public override bool Equals(object? obj)
{
	return Equals(obj as Derived);
}
public sealed override bool Equals(R? other)
{
	return Equals((object?)other);
}
public virtual bool Equals(Derived? other)
{
	return base.Equals(other) && EqualityComparer<int>.Default.Equals(X, other!.X);
}
```

#### GetHashCode

Generated unless user-declared.
```
// R:
public override int GetHashCode()
{
	return ((EqualityComparer<Type>.Default.GetHashCode(EqualityContract) * -1521134295 + EqualityComparer<int>.Default.GetHashCode(A)) * -1521134295 + EqualityComparer<int>.Default.GetHashCode(B)) * -1521134295 + EqualityComparer<int>.Default.GetHashCode(C);
}
// Derived:
public override int GetHashCode()
{
	return base.GetHashCode() * -1521134295 + EqualityComparer<int>.Default.GetHashCode(X);
}
```

#### operator==

Always generated; error if also user-declared.

```
public static bool operator ==(R? r1, R? r2)
{
	return (object)r1 == r2 || (r1?.Equals(r2) ?? false);
}

public static bool operator !=(R? r1, R? r2)
{
	return !(r1 == r2);
}

public static bool operator ==(Derived? r1, Derived? r2)
{
	return (object)r1 == r2 || (r1?.Equals(r2) ?? false);
}

public static bool operator !=(Derived? r1, Derived? r2)
{
	return !(r1 == r2);
}
```

#### ToString

Generated if not user-declared.

```
public override string ToString()
{
	StringBuilder stringBuilder = new StringBuilder();
	stringBuilder.Append("R");
	stringBuilder.Append(" { ");
	if (PrintMembers(stringBuilder))
	{
		stringBuilder.Append(" ");
	}
	stringBuilder.Append("}");
	return stringBuilder.ToString();
}
public override string ToString()
{
	StringBuilder stringBuilder = new StringBuilder();
	stringBuilder.Append("Derived");
	stringBuilder.Append(" { ");
	if (PrintMembers(stringBuilder))
	{
		stringBuilder.Append(" ");
	}
	stringBuilder.Append("}");
	return stringBuilder.ToString();
}
```

#### PrintMembers

Generated if not user-declared.

```
// R:
protected virtual bool PrintMembers(StringBuilder builder)
{
	builder.Append("A");
	builder.Append(" = ");
	builder.Append(A.ToString());
	builder.Append(", ");
	builder.Append("B");
	builder.Append(" = ");
	builder.Append(B.ToString());
	builder.Append(", ");
	builder.Append("C");
	builder.Append(" = ");
	builder.Append(C.ToString());
	return true;
}
// Derived:
protected override bool PrintMembers(StringBuilder builder)
{
	if (base.PrintMembers(builder))
	{
		builder.Append(", ");
	}
	builder.Append("X");
	builder.Append(" = ");
	builder.Append(X.ToString());
	return true;
}
```

#### Primary Constructor

Generated only for positional records. Error if also user-declared.
```
public R(int A)
{
	this.A = A;
	base..ctor();
}
public Derived(int X)
{
	this.X = X;
	base..ctor(X * 2);
}
```

#### Deconstruct

Generated only for positional records. Error if also user-declared.
```
public void Deconstruct(out int A)
{
	A = this.A;
}
public new void Deconstruct(out int X)
{
	X = this.X;
}
```

dgrunwald:
See #2217.

The decompiler now omits methods inside a record if the body looks like the default implementation that would be emitted by the C# compiler.

TODO:
 * [x] Tests
 * [x] Equals() in derived records
 * [ ] Primary constructor syntax
 * [ ] with-expressions


ICSharpCode.Decompiler/DecompilerSettings.cs
```diff
@@ -206,6 +207,24 @@ public CSharp.LanguageVersion GetMinimumRequiredVersion()
 			}
 		}
 
+		bool recordClasses = true;
+
+		/// <summary>
+		/// Use C# 9 <c>init;</c> property accessors.
```
siegfriedpammer, line 213: This should probably be fixed before the merge...

Issue #2288, opened 2021-01-30
Wrong nullable value type expression decompilation


wwh1004:
input:
``` cs
Guid? a = null;
Guid? b = null;
var expr = (Expression<Func<bool>>)(() => a == b);
```

ilspy output:
``` cs
Guid? a = null;
Guid? b = null;
Expression<Func<bool>> expr = () => (Guid)a == (Guid)b;
```

they are different semantic

```
stloc expression(ILFunction.ET[System.Linq.Expressions.Expression`1[[System.Func`1[[System.Boolean]]]]] {
	
	BlockContainer {
		Block IL_0000 (incoming: 1) {
			leave IL_0000 (call op_Equality(ldloc a, ldloc b))
		}

	}
}
)
stloc expression2(ILFunction.ET[System.Linq.Expressions.Expression`1[[System.Func`1[[System.Boolean]]]]] {
	
	BlockContainer {
		Block IL_0000 (incoming: 1) {
			leave IL_0000 (call op_Equality(expression.tree.cast System.Guid(ldloc a), expression.tree.cast System.Guid(ldloc b)))
		}

	}
}
)
```

but ilspy will decompile them to the same code

``` cs
Expression<Func<bool>> expr = () => (Guid)a == (Guid)b;
Expression<Func<bool>> expr2 = () => (Guid)a == (Guid)b;
```

ilspy version: 7.0.0.6306-preview2

wwh1004:
https://docs.microsoft.com/en-us/dotnet/api/system.linq.expressions.binaryexpression.isliftedtonull?view=net-5.0
liftToNull is different with lift
liftToNull means return value is also nullable
https://github.com/icsharpcode/ILSpy/blob/06bae54b91a61f0ec12381c893f6b64f4afa7026/ICSharpCode.Decompiler/IL/Transforms/TransformExpressionTrees.cs#L734-L738
i think this should be updated

PR #2289, opened 2021-01-30
Fix wrong nullable lift

wwh1004:
Link to issue(s) this covers
https://github.com/icsharpcode/ILSpy/issues/2288

### Problem
Wrong nullable value type expression decompilation


siegfriedpammer:

PR #2301, opened 2021-02-13
async IAssemblyResolver

dgrunwald:
I encountered a problem where sometimes the ILSpy UI freezes while decompiling a whole solution.
The issue was that the `lock (loadingAssemblies)` was held for too long (due to the blocking `GetPEFileOrNull()` calls inside the lock), and was also used on the UI thread (for the `App.Current.Dispatcher.BeginInvoke((Action)delegate () { ... lock (loadingAssemblies) { loadingAssemblies.Remove(file); }}`).

This PR eliminates the `loadingAssemblies` lock, instead handling multiple load requests for the same file within the `AssemblyList` itself.
It also changes the `DecompilerTypeSystem` to use the new `ResolveAsync` overload in order to parallelize the automatic assembly-loading.

Implementation changes:
* We no longer maintain the weird `loadingAssemblies` global state.
  * `AssemblyList` now internally handles multiple concurrent load requests for the same filename.
  * `AssemblyList.assemblies` and its lock is now private to the AssemblyList.
* `AssemblyList.OpenAssembly` is now usable on background threads; which dramatically simplifies the loading-logic in `LoadedAssembly`.
* Removed a questionable caching layer (cache was per-AssemblyList, but was caching the result of a per-LoadedAssembly lookup function.
* Replaced static `DisableAssemblyLoad()` with bool-parameter on `GetAssemblyResolver()` call.

### DIFF
--- a/ICSharpCode.Decompiler/Metadata/AssemblyReferences.cs
+++ b/ICSharpCode.Decompiler/Metadata/AssemblyReferences.cs
@@ -23,6 +23,7 @@ using System.Reflection;
 using System.Reflection.Metadata;
 using System.Security.Cryptography;
 using System.Text;
+using System.Threading.Tasks;
 
 namespace ICSharpCode.Decompiler.Metadata
 {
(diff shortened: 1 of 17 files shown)


dgrunwald:
For testing I added a `Thread.Sleep(1000);` to `LoadedAssembly.LoadAssembly`

siegfriedpammer (approved):

PR #2308, opened 2021-02-21
C# decompiler doesn't support WinRT metadata

### LINKED ISSUE #2306: C# decompiler doesn't support WinRT metadata
### Steps to reproduce
1. Open a WinRT winmd like C:\Windows\System32\WinMetadata\Windows.Foundation.winmd
2. Search for an interface like "IAsyncAction"

### Error message shown

```
System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at ICSharpCode.Decompiler.TypeSystem.MetadataModule.ResolveModule(AssemblyReferenceHandle handle)
   at ICSharpCode.Decompiler.TypeSystem.MetadataModule.GetDeclaringModule(TypeReferenceHandle handle)
   at ICSharpCode.Decompiler.TypeSystem.TypeProvider.GetTypeFromReference(MetadataReader reader, TypeReferenceHandle handle, Byte rawTypeKind)
   at ICSharpCode.Decompiler.TypeSystem.MetadataModule.ResolveType(EntityHandle typeRefDefSpec, GenericContext context, TypeSystemOptions customOptions, Nullable`1 typeAttributes, Nullability nullableContext)
   at ICSharpCode.Decompiler.TypeSystem.Implementation.MetadataTypeDefinition.get_DirectBaseTypes()
   at ICSharpCode.Decompiler.TypeSystem.Implementation.BaseTypeCollector.CollectBaseTypes(IType type)
   at ICSharpCode.Decompiler.TypeSystem.TypeSystemExtensions.GetAllBaseTypes(IType type)
   at ICSharpCode.Decompiler.CSharp.RequiredNamespaceCollector.CollectNamespacesForTypeReference(IType type)
   at ICSharpCode.Decompiler.CSharp.RequiredNamespaceCollector.CollectNamespaces(IEntity entity, MetadataModule module, CodeMappingInfo mappingInfo)
   at ICSharpCode.Decompiler.CSharp.RequiredNamespaceCollector.CollectNamespaces(IEntity entity, MetadataModule module, CodeMappingInfo mappingInfo)
   at ICSharpCode.Decompiler.CSharp.RequiredNamespaceCollector.CollectNamespaces(IEntity entity, MetadataModule module, CodeMappingInfo mappingInfo)
   at ICSharpCode.Decompiler.CSharp.RequiredNamespaceCollector.CollectNamespaces(EntityHandle entity, MetadataModule module, HashSet`1 namespaces)
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.Decompile(IEnumerable`1 definitions)
   at ICSharpCode.ILSpy.CSharpLanguage.DecompileType(ITypeDefinition type, ITextOutput output, DecompilationOptions options)
   at ICSharpCode.ILSpy.TextView.DecompilerTextView.DecompileNodes(DecompilationContext context, ITextOutput textOutput)
   at ICSharpCode.ILSpy.TextView.DecompilerTextView.<>c__DisplayClass51_0.<DecompileAsync>b__0()
```

image

steamos:
Fixes #2306

### Problem
When calling the ResolveModule(AssemblyReferenceHandle handle) method in MetadataModule.cs line 316 was returning -1, which causes an IndexOutOfRangeException when used in referencedAssemblies[row].

### Solution
* In theory I think we could only use `MetadataTokens.GetRowNumber(metadata, handle);` instead of that being the backup case but I wasn't 100% confident on that so I've added it as an edge case for now.
* [x] At least one test covering the code changed - it looks like existing tests but if you're unhappy with this I'll add a specific test case.


ICSharpCode.Decompiler/TypeSystem/MetadataModule.cs
```diff
@@ -324,6 +324,14 @@ public IModule ResolveModule(AssemblyReferenceHandle handle)
 			return LazyInit.GetOrSet(ref referencedAssemblies[row], module);
 		}
 
+		int GetRowNumber(AssemblyReferenceHandle handle)
+		{
+			var row = MetadataTokens.GetRowNumber(handle);
+			if (row == -1 )
+				row = MetadataTokens.GetRowNumber(metadata, handle);
```
siegfriedpammer, line 331: Wouldn't it be faster to always use the `MetadataReader` overload?

steamos: > Wouldn't it be faster to always use the MetadataReader overload?

It would, this is what I was talking about here :) : 
> * In theory I think we could only use `MetadataTokens.GetRowNumber(metadata, handle);` instead of that being the backup case but I wasn't 100% confident on that so I've added it as an edge case for now.

steamos: If you're more confident than I am that using the overload won't cause problems I'll refactor and remove the new method.

siegfriedpammer: I don't see why this would cause problems.

ICSharpCode.Decompiler/TypeSystem/MetadataModule.cs
```diff
@@ -313,7 +313,7 @@ public IModule ResolveModule(AssemblyReferenceHandle handle)
 
 			if (referencedAssemblies == null)
 				return ResolveModuleUncached(handle);
-			int row = MetadataTokens.GetRowNumber(handle);
+			int row = GetRowNumber(handle);
```
siegfriedpammer, line 316:
```suggestion
			int row = metadata.GetRowNumber(handle);
```

siegfriedpammer:
Thank you for investigating and fixing this issue!

kennykerr:
Is it possible to download the CI build to test the fix while I wait for the next release?

siegfriedpammer:
@kennykerr You should be able to find pre-built artifacts here: https://github.com/icsharpcode/ILSpy/actions/runs/587272614

>  At least one test covering the code changed - it looks like existing tests but if you're unhappy with this I'll add a specific test case.
@Matt-Rouse I completely overlooked this line... would it be possible for you to share which test case changed?

dgrunwald:
@siegfriedpammer That's the line from our template asking for changes to be covered by tests.
No test was affected by this change, but we should add a .winmd test to find regressions like this in the future.

kennykerr:
Thanks, that's fantastic!

Matt-Rouse:
> > At least one test covering the code changed - it looks like existing tests but if you're unhappy with this I'll add a specific test case.
> 
> @Matt-Rouse I completely overlooked this line... would it be possible for you to share which test case changed?

Apologies I worded that really badly, I meant that there's an existing test that covers this code.

> @siegfriedpammer That's the line from our template asking for changes to be covered by tests.
> No test was affected by this change, but we should add a .winmd test to find regressions like this in the future.

Agreed, I'll get a test added and PR raised to cover this case :)

PR #2309, opened 2021-02-22
Update nuget packages

siegfriedpammer:
* Dirkster.AvalonDock
* System.Composition
* OSVersionHelper
* DataGridExtensions
* Microsoft.Xaml.Behaviors.Wpf


christophwille (approved):
If something, it might be runtime behavior. Worst case: we have to revert one package upgrade.

Issue #2220, opened 2020-11-15
Mechanism to checksum assembly-compilation to determine if it has changed or not


tapika:
I have raised same issue on dotnet / core, ticket 5577.

But was wondering if it's possible to support the same functionality using ilspy , reusable functionality e.g. using ICSharpCode.Decompiler. Please note that I don't need full decompilation support, but as quick as possible mechanism to checksum assembly. See full description.

It should be possible to calculate hash or checksum over compiled assembly, so same input files would produce the same checksum.

This is useful to identify whether assembly has changed or not - and act upon that - for example execute correspondent tests or release nuget package.

Doing something like this was attempted in several ways - for example using ildasm:

1. https://www.codeproject.com/Articles/501631/Compare-Two-DLL-Files-Programmatically-Using-Hash
dumpbin:

2. https://stackoverflow.com/questions/3316505/how-can-i-know-if-an-assembly-really-did-change
mono.Cecil - programmatically:

3. https://github.com/jbevain/cecil/pull/618/files#diff-0e0bdb24805aa1a95d31f9d8561d21b80a400756610cf1d20e4392359589e2ef
GetBuildIndependentHash.

There also exists an option for Roslyn to use deterministric builds:

4. https://blog.paranoidcoding.com/2016/04/05/deterministic-builds-in-roslyn.html
To my best understanding using approach 1 or 2 is slow, because it's call to external tool. Which also might or might not work depending on version of compiler and / or tool.

Brief analysis of 4 - showed that timestamp still changes within a compilation. Suspect if I'll sign assembly - it will change even more. 3 could be working solution, but my brief trials showed the differences to appear in a middle of byte code (.text section, as a platform I have used .net framework).

.net should provide mechanism / api to be able to reliably checksum assembly to identify whether it changed or not.

Checksuming should be able to exclude:

- Assembly signature
- Build timestamp ('Time-date stamp:' in ildasm)
- GUID of debug data ('MVID' in ildasm ?)
- 'Image base' is ildasm
- Anything else that changes from build to build.

It should be possible to compile same input sources on two different machines, and then run through assembly checksum and get identical checksum after that.

dgrunwald:
The `/deterministic` switch of the C# compiler should do what you want. If it doesn't, that's a bug in the C# compiler.

Sure, you can alternatively use ILSpy to decompile the assembly and compare the decompiled code.
But I don't see what kind of change to ILSpy you are requesting here?

tapika:
[...]

PR #2320, opened 2021-02-28
feature: Initial implementation of 'disasm' command line tool

tapika:
### Problem

Initial implementation of command line tool for using ilspy. Planned to be used for calculating checksum over assembly later on ("front end" command line tool)

Relates to ticket https://github.com/icsharpcode/ILSpy/issues/2220.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
+++ b/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
@@ -103,7 +103,6 @@
     <Compile Include="Helpers\RoslynToolset.cs" />
     <Compile Include="Output\InsertParenthesesVisitorTests.cs" />
     <Compile Include="ProjectDecompiler\TargetFrameworkTests.cs" />
-    <Compile Include="TestAssemblyResolver.cs" />
     <Compile Include="TestCases\Correctness\DeconstructionTests.cs" />
     <Compile Include="TestCases\Correctness\StringConcat.cs" />
     <None Include="TestCases\ILPretty\Issue2260SwitchString.cs" />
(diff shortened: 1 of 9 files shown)


siegfriedpammer:
Thank you for your interest in contributing... May I ask how this is different from the "Show IL" option of `ilspycmd`?

https://github.com/icsharpcode/ILSpy/blob/ee10d2942b381a32efd527bcc723d366e397d299/ICSharpCode.Decompiler.Console/IlspyCmdProgram.cs#L171-L182

I wonder if it would be better to integrate the missing parts of this feature into `ilspycmd` instead of creating a new command-line application. What do you think?

siegfriedpammer:
Oh sorry, I was mislead by the name `disasm`. This is not actually related to IL disassembly but uses the C# decompiler engine. Then the question is: how is this different from the `ilspycmd` application as a whole?

https://github.com/icsharpcode/ILSpy/tree/ee10d2942b381a32efd527bcc723d366e397d299/ICSharpCode.Decompiler.Console shows the list of available options. Could you tell us, what is missing there? Thank you!

tapika:
That's a good question - I just now noticed that you have this project, but it was in separate solution - that's why I did not detect it - otherwise would probably start from it.

But now also opened Frontends.sln, which seems to be vs2017 project - it does not compiles out of box.

May I propose that we start from disasm tool, as it should be buildable and compliable - but I can change the project name and/or front end behavior as seen necessary.

siegfriedpammer:
> But now also opened Frontends.sln, which seems to be vs2017 project - it does not compiles out of box.

Works just fine with VS 2019 16.8.6 (latest). Which errors are you seeing?

> May I propose that we start from disasm tool, as it should be buildable and compliable - but I can change the project name and/or front end behavior as seen necessary.

I don't think this is a good idea, as `ilspycmd` is already deployed as dotnet tool. Getting it to compile on your machine most likely is just a matter of installing the required work loads/frameworks using VS 2019 installer.

Could you please share, which errors you are receiving when opening Frontends.sln? Thanks!

siegfriedpammer:
Did you try following our "How to build" guide? https://github.com/icsharpcode/ILSpy#how-to-build

If the guide does not fix your problem, tell us so we are able to improve it! Thanks!

tapika:
I've cleaned up git, and compilation of front end solution started to work as well.
I'll try to patch ilspycmd then, let me delete this fork / pull request, and re-create my changes in another project.

PR #2325, opened 2021-03-02
Add capability to checksum assembly

tapika:
Allow to retarget to net472 platform
Make '-p' option default, autoprobe output directory if not specified
Make class public (can be later on used as API)

### Problem
Slightly simplify usage of ilspycmd

### Solution
Easier to check ilspycmd usage (run without arguments), no output directory is required to be provided.
Project recompile is default option now (-p is not needed anymore).
Main class made public (preparing for checksum API usage)

### DIFF
--- a/ICSharpCode.Decompiler/IL/ILReader.cs
+++ b/ICSharpCode.Decompiler/IL/ILReader.cs
@@ -575,10 +575,26 @@ namespace ICSharpCode.Decompiler.IL
 		/// <summary>
 		/// Decodes the specified method body and returns an ILFunction.
 		/// </summary>
-		public ILFunction ReadIL(MethodDefinitionHandle method, MethodBodyBlock body, GenericContext genericContext = default, ILFunctionKind kind = ILFunctionKind.TopLevelFunction, CancellationToken cancellationToken = default)
+		public ILFunction ReadIL(MethodDefinitionHandle method, MethodBodyBlock body, GenericContext genericContext = default, 
+			ILFunctionKind kind = ILFunctionKind.TopLevelFunction, CancellationToken cancellationToken = default,
+			IncrementalChecksum checksumcalc = null)
 		{
 			cancellationToken.ThrowIfCancellationRequested();
 			Init(method, body, genericContext);
+
+			if (checksumcalc != null && checksumcalc.Enabled)
+			{
+				var bytes = body.GetILBytes();
+
+				if (bytes != null)
+				{
+					checksumcalc.AppendString(this.method.FullName);
+					checksumcalc.AppendAndLogBinaryData(bytes);
+				}
+
+				return null;
+			}
+
 			ReadInstructions(cancellationToken);
 			var blockBuilder = new BlockBuilder(body, variableByExceptionHandler);
 			blockBuilder.CreateBlocks(mainContainer, instructionBuilder, isBranchTarget, cancellationToken);
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ _ReSharper*/
 /ICSharpCode.Decompiler.Tests/TestCases/Correctness/*.exe
 /ICSharpCode.Decompiler.Tests/TestCases/Correctness/*.exe.config
 /ICSharpCode.Decompiler/ICSharpCode.Decompiler.nuspec
+/ICSharpCode.Decompiler.Console/retarget.props
--- a/Frontends.sln
+++ b/Frontends.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27004.2005
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.31005.135
 MinimumVisualStudioVersion = 15.0
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "doc", "doc", "{F45DB999-7E72-4000-B5AD-3A7B485A0896}"
 	ProjectSection(SolutionItems) = preProject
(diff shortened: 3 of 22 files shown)


christophwille:
Changing existing command line parameters in a non-backwards compatible way should not be done lightly. Why didn't you open an issue first to propose your changes?

tapika:
I've only changed system so that

`ilspycmd -p test.dll -o test`

would be just 

`ilspycmd test.dll`

Older variant like:

`ilspycmd test.dll -o test`

would be like:

`ilspycmd -f test.dll -o test`

Just to simplify usage. I can change current behavior to old syntax, but I highly doubt that someone uses it.

tapika:
`ilspycmd` is not autobuilt with ilspy.sln solution, and it also had separate nuget package than ilspy.

I would prefer to built with ilspy.sln and released with ilspy - it's small utility after all compared to ilspy.

christophwille:
There is a good reason for it being separate - and it having DEBUG dep on the decompiler project vs RELEASE dependency on NuGet: it started it's life as a x-platform test balloon for the decompiler engine, then turned into a dotnet tool. But still being a solution that can be opened on Mac and Linux; that is why in DEBUG it references the csproj to be able to actually develop on Linux.

tapika:
Theoretically project can be reconfigured to be compliable from ilspy.sln with net472 platform, but there still would be separate solution for linux and mac, and for .net tool. I can enable if necessary.

I want this pull request to go through, so I can prepare further work for checksum calculation (I have initial draft, need to finalize it)

christophwille:
No, we don't want that project to compile for net472. Also, why did you change DecompilerSettings?

christophwille:
The change to the sln file -why? The change for ILSpyCmdProgram to public - why? Please open an issue so we can discuss your approach to checksum calculation.

tapika:
Sln file - Im testing by myself using .net framework, if someone wants to do it as well. Easier to run exe than dll.

tapika:
Public - Im planning to use this as Api.

tapika:
Ticket was this one : https://github.com/icsharpcode/ILSpy/issues/2220

tapika:
> No, we don't want that project to compile for net472. Also, why did you change DecompilerSettings? 

Additional flag will be used by additional functionality, which I will integrate. For this commit could be left out if you want.

tapika:
If you want to instruct me better how and where I will make changes, also testing part, we can chat about it. Mail/skype/audio?!

ICSharpCode.Decompiler/IL/ILReader.cs
```diff
@@ -575,10 +575,26 @@ void InsertStackAdjustments()
...
 		/// </summary>
-		public ILFunction ReadIL(MethodDefinitionHandle method, MethodBodyBlock body, GenericContext genericContext = default, ILFunctionKind kind = ILFunctionKind.TopLevelFunction, CancellationToken cancellationToken = default)
+		public ILFunction ReadIL(MethodDefinitionHandle method, MethodBodyBlock body, GenericContext genericContext = default, 
+			ILFunctionKind kind = ILFunctionKind.TopLevelFunction, CancellationToken cancellationToken = default,
+			IncrementalChecksum checksumcalc = null)
 		{
 			cancellationToken.ThrowIfCancellationRequested();
 			Init(method, body, genericContext);
+
+			if (checksumcalc != null && checksumcalc.Enabled)
```
siegfriedpammer, line 585: This should not be part of ILReader! Why are you adding unrelated functionality in random places?

tapika: I need to know function name, so I can checksum it. Init in that phase determines function name, after that further decompilation is not needed - that's why breaking there.

But my test coverage is rather low right now, need to check deeper what kind of changes there are expected and what and where to checksum.

Do you know correct place where strings reside ?
https://github.com/tapika/ILSpy/blob/master/ICSharpCode.Decompiler.Tests/ChecksumTest.cs#L78

this test succeeds, but it should fail. Probably need to checksum data as well.

dgrunwald:
I don't understand why you're extending ILSpy here.
This feature doesn't belong in the `ILReader` class; in fact I think it doesn't belong in ILSpy at all.
You don't even need any ILSpy code to compute your checksum -- just directly use `System.Reflection.Metadata` in a separate tool.

I'm going to close this PR as we're not going to accept a checksum feature into ILSpy.

tapika:
I'm not familiar well enough with ilspy to detect best integration approach into it.

I think I've started to use ILReader just because it gets more or less decrypted logical name of function, which in a turn can be just logged out as such - easier to track down what was changed on logical level. 

I agree that maybe running checksum over metadata would get identical results, but comparing checksum result is more complex - as it would require some binary knowledge on metadata.

If you guide me and and where to change, I can direct changes into right places.

ICSharpCode.Decompiler/IL/ILReader.cs
siegfriedpammer: I suggest you take a look at how .NET assemblies are structured (see ECMA-335 https://www.ecma-international.org/publications-and-standards/standards/ecma-335/). The .NET metadata is generally organized in tables, except for strings, blobs and GUIDs, which are stored in a separate stream (called heap). Also, the IL code is stored in another section of the executable.

There are tables for type definitions, type references, method definitions, etc. Each method/type defined in an assembly has a row in the respective table. The MethodDef table for example has a column for the method name.

What ILSpy does is the following: it uses the SRM PEReader class to read the PE file (You can use this nuget package provided by Microsoft: https://www.nuget.org/packages/System.Reflection.Metadata) and then instantiates a MetadataReader, which then can be used to browse all metadata tables. For example:

```csharp
PEReader reader = ...;
MetadataReader metadata = reader.GetMetdataReader();

MethodDefinition method = metadata.GetMethodDefinition(MetadataTokens.MethodDefinitionHandle(1)); // this retrieves the data of the "first" method defined in the assembly.
string methodName = metadata.GetString(method.Name); // retrieves the name of the first method as string...
```

If you want to look at the structure of an assembly, you can use the metadata explorer included in ILSpy (beneath the assembly tree node, there is the "Metadata" node, which offers access to all metadata tables in the assembly).

I hope this helps!

tapika: It's indeed much easier to get into magical world of ECMA-335 specs without ECMA-335 specs.

I've checked through - documentation itself is ~ 500 pages long , and finding right information seems to be almost impossible.

With ilspy I have in several rounds get into point what I want to check in assembly and how it works.

Strings which I'm trying to find at the moment are called user strings, and ilspy have code snipet
`return metadata.GetUserString(MetadataTokens.UserStringHandle(blob.ReadInt32()));`

which obtains that string.

But if I want to enumerate all user strings - I was not able to find any code, which can perform something like this.

Even in test application `metadata.GetUserString` works with `UserStringHandle.FromOffset(1)` and `UserStringHandle.FromOffset(0x1B)` - so suspect strings are just serialized one after another without being able to enumerate them.

Concerning long term development strategy - what if I'll split my changes into smaller parts and then make changes in step by step manner. Maybe hacking ILReader is not best approach after all, but I need to identify best strategy of adding check summing functionality into ilspy.

siegfriedpammer: From the spec:
image

siegfriedpammer: So, you will have to either scan all method bodies for ldstr instructions and inspect their referenced user strings, or scan the heap and filter out the garbage, but there is no table for user strings, which could be read in a fast way.

Take a look at https://github.com/icsharpcode/ILSpy/blob/e08057a4a4d7f60458925765d7e06223c8ca70e0/ILSpy/Search/LiteralSearchStrategy.cs for an example on how to look for strings/user-strings. Note that the code handles all instructions individually, but you could also use ILParser.GetOperandType to simplify the code.

Hope this helps!

tapika: Btw - it's also possible to use

https://github.com/0xCM/z0/blob/fbeab69df16d550c88e65d3e402a6d62d942e068/src/cli/src/pe/reader/UserStrings.cs

`reader.GetNextHandle(handle);`

will try this later on maybe.

PR #2326, opened 2021-03-09
Expose assemblyReferences for plugin developer

CreateAndInject:


siegfriedpammer:
So basically you are asking us to provide a public API to the BAML decompiler, similar to the CSharpDecompiler API?

CreateAndInject:
@siegfriedpammer Yes

PR #2333, opened 2021-03-13
Fix insertion of result at end position

MikeFH:
### Problem
When using "sort result by fitness", some results are not sorted in correct order.

### Solution
* Fix insertion when result has to be inserted at the end (it was inserted at the second last position before)

### DIFF
--- a/ILSpy/Search/SearchPane.cs
+++ b/ILSpy/Search/SearchPane.cs
@@ -269,7 +269,7 @@ namespace ICSharpCode.ILSpy
 						return;
 					}
 				}
-				results.Insert(results.Count - 1, result);
+				results.Add(result);
 			}
 			else
 			{


siegfriedpammer:
What's the reason for this change? I don't see a bug that needs fixing on that line... please explain. Thanks!

siegfriedpammer:
Thank you for the fix... however, I wonder, wouldn't it make sense to refactor that part of the code? ... I think, it would be easy to use BinarySearch just as in the "Original Code" path. Also O(log N) vs O(N) for each result is a nice improvement.

siegfriedpammer:
I will merge this and do the refactoring on top of that... so your contribution does not get lost. Thanks!

MikeFH:
I didn't want to change too many things, but I don't mind refactoring the method as you suggested.

Issue #665, opened 2015-12-16
Dark theme


exyi:
I have spent lot of time in ILSpy last week and I'm really missing the option to set the dark theme.

LPayton314:
Agreed, theming options would be a welcome addition

AraHaan:
I dont care about theming however I do admit it would be nice if it can save Managed C++ .Net Assemblies. So that way it be easier on my tons of C# projects on my quest to translate the crap from Japanese to English. Which is a pain in the ass to do as it requires a lot of time. At least the pro to ILSpy is that I do not have to remake the stuff from scratch. However I hate having to find a way to fix the code just to get the stuff to compile. Not only that but I do not have ``PmdForm.cs`` because it failed to ``read it`` but oddly it did read everything else. I hate having to try to generate that one manually.

webmasterMeyers:
+1 A dark theme would be awesome

AraHaan:
Ok, well reading back. I should have worded it ``I dont mind about theming however ...``

benda:
Hi, I'm interested in working on this issue, can it be assigned to me ? Is this issue still needed / wanted? Just planning on adding a dropdown to the Options dialog in the "Display" tab to toggle between standard and dark themes.

christophwille:
For the WPF client, yes, feel free to work on this (no one is working on that atm).

benda:
One thing that complicates this is that you are manually drawing the tree view items (SharpTreeViewItem) - so in order to style that we'd have to add style support to the ICSharpCode.TreeView project and do that all in code which is not standard practice in the WPF world - is that OK? Not sure why that project exists really, seems like a standard tree view would work fine here? Is there something I'm missing as to what need that project serves?

AraHaan:
What is wrong with providing a version of the xaml that is dark on it and have a mode switcher that sets the style of the control based on a property (like "IsDark")?

benda:
Drawing a control to a series of images is not feasible and completely outside of all standard practice for a simple tree view.

dgrunwald:
Does the standard WPF tree view do multiple selection?

Also we use:
 * drag'n'drop to reorder
 * potentially extremely deep nesting (esp. in the analyzer) -- afaik the WPF treeview crashes once you get to ca. 130 nesting levels

benda:
[...]

PR #2334, opened 2021-03-13
Dark theme #665

tom-englert:
Link to issue(s) this covers

### Problem
Dark theme #665

### Solution
* just using the themable styles from https://www.nuget.org/packages/TomsToolbox.Wpf.Styles/
* fixed some style references and replaced hardcoded colors
* added syntax highlighting schemes for dark mode


Dirkster99:
Pretty cool work :-) but why don't you use the same AvalonDock Style Theme (eg VS 2013) for Dark and Light? This way you would end up using a consistent look and feel for document and tool windows and docking buttons - look at the image below to see the inconsistencies:

Untitled4

Untitled3

christophwille:
@tom-englert Once you are done, how about the following: we tweet with request for comment on this PR for early testing / feedback from devs that did dark theme in their apps with artifacts from https://github.com/icsharpcode/ILSpy/actions/workflows/build-ilspy.yml?query=workflow%3A%22Build+ILSpy%22+branch%3ADarkTheme%23665+is%3Asuccess

siegfriedpammer:
Note: The AvalonEdit fold margin will need some adjustments as well...

Also: inline buttons need extra styling...
image

siegfriedpammer:
Nit: the light theme should retain the typical ILSpy look (light yellowish background of the decompiler view).

tom-englert:
@Dirkster99 missed that, thanks for mentioning...

tom-englert:
> Note: The AvalonEdit fold margin will need some adjustments as well...

Where exactly?

siegfriedpammer:
I just tried adding the following lines to DecompilerTextView::.ctor

			textEditor.SetValue(FoldingMargin.SelectedFoldingMarkerBackgroundBrushProperty, Brushes.LightGreen);
			textEditor.SetValue(FoldingMargin.FoldingMarkerBackgroundBrushProperty, Brushes.Green);

and it seems to work... not sure how you are able to use this with theming...

tom-englert:
Generally not a problem, but these properties are implemented wrong in the `FoldingMargin` class, so they are not usable in XAML.

siegfriedpammer:
As AvalonEdit is maintained by @dgrunwald and me (just as ILSpy), it would suffice if you open an issue describing the problem and the fix. I will then apply your suggested fix, so you don't have to provide two PRs... thanks!

Dirkster99:
@siegfriedpammer Maybe its time for an update/extension in AvalonEdit where you support {themable highlightings](https://github.com/Dirkster99/AvalonEditHighlightingThemes) out of the box :-) Looking at @tom-englert stuff I am sure he can take the idea and push it even further (maybe for a later release). It would make your ILSpy live easier and I am sure there are a few people (including myself) waiting around on this one to be integrated in AvalonEdit :-)

siegfriedpammer:
@christophwille @dgrunwald what do you think?

christophwille:
For AvalonEdit, our mantra is low-risk changes because ILSpy as the only inhouse consumer of AvalonEdit doesn't test all areas like SharpDevelop did back in the day. Also, it is a question of "bang for the buck" vs time spent: how big a customer base is this going to be helping? (and maybe we should be moving that part of the discussion over to a new issue in AE repo)

christophwille:
I take that @tom-englert needs to adapt the fold margins to the fixed attached properties in AvalonEdit 6.1.1?

tom-englert:
Actually they are working fine, so it can be tested.
I can clean up the code to use proper Xaml notation probably next weekend.

xoofx:
If you could make it also optionally OS settings dependent - that would be amazing

siegfriedpammer:
Thank you, that's a good idea. However, ILSpy 7.0 is really close to being released and dark mode is considered an experimental feature, which still needs a lot of testing. For this reason, we will not add an automatic switch to dark mode for 7.0. For 7.1, we will consider adding it. Thank you for your understanding!

christophwille:
@tom-englert What is the status in terms of things to do? Asking to get an idea about when the merge might happen.

tom-englert:
From MPOV it's done

tom-englert:
@siegfriedpammer p.s. the update of Avalon Edit worked fine

siegfriedpammer:
More feedback...

image

Highlighting of local variable references and search results needs adjustment in dark mode.

image

Also the "wait bar" shown when decompiling is not adjusted for dark mode yet.

The SearchBar of AvalonEdit might need adjustments as well...

Not sure which of these are really high-priority, as I am not a dark mode user myself...

siegfriedpammer:
image

Hyperlinks might need a different color to make them look good.

tom-englert:
> Hyperlinks might need a different color to make them look good.

@siegfriedpammer any hint which highlighting scheme is used for the about view?

siegfriedpammer:
Thanks you very much for adding this!

siegfriedpammer:
@tom-englert well, I found something else, which should be fixed before the release:

image

image

Do you know why https://github.com/icsharpcode/ILSpy/blob/60035befbfc3b08081f86ea0ef339400450cf631/ILSpy/Metadata/Helpers.cs#L57 is working no longer?

Also changing the borders around the column headers to be less visible, would be nice... If you have an idea on how to fix this, let me know. Thanks!

christophwille:
How about we open an issue for "Dark Theme Improvements / Bugs" so Tom has only one (current) item to monitor? We could even pin it for others to know that this is where comments / questions should go.

PR #2335, opened 2021-03-13
Refactor searchresult insert

MikeFH:
Refactoring of the InsertResult method of the SearchPane to use BinarySearch in all cases for sorted insertion, as discussed in #2333
Also contains a small typo fix.

### DIFF
--- a/ILSpy/Search/SearchPane.cs
+++ b/ILSpy/Search/SearchPane.cs
@@ -49,6 +49,7 @@ namespace ICSharpCode.ILSpy
 		const int MAX_REFRESH_TIME_MS = 10; // More means quicker forward of data, less means better responsibility
 		RunningSearch currentSearch;
 		bool runSearchOnNextShow;
+		IComparer<SearchResult> resultsComparer;
 
 		public static readonly DependencyProperty ResultsProperty =
 			DependencyProperty.Register("Results", typeof(ObservableCollection<SearchResult>), typeof(SearchPane),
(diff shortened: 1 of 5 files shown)


ILSpy/Search/SearchPane.cs
```diff
@@ -253,30 +254,11 @@ async void StartSearch(string searchTerm)
...
-				}
-				results.Add(result);
-			}
-			else
-			{
-				// Original Code
-				int index = results.BinarySearch(result, 0, results.Count - 1, SearchResult.Comparer);
-				results.Insert(index < 0 ? ~index : index, result);
-			}
+			var comparer = DisplaySettingsPanel.CurrentDisplaySettings.SortResults ?
```
siegfriedpammer, line 257: this could be moved to the start of the search

MikeFH:
I noticed a small problem in the fitness calculation when the search result is a destructor since the method name is always Finalize.
This can lead to a weird sorting of results, for example in a search for "memorystream".

I propose to replace
https://github.com/icsharpcode/ILSpy/blob/97334ef5a7142e88b0d1662fd267753c28db46d4/ILSpy/Search/AbstractEntitySearchStrategy.cs#L82
by
`if (member.SymbolKind == SymbolKind.Constructor || member.SymbolKind == SymbolKind.Destructor)`

Do you think this should be the subject of another issue or another PR or should I include the modification in this PR ?

siegfriedpammer:
> I noticed a small problem in the fitness calculation when the search result is a destructor since the method name is always Finalize.
> This can lead to a weird sorting of results, for example in a search for "memorystream".
> 
> I propose to replace
> https://github.com/icsharpcode/ILSpy/blob/97334ef5a7142e88b0d1662fd267753c28db46d4/ILSpy/Search/AbstractEntitySearchStrategy.cs#L82
> 
> 
> by
> `if (member.SymbolKind == SymbolKind.Constructor || member.SymbolKind == SymbolKind.Destructor)`
> Do you think this should be the subject of another issue or another PR or should I include the modification in this PR ?

Ah, yes, sorry, I just realized you had this question posted... I think, using symbol kind instead of the name is a good idea. If you want, you can provide another PR.

PR #2343, opened 2021-03-19
Fix url of SearchMsdn when do search with generic types and .ctor methods

yyjdelete:
Before: https://docs.microsoft.com/dotnet/api/system.collections.generic.dictionary.keycollection..ctor
After: https://docs.microsoft.com/dotnet/api/system.collections.generic.dictionary-2.keycollection.-ctor

### DIFF
--- a/ILSpy/Commands/SearchMsdnContextMenuEntry.cs
+++ b/ILSpy/Commands/SearchMsdnContextMenuEntry.cs
@@ -56,13 +56,13 @@ namespace ICSharpCode.ILSpy
 				if (node is EventTreeNode eventNode && (!eventNode.IsPublicAPI || !IsAccessible(eventNode.EventDefinition)))
 					return false;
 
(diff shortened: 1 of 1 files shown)


siegfriedpammer:
Maybe this will need adjustments for static constructors (.cctor) and operators (op_) as well?

yyjdelete:
operators already works
https://docs.microsoft.com/dotnet/api/system.decimal.op_explicit

`.cctor` should always be private and already be disabled by checks in `bool IsEnabled(TextViewContext context)`.

----

But I'm not sure which is better for enum and delegate, add special logic for them to use DeclaringType instead, or simply disable it in IsEnabled, or just leave it as it is before.
There is only
https://docs.microsoft.com/en-us/dotnet/api/system.dayofweek
but not
https://docs.microsoft.com/en-us/dotnet/api/system.dayofweek.sunday
https://docs.microsoft.com/en-us/dotnet/api/system.dayofweek.value__

siegfriedpammer:
Enum values and delegate members should just return false for IsEnabled, I think.

siegfriedpammer:

PR #2344, opened 2021-03-20
Adjust fitness calculation for desctructors

MikeFH:
As discussed in #2335, adjust calculation of fitness for destructors.


siegfriedpammer:
Thank you very much!

PR #2345, opened 2021-03-20
Fix warning CS8625

riQQ:
### Problem
Warning CS8625: Cannot convert null literal to non-nullable reference type.

### Solution
Fix warning by making the parameter nullable.


siegfriedpammer:

PR #2347, opened 2021-03-21
Dark theme #665 fixes

tom-englert:
@siegfriedpammer as requested in #2334

### DIFF
--- a/ILSpy/ILSpy.csproj
+++ b/ILSpy/ILSpy.csproj
@@ -48,7 +48,7 @@
     <PackageReference Include="OSVersionHelper" Version="1.1.24" />
     <PackageReference Include="DataGridExtensions" Version="2.5.0" />
     <PackageReference Include="Microsoft.Xaml.Behaviors.Wpf" Version="1.1.31" />
-    <PackageReference Include="TomsToolbox.Wpf.Styles" Version="2.4.2" />
+    <PackageReference Include="TomsToolbox.Wpf.Styles" Version="2.4.3" />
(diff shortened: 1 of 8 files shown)


siegfriedpammer:
Wow... you are really fast! Thanks for fixing this...

I just found another problem:

image

I am unable to find the correct location to change the background color of the tab that the cursor is currently hovering over and also the text color of the explanation looks a bit strange if everything else is black.

Sorry for throwing so many things at you currently. If I were more versed in WPF theming I would fix it myself. Thank you so much!

tom-englert:
@siegfriedpammer no problem, better you find them before they make it into the release 😏

siegfriedpammer:
I found another issue: The access keys of the buttons in the "Manage assembly lists..." dialog are no longer working properly... Seems the `ContentPresenter` is not setting the `RecognizesAccessKey` property to `true` somewhere.

christophwille:
> 
> 
> I found another issue: The access keys of the buttons in the "Manage assembly lists..." dialog are no longer working properly... Seems the `ContentPresenter` is not setting the `RecognizesAccessKey` property to `true` somewhere.

@tom-englert Are you going to add the fix for this to this PR?

tom-englert:
Just pushed...

siegfriedpammer:
Thank you for fixing these issues so quickly!

tom-englert:
@siegfriedpammer P.S. thanks for the thorough testing, pointing me to views and features I missed to test - most of them I even didn't know they exist. 😄

PR #2346, opened 2021-03-20
modernize ILSpy.AddIn.csproj

tom-englert:
Link to issue(s) this covers

### Problem
#2323 => ILSpy.AddIn project not in new SDK style

### Solution
Update this project, too.

PR #2350, opened 2021-03-22
Net5take2 b

tom-englert:
2nd part for #2346


christophwille:
@sharwell we'd like to cherry-pick that PR as well - which changes should be included if we wanted to remove the workaround as pointed out by Tom in #2346? (note that today it seems to be a mix of #2346 and the "second part")

tom-englert:
I think there's not much cherry picking left, it's more like take it or leave it - unless you want to ignore DRY and SPOM and resemble what the VSIX-SdkProjectAdapter does, implementing the same by your own.

christophwille:
Sorry, I know nothing about the specifics in the VS addin csproj file format. Therefore I have to ask.

sharwell:
I'm not a huge fan of hiding the VSIX build configuration behind a new package. For me, it make it harder to understand what the build is doing because it no longer follows the default behavior for VSIX builds.

Aside from that, I'm _very much not_ a fan of the way [PowerShell is used](https://github.com/tom-englert/VSIX-SdkProjectAdapter/blob/ab497047b8fbe65c7b0e226734dbc611630c0fcb/build/VSIX-SdkProjectAdapter.targets#L10-L13) in the VSIX-SdkProjectAdapter build targets. If you want to inject a value for `Version` in the vsixmanifest, all you need to do is set `Version="|%CurrentProject%;GetVsixVersion|"` in your vsixmanifest and define the [`GetVsixVersion` target in your project](https://github.com/tunnelvisionlabs/InheritanceMargin/blob/7c3ebf3d71d0185d8b6f74a4b4c1c99f3f9217e6/Tvl.VisualStudio.InheritanceMargin/Tvl.VisualStudio.InheritanceMargin.csproj#L94-L98).

tom-englert:
> I'm not a huge fan of hiding the VSIX build configuration behind a new package. For me, it make it harder to understand what the build is doing because it no longer follows the default behavior for VSIX builds.

- what exactly do you mean by the "default behavior for VSIX builds"? Where is this defined? Officially there is no support for VSIX in SDK-style projects 
- the paradigm of SDK-style projects is to stop duplicating all the defaults in every project (DRY), so you should not use SDK-style projects at all, they are hiding so many defaults...

> all you need to do is set [`Version="|%CurrentProject%;GetVsixVersion|"`]
if that would have been documented somewhere, I'd used this from the beginning. I'll definitely gonna change this!

christophwille:
So, provocatively speaking, we'd "paper over" a few cracks in project format and exchange the devil we know (the workaround) for a NuGet package that is the "filler". Did I get that options comparison close to right?

tom-englert:
@sharwell: updated the VSIX-SdkProjectAdapter, the PowerShell is now gone...
@christophwille: the devil always will catch you earlier or later, and the NuGet package is the "exorzist" 😏; I'm using that e.g. here [dotnet/ResXResourceManager](https://marketplace.visualstudio.com/items?itemName=TomEnglert.ResXManager), which uses quite a lot of runtime NuGet packages, so this is proven to not break anything

sharwell:
I'm not going to _block_ this if someone wants to merge it, but I would caution the maintainers to unzip the resulting VSIX before and after this change and do a directory diff to make sure there aren't any unintended consequences.

tom-englert:
I have already tested this, the only difference is that it now contains all  binaries from the mono.cecil package, so that is now on the safe side

sharwell:
@tom-englert the concept seems cool. I'm not familiar with the specific way NuGet is being used to resolve dependencies. How do you envision the automatic assembly inclusion to play with `[ProvideCodeBase]`, considering packages need a 1:1 mapping of included assemblies to this attribute?

tom-englert:
@sharwell VSIX-SdkProjectAdapter only ensures that all the necessary targets are executed in the same order as when using the old style project format, so the VSIX container is generated and the content is the same.

It does not deal with the `[ProvideCodeBase]` attribute. Which use case do you have in mind?

sharwell:
See here for an example of registering dependencies that are included in a VSIX package:
https://github.com/laurentkempe/GitDiffMargin/blob/05cc232d8e6507302ea1d2958c50a68893a6436a/GitDiffMargin/Properties/AssemblyInfo.cs#L39-L53

tom-englert:
@sharwell I know, I'm using this in my extensions, too: https://github.com/dotnet/ResXResourceManager/blob/master/src/ResXManager.VSIX/Properties/AssemblyInfo.cs

However VSIX-SdkProjectAdapter does not automate the inclusion of NuGet dependencies, it just restores the build behavior as it was with the original old-style project format, where including NuGet packages already worked the way it is now.

Automating the ProvideCodeBase behavior is a good point I was already thinking about, but not really related to this PR. Can we move this discussion somewhere else?

christophwille:
How about to https://github.com/tom-englert/VSIX-SdkProjectAdapter/issues with a link back to where it started?

tom-englert:
Closed as stale

PR #2357, opened 2021-03-26
Modernize Visual Studio AddIn csproj part 1

siegfriedpammer:
Cherry-picking changes from #2346 as requested by @sharwell.

@sharwell  are you OK with these changes?
@tom-englert fyi

### DIFF
--- a/ILSpy.AddIn/ILSpy.AddIn.csproj
+++ b/ILSpy.AddIn/ILSpy.AddIn.csproj
@@ -2,9 +2,6 @@
 <Project>
   <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.props" />
 
-  <PropertyGroup>
-    <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
-  </PropertyGroup>
 
   <PropertyGroup>
     <TargetFramework>net472</TargetFramework>
@@ -18,6 +15,7 @@
 
     <EnableDefaultItems>False</EnableDefaultItems>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <DefineConstants>TRACE;VSADDIN</DefineConstants>
 
     <SignAssembly>True</SignAssembly>
     <AssemblyOriginatorKeyFile>Key.snk</AssemblyOriginatorKeyFile>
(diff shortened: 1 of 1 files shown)


ILSpy.AddIn/ILSpy.AddIn.csproj
```diff
@@ -1,10 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project>
-  <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.props" />
-
-  <PropertyGroup>
-    <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
-  </PropertyGroup>
+<Project Sdk="Microsoft.NET.Sdk">
```
sharwell, line 2:
```suggestion
<Project>
  <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.props" />
```

sharwell (commented):
I'm fine with this except for the one suggested change

ILSpy.AddIn/ILSpy.AddIn.csproj
sharwell: 📝 The shorthand syntax (where these two lines are combined) implicitly adds two imports:

* It imports **Sdk.props** as the first item in the file
* It imports **Sdk.targets** as the last item in the file

Due to the requirement that **Microsoft.VsSDK.targets** be imported _after_ **Sdk.targets**, we cannot use the shorthand form and must instead explicitly import both **Sdk.props** and **Sdk.targets**.

siegfriedpammer: Fixed! OK?

sharwell (approved):

ILSpy.AddIn/ILSpy.AddIn.csproj
sharwell: Yep!

tom-englert: > The shorthand syntax (where these two lines are combined) implicitly adds two imports:
> 
> * It imports **Sdk.props** as the first item in the file
> * It imports **Sdk.targets** as the last item in the file
> 
> Due to the requirement that **Microsoft.VsSDK.targets** be imported _after_ **Sdk.targets**, we cannot use the shorthand form and must instead explicitly import both **Sdk.props** and **Sdk.targets**.

=> This is exactly what the VSIX-SdkProjectAdapter package fixes.

Issue #2363, opened 2021-04-01
Decompile multiple dlls / build sln


marwie:
**Is your feature request related to a problem? Please describe.**
related to https://github.com/icsharpcode/ILSpy/issues/972

**Describe the solution you'd like**
I have multiple dlls that I want to build csprojs for (or resolve into one project)
Is it possible to give the CLI more than one dll to decompile and correctly build references? 
Could the process also generate a solution?

**Additional context**
https://github.com/icsharpcode/ILSpy/tree/master/ICSharpCode.Decompiler.Console

marwie:
My basic questions are:
How would I go about decompiling one or multiple projects from a couple of dlls in a folder.

Should decompiling every single dll with the same output folder result in a correct project? 
Should I just decompile each dll into its own output folder?
Should ``-r|--referencepath <path>`` point to the path of the dll that is being decompiled (assuming it contains all its dependencies)?

PR #2364, opened 2021-04-03
Cli support for generating solution for multiple projects

marwie:
Link to issue(s) this covers
https://github.com/icsharpcode/ILSpy/issues/2363
https://github.com/icsharpcode/ILSpy/pull/1550

### Problem
Commandline interface does not expose a feature to compile multiple assemblies into a solution

### Prefix
This PR should be considered as a possible basic solution that I need feedback on to properly fix/implement :)

### Solution
I added support to use multiple dll files as ``InputAssemblyName`` arg (used like this ``ilspycmd "lib1.dll lib2.dll" <other arguments>``)
And provided with a output solution path as ``--sln`` it would call ``WriteSolutionFile`` as implemented in https://github.com/icsharpcode/ILSpy/issues/2363

But generally speaking I'm not sure if this is the correct approach to decompile multiple assemblies/projects. For example when decompiling both UnityEditor.dll and UnityEngine.dll to the same output path both projects would contain the same files (as everything is in one directory)

So the problem for that is is that the output path for an assembly must be empty/multiple projects must be output to separate output directories, maybe implicitly generated in the output directory from the dll names?

### Alternative ideas
Another idea would be to provide an --sln argument to multiple generated ``.csproj`` files which then would invoke the ``WriteSolutionFile`` call. But this would require to collect the necessary meta information from previous compiled/the csproj(?)
which currently is ``string projectFile, string projectPlatform, Guid projectGuid, Guid typeGuid``.

Would be interested to hear what you think/prefer and/or if there are better solutions already or in process?! :)


siegfriedpammer:
05eb2cdf049abfadb975a2eb6a403848fa3f0bd7 implements this feature by migrating your code to ILSpy 8.0.

ilspycmd now accepts multiple assembly file names. If more than one filename and the "-p" option is given, a solution file is automatically generated in the root output directory. For every project a sub-directory is created and the project file and source files are placed within. Sorry for taking so long to get to this.

PR #2373, opened 2021-04-19
Support for single-file bundle from .NET 6.

vitek-karas:
Adds support for the new version and also the new compressed files.

Reacts to .NET changes:
https://github.com/dotnet/runtime/pull/49855
https://github.com/dotnet/runtime/pull/50817

### DIFF
--- a/ILSpy/LoadedPackage.cs
+++ b/ILSpy/LoadedPackage.cs
@@ -197,7 +197,24 @@ namespace ICSharpCode.ILSpy
 			public override Stream TryOpenStream()
 			{
 				Debug.WriteLine("Open bundle member " + Name);
-				return new UnmanagedMemoryStream(view.SafeMemoryMappedViewHandle, entry.Offset, entry.Size);
+
+				if (entry.CompressedSize == 0)
+				{
+					return new UnmanagedMemoryStream(view.SafeMemoryMappedViewHandle, entry.Offset, entry.Size);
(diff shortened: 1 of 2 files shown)


christophwille:
In case you missed the milestone being added - we will merge this PR once 7.0 is released.

ILSpy/LoadedPackage.cs
```diff
@@ -197,7 +197,24 @@ public BundleEntry(string bundleFile, MemoryMappedViewAccessor view, SingleFileB
...
+				{
+					Stream compressedStream = new UnmanagedMemoryStream(view.SafeMemoryMappedViewHandle, entry.Offset, entry.CompressedSize);
+					using var deflateStream = new DeflateStream(compressedStream, CompressionMode.Decompress);
+					Stream decompressedStream = new MemoryStream((int)entry.Size);
+					deflateStream.CopyTo(decompressedStream);
+					if (decompressedStream.Length != entry.Size)
+					{
+						throw new InvalidDataException($"Corrupted single-file entry '${entry.RelativePath}'. Declared decompressed size '${entry.Size}' is not the same as actual decompressed size '${decompressedStream.Length}'.");
+					}
+
```
dgrunwald, line 215: Don't we have `decompressedStream.Position == decompressedStream.Length` at this point?
I think we need to seek back to the start before returning the stream.

Issue #2359, opened 2021-03-28
UI and theming issues


siegfriedpammer:
This issue is to track all the remaining questions/issues regarding UI and theming:

/cc @tom-englert fyi

* [x] When hovering over the scrollbars the cursor changes in an unexpected way. It should display the default cursor, however, currently the used cursor is the one for vertical/horizontal resize.
![scrollbar_cursor](https://user-images.githubusercontent.com/344204/112747585-9ca7e200-8fb6-11eb-8d14-6e127674ac36.gif)
~~Will have to investigate where the alternate cursors for the scrollbar thumb are defined.~~ found it:
https://github.com/tom-englert/TomsToolbox/blob/master/src/TomsToolbox.Wpf.Styles/Themes/ScrollViewerStyle.xaml#L55-L117

siegfriedpammer:
* [ ] For me it's a bit irritating that the currently hovered menu item does not change its background color - the slight change in the text color is not enough visual feedback in my opinion.
* [ ] Another problem is that the shortcuts are no longer visible in the menu
![image](https://user-images.githubusercontent.com/344204/113197807-6f349000-9265-11eb-81c8-7b7d09c05fc8.png)

christophwille:
Showing the shortcuts is definitely necessary (how would you learn about the shortcut otherwise?). And yes, similarity to VS is a good thing.

tom-englert:
https://github.com/icsharpcode/ILSpy/commit/6840adb2fd08cc2f876f4b3bea303b17a71e6e27 fixes the menu style

christophwille:
![image](https://user-images.githubusercontent.com/344208/115513635-90e0d000-a283-11eb-9d99-7738d5c6cc78.png)

@tom-englert would we happen to have radio buttons too? (because it is a radio and not checkbox scenario)

tom-englert:
Not out of the box - there is no such design in the standard...

tom-englert:
We would need to have a special style for this

christophwille:
It is only irritating because check boxes indicate multi-selection (well-aware that this is menu default, but a checkmark doesn't look as distracting as a checkbox).

tom-englert:
I'll have a look

tom-englert:
Got it - that should be only a check-mark, without the border making it look like a check-box

tom-englert:
@christophwille checkmark is fixed

christophwille:
One thing that I noticed already way back - Notepad focused:

[...]

PR #2375, opened 2021-04-21
#2359: Adjust checkable style of MenuItem

tom-englert:


christophwille:
Thank you very much! The checkmark looks way nicer.

PR #2376, opened 2021-04-25
Added .vsconfig to solution

### LINKED ISSUE #2200: Provide a .vsconfig
Basically "codify" what we have in our README/How to build. See https://devblogs.microsoft.com/setup/configure-visual-studio-across-your-organization-with-vsconfig/

joshuacookdev:
### Problem
Resolves #2200 - Provide a .vsconfig

### Solution
* Added .vsconfig with the 3 workloads/components listed in the readme
  * Selected _(Latest)_ for the "MSVC" component, can be changed to the _(v14.23)_ listed if desired.
* Tested on clean install of VS in Windows Sandbox to verify it built after correctly prompting to install components


joshuacookdev:
VS defaulted this to spaces when I created it, and tidy didn't pick it up for some reason. I can adjust if that's preferred, just let me know.

dgrunwald:
There's a whole bunch of components in that list that seem unnecessary for ILSpy.
E.g. im pretty sure I don't have `Component.Microsoft.VisualStudio.LiveShare` installed anymore, because that component used to cause trouble a few updates ago.

This list needs to be reduced to the bare minimum that's relevant for the ILSpy build.

joshuacookdev:
That's odd - when VS built the vsconfig, the only boxes that were checked were the 3 listed. I'll go through and verify those again @dgrunwald! Thanks for catching that.

joshuacookdev:
@dgrunwald - do you want me to include:

* C++ component (for editbin.exe)
* VS Extension workload (for extension project)

or do you just want the default .NET Desktop development (which is enough to run the regular build of ILSpy)?

christophwille:
I'd say extension dev is necessary because that would mean build errors for ilspy.sln  - whereas editbin doesn't fail the build.

joshuacookdev:
@christophwille - I'll say that when I ran a pared-down build in Sandbox, I was able to build the solution without the extensions workload at all. I'll include it anyway (minus the optional components, obviously) since you currently mark it as required.

Also curious why the ILSpy.WithPackage requires the UWP workload when I try to run it, if you have the time to answer that for me

christophwille:
ILSpy.WithPackage is separate because it builds the Microsoft Store package of ILSpy. It is split out because we want to keep the requirements down for ILSpy development (and the build times as well).

joshuacookdev:
@christophwille @dgrunwald 

Updated the components and readme as requested. Let me know if either of you have any further change requests. Thanks again for all of the feedback thus far!

PR #2377, opened 2021-04-26
Switch to dotnet6-transport

cshung:
The change uses the `dotnet6-transport` feed to replace my personal feed to obtain `ILCompiler.Reflection.ReadyToRun.dll`.


christophwille:
@cshung is there any timeline when that switch might happen? (we are post-rtm on 7.0 so we can take new things)

cshung:
> @cshung is there any timeline when that switch might happen? (we are post-rtm on 7.0 so we can take new things)

I think we can do the switch now. I have just confirmed with the engineering team that this feed will practically last forever. (Until the engineering team moves off Azure DevOps, which I don't think is likely)

christophwille:
One thing - please keep your existing feed online because otherwise source builds of existing older versions will break for r2r.

PR #2385, opened 2021-05-01
Translate new words to Chinese

maikebing:


siegfriedpammer:
Thanks for updating the translation!

PR #2386, opened 2021-05-01
Fix invalid il from identifiers containing repeating dots

DarkOoze:
### Problem
Decompiling of assembly containing a identifier with two dots result of il that is not valid by ilasm.exe.

Test code used [test.zip](https://github.com/icsharpcode/ILSpy/files/6409584/test.zip).

```
ilasm.exe /DLL .\test_decompiled.il
```

```
.\test_decompiled.il(28) : error : syntax error at token '.' in:        .field private static int32 foo..bar
```

### Solution

Add quotes to any identifier that contains repeating dots (i.e. "..").

### DIFF
--- a/ICSharpCode.Decompiler/Disassembler/DisassemblerHelpers.cs
+++ b/ICSharpCode.Decompiler/Disassembler/DisassemblerHelpers.cs
@@ -52,6 +52,8 @@ namespace ICSharpCode.Decompiler.Disassembler
 
 	public static class DisassemblerHelpers
 	{
+		static readonly char[] _validNonLetterIdentifierCharacter = new char[] { '_', '$', '@', '?', '`', '.' };
+
 		public static string OffsetToString(int offset)
 		{
 			return string.Format("IL_{0:x4}", offset);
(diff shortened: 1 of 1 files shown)


ICSharpCode.Decompiler/Disassembler/DisassemblerHelpers.cs
```diff
@@ -317,7 +319,7 @@ public static string EscapeString(string str)
 						// print control characters and uncommon white spaces as numbers
 						if (char.IsControl(ch) || char.IsSurrogate(ch) || (char.IsWhiteSpace(ch) && ch != ' '))
 						{
-							sb.Append("\\u" + ((int)ch).ToString("x4"));
+							sb.AppendFormat("\\u{0:x4}", ch);
```
siegfriedpammer, line 322: This change introduces a bug. There is a good reason for the cast to `int` here: `char.ToString()` is not the same as `int.ToString()`.

See this for example:

```
> string.Format("\\u{0:x4}", '\u0041')
"\\uA"
> string.Format("\\u{0:x4}", (int)'\u0041')
"\\u0041"
```

image


```
System.Text.EncoderFallbackException: 'Unable to translate Unicode character \uDB80 at index 309 to specified code page.'
```

Other than that your changes seem fine...

DarkOoze: Thanks, will fix that and rebase to master.

siegfriedpammer:
Thank you for providing this bugfix!

PR #2408, opened 2021-05-11
Fix #2406: for readonly property with only getter, move readonly to property instead of getter

### LINKED ISSUE #2406: CS8664: 'readonly' can only be used on accessors if the property or indexer has both a get and a set accessor
### Input code
```cs
    public struct C
    {
//Or disable `Use expression-bodied member syntax for get-only properties` in C# 6.0
        public readonly int X
        {
            get
            {
                int j = 0;
                for (int i = 0; i < 10; ++i)
                {
                    j += i;
                }
                return j;
            }
        }
    }
```

### Erroneous output
```c#
public struct C
{
	public int X
	{
		readonly get
		{
			int j = 0;
			for (int i = 0; i < 10; i++)
			{
				j += i;
			}
			return j;
		}
	}
}
```
>error CS8664: 'readonly' can only be used on accessors if the property or indexer has both a get and a set accessor

### Details
* Product in use: ILSpy
* Version in use: 7.0.0

yyjdelete:
Link to issue(s) this covers

### Problem
#2406

### Solution

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/RefLocalsAndReturns.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/RefLocalsAndReturns.cs
@@ -98,6 +98,13 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 				}
 			}
 
+			public readonly int ReadOnlyPropertyWithOnlyGetter {
+				get {
+					Console.WriteLine("No inlining");
+					return 1;
+				}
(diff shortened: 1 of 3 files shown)


yyjdelete:
Seems another thing is also affected, auto property with only getter is marked as `readonly` in the left tree now.

`public int Y { get; }`

Before
>Y : int

After
>readonly Y : int

siegfriedpammer:
Thank you very much for contributing this fix!

PR #2412, opened 2021-05-17
Fix #2409: Styling of windows title bar is now optional.

### LINKED ISSUE #2409: ILSpy's titlebar does not differentiate between focused/active and inactive
## Steps to reproduce
1. Open ILSpy 7+ as a maximized window on one monitor.
2. Open Notepad on another monitor.
3. In ILSpy, click around on things
4. Alt-tab to notepad
5. Observe that the Notepad titlebar changed color when Notepad gained focus (or at least its text color did with default Windows 10 settings), but ILSpy's titlebar, nor _any_ part of its window, changed appearance when ILSpy lost focus.
    * It's even worse than that: the current document-tab in ILSpy retains the "active" color even when the window doesn't have focus, which is misleading.
      * Many other programs will use a different color for the active document tab, such as a slightly desaturated version of the active color, but not grey either. 

(After I wrote this, I realized that ILSpy7 _does_ very subtly change the opacity of its custom window drop-shadow, but this is almost invisible on a high-DPI display or if you have a dark desktop wallpaper that doesn't have sufficient contrast to make drop-shadows visible).

## Expected results

* ILSpy's titlebar should still respect the user's color and theming settings to indicate if it has focus or not.

## Actual results:

* **It is currently almost impossible to tell visually if ILSpy has focus or not** (and it may as well be fully impossible if you have impaired vision). This is a pain if I start typing when I think that ILSpy has focus when it doesn't - or vice-versa - and my keystrokes end-up in the wrong program.

### Preferred solution

* Bring back the Windows default rendering of the nonclient area of the ILSpy application window.
  * I'm not saying to go ditch the current WPF application-wide skin that ILSpy has started using - I just don't think it's a good idea to skin or customize the nonclient area as right now it not only adds zero value, but takes away useful visual hints for the user.

### Alternative solution

* Set the WPF skin's nonclient area's customized titlebar's background color to use the user's system color for active and inactive windows as per the user's settings.

tom-englert:
Fixes #2409: 
If you like to have the title bar styled along with the theme you can now opt via View-Options-Display. 
Default is no styling.
It also fixes the border color mentioned in #2359

### DIFF
--- a/ILSpy/AboutPage.cs
+++ b/ILSpy/AboutPage.cs
@@ -34,6 +34,7 @@ using ICSharpCode.AvalonEdit.Rendering;
 using ICSharpCode.Decompiler;
 using ICSharpCode.ILSpy.Properties;
 using ICSharpCode.ILSpy.TextView;
+using ICSharpCode.ILSpy.Themes;
 
 namespace ICSharpCode.ILSpy
 {
--- a/ILSpy/App.xaml
+++ b/ILSpy/App.xaml
@@ -2,12 +2,22 @@
              xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
              xmlns:styles="urn:TomsToolbox.Wpf.Styles"
+             xmlns:toms="urn:TomsToolbox"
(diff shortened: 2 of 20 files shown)


siegfriedpammer:
LGTM -- thank you very much!

PR #2451, opened 2021-07-13
Reproducible builds

christophwille:
Before:
image

After:
image


christophwille:
From https://docs.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-properties?view=vs-2019

<html><body>


DebugSymbols | All | A boolean value that indicates whether symbols are generated by the build. Setting -p:DebugSymbols=false on the command line disables generation of program database (.pdb) symbol files.
-- | -- | --


</body>
</html>

@dgrunwald that should not influence optimization as discussed yesterday. How this influences the unit test though is still beyond me.

siegfriedpammer:
Should we merge this then?

christophwille:
Given that DebugSymbols was in there before, I'd vote for yes.

Issue #2446, opened 2021-07-05
VS-Extension: Support for PackageReference


paukr:
Currently it seems one can only right click → _Open in ILSpy_ for normal references in Visual Studio. Please also support this very useful shortcut for PackageReferences.

PR #2454, opened 2021-07-17
Solution for #2446 adding support for package and project references in new-style "Dependencies" node

Rpinski:
Issue: https://github.com/icsharpcode/ILSpy/issues/2446 (for package references under "Dependencies -> Packages")

Also added support for project references under "Dependencies -> Projects" node, which were not covered yet, either.

### Solution
* Add ILSpy context menu item to additional groups (PACKAGEREFERENCE and PROJECTREFERENCE)
* Extend detection of NuGet packages for the new nodes

### DIFF
--- a/ILSpy.AddIn/Commands/NuGetReferenceForILSpy.cs
+++ b/ILSpy.AddIn/Commands/NuGetReferenceForILSpy.cs
@@ -8,6 +8,8 @@ using EnvDTE;
 
 using Microsoft.VisualStudio.Shell;
 
+using VSLangProj;
+
 namespace ICSharpCode.ILSpy.AddIn.Commands
 {
 	/// <summary>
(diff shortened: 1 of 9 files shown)


christophwille:
Verified with VS2022.

One thing of note is an ILSpy UI bug: when right-clicking on a Project ref / Open in ILSpy, the node is activated in ILSpy. With a NuGet package, this "scroll into view + activate node" doesn't happen.

Rpinski:
Yeah, noticed that, too.
I'll have a look at this.

Rpinski:
BTW, hast du noch ein VS 2017 irgendwo auf den Rechnern rumliegen? Die Nodes im Solution Explorer kamen wohl erst mit VS 2019 dazu. Wichtig wäre, dass mit dem PR keine Errors fliegen. 
Mein VS 2017 musste ich aus Platzgründen wegputzen.

christophwille:
With VS2022 on the horizon, we should discuss ending support for VS2017 (it will be in support from MS for a long time according to https://docs.microsoft.com/en-us/visualstudio/releases/2019/servicing, however, I don't think too many people are still using it). This also would allow us to rev Roslyn dependencies to 3.

siegfriedpammer:
Do we want to wait for a decision with this PR? Should I still test Dev16 and Dev17? If necessary I can create a VM and test Dev15. What do you think?

christophwille:
Decision reached at DevDays 2021: we will keep the VS2017 support until it breaks (either due to VS changes or Roslyn lowest common denonminator no longer working)

PR #2461, opened 2021-07-25
Pattern matching

siegfriedpammer:


siegfriedpammer:
```
public static void StructBasicTest(TS? a, TS? b)
{
	if (!(a == b))
	{
		Console.WriteLine();
	}
}
=>
stloc V_0(ldloc a)
stloc V_1(ldloc b)
if (
	if (comp.i4(comp.i4(comp.i4(call get_HasValue(ldloca V_0) != call get_HasValue(ldloca V_1)) == ldc.i4 0) == ldc.i4 0))
		ldc.i4 1
	else if (call get_HasValue(ldloca V_0))
		comp.i4(call op_Equality(call GetValueOrDefault(ldloca V_0), call GetValueOrDefault(ldloca V_1)) == ldc.i4 0)
	else
		ldc.i4 0) Block IL_003e {
	call WriteLine()
}
```
LiftCSharpUserEqualityComparison: extra LogicNot in trueInst

siegfriedpammer:
```
    class Program
    {
        int f;

        {            
        }

        static void TestCase(object x)
        {
            var program = x as Program;
            if (program != null && program.f > 5)
            {
                Console.WriteLine("pattern matches");
            }
            if (program != null && program.f > 10)
            {
                Console.WriteLine("other use!");
            }
        }
    }

```

This is transformed and produces invalid C# code. The problem is that `var program = x as Program;` definitely assigns `program` in both cases, but when using pattern matching it is only definitely assigned when not null.

dgrunwald:
Our current check that the variable is single-definition guarantees that all uses are dominated by the `program = x as Program` assignment, but that's not good enough -- I think we will need to check that all uses are dominated by the control flow edge taken after the `!= null` comparison that gets turned into the pattern match.

This will require changing the transform in some way, because such a check seems really difficult for the current statement transform. Maybe it could become a block transform running prior to ConditionDetection?

dgrunwald:
After rewriting the transform and implementing the dominance check; we discovered that `RemoveInfeasiblePathTransform` doesn't yet remove all infeasible paths, which causes a dominance check to fail for one of our existing test cases.

PR #2485, opened 2021-08-19
Added null checks, for using ilspy appless/windowless

HBerni:
I am using ILSpy as unpacker in WinMerge and therefore
linking it as a library from c++/cli code running it without any
window/app, resulting in 2 NullReferenceExceptions which I could not
workaround.

### DIFF
--- a/ILSpy/ILSpySettings.cs
+++ b/ILSpy/ILSpySettings.cs
@@ -126,7 +126,7 @@ namespace ICSharpCode.ILSpy
 
 		static string GetConfigFile()
 		{
-			if (App.CommandLineArguments.ConfigFile != null)
+			if (App.CommandLineArguments?.ConfigFile != null)
 				return App.CommandLineArguments.ConfigFile;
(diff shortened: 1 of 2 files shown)


siegfriedpammer:
How difficult would it be to use ilspycmd/ICSharpCode.Decompiler.Console instead?

HBerni:
Hi @siegfriedpammer,

thanks for your hint.

I rewrote my unpacker inspired by ilspycmd (but without using ilspycmd/ICSharpCode.Decompiler.Console)
and it works well with the official 7.1 libraries.

Now I just miss some header information in the output that was generated by CSharpLanguage::DecompileAssembly like
  // U:\xxx.dll
  // xxx, Version=3.0.0.0, Culture=neutral, PublicKeyToken=xxxx
  // Global type: <Module>
  // Architecture: AnyCPU (64-bit preferred)
  // Runtime: .NET 4.0
but that is no issue at all.

Anyway, I think the two additional checks in my commit wouldn't hurt any, but
are not required anymore, so you may close my pull request.

Thanks,
Bernhard

PR #2491, opened 2021-08-26
Fix #2337: Expose simple public XamlDecompiler API.

### LINKED ISSUE #2337: BAML Decompiler public API design
For ILSpy 7.1 we are planning to expose a reusable API to the BAML decompiler. This is a brain storming issue to develop a minimal API that covers most immediate known use-cases.

@CreateAndInject since you asked for this in the first place. What is necessary to get the BAML decompiler into shape for public use?

siegfriedpammer:
/cc @CreateAndInject does this work for you?


CreateAndInject:
Thanks! This is what I want!

PR #2496, opened 2021-09-03
Fix attribute application analysis could fail if there are wrong assemblies loaded

zvirja:
Link to issue(s) this covers

### Problem
Scenario:
- Load `System` assembly together with e.g. unknown assembly which IL Spy cannot read (so it shows error in a tree)
- Find any attribute and Analyze it
- In the Analysis window expand "Applied To" analysis.

The following error occurs:
image

### Solution

The `AnalyzerScope.GetAllModules()` was not expected to return `null` modules as I can see by the usages around. So I fixed that by filtering out invalid modules from Assembly List


siegfriedpammer:
Thank you very much for fixing this!

PR #2509, opened 2021-09-27
Fix null check in BamlDecompilerTypeSystem.HasType

yyjdelete:
Fix CS8073: `The result of the expression is always 'true' since a value of type 'TypeDefinitionHandle' is never equal to 'null' of type 'TypeDefinitionHandle?'`

`TypeDefinitionHandle` is an struct and can never be null.


siegfriedpammer:
Thank you very much!

PR #2521, opened 2021-10-13
What-if we split the addin by introducing a property (target?)

christophwille:
TODO: It should build into separate directories so we can easily collect the build output.


christophwille:
To be closed in favor of split projects approach

Issue #2539, opened 2021-11-08
Some C#3 restrictions are ignored and slight fail on an IEnumerable method


Sekoree:
I tried to decompile the Assembly-CSharp of an old Unity (5.5) game (Audiosuf2) which uses .Net FX 3.5 and is limited to C#3 cause of that. Most stuff decompiled flawlessly. But for some reason it added optional method parameters which are only valid for C#4+ and a few ``ref`` keywords in methods which are C#7+. It also had an issue with ``IEnumerable<KeyValuePair<TKey,TValue>>``

Not sure if I'm allowed to share the file, but here's the code output from ILSpy which has issues and working results from dnSpy
The optional method parameters happen on both.
#### Optional Method Parameters

![image](https://user-images.githubusercontent.com/37484974/140703409-90c23ecb-10ff-4366-8772-36e5042c717a.png)

#### ``ref`` Keyword Used

ILSpy:
![image](https://user-images.githubusercontent.com/37484974/140703636-56415488-f72c-4590-8271-2707dacd2a8f.png)
dnSpy:
![image](https://user-images.githubusercontent.com/37484974/140703711-8fe7a83c-f42d-46e6-ae76-b7d34df25a08.png)

#### Issue with IEnumerable<KeyValuePair<TKey,TValue>>

ILSpy:
![image](https://user-images.githubusercontent.com/37484974/140703919-d2c5a85a-0730-4784-9be5-f0b57673254c.png)
dnSpy:
![image](https://user-images.githubusercontent.com/37484974/140703955-904c1c03-6957-42af-8b35-70fa6a720a45.png)

### Details
* Product in use: ILSpy VS 2022 extension (Preview 2)
* Version in use: ILSpy version 7.2.0.6702-preview2

siegfriedpammer:
Problem 1) It seems that the original binary was not compiled using C# 3, but C# 4, as the optional parameter is defined *in metadata*. It might be that this is another quirk of the Mono C# compiler that it allowed optional parameters already in earlier language versions - nonetheless, this is *a bug* in the Mono C# compiler. If you want to stay close to standard C#, you should try using C# 4 with the whole codebase. You can use C# 4 with .NET Framework 3.5 as long as you do not try to use the dynamic language feature or other features that require framework support.

I see two options for us here:
1. If C# 3 is selected ignore the parameter metadata and remove the default value from the parameter declaration.
2. Just like with other features that cannot be represented differently (think generics): leave it as-is

[...]

PR #2546, opened 2021-11-09
Mcs5

siegfriedpammer:
Adds mcs 5.23 to our test-suite, fixes part 3 of #2539

### DIFF
--- a/ICSharpCode.Decompiler/IL/ControlFlow/YieldReturnDecompiler.cs
+++ b/ICSharpCode.Decompiler/IL/ControlFlow/YieldReturnDecompiler.cs
@@ -172,7 +172,11 @@ namespace ICSharpCode.Decompiler.IL.ControlFlow
 			newBody.SortBlocks(deleteUnreachableBlocks: true);
 			function.CheckInvariant(ILPhase.Normal);
 
-			if (!isCompiledWithMono)
+			if (isCompiledWithMono)
+			{
+				CleanSkipFinallyBodies(function);
+			}
+			else
 			{
 				DecompileFinallyBlocks();
 				ReconstructTryFinallyBlocks(function);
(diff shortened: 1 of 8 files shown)


ICSharpCode.Decompiler/IL/Transforms/SwitchOnStringTransform.cs
```diff
@@ -702,6 +702,8 @@ bool HasLabel(SwitchSection section)
 			{
 				if (section == defaultSection)
 					continue;
+				if (nullValueCaseBlock != null && section.Body.MatchBranch(nullValueCaseBlock))
```
dgrunwald, line 705: This if seems to say "sections that join up with `case null` cannot be used as the default section".
Why would that make any sense?

ICSharpCode.Decompiler/IL/ControlFlow/YieldReturnDecompiler.cs
```diff
@@ -1252,8 +1255,38 @@ private void CleanSkipFinallyBodies(ILFunction function)
 						entryPoint.Instructions.RemoveRange(1, entryPoint.Instructions.Count - 1);
 					}
 				}
+				entryPoint = AsyncAwaitDecompiler.GetBodyEntryPoint(container);
```
dgrunwald, line 1258: Redundant assignment? `entryPoint` should already have this value?

siegfriedpammer: GetBodyEntryPoint returns the first block that does not have an unconditional branch as its first instruction. After the "skip finally" if-instruction is removed, we reuse the logic to get to the block that contains the generated finally-method.

ICSharpCode.Decompiler/IL/Transforms/SwitchOnStringTransform.cs
siegfriedpammer: At that stage of the transform the first section is usually the default section (that section's label contains all "unused" values). We want to exclude "case null" because it is treated specially in the transform. mcs 5 (unlike mcs 2 or any MS compiler) seems to generate an entry for case null in the jump-table, which causes "FixCasesWithoutValue" to fail.

Issue #2551, opened 2021-11-15
Tabs not being properly closed on assemblies deletion from treeview


greenozon:
### Steps to reproduce
1. Drop assembly to empty treeview
2. Go to References and click on "Decompile to new tab" on some assemblies so they will open in new tabs
3. on treeview click CTRL+A and Del, thus empty it.
4. Click on any old tabs

### Error message shown

![image](https://user-images.githubusercontent.com/765343/141851878-b31e44db-71e5-4765-bd82-7c90db6cc7c4.png)

### Details
ILSpy version 7.2.0.6719-preview2
.NET version 4.8.3928.0

Issue #2550, opened 2021-11-15
Menu "Save Code" enable issue on changing C# version


greenozon:
### Steps to reproduce
1. Drop assembly onto empty treeview
2. Check that Save Code button is enabled
3. Change C# eg from default to say C#9

### Error message shown

Save Code button gets disabled

![image](https://user-images.githubusercontent.com/765343/141848921-9704cf53-dcca-4917-b2f9-cb2ee6dcd82e.png)

If you change C# version again, issue is gone.

### Details
ILSpy version 7.2.0.6719-preview2
.NET version 4.8.3928.0

dgrunwald:
The problem here is that (since the fix for #1504) changing the language version results in all treeview items being re-created from scratch.
This loses not only their IsExpanded state, but also makes it so that the logical focus gets lost. Thus the command bindings lose their target and end up being disabled (until the treeview gets the keyboard focus again).

Issue #2404, opened 2021-05-10
Feature Request: Cache decompilation result for tabs


sgkoishi:
**Is your feature request related to a problem? Please describe.**
Decompilation may take some time from seconds to minutes, switching between tabs and repeatedly waiting for result is annoying.

**Describe the solution you'd like**
Cache the decompilation result to avoid the waiting.

**Additional context**
The result is already cached but a re-decompilation is triggered.

Issue #1948, opened 2020-03-07
Toolbar loses focus when changing language/version


dgrunwald:
In ILSpy 5.0 it was possible to select one of the comboboxes in the toolbar, then use the arrow keys or the mouse wheel to cycle through the language versions.
In master this no longer works; any change to the language settings causes the focus to jump to the tree view.

PR #2554, opened 2021-11-17
Fix typesystems used in the UI

siegfriedpammer:
Proper fix for
* #1948
* #2404
* #2550
* #2551

### DIFF
--- a/ILSpy/Docking/DockWorkspace.cs
+++ b/ILSpy/Docking/DockWorkspace.cs
@@ -94,8 +94,6 @@ namespace ICSharpCode.ILSpy.Docking
 				if (_activeTabPage != value)
 				{
 					_activeTabPage = value;
-					this.sessionSettings.FilterSettings.Language = value.Language;
-					this.sessionSettings.FilterSettings.LanguageVersion = value.LanguageVersion;
 					var state = value.GetState();
 					if (state != null)
 					{
@@ -173,23 +171,6 @@ namespace ICSharpCode.ILSpy.Docking
 		internal void LoadSettings(SessionSettings sessionSettings)
 		{
 			this.sessionSettings = sessionSettings;
-			sessionSettings.FilterSettings.PropertyChanged += FilterSettings_PropertyChanged;
-		}
-
-		private void FilterSettings_PropertyChanged(object sender, PropertyChangedEventArgs e)
-		{
-			if (e.PropertyName == "Language")
-			{
-				ActiveTabPage.Language = sessionSettings.FilterSettings.Language;
-				if (sessionSettings.FilterSettings.Language.HasLanguageVersions)
-				{
-					sessionSettings.FilterSettings.LanguageVersion = ActiveTabPage.LanguageVersion;
-				}
-			}
-			else if (e.PropertyName == "LanguageVersion")
-			{
-				ActiveTabPage.LanguageVersion = sessionSettings.FilterSettings.LanguageVersion;
-			}
 		}
 
 		internal void CloseAllTabs()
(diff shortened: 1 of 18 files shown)


ILSpy/TreeNodes/EventTreeNode.cs
```diff
@@ -45,14 +46,21 @@ public EventTreeNode(IEvent @event)
...
 		public IEvent EventDefinition { get; }
 
-		public override object Text => GetText(EventDefinition, this.Language) + EventDefinition.MetadataToken.ToSuffixString();
+		public override object Text => GetText(GetEventDefinition(), this.Language) + EventDefinition.MetadataToken.ToSuffixString();
+
+		private IEvent GetEventDefinition()
+		{
+			return ((MetadataModule)EventDefinition.ParentModule.PEFile
+				?.GetTypeSystemWithCurrentOptionsOrNull()
+				?.MainModule).GetDefinition((EventDefinitionHandle)EventDefinition.MetadataToken);
```
dgrunwald, line 55: `(a?.b).c()` is likely to result in `NullReferenceException`s.
It would be better to use the original member definition if the `OrNull` suddenly starts returning `null`

greenozon:
sure thing!

siegfriedpammer:
@greenozon added a fix for #2551

greenozon:
I've downloaded built binaries - ILSpy 7.2.0.6737-fix-ui-typesystem-preview2 (debug).zip
spent like 10-15 min utes testing two issues - save menu item + closing tabs on assembly removal
I give 5 out of 5 stars, 
🌟🌟🌟🌟🌟
well done!

greenozon:
Found  one issue with tabs closing:

if you do the "Decompile to new tab" on resources of the assembly, then close assembly - the resource(s) tabs are not closed,
thus when clicking on it - shows this

image

greenozon:
@siegfriedpammer  ^^^

sgkoishi:
The release build 639 (`ILSpy 7.2.0.6737-fix-ui-typesystem-preview2 (release)`) failed to launch, the debug build works correctly.
image

greenozon:
tried to run build artifact ILSpy 7.2.0.6738-preview2-pr2554 (release).zip
it has same issue on start as reported above

siegfriedpammer:
> tried to run build artifact ILSpy 7.2.0.6738-preview2-pr2554 (release).zip it has same issue on start as reported above

@greenozon @sgkoishi a fix for this is almost done. Thanks a lot for your help!

siegfriedpammer:
> A minor issue: if you switch to a tab before the decompilation is completed, the selected item in the treeview is not updated (the previous item was selected instead of the new-selected decompiling item)

Should be fixed now.

greenozon:
Tested both debug & release - all looks pretty cool

siegfriedpammer:
@dgrunwald can you please take another look at this?

ILSpy/Docking/DockWorkspace.cs
```diff
@@ -45,14 +46,46 @@ public class DockWorkspace : INotifyPropertyChanged, ILayoutUpdateStrategy
...
+			{
+				var state = tab.GetState();
+				if (state == null || state.DecompiledNodes == null)
+				{
+					continue;
+				}
+				bool found = false;
+				foreach (var node in state.DecompiledNodes)
+				{
+					var assemblyNode = node.Ancestors().OfType<TreeNodes.AssemblyTreeNode>().FirstOrDefault();
```
dgrunwald, line 74: This will find the innermost `AssemblyTreeNode`, which may be a bundle member.
I think `LastOrDefault` would be more appropriate as it would handle the case where a bundle is removed from the assembly list.

PR #2566, opened 2021-12-02
Use Directory.Build.props.template

christophwille:
dotnet build -> Version 1.0.0 via ilspycmd.exe. Build again, it picks up the newly written Directory.Build.props. How to fix?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ _ReSharper*/
 multitargeting.props
 ILSpy.Installer/wix/
 ILSpy.Installer/AppPackage.cs
+/ICSharpCode.Decompiler.Console/Directory.Build.props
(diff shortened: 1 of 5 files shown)


christophwille:
```
 Note

The MSBuild engine reads in all imported files during evaluation, before starting build execution for a project (including any PreBuildEvent), so these files are not expected to be modified by the PreBuildEvent or any other part of the build process. Any modifications do not take effect until the next invocation of MSBuild.exe or the next Visual Studio build. Also, if your build process contains many project builds (as with multitargeting or building dependent projects), then imported files, including Directory.build.props are read when evaluation occurs for each individual project build.
```

explains why that approach doesn't work.

Alternative (that doesn't work in VS):

https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022#customize-the-solution-build

christophwille:
Superseded by https://github.com/icsharpcode/ILSpy/commit/8b12ad23a64df338591eaeed7ccc14fc4bc9be87

PR #2567, opened 2021-12-03
Improve file name sanitizer

siegfriedpammer:
- Replace DecompilerTextView.CleanUpName with direct uses of WholeProjectDecompiler.CleanUpFileName.
- Add file extension handling to WholeProjectDecompiler.CleanUpName
- Add WholeProjectDecompiler.SanitizeFileName, which respects file extensions.

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
+++ b/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
@@ -269,7 +269,7 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 						{
 							foreach (var (name, value) in resourcesFile)
 							{
-								string fileName = CleanUpFileName(name)
+								string fileName = SanitizeFileName(name)
 									.Replace('/', Path.DirectorySeparatorChar);
 								string dirName = Path.GetDirectoryName(fileName);
 								if (!string.IsNullOrEmpty(dirName) && directories.Add(dirName))
@@ -571,14 +571,25 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 		/// </summary>
 		public static string CleanUpFileName(string text)
 		{
-			return CleanUpFileName(text, separateAtDots: false);
+			return CleanUpName(text, separateAtDots: false, treatAsFileName: false);
+		}
+
+		/// <summary>
+		/// Removes invalid characters from file names and reduces their length,
+		/// but keeps file extensions intact.
+		/// </summary>
+		public static string SanitizeFileName(string fileName)
+		{
+			return CleanUpName(fileName, separateAtDots: false, treatAsFileName: true);
 		}
 
 		/// <summary>
 		/// Cleans up a node name for use as a file system name. If <paramref name="separateAtDots"/> is active,
-		/// dots are seen as segment separators. Each segment is limited to 255 characters.
+		/// dots are seen as segment separators. Each segment is limited to maxSegmentLength characters.
+		/// (see <see cref="GetLongPathSupport"/>) If <paramref name="treatAsFileName"/> is active,
+		/// we check for file a extension and try to preserve it, if it's valid.
 		/// </summary>
-		static string CleanUpFileName(string text, bool separateAtDots)
+		static string CleanUpName(string text, bool separateAtDots, bool treatAsFileName)
 		{
 			int pos = text.IndexOf(':');
 			if (pos > 0)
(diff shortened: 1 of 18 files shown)


ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
```diff
@@ -587,23 +598,47 @@ static string CleanUpFileName(string text, bool separateAtDots)
...
+				if (lastDot >= 0 && text.Length - lastDot < maxSegmentLength)
+				{
+					extension = text.Substring(lastDot);
+					text = text.Remove(lastDot);
+					foreach (var c in extension)
+					{
+						if (!(char.IsLetterOrDigit(c) || c == '-' || c == '_' || c == '.'))
+						{
+							// extension contains an invalid character, therefore cannot be a valid extension.
+							extension = null;
```
dgrunwald, line 621: We just drop the whole extension?
Don't we need to revert to the original `text`?

siegfriedpammer: ah, yes... thanks!

greenozon:
@siegfriedpammer  so what is a segment meaning? I understand MAX_PATH limitation, but never heard about segment...
also what platforms are we talking about, like OS (win/lin/max) or different .net versions

siegfriedpammer:
> @siegfriedpammer so what is a segment meaning? I understand MAX_PATH limitation, but never heard about segment...

See https://docs.microsoft.com/en-us/archive/blogs/jeremykuhne/net-4-6-2-and-long-paths-on-windows-10

A segment is one part of a path, e.g., directory name, file name... these are limited to 255 characters.

Jeremy Kuhne writes (in a comment at the bottom of the page): 

> "You are correct that segment limits are part of file system specifications and have nothing to do with MAX_PATH. There are no file systems in common usage on any OS (Windows, Mac, Unix, etc.) that allow more than 255 for segments. There are a few that allow less (optical media notably). I'm unaware of any limitations where you can't use the full segment length outside of a wacky max path limitation of 248 in CreateDirectory() that is also removed with the support enabled."

For example:

```
C:\temp>mkdir 012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234
```

Creates a path with a single directory that has a length of 255 characters, if you add one more, it fails:

```
C:\temp>mkdir 0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345
```

You can create very very long paths

```
C:\temp\012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234\012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234>dir
 Volume in drive C has no label.
 Volume Serial Number is 4EFF-A111

 Directory of C:\temp\012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234\012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234

04.12.2021  13:34    <DIR>          .
04.12.2021  13:34    <DIR>          ..
04.12.2021  13:34    <DIR>          012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234
               0 File(s)              0 bytes
               3 Dir(s)  1 404 808 548 352 bytes free
```

That's the long directory name nested 3 times.

> also what platforms are we talking about, like OS (win/lin/max) or different .net versions

This is available since Windows 10 Version 1607 (Anniversary Update), internal version 10.0.14393, released in 2016. Support was added to .NET in .NET Framework 4.6.2 (classic framework), as pointed out in the above blog post. As ILSpy is WPF/Windows-only, I have no idea whether that would work on other platforms. Support was first added to ILSpy in https://github.com/icsharpcode/ILSpy/commit/f1d5f9d7ec9dd751aa1cfbce21c626b9fcf30239.

Hope this helps!

christophwille:
> Open question: should we increase the maximum segment length for platforms that do not support long paths? If yes, what is a good limit? 50 characters?

As the current number is arbitrary we can either keep it or use 42 (at least that would tip everybody off that we plucked it out of the air)

dgrunwald:
> As ILSpy is WPF/Windows-only, I have no idea whether that would work on other platforms.

For Linux the typical limits are:
 * segment (NAME_MAX): 255 bytes
 * total path length: PATH_MAX is 4096 but that's [not a hard limit](https://eklitzke.org/path-max-is-tricky)

greenozon:
Thanks a lot for details & explanations!
42 is definitely a magic all questions answer number :+1: 
(joking)

siegfriedpammer:
@dgrunwald Any final comments?

greenozon:
I've tested ILSpy 7.2.0.6781-filename-sanitizer-preview3 (debug).zip  on my sample - it produces 100% xamls (no corruption in names, no bamls) + puts into (sub)dirs which is very nice
kudos!

PR #2568, opened 2021-12-03
C# 10 abstract static interface members

siegfriedpammer:


ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
```diff
@@ -1807,6 +1816,10 @@ EntityDeclaration DoDecompile(IProperty property, DecompileRun decompileRun, ITy
 				{
 					propertyDecl.Modifiers |= Modifiers.Extern;
 				}
+				if (property.DeclaringType.Kind == TypeKind.Interface && property.IsStatic)
```
dgrunwald, line 1819: Marking all static properties in interfaces as abstract?
Aren't these already `abstract` in the IL, and the modifier is only missing because TSAB suppresses `abstract` on interface members?

siegfriedpammer: Yes, would be better to fix this in the TSAB... thanks for pointing this out.

Issue #2572, opened 2021-12-08
Shallow clone is not supported


shurivich:
### Steps to reproduce
1. Clone with flag `--depth 1`, e.g. `git clone --depth 1 https://github.com/icsharpcode/ILSpy.git`
2. Try to compile solution or test project, e.g. `dotnet build .\ILSpy\ICSharpCode.Decompiler.Tests`

Build is failed with error:
### Error message shown

```
PS C:\temp> dotnet build .\ILSpy\ICSharpCode.Decompiler.Tests
Microsoft (R) Build Engine version 17.0.0+c9eb9dd64 for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...
  Restored C:\temp\ILSpy\ICSharpCode.Decompiler\ICSharpCode.Decompiler.csproj (in 155 ms).
  Restored C:\temp\ILSpy\ICSharpCode.Decompiler.Tests\ICSharpCode.Decompiler.Tests.csproj (in 155 ms).
  0f3e8b79be6019de6145a52b4ee7002a1772e5c6
  ILSpyUpdateAssemblyInfo1628202346
  fatal: Invalid revision range d779383cb85003d6dabeb976f0845631e07bf463..HEAD
  Exception calling "Parse" with "1" argument(s): "Value cannot be null.
  Parameter name: String"
  At C:\temp\ILSpy\BuildTools\update-assemblyinfo.ps1:49 char:12
  +     return [Int32]::Parse((git rev-list --count "$baseCommit..HEAD")) ...
  +            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      + CategoryInfo          : NotSpecified: (:) [], ParentContainsErrorRecordException
      + FullyQualifiedErrorId : ArgumentNullException

C:\temp\ILSpy\ICSharpCode.Decompiler\ICSharpCode.Decompiler.csproj(677,5): error MSB3073: The command "powershell -NoProfile -ExecutionPolicy Bypass -File BuildTools/update-assemblyinfo.ps1 Debug" exited with code 1.

Build FAILED.

C:\temp\ILSpy\ICSharpCode.Decompiler\ICSharpCode.Decompiler.csproj(677,5): error MSB3073: The command "powershell -NoProfile -ExecutionPolicy Bypass -File BuildTools/update-assemblyinfo.ps1 Debug" exited with code 1.
[...]

PR #2573, opened 2021-12-08
[FIX] Shallow clone is not supported

shurivich:
This PR fixes issue https://github.com/icsharpcode/ILSpy/issues/2572

### DIFF
--- a/BuildTools/update-assemblyinfo.ps1
+++ b/BuildTools/update-assemblyinfo.ps1
@@ -46,7 +46,7 @@ function gitVersion() {
     if (No-Git) {
         return 0;
     }
-    return [Int32]::Parse((git rev-list --count "$baseCommit..HEAD")) + $baseCommitRev;
+	return ((git rev-list --count "$baseCommit..HEAD") -as [Int32]) + $baseCommitRev;
 }
 
 function gitCommitHash() {


christophwille:
Does that fix or mask the problem?

siegfriedpammer:
We won't be able to calculate the correct revision from a shallow clone, as it only contains the most recent revision.

```
ILSpy-shallow>git rev-list --count d779383cb85003d6dabeb976f0845631e07bf463..HEAD
fatal: Invalid revision range d779383cb85003d6dabeb976f0845631e07bf463..HEAD
```

A possible fix is to ignore the output, if it's not numeric, as this PR proposes. I would say this is an acceptable solution.

siegfriedpammer:
One more thing, I would prefer the function to return 0 in this case... just like in the No-Git case...

shurivich:
> One more thing, I would prefer the function to return 0 in this case... just like in the No-Git case...

It's may be not clear from code, but this is exactly what happening:
`((git rev-list --count "$baseCommit..HEAD") -as [Int32]) + $baseCommitRev`
Is recalculated to be zero.

As more robust solution, I can suggest surrounding `gitVersion` (and probably `gitCommitHash)`?) with `try..catch`, that will return 0 in case of exception.

siegfriedpammer:
I have fixed this by adding try-catch handling to the respective parts of the powershell script. Thanks for reporting!

PR #2578, opened 2021-12-13
Add TypeDefinitionDocuments to custom debug information data

davidwengier:
For .NET 6/C# 10 we added a new CustomDebugInformation type for linking a TypeDef entry to a Document entry, in https://github.com/dotnet/roslyn/pull/56278
This updates ILSpy to correctly display the information.

Before:
image

After:
image

The `02` in the value is correct, it refers to the 2nd row of the Document table.
 
Sorry I took so long to submit this PR :)

### DIFF
--- a/ICSharpCode.Decompiler/DebugInfo/KnownGuids.cs
+++ b/ICSharpCode.Decompiler/DebugInfo/KnownGuids.cs
@@ -21,6 +21,7 @@ namespace ICSharpCode.Decompiler.DebugInfo
 		public static readonly Guid CompilationOptions = new Guid("B5FEEC05-8CD0-4A83-96DA-466284BB4BD8");
 		public static readonly Guid CompilationMetadataReferences = new Guid("7E4D4708-096E-4C5C-AEDA-CB10BA6A740D");
 		public static readonly Guid TupleElementNames = new Guid("ED9FDF71-8879-4747-8ED3-FE5EDE3CE710");
+		public static readonly Guid TypeDefinitionDocuments = new Guid("932E74BC-DBA9-4478-8D46-0F32A7BAB3D3");
 
 		public static readonly Guid HashAlgorithmSHA1 = new Guid("ff1816ec-aa5e-4d10-87f7-6f4963833460");
 		public static readonly Guid HashAlgorithmSHA256 = new Guid("8829d00f-11b8-4213-878b-770e8597ac16");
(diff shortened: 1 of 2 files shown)


siegfriedpammer:
Thank you for your contribution! Just one question: wouldn't it be better to show the contents of the Document table entry as detail view?

davidwengier:
Certainly can do that if you'd prefer, it wasn't clear to me whether the detail view should show just the decoded value, or do more. I will update

siegfriedpammer:
You are of course right. Now I am no longer sure whether adding that information to the details view is a good idea. Maybe it will cause confusion, as one might think that all of the information is present in the blob.

Probably just add a link to the output, which will open the Documents table in a new tab?

Would you want me to add this or do you want to have a look yourself?

davidwengier:
Is there any example where links are used to navigate between nodes now? I couldn't see anything from a quick look through the app.

In general I think more linking would be great though, for things like values in the "Document" field of the MethodDebugInformation table, or the Parent column of any metadata tables etc. Obviously not as part of this PR though :)

davidwengier:
Showing the document record is not terrible thought:
image

siegfriedpammer:
> Is there any example where links are used to navigate between nodes now? I couldn't see anything from a quick look through the app.
> 
> In general I think more linking would be great though, for things like values in the "Document" field of the MethodDebugInformation table, or the Parent column of any metadata tables etc. Obviously not as part of this PR though :)

See https://github.com/icsharpcode/ILSpy/blob/master/ILSpy/Metadata/GoToTokenCommand.cs#L39 for an example. I can take a closer look tomorrow.

davidwengier:
> > Is there any example where links are used to navigate between nodes now? I couldn't see anything from a quick look through the app.
> > In general I think more linking would be great though, for things like values in the "Document" field of the MethodDebugInformation table, or the Parent column of any metadata tables etc. Obviously not as part of this PR though :)
> 
> See https://github.com/icsharpcode/ILSpy/blob/master/ILSpy/Metadata/GoToTokenCommand.cs#L39 for an example. I can take a closer look tomorrow.

😮

I had no idea that context menu existed!

siegfriedpammer:
DataGridHyperlinkColumn would probably be a good idea to use for all fields that reference other tables.

siegfriedpammer:
I have prototyped this... the result is quite good, I think. What do you think?

hyperlinks

davidwengier:
> I have prototyped this... the result is quite good, I think. What do you think?

Love it!

siegfriedpammer:

PR #2585, opened 2021-12-24
Fix MainWindow.OpenLink for net6.0

yyjdelete:


siegfriedpammer:
Would you mind explaining why this change is necessary? I didn't yet have a closer look at the changes required to get ILSpy working with net6.0, hence the question. Thanks!

yyjdelete:
It's a broken change from .NET Framework to .NET Core, and `UseShellExecute=true` is required to open URL with browser in windows.
https://docs.microsoft.com/en-us/dotnet/core/compatibility/fx-core#change-in-default-value-of-useshellexecute
`ProcessStartInfo.UseShellExecute has a default value of false on .NET Core. On .NET Framework, its default value is true.`

And the same change is already in AvalonEdit.
https://github.com/icsharpcode/AvalonEdit/commit/9ca474a1d5f36d880cf86776781653f05a9fa6e3

siegfriedpammer:
Thank you for the explanation!

PR #2595, opened 2021-12-31
Fix displaying icons from embedded resources

AustinWise:
### Problem

When `BitmapCacheOption.None` is used, icons will not be loaded from the
stream until they are needed to be displayed. In this case, the icon
loading will be triggered when the text box is rendered from the UI thread.
However by that time the steam object will have already been disposed.
This results in no icons being visible.

### Solution

The fix is to eager load the icons while the stream is still alive.

An alternative fix would be to defer disposing the stream to a later time
or not dispose it at all.


siegfriedpammer:

Issue #2605, opened 2022-01-07
Add possibility to add submenu items


siegfriedpammer:
### Discussed in https://github.com/icsharpcode/ILSpy/discussions/2604

<div type='discussions-op-text'>

<sup>Originally posted by **psmacchia** January  7, 2022</sup>
We'd like to have a popup context menu with several child menus like we do in our Reflector plugin (see the **NDepend >** popup menu in the screenshot below):

![image](https://user-images.githubusercontent.com/511445/148546934-f35b4b43-546d-4774-8263-c7d84aff720e.png)

It looks like it is not feasible, am I right?

This would be useful for sophisticated plugins. For example [the ILSpy Reflexil plugin](https://github.com/sailro/Reflexil/releases) adds a dozen of context menus. Wouldn't it be preferable to have a single **Reflexil >** popup menu with its child menus?

![image](https://user-images.githubusercontent.com/511445/148547161-8c7ab14b-4014-4de4-adf0-cd44494926e7.png)
</div>

PR #2608, opened 2022-01-14
Submenu support

siegfriedpammer:
### Problem
See #2605 and #2604

### Solution

I added one resp. two new properties to our main menu/context menu attributes:

Main menus now can define IDs, which should be unique across the whole application. And by using the already existing Menu property, the submenu item can be connected to its parent item.
Context menus can do the same:

Examples:

	// Adds a sub menu to the file menu
	[ExportMainMenuCommand(Menu = "_File", Header = "Sub Menu", ID = "SubMenuParent")]
	sealed class DummyMenuItem
	{
	}

	[ExportMainMenuCommand(Menu = "SubMenuParent", Header = "SubItem")]
	sealed class TestCommand : SimpleCommand
	{
		public override void Execute(object parameter)
		{
			throw new NotImplementedException();
		}
	}

	// Adds a sub menu to all context menus
	[ExportContextMenuEntry(Header = "Sub Menu", ID = "SubMenuParent")]
	sealed class ContextDummyMenuItem : IContextMenuEntry
	{
		public void Execute(TextViewContext context)
		{
		}

		public bool IsEnabled(TextViewContext context)
		{
			return true; // display menu in all context menus
		}

		public bool IsVisible(TextViewContext context)
		{
			return true; // display menu in all context menus
		}
	}

	[ExportContextMenuEntry(Menu = "SubMenuParent", Header = "SubItem")]
	sealed class ContextTestCommand : IContextMenuEntry
	{
		public void Execute(TextViewContext context)
		{
		}

		public bool IsEnabled(TextViewContext context)
		{
			return true;
		}

		public bool IsVisible(TextViewContext context)
		{
			return true;
		}
	}

@sailro @psmacchia would this solution work for you? Please provide feedback. Thanks!

### DIFF
--- a/ILSpy/AboutPage.cs
+++ b/ILSpy/AboutPage.cs
@@ -38,7 +38,7 @@ using ICSharpCode.ILSpy.Themes;
 
 namespace ICSharpCode.ILSpy
 {
-	[ExportMainMenuCommand(Menu = nameof(Resources._Help), Header = nameof(Resources._About), MenuOrder = 99999)]
+	[ExportMainMenuCommand(ParentMenuID = nameof(Resources._Help), Header = nameof(Resources._About), MenuOrder = 99999)]
 	sealed class AboutPage : SimpleCommand
 	{
 		public override void Execute(object parameter)
--- a/ILSpy/Commands/CheckForUpdatesCommand.cs
+++ b/ILSpy/Commands/CheckForUpdatesCommand.cs
@@ -21,7 +21,7 @@ using ICSharpCode.ILSpy.Properties;
 
 namespace ICSharpCode.ILSpy
 {
-	[ExportMainMenuCommand(Menu = nameof(Resources._Help), Header = nameof(Resources._CheckUpdates), MenuOrder = 5000)]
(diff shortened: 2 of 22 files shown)


psmacchia:
Thanks @siegfriedpammer, looks like it'd work out, I assume you could use it recursively (I mean with child menus being parent themselves). 

I find those 2 new property names confusing but I am not proficient into the ILSpy terminology as you are. I might have used instead `ChildOf` and `ParentID`.

```
	public interface IContextMenuEntryMetadata
	{
		string ID { get; }
		string Menu { get; }
```

siegfriedpammer:
Thank you for your feedback. Self-recursion is not supported. The top-level item of a context menu must have `Menu == null` to be even considered, however, if you reuse the same `ID` for a child of the top-level item, then a stack overflow will happen, so it is important to keep the IDs unique.

> I might have used instead ChildOf and ParentID.

Yes, these would have been better names, but renaming them (for consistency with the existing `Menu` property of `ExportMainMenuCommandAttribute`) would be a breaking change. I think adding proper documentation to these interfaces/attributes is long overdue and might help mitigate the issue.

psmacchia:
Thanks @siegfriedpammer for the explanations. Is this commit planned for upcoming 7.2 RTM ?

siegfriedpammer:
Yes.

Issue #2623, opened 2022-02-08
ILSpy 7.2.0.6839-preview4 Failed to start in zh-Hans


czphoenix:
### Steps to reproduce
1. Set Windows locale to zh-CN
2. Run program

### Error message shown

```
System.Windows.Markup.XamlParseException: 对类型“ICSharpCode.ILSpy.MainWindow”的构造函数执行符合指定的绑定约束的调用时引发了异常。 ---> System.InvalidOperationException: 序列不包含任何匹配元素
   在 System.Linq.Enumerable.First[TSource](IEnumerable`1 source, Func`2 predicate)
   在 ICSharpCode.ILSpy.MainWindow.InitWindowMenu()
   在 ICSharpCode.ILSpy.MainWindow..ctor()
   --- 内部异常堆栈跟踪的结尾 ---
   在 System.Windows.Markup.XamlReader.RewrapException(Exception e, IXamlLineInfo lineInfo, Uri baseUri)
   在 System.Windows.Markup.WpfXamlLoader.Load(XamlReader xamlReader, IXamlObjectWriterFactory writerFactory, Boolean skipJournaledProperties, Object rootObject, XamlObjectWriterSettings settings, Uri baseUri)
   在 System.Windows.Markup.WpfXamlLoader.LoadBaml(XamlReader xamlReader, Boolean skipJournaledProperties, Object rootObject, XamlAccessLevel accessLevel, Uri baseUri)
   在 System.Windows.Markup.XamlReader.LoadBaml(Stream stream, ParserContext parserContext, Object parent, Boolean closeStream)
   在 System.Windows.Application.LoadBamlStreamWithSyncInfo(Stream stream, ParserContext pc)
   在 System.Windows.Application.DoStartup()
   在 System.Windows.Application.<.ctor>b__1_0(Object unused)
   在 System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   在 System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)

```

### Details
* Product in use: ILSpy
* Version in use: 7.2.0.6839-preview4 41c99e4727beb04b876a76f280bbd74bec20324b

* in MainWindow.xaml:
    * MenuItem.Header is Properties.Resources._Window
    * MenuItem.Tag is fixed string "_Window"
[...]

PR #2626, opened 2022-02-08
#2623 Fix error in start up in zh-Hans

czphoenix:
Link to issue(s) this covers

### Problem
#2623

### Solution
* use nameof(Properties.Resources._Window) instead of Properties.Resources._Window when comparing


siegfriedpammer:
Thank you for contributing this fix. Sorry for the inconvenience!

Issue #2638, opened 2022-03-05
COM interface method/property ordering is not preserved in decompiled output


zhuman:
### Input code
Any interface/class with properties and methods interleaved will result in ILSpy reordering the methods. For COM interfaces, this could be seen with something like this:

```c#
using System;
using System.Reflection;
using System.Runtime.CompilerServices;
using System.Runtime.InteropServices;
using Dia2Lib;

[ComImport]
[Guid("CB787B2F-BD6C-4635-BA52-933126BD2DCD")]
[InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
[DefaultMember("symIndexId")]
public interface IDiaSymbol
{
	[DispId(0)]
	uint symIndexId
	{
		[MethodImpl(MethodImplOptions.InternalCall)]
		get;
	}

// .... Lots of properties

	[MethodImpl(MethodImplOptions.InternalCall)]
	void get_dataBytes([In] uint cbData, out uint pcbData, out byte pbData);

// ... Lots more methods

	[DispId(84)]
	uint targetSection
	{
		[MethodImpl(MethodImplOptions.InternalCall)]
		get;
	}

// ... Lots more properties, methods, etc.
}
```

### Erroneous output
When ILSpy decompiles such an interface, it will first list all the properties, then all the methods, but this breaks the V-table for the COM interface. Calling any out of order method will jump to the wrong code location and hard-crash the application in most instances, or produce totally unexpected behavior in others.

### Details
Any interface that gets consumed or generated for interop needs to perfectly preserve the ordering of function declarations so that the V-table doesn't change. ILSpy seems to use its own sorting order, putting fields, then properties, and then methods. While this may be reasonable for classes and non-interop interfaces, this results in code that crashes if run after recompiling the decompiled results. It's possible that ILSpy already preserves ordering for interfaces tagged with the `StructLayoutAttribute`, but this attribute has never been applied by the .NET SDK tools which generate COM interface definitions from COM Type Libraries. Arguably `StructLayoutAttribute` may not have been deemed appropriate since it typically refers to the layout of fields and isn't as obviously applicable to the V-table.

I would propose one of three options:
* Always preserve the ordering of the IL when running "Decompile to source". This ensures that even non-interop V-tables have exactly the same layouts for maximum compat with the original IL. There could be other scenarios where ordering makes some small difference such as serialization scenarios. It appears there is some debate online about whether serializers should take dependencies on ordering, but some clearly do.
[...]

PR #2639, opened 2022-03-05
Fix the ordering of COM interface methods, properties, and events to appear in their originally defined order

zhuman:
… This was resulting in fatal crashes while running decompiled COM interop code.

#2638

### Problem
See #2638 for more details on the problem being solved

### Solution
* I have factored out the problem of determining whether native member ordering is required into its own function. If it's not required, ILSpy will still group the members by kind (fields, properties, methods, events).
* Added a new correctness unit test that validates COM V-table ordering does not change across a decompile loop


zhuman:
Ah, I see ILSpy just moved to .NET 6 in the intervening week and broke my test haha. I will update it to something that works with .NET 6.

siegfriedpammer:
Thank you for your contribution! Sorry for taking so long to get to this.

PR #2649, opened 2022-03-10
Fix output artifacts for CI builds

yyjdelete:
1. `app.config` should be no longer required for net6.0;
2. include `*.deps.json` and `*.runtimeconfig.json` in output zip;
3. move gcServer setting from `app.config` to `ILSpy.runtimeconfig.json` for net6.0;


siegfriedpammer:

siegfriedpammer:
Looks like we still need app.config as I am now seeing CS1705 when working locally.

@yyjdelete do you have any documentation that states that app.config or rather *binding redirects* are no longer needed in .NET 6? Would love to read more about this.

I guess, I will put back app.config for the time being.

yyjdelete:
@siegfriedpammer
It's strange I can only see some warn [MSB3277](https://docs.microsoft.com/en-us/visualstudio/msbuild/errors/msb3277?view=vs-2022) for ILSpy.Addin.*, but no CS1705 when build master with net6.0.

I think net472 support is dropped in ILSpy 8.0 and it only support net6.0(and newer) now, is it right? If any version of net framework is still need to be support, `app.config` should be kept.

Seems `app.config` is actual no long needed by netcore
https://docs.microsoft.com/en-us/dotnet/core/porting/upgrade-assistant-winforms-framework#upgrade-the-project-file
https://docs.microsoft.com/en-us/dotnet/core/runtime-config/
https://docs.microsoft.com/en-us/dotnet/core/runtime-config/garbage-collector (for ServerGarbageCollection)

For `binding redirects` I remember it's removed, but can't found the doc for it.

siegfriedpammer:
We have two projects (ILSpy.AddIn and ILSpy.AddIn2022 for VS) which still use net472 and these are the projects that reported CS1705 on my machine. I don't think it's much of a problem to keep the app.config for now.

PR #2665, opened 2022-04-03
Convert solution-level dependencies to project-level dependencies.

siegfriedpammer:
Solution-level dependencies are incompatible with slnf.


christophwille:
lgtm

PR #2677, opened 2022-04-26
Assume conventionally named unresolved method references are properties or events

fowl2:
eg. decompile to `.Property` instead of `.get_Property()`

### Problem

Two assemblies:
``` csharp 
using System;

namespace UnknownNamespace
{
    public class UnknownClass<T>
    {
        public T UnknownProperty { get; set; }
        public event EventHandler<T> OnEvent;

        public T this[T index]
        {
            get => default;
            set { }
        }

        public T this[T x, T y]
        {
            get => default;
            set { }
        }

    }
}

```

And

``` csharp
using System;

namespace ClassLibrary1
{
    public class Class1 : EventArgs
    {
        public void Method()
        {
            var instance = new UnknownNamespace.UnknownClass<Class1>();

            var valueOfProp = instance.UnknownProperty;
            instance.UnknownProperty = valueOfProp;

            instance.OnEvent += Instance_OnEvent;
            instance.OnEvent -= Instance_OnEvent;
            
            var valueOfIndexer = instance[this];
            instance[this] = valueOfIndexer;

            var valueOfMultiIndexer = instance[this, this];
            instance[this, this] = valueOfMultiIndexer;
        }

        private void Instance_OnEvent(object sender, EventArgs e)
        {
            throw new NotImplementedException();
        }
    }
}
```

If the first is unresolvable/missing, the second can't desugar properties or event handlers and decompiles to:
``` csharp
public void Method()
{
	UnknownClass<Class1> instance = new UnknownClass<Class1>();
	Class1 valueOfProp = instance.get_UnknownProperty();
	instance.set_UnknownProperty(valueOfProp);
	instance.add_OnEvent((EventHandler<Class1>)Instance_OnEvent);
	instance.remove_OnEvent((EventHandler<Class1>)Instance_OnEvent);
	Class1 valueOfIndexer = instance.get_Item(this);
	instance.set_Item(this, valueOfIndexer);
	Class1 valueOfMultiIndexer = instance.get_Item(this, this);
	instance.set_Item(this, this, valueOfMultiIndexer);
}
```

### Solution

While technically we can't be 100% sure, it's a pretty safe bet to assume calls to `get_`, `add_`, etc. are properties and event handers.

With this PR, the method decompiles to this:
``` csharp
public void Method()
{
	UnknownClass<Class1> instance = new UnknownClass<Class1>();
	Class1 valueOfProp = (instance.UnknownProperty = instance.UnknownProperty);
	instance.OnEvent += Instance_OnEvent;
	instance.OnEvent -= Instance_OnEvent;
	Class1 valueOfIndexer = (instance[this] = instance[this]);
	Class1 valueOfMultiIndexer = (instance[this, this] = instance[this, this]);
}
```

> * Any comments on the approach taken, its consistency with surrounding code, etc.

1. This creates a new `FakeProperty` or `FakeEvent` for each method reference, which means that eg. getters and setter won't be associated with each other. It's consistent in that `UsedBy` still won't work, I guess.

2. I'm not sure how `raise_` (`.fire`) works, but I implemented it anyway to fit the pattern. 

3. I don't think it needs a configuration to disable/enable, as the regular desugaring option should work.

> * Which part of this PR is most in need of attention/improvement?
> * [x] At least one test covering the code changed

### DIFF
--- a/ICSharpCode.Decompiler/TypeSystem/MetadataModule.cs
+++ b/ICSharpCode.Decompiler/TypeSystem/MetadataModule.cs
@@ -626,7 +626,110 @@ namespace ICSharpCode.Decompiler.TypeSystem
 				parameters.Add(new DefaultParameter(type, ""));
 			}
 			m.Parameters = parameters;
+
+			GuessFakeMethodAccessor(declaringType, name, signature, m, parameters);
+
 			return m;
+		}
+
+		private void GuessFakeMethodAccessor(IType declaringType, string name, MethodSignature<IType> signature, FakeMethod m, List<IParameter> parameters)
+		{
+			if (signature.GenericParameterCount > 0)
+				return;
+
+			var guessedGetter = name.StartsWith("get_", StringComparison.Ordinal);
+			var guessedSetter = name.StartsWith("set_", StringComparison.Ordinal);
+			if (guessedGetter || guessedSetter)
+			{
+				var propertyName = name.Substring(4);
+
+				var fakeProperty = new FakeProperty(Compilation) {
+					Name = propertyName,
+					DeclaringType = declaringType,
+					IsStatic = m.IsStatic,
+				};
+
+				if (guessedGetter)
+				{
+					if (signature.ReturnType.Kind == TypeKind.Void)
+						return;
+
+					m.AccessorKind = MethodSemanticsAttributes.Getter;
+					m.AccessorOwner = fakeProperty;
+					fakeProperty.Getter = m;
+					fakeProperty.ReturnType = signature.ReturnType;
+					fakeProperty.IsIndexer = parameters.Count > 0;
+					fakeProperty.Parameters = parameters;
+					return;
+				}
+
+				if (guessedSetter)
+				{
+					if (parameters.Count < 1 || signature.ReturnType.Kind != TypeKind.Void)
+						return;
+
+					m.AccessorKind = MethodSemanticsAttributes.Setter;
+					m.AccessorOwner = fakeProperty;
+					fakeProperty.Getter = m;
+					fakeProperty.ReturnType = parameters.Last().Type;
+					fakeProperty.IsIndexer = parameters.Count > 1;
+					fakeProperty.Parameters = parameters.SkipLast(1).ToArray();
+					return;
+				}
+			}
+
+			const string addPrefix = "add_";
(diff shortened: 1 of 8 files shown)


ICSharpCode.Decompiler/TypeSystem/MetadataModule.cs
```diff
@@ -626,8 +626,70 @@ IMethod CreateFakeMethod(IType declaringType, string name, MethodSignature<IType
...
+			{
+				var propertyName = name.Substring(4);
+				m.AccessorKind = guessedGetter ? MethodSemanticsAttributes.Getter :
+								 guessedSetter ? MethodSemanticsAttributes.Setter :
+								 default;
+				m.AccessorOwner = new FakeProperty(Compilation) {
+					Name = propertyName,
+					DeclaringType = declaringType,
+					ReturnType = guessedGetter ? signature.ReturnType
+								: guessedSetter ? parameters.Last().Type
```
dgrunwald, line 649: I believe this will crash when calling a parameterless method `void set_x() {}`

ICSharpCode.Decompiler/TypeSystem/MetadataModule.cs
```diff
@@ -626,8 +626,70 @@ IMethod CreateFakeMethod(IType declaringType, string name, MethodSignature<IType
...
+							   guessedRemove ? name.Substring(removePrefix.Length) :
+							   guessedRaise ? name.Substring(raisePrefix.Length)
+							 : throw new InvalidOperationException(),
+						DeclaringType = declaringType,
+						IsStatic = m.IsStatic,
+						AddAccessor = guessedAdd ? m : null,
+						RemoveAccessor = guessedRemove ? m : null,
+						InvokeAccessor = guessedRaise ? m : null,
+						ReturnType = guessedAdd ? parameters.Single().Type :
+									 guessedRemove ? parameters.Single().Type :
```
dgrunwald, line 681: There's no guarantee that `parameters.Count == 1`; this needs to be validated before attempting to create a fake event.

dgrunwald:
For tests, I think `ILPrettyTests` would be the best fit -- with an .il input file, there's no need for the referenced assemblies to exist.
e.g. see the existing `UnknownTypes.il` test

fowl2:
resolved review comment and added a test :)

PR #2678, opened 2022-04-27
Allow user to provide ID when generating a PDB

andrewcrawley:
### Problem

The Visual Studio debugger's decompiler integration needs to be able to control the ID used for generated PDBs.  There are a couple of scenarios that are interesting to us:

1. Generating a PDB for an assembly that was built without debug info. The PDB writer currently fails in this case, since the input assembly has no debug directory from which to extract the relevant info.  The debugger can provide values that will allow us to load the generated PDB.
2. Generating a PDB for an assembly that has multiple debug directories.  The PDB writer currently uses the first debug directory it finds, but this isn't necessarily the correct one.  The debugger can disambiguate these and provide the correct values.

### Solution

This commit adds a new parameter to `PortablePdbWriter.WritePdb` that allows the caller to specify the exact Guid and timestamp that should be used in the generated PDB.


siegfriedpammer:

PR #2679, opened 2022-04-28
Add Debug Directory raw data and sub tree nodes

fowl2:
Add Debug Directory sub tree nodes, with special handling for:
* CodeView
* EmbeddedPortablePdb
* PdbChecksum

image

Raw data column:
image


siegfriedpammer:
Thank you for your contribution.

PR #2687, opened 2022-05-10
[BIOC-6590] dotnet analyzer - icsharp decompiler throws an"failed to resolve assembly" exception on system library

lilach-freedberg:
Link to issue(s) this covers

### Problem
entire decompilation fails when failing to locate an assembly

### Solution
Catch exception + remove assertion that tends to fail (separate issue)

### DIFF
--- a/ICSharpCode.Decompiler/TypeSystem/ApplyAttributeTypeVisitor.cs
+++ b/ICSharpCode.Decompiler/TypeSystem/ApplyAttributeTypeVisitor.cs
@@ -220,8 +220,8 @@ namespace ICSharpCode.Decompiler.TypeSystem
 
 		void ExpectDummyNullabilityForGenericValueType()
 		{
-			var n = GetNullability();
-			Debug.Assert(n == Nullability.Oblivious);
+			//var n = GetNullability();
+			//Debug.Assert(n == Nullability.Oblivious);
 		}
 
 		public override IType VisitArrayType(ArrayType type)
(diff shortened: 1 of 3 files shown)


siegfriedpammer:
> entire decompilation fails when failing to locate an assembly

There is already a solution for this: setting `DecompilerSettings.ThrowOnAssemblyResolveErrors` to `false` or passing `false` as the `throwOnErrors` argument of the `UniversalAssemblyResolver` ctor.

ICSharpCode.Decompiler/TypeSystem/ApplyAttributeTypeVisitor.cs
```diff
@@ -220,8 +220,8 @@ Nullability GetNullability()
 
 		void ExpectDummyNullabilityForGenericValueType()
 		{
-			var n = GetNullability();
```
siegfriedpammer, line 223: What's the point of this? You cannot just comment out code that does not work in your case!

siegfriedpammer:
Please open an issue describing the errors you are seeing with thorough reproduction steps. Thanks!

ICSharpCode.Decompiler/TypeSystem/DecompilerTypeSystem.cs
```diff
@@ -260,19 +260,14 @@ private async Task InitializeAsync(PEFile mainModule, IAssemblyResolver assembly
 				{
 					referencedAssemblies.Add(asm);
 					var metadata = asm.Metadata;
-					foreach (var h in metadata.ExportedTypes)
```
siegfriedpammer, line 263: why would you remove support for exported types? Essentially now the code just visits all assembly references twice...

PR #2693, opened 2022-05-18
Add 'Jump To Entry Point' context menu item

1337vida:
This adds an function to the context menu item which automatically finds the entry point and jumps to it.

### DIFF
--- a/ILSpy/TreeNodes/AssemblyTreeNode.cs
+++ b/ILSpy/TreeNodes/AssemblyTreeNode.cs
@@ -687,4 +687,35 @@ namespace ICSharpCode.ILSpy.TreeNodes
 		}
 	}
 
+	[ExportContextMenuEntry(Header = nameof(Resources._JumpToEntryPoint))]
+	sealed class JumpToEntryPoint : IContextMenuEntry
+	{
+		public bool IsVisible(TextViewContext context)
+		{
+			return GetEntryPoint() != null;
+		}
+
+		public bool IsEnabled(TextViewContext context) 
+		{
+			return true;
+		}
+
+		public void Execute(TextViewContext context)
+		{
+			MethodDefinition entryPoint = GetEntryPoint();
+			if (entryPoint != null)
+			{
+				MainWindow.Instance.JumpToReference(entryPoint);
+			}
+		}
+
+		public MethodDefinition GetEntryPoint(TextViewContext context)
+		{
+			if (context.SelectedTreeNodes == null || context.SelectedTreeNodes.Length != 1)
+				return null;
+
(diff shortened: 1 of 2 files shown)


ILSpy/TreeNodes/AssemblyTreeNode.cs
```diff
@@ -687,4 +687,35 @@ public void Execute(TextViewContext context)
...
+			}
+		}
+
+		public MethodDefinition GetEntryPoint(TextViewContext context)
+		{
+			if (context.SelectedTreeNodes == null || context.SelectedTreeNodes.Length != 1)
+				return null;
+
+			AssemblyTreeNode asmNode = context.SelectedTreeNodes[0] as AssemblyTreeNode;
+			return asmNode?.LoadedAssembly?.AssemblyDefinition?.EntryPoint;
```
siegfriedpammer, line 718: This code will not work with the latest version of ILSpy. There is no `AssemblyDefinition` property on `LoadedAssembly`.

ILSpy/TreeNodes/AssemblyTreeNode.cs
```diff
@@ -687,4 +687,35 @@ public void Execute(TextViewContext context)
 		}
 	}
 
+	[ExportContextMenuEntry(Header = nameof(Resources._JumpToEntryPoint))]
+	sealed class JumpToEntryPoint : IContextMenuEntry
+	{
+		public bool IsVisible(TextViewContext context)
+		{
+			return GetEntryPoint() != null;
```
siegfriedpammer, line 695: The method is defined as taking one parameter.

ILSpy/TreeNodes/AssemblyTreeNode.cs
```diff
@@ -687,4 +687,35 @@ public void Execute(TextViewContext context)
...
+		}
+
+		public bool IsEnabled(TextViewContext context) 
+		{
+			return true;
+		}
+
+		public void Execute(TextViewContext context)
+		{
+			MethodDefinition entryPoint = GetEntryPoint();
```
siegfriedpammer, line 705: Here the `context ` parameter is missing as well...
It seems like you did not try to run this code before creating a pull request...

siegfriedpammer:
Honestly I am not sure, if this feature is really necessary... It takes two clicks to use this feature: Right click + Left click on the menu item.
The same is necessary to click on the Assembly node, then move to the decompiler view and click on the entry point link:

image

1337vida:
Yes, but i think its an nice addition to have. I think its more intuitive to right click on assembly and click "Jump to Entry Point". Sorry for the broken code, I will fix it when I'm home.

siegfriedpammer:
Are you going to open another pull request?

Issue #2707, opened 2022-06-11
ILSpyCmd: Empty solution name if output is to current directory


avengerx:
### Steps to reproduce
Decompile a dll with `ilspycmd` using `--outputdir .` and `--project`

### Result

On *nix systems, "no solution file would be created". Because it would be named `.sln`.

### Details
- ILSpyCmd 8.0.0.7047

The fix is pretty simple: [IlspyCmdProgram.cs:137](https://github.com/icsharpcode/ILSpy/blob/e2d79774cd3eb98b2b14f2a1fbeda1d9d0089d60/ICSharpCode.ILSpyCmd/IlspyCmdProgram.cs#L137) will return an empty name if `OutputDirectory` is `.`.

I have this suggestion. Transform that line, that reads as:

```csharp
SolutionCreator.WriteSolutionFile(Path.Combine(Environment.CurrentDirectory, OutputDirectory, Path.GetFileNameWithoutExtension(OutputDirectory) + ".sln"), projects);
```

Into this set of lines:

```csharp
var fullOutputPath = Path.Combine(Environment.CurrentDirectory, OutputDirectory);
fullOutputPath = new DirectoryInfo(fullOutputPath).FullName;
var solutionName = Path.GetFileNameWithoutExtension(fullOutputPath);
SolutionCreator.WriteSolutionFile(Path.Combine(fullOutputPath, solutionName + ".sln"), projects);
```

This should always resolve the output directory to its actual destination, even with relative paths and `.` in it.

siegfriedpammer:
Thank you for reporting this issue!

PR #2708, opened 2022-06-11
Fixes solution name when output directory is a relative path.

avengerx:
It will always use the output directory name as the solution name.

#2707 

### Problem
Solution file name becomes `.sln` if `--outputdir` is `.`.

### Solution

Uses `System.IO.DirectoryInfo` to expand and resolve the output path before trying to infer what file name to use as solution name.


siegfriedpammer:
I have pushed another solution which does not use `DirectoryInfo` but uses `Path.GetFullPath` instead, also I have added the resolving before any other action is performed, so all code that depends on correct path names uses the one we resolve at the beginning.

Issue #2038, opened 2020-06-14
Not possible to save project with very long names


greenozon:
ILSpy version 6.0.0.5800-preview4

Proposal is to trim down such long names when exporting into project (as the OS has the limitation of 256 chars total)

```
System.IO.DirectoryNotFoundException: Could not find a part of the path 'C:\------------------------------------------------------------------------------------------------------------------------------------------------------------\-lambda_d951b2ca66984381cd319dfda8800370-.cs'.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.StreamWriter.CreateFile(String path, Boolean append, Boolean checkHost)
   at System.IO.StreamWriter..ctor(String path, Boolean append, Encoding encoding, Int32 bufferSize, Boolean checkHost)
   at System.IO.StreamWriter..ctor(String path)
   at ICSharpCode.Decompiler.CSharp.ProjectDecompiler.WholeProjectDecompiler.<>c__DisplayClass39_0.<WriteCodeFilesInProject>b__2(IGrouping`2 file)
   at System.Threading.Tasks.Parallel.<>c__DisplayClass17_0`1.<ForWorker>b__1()
   at System.Threading.Tasks.Task.InnerInvokeWithArg(Task childTask)
   at System.Threading.Tasks.Task.<>c__DisplayClass176_0.<ExecuteSelfReplicating>b__0(Object <p0>)

```

https://www.sendspace.com/file/koowft

siegfriedpammer:
This does not crash on my system. The path is longer than 250 characters. Which operating system are you using. If I recall correctly, ILSpy already supports paths longer than 250 characters.

greenozon:
W7 x64
the question is to WInAPI I guess which allows up to MAX_PATH symbolx (260?)

dgrunwald:
[...]

PR #2709, opened 2022-06-11
Adjusts Max path lengths

avengerx:
#2707 (reviews over #2038)

### Problem
I believe the registry setting was misunderstood and the original problem triggered because windows could store 260-characters-long file names and linux/macos would only be able to handle 255.

In fact, the registry setting being used to set 255 limit on Windows actually [raises limits to over 30,000 characters](https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=cmd)!

This change should make ILSpy handle it nicely across OSes. Mac M1's (using Apple silicon, or ARM architecture) should fit this limitation nicely, I really think that we would need to look up .NET supported platforms and their respective file systems whether a given filesystem-architecture combination allows less than 255 chars. It really looks like 255 is a safe value for the time being.

### Solution
- Raised max file length limit to 260 when `LongPathsEnabled` is not set in registry in Windows Systems.
- Raised max file length limit to 32,767 when `LongPathsEnabled` is set in registry in Windows Systems.
- Set default/failover file length limit to 255. Should be safe in other supported .NET architectures.

### To-do!

Given the [Microsoft documentation on maximum path limit feature](https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=cmd):

1. it suggests the feature was supported prior to Windows 10, just in another mechanism, so perhaps checking by the registry key alone wouldn't do.
2. application manifest's `LongPathAware` should be used in order to allow the built Windows binaries to fully use the real long path support feature (or to correctly use it?)

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
+++ b/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
@@ -554,16 +554,16 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 						var value = (int?)fileSystem.GetValue("LongPathsEnabled");
 						if (value == 1)
 						{
-							return (true, int.MaxValue, 255);
+							return (true, int.MaxValue, 32767);
 						}
-						return (false, 200, 30);
+						return (false, 200, 260);
 					default:
-						return (false, 200, 30);
+						return (false, 200, 255);
 				}
 			}
 			catch
 			{
-				return (false, 200, 30);
+				return (false, 200, 255);
 			}
 		}
 


siegfriedpammer:
> 1. it suggests the feature was supported prior to Windows 10, just in another mechanism, so perhaps checking by the registry key alone wouldn't do.

Yes, but only via the "extended path syntax" enabled by `\\?\`, which is another can of worms, were not interested in opening.

siegfriedpammer:
> 2\. application manifest's `LongPathAware` should be used in order to allow the built Windows binaries to fully use the real long path support feature (or to correctly use it?)

That's what we already do in ILSpy:
https://github.com/icsharpcode/ILSpy/blob/e2d79774cd3eb98b2b14f2a1fbeda1d9d0089d60/ILSpy/app.manifest#L58-L59

There is nothing to do here.

siegfriedpammer:
> * Raised max file length limit to 32,767 when `LongPathsEnabled` is set in registry in Windows Systems.

Believe me, when implementing these changes I did a lot of research and testing. There is no evidence at all that Windows support **segment lengths** > 260 characters.
Did you even once try whether this actually works? I just did and guess what: it DID NOT work at all!

My system:
Windows [Version 10.0.19043.1706]
With long path support enabled in the registry.

Example 1: cmd.exe

The following directory name is 100 times "abcdefg....wxyz" (i.e. 26 characters times 100 == 2600 characters)

```
C:\temp>mkdir abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz

```

If I lower it to 9 times (= 234 characters):

```
mkdir abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz
```
No error occurs and it works as expected.

Example 2: Windows Explorer

Trying to create a directory or file with the same 2600 character name just results in Windows Explorer truncating the directory/file name to something shorter.

siegfriedpammer:
Until you provide more evidence that this actually works, I will have to decline this request.

avengerx:
The solution works for me and fixed the problem I had extracting embedded resources with deep paths (which are turned into file names, as it is not obvious how many dots should become slashes in that case). This helped as well if I don't expand namespaces to directories while extracting the project (in projects with lots of deep directories).

It works fine on my system for up to more than 30 chars (Windows), but in all honesty I didn't try edge cases. I don't have (and never had) the registry key `LongPathsEnabled` as `1`.

I provided the scenario in the referenced issue. It fixed that scenario. I provided this pull request as "works for me" but if you think it's broken and it broken on your tests, I can't argue. It's a good thing it is open source so I can at least benefit from the fix for my use case and a requirement for me. :)

> Believe me, when implementing these changes I did a lot of research and testing. There is no evidence at all that Windows support segment lengths > 260 characters.

I'm sorry this confronts your deep research, before making the pr I checked up the changes in the affected file and found the issue that introduced the limits. In both the issue and file I didn't find hints as to why the limit was 30, so I felt like sharing my solution in case it was welcome.

avengerx:
I found this fiddling on **cmd**. Interestingly enough, **cygwin** does create files longer than that... beats me.

```
In a quick test here using `cmd` I could create these files: (length then name)

```
- 225: 00_456789_01_456789_02_456789_03_456789_04_456789_05_456789_06_456789_07_456789_08_456789_09_456789_10_456789_11_456789_12_456789_13_456789_14_456789_15_456789_16_456789_17_456789_18_456789_19_456789_20_456789_21_456789_2.txt
- 225: 00_456789_01_456789_02_456789_03_456789_04_456789_05_456789_06_456789_07_456789_08_456789_09_456789_10_456789_11_456789_12_456789_13_456789_14_456789_15_456789_16_456789_17_456789_18_456789_19_456789_20_456789_21_456789_22_.t
- 224: 00_456789_01_456789_02_456789_03_456789_04_456789_05_456789_06_456789_07_456789_08_456789_09_456789_10_456789_11_456789_12_456789_13_456789_14_456789_15_456789_16_456789_17_456789_18_456789_19_456789_20_456789_21_456789_22_4
- 225: 00_456789_01_456789_02_456789_03_456789_04_456789_05_456789_06_456789_07_456789_08_456789_09_456789_10_456789_11_456789_12_456789_13_456789_14_456789_15_456789_16_456789_17_456789_18_456789_19_456789_20_456789_21_456789_22_45
```

But then, in order to be able to create such a file, say, within `my_directory\my_other_directory`, I suspect we should avoid just appending the path. There should happen some kind of "chdir" and only then "fopen()" with no path. I will test this in a project and see what I can trim it down to.

Issue #2706, opened 2022-06-11
Filenames and directories truncated to 30 characters


avengerx:
### Input code
Just make a simple project with an embedded resource file name with more than 30 characters, example:
`mylongfilenameherebuthereandherealso.txt`

### Erroneous output
The file will be output as:
`mylongfilenameherebuthereandhe.txt` (if I counted right, 30 characters)

### Details
* Product in use: ILSpyCmd (dotnet tool),
* Version in use: 8.0.0.7047 (prerelease, pulled from latest `master` build workflow)

This was also reproduced in 7.2.0.0.rc version of the `Our Linux/Mac/Windows ILSpy UI based on [Avalonia](http://www.avaloniaui.net/) - check out https://github.com/icsharpcode/AvaloniaILSpy` (found in [README.md](https://github.com/icsharpcode/ILSpy/blob/v8.0-preview1/README.md?plain=1#L18)

Also reproduced in the windows GUI I downloaded, version: ILSpy version 7.2.1.6856. (sorry, I'm not sure, was it from MS Store?.. not sure... it works better than AvalonUI one).

Besides embedded resource filenames, for instance, if we don't `--nested-directories`, and the namespace goes beyond 30 characters, the same happens to the output directory name: truncated at 30 characters. The actual namespace in the output file is complete.

While all links in the end project works well, in case the project has a reference or path to that embedded resource (if using string/reflection for instance), that would break the decompiled code (thus I believe this is a bug).

avengerx:
I see... I ... think... I see.. :P

But if `HKLM/System/CurrentControlSet/Control/FileSystem/LongPathsEnabled` is 0 on windows (95+?) this just means file name length is limited to 260 characters, not just 30. When this setting is 1, it means Windows FS will support filenames with _over_ 260 chars. That's what I read here: [LongPathsEnabled in Windows docs Q&A site](https://docs.microsoft.com/en-us/answers/questions/293227/longpathsenabled.html)

For the long-term record, a summary of the question:
> _"Is the registry setting [HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem] LongPathsEnabled supported in Windows Server 2019? Is it supposed to work?"_

And the answer:
> _"By setting the registry key LongPathsEnabled to 1, most of the common Win32 file and directory functions will be able to use file paths exceeding 260 characters._
> 
[...]

Issue #2622, opened 2022-02-05
Crash when Attempting to Decompile Microsoft.Exchange.Monitoring.ActiveMonitoring.Local.Components.dll And Stops All Decompilation


gwillcox-r7:
### Steps to reproduce
1. Get a copy of Microsoft.Exchange.Monitoring.ActiveMonitoring.Local.Components.dll from one of the Exchange 2019 patch files such as https://download.microsoft.com/download/b/a/1/ba1bf2f2-91f1-4670-882f-2d62506444e0/Exchange2019-KB5007409-x64-en.msp. Sorry not sharing it here due to potential copywrite issues, but if you have trouble finding it, let me know.
2. Load it into ILSpy and attempt to export this code along with other DLLs from the patch into a project file.
3. Find that not only does it crash, but also all of the decompilation stops at this point, leaving the decompiled state in an half finished state instead of fully completing despite the error.

### Error message shown

```
Failed to decompile the assembly 'C:\Users\gwillcox\Documents\CVE-2021-42321\Before\Microsoft.Exchange.Monitoring.ActiveMonitoring.Local.Components.dll':
System.AggregateException: One or more errors occurred. ---> System.IO.DirectoryNotFoundException: Could not find a part of the path 'C:\Users\gwillcox\Documents\CVE-2021-42321\Before\Dec IlSpy\Microsoft.Exchange.Monitoring.ActiveMonitoring.Local.Components\Microsoft\Exchange\Monitoring\ActiveMonitoring\Local\CafeV2\Probes\Detectors\HttpRequestDetectors\SmokeTestDetector\SmokeTestDetector.cs'.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.StreamWriter.CreateFile(String path, Boolean append, Boolean checkHost)
[...]

PR #2711, opened 2022-06-12
Checks maximum path length using .NET features.

avengerx:
icsharpcode/ILSpy#2706.

### Problem
Okay, this is my best shot. I have heavily simplified `CleanUpName()` method, sacrificing the file truncation functionality in favor of the much more relaxed file name length.

I have tested the edge cases, but by using `new DirectoryInfo(path).FullName` basically I enable any long path support .NET might like, as I could create the files with much longer than 260 characters path during my tests. The problems turns in limitations with Windows applications like Visual Studio, Notepad++, etc. Surprisingly though, puny stock Windows Notepad 
 (`%windir%\system32\notepad.exe`) could read files with very long paths without issues.

Adding the path limit to 258 characters allowed all applications I have to read the generated files (I tested the longest file path the project generated of course), such that if I increased the limit to 259 the applications won't be able to (Visual Studio 2022 actually could, but no longer if I changed it to 260).

I tried also to make very descriptive exceptions when the limit is reached to help users identify it and report if this ever gets accepted mainstream.

### Solution

- rely in `DirectoryInfo` and `FileInfo` classes to validate paths in every write operation
- raise windows hardcoded limits to the absolute maximum usual applications would work with
- no longer truncate -- throw descriptive exceptions pointing the attempted path and its actual length
- use absolute paths on all file operations (expand `TargetDirectory`)

### Disclaimer

This works for me. If you don't fancy the feature or wasting more time on it, feel free to close the pull request on spot.

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
+++ b/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
@@ -132,20 +132,36 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 
 		public void DecompileProject(PEFile moduleDefinition, string targetDirectory, CancellationToken cancellationToken = default(CancellationToken))
 		{
-			string projectFileName = Path.Combine(targetDirectory, CleanUpFileName(moduleDefinition.Name) + ".csproj");
+			var targetDirectoryInfo = new DirectoryInfo(targetDirectory);
+
+			if (TargetDirectory != targetDirectoryInfo.FullName)
+			{
+				TargetDirectory = targetDirectoryInfo.FullName;
+			}
+
+			string projectFileName = Path.Combine(TargetDirectory, CleanUpFileName(moduleDefinition.Name) + ".csproj");
+
 			using (var writer = MakeStreamWriter(projectFileName))
 			{
-				DecompileProject(moduleDefinition, targetDirectory, writer, cancellationToken);
+				DecompileProject(moduleDefinition, targetDirectoryInfo, writer, cancellationToken);
 			}
 		}
 
-		public ProjectId DecompileProject(PEFile moduleDefinition, string targetDirectory, TextWriter projectFileWriter, CancellationToken cancellationToken = default(CancellationToken))
+		public ProjectId DecompileProject(PEFile moduleDefinition, string targetDirectory, TextWriter projectFileWriter, CancellationToken cancellationToken = default(CancellationToken)) =>
+			DecompileProject(moduleDefinition, new DirectoryInfo(targetDirectory), projectFileWriter, cancellationToken);
+
+		public ProjectId DecompileProject(PEFile moduleDefinition, DirectoryInfo targetDirectory, TextWriter projectFileWriter, CancellationToken cancellationToken = default(CancellationToken))
 		{
-			if (string.IsNullOrEmpty(targetDirectory))
+			if (targetDirectory == null)
 			{
 				throw new InvalidOperationException("Must set TargetDirectory");
 			}
-			TargetDirectory = new DirectoryInfo(targetDirectory).FullName;
+
+			if (TargetDirectory != targetDirectory.FullName)
+			{
+				TargetDirectory = targetDirectory.FullName;
+			}
+
 			directories.Clear();
 			var files = WriteCodeFilesInProject(moduleDefinition, cancellationToken).ToList();
 			files.AddRange(WriteResourceFilesInProject(moduleDefinition));
--- a/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
+++ b/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
@@ -758,7 +758,7 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 
 		#region Full path length checking helpers
 
-		static bool UnderWindows = Environment.OSVersion.Platform == PlatformID.Win32NT;
+		static readonly bool UnderWindows = Environment.OSVersion.Platform == PlatformID.Win32NT;
 
 		private DirectoryInfo CreateDir(string path) => Directory.CreateDirectory(ValidatePath(path, true));
 
@@ -777,8 +777,6 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 			var (supportsLongPaths, maxPathLength, maxSegmentLength) = longPathSupport.Value;
 			if (!Path.IsPathRooted(path))
 				throw new Exception("Non-root path passed to ValidatePath().");
-
-			var onWin = Environment.OSVersion.Platform == PlatformID.Win32NT;
 
 			string dotnetPath;
 			try
@@ -821,7 +819,7 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 
 			if (dotnetPath.Length > maxPathLength)
 			{
-				if (onWin && !supportsLongPaths)
+				if (UnderWindows && !supportsLongPaths)
 				{
 					throw new PathTooLongException("Path is too long. Files could be created, but they won't be accessible by most applications." + Environment.NewLine +
 						"Path: " + dotnetPath + Environment.NewLine +
--- a/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
+++ b/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
@@ -133,7 +133,7 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 		public void DecompileProject(PEFile moduleDefinition, string targetDirectory, CancellationToken cancellationToken = default(CancellationToken))
 		{
 			string projectFileName = Path.Combine(targetDirectory, CleanUpFileName(moduleDefinition.Name) + ".csproj");
-			using (var writer = new StreamWriter(projectFileName))
+			using (var writer = MakeStreamWriter(projectFileName))
 			{
 				DecompileProject(moduleDefinition, targetDirectory, writer, cancellationToken);
 			}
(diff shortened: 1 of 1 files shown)


ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
```diff
@@ -581,115 +581,48 @@ public static string CleanUpFileName(string text)
...
 		/// <summary>
 		/// Cleans up a node name for use as a file system name. If <paramref name="separateAtDots"/> is active,
-		/// dots are seen as segment separators. Each segment is limited to maxSegmentLength characters.
-		/// (see <see cref="GetLongPathSupport"/>) If <paramref name="treatAsFileName"/> is active,
-		/// we check for file a extension and try to preserve it, if it's valid.
+		/// dots are seen as segment separators. If <paramref name="lookForExtension"/> is active,
+		/// we check for file a extension and try to preserve it when separating paths at dots.
 		/// </summary>
-		static string CleanUpName(string text, bool separateAtDots, bool treatAsFileName)
+		static string CleanUpName(string name, bool separateAtDots, bool lookForExtension)
```
siegfriedpammer, line 592: You have modified `CleanUpName` to the point, where it no longer fulfills its actual main responsibility. First and foremost, it was meant to **clean up** names (hence the name) used in managed assemblies (which can be arbitrary strings) so that all characters that are not human readable are replaced by '-'.
We added the trimming functionality, because we wanted to avoid having to scan the string multiple times, when creating filenames for resource or type names.

dgrunwald: Note that this cleaning functionality was *security critical*.
Without it is was possible that weird enough class names (e.g. involving `\..\`) cause ILSpy to write outside the output directory.

avengerx: I will review it to restore its functionality. May we at least get rid of the trimming? Make it selectable maybe? This is making embedded resources most likely broken when trimming happen to them.

avengerx: The `new DirectoryInfo` (or `FileInfo`) introduced with `ValidatePath()` nicely blocks any leftover invalid character in path from making it to an actual write. I believe all that's left as far as path _validation_ is concerned, is to add a

```csharp
if (!dotnetPath.StartsWith(TargetDirectory))
  throw new Exception("Path resolved outside of output directory: " + path + Environment.NewLine + "Resolved to: " + dotnetPath);
```

To the validation, ensuring the path is within `TargetDirectory`. At that point, any `../` or relative paths should have already been expanded by `FileSystemInfo`. This would ultimately avoid accidents, right? Now, if a file file has a backtick or dollar sign, it would be another problem (if somebody processes the output files and does not properly treat it as a string, for instance). 

As for the cleanup, something I'm worried about is that characters like `]`, `{`, and many others, including accent, UTF-8, should be accepted as path strings. For instance, the cyrillic `выход` is a valid path name that, as far as I can see, the current implementation of `CleanUp()` wouldn't allow. I want to think a way to effectively allow valid characters go, and clean up only what's clearly valid in managed assembly handles but not in paths. I.e. newlines, tabs, and other control characters.

siegfriedpammer: > The new DirectoryInfo (or FileInfo) introduced with ValidatePath() nicely blocks any leftover invalid character in path from making it to an actual write.

IIRC DIrectoryInfo and friends will do *a lot* more than just cleaning up the path and in order to accomplish that will use disc IO, which we **don't want** at all in that method, as it is called in many places, for example the ILSpy UI for TreeNode labels, not just the project decompiler.

> As for the cleanup, something I'm worried about is that characters like ], {, and many others, including accent, UTF-8, should be accepted as path strings. For instance, the cyrillic выход is a valid path name that, as far as I can see, the current implementation of CleanUp() wouldn't allow.

Then you need to take another careful look: The original implementation uses [`char.IsLetterOrDigit`](https://docs.microsoft.com/en-us/dotnet/api/system.char.isletterordigit?view=net-6.0) (see line 635 below),

https://github.com/icsharpcode/ILSpy/blob/107a5b50442a098e98e62068c29efd32c7b28c2e/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs#L630-L636

which follows the Unicode Standard and thus supports all characters that are widely in use. Also, .NET Strings are UTF-16 in memory anyway, so UTF-8 is of no concern at this point. Any string that can be stored in the ECMA-335 metadata tables, may be either UTF-8 or UTF-16 and is supported by .NET's built-in functions.

```csharp
> "выход".All(char.IsLetterOrDigit)
true
> "출구".All(char.IsLetterOrDigit)
true
> "saída".All(char.IsLetterOrDigit)
true
```

You see, **it properly supports** Cyrillic script, Hangul (Korean) script and Portuguese (see the acute accent on the i?) and I could go on with Japanese, Chinese and many others. Thank you!

avengerx: > IIRC DIrectoryInfo and friends will do a lot more than just cleaning up the path and in order to accomplish that will use disc IO, which we don't want at all in that method, as it is called in many places, not just the project decompiler.

You are right, and as I'm developing here I'm seeing the alternative _without_ using `DirectoryInfo` _except_ when we really want to write a file. I'll keep that in mind, so thanks for the heads up.

> Then you need to take another careful look: The original implementation uses [char.IsLetterOrDigit](https://docs.microsoft.com/en-us/dotnet/api/system.char.isletterordigit?view=net-6.0) (see line 635 below),

Sorry I really made you angry here again... My argument was with this in mind:

```
Console.WriteLine("Is bracket valid: " + char.IsLetterOrDigit('['));
Is bracket valid: False
```

So instead, I'm thinking in looking up _what is not valid_ instead of _what is valid_ in the cleanup. I guess it would work best if I show you the code when it's done. What's your position about regular expressions at that point?

fun fact: this came to mind because I recently noticed I have installed Linqpad in `c:\apps\linqpad5]`, and it's been working like a charm!

avengerx: @siegfriedpammer what's the plot behind [this positioning after colon](https://github.com/icsharpcode/ILSpy/blob/e2d79774cd3eb98b2b14f2a1fbeda1d9d0089d60/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs#L595-L597)? Just our usual `c:\path`? Or is there any other situation where a node will have a colon and you'd want to skip the string until it? Any that comes on top of your head so I can ensure the case is safe?

siegfriedpammer: > Sorry I really made you angry here again... 

I am sorry for being unfair to you. I can see your good intent, but so far your suggestions seem to be aimed only at solving your specific problem, while introducing a lot of other problems/bugs in other cases at the same time. I am sure, we will find a satisfying solution for your problem as well.

> So instead, I'm thinking in looking up what is not valid instead of what is valid in the cleanup.

Sorry, but "whitelisting" (only allowing a defined set of things) is considered more secure than "blacklisting" (only removing a defined set of things). As @dgrunwald already pointed out, this code must be as secure as possible, because it is used when writing files. Many of our users are security researchers, who are analyzing .NET malware with ILSpy. We need to make sure that there is absolutely no unexpected behavior when working with ILSpy. And only a allowing a "known good/harmless set of characters" is the best approach with regards to security.

Regarding brackets: Just like we allow "_" in names, we could add other valid characters to the list as well, however, it quickly will get very complicated, because there are different file systems that have different restrictions (see also https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words for more details).

We want our code to be as portable as possible, hence we have a very limited list of characters we allow in filenames.

> What's your position about regular expressions at that point?

Why use regular expressions when a simple O(n) scan is sufficient?

> what's the plot behind [this positioning after colon](https://github.com/icsharpcode/ILSpy/blob/e2d79774cd3eb98b2b14f2a1fbeda1d9d0089d60/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs#L595-L597)?

`:` is only valid in Windows paths, when the path is an absolute path, `:` is the second character and the first character is `A-Za-z`. It can only be used to denote a "drive-letter" at start of an absolute path.

```
❯ mkdir ab:
```

But you would have to ask @dgrunwald why we are stripping that part completely instead of replacing it with '-'.

avengerx: Okay, I see... Let me start this from scratch once again :rofl: now with a more conservative approach. And with clear goals:
- allow file name characters over 30 in length on Windows without long paths enabled
- check full path length only when I'm about to write files (thus properly checking full path limits when we actually have that information)

So perhaps from the next attempt we could work towards having the best of security without sacrificing path lengths, thus increasing the odds of the ability to extract embedded resources that can actually match references in the code, given the constraints that will always be there.

At least by the comprehensiveness of your responses I assume this is something you'd also want to see addressed, right?

dgrunwald: "check full path length only when I'm about to write files" -- I'm not sure why you feel the need to check this yourself when the operating system will already do so for us. 
If you don't use the total path length for any sort of truncation then there isn't at any point in checking it, is there?

dgrunwald: Assuming that .NET 6 can write long paths even if the registry is key is not set, then I would suggest just removing the registry check and proceeding as if long path support is enabled.

avengerx: > "check full path length only when I'm about to write files" -- I'm not sure why you feel the need to check this yourself when the operating system will already do so for us. If you don't use the total path length for any sort of truncation then there isn't at any point in checking it, is there?

Reasons I feel the need:
1. produce more meaningful error messages for the user, allowing them to know what exact path is breaking and by how many characters
2. avoid an issue where files can be created yet most Windows applications won't open them (with "cannot find part of path" much like #2622 initial report)

In my tests, simply letting the OS tell the length (at least on Windows 10, no long paths enabled), files generated on very long paths are able to be opened with Windows Notepad (%windir%\System32\notepad.exe).

At least Visual Studio 2022, WordPad, and Notepad++, they won't open these edge cases:
- VS2022 tells it exceeds 259 chars length,
- notepad++ silently ignores it
- WordPad closes on spot

> Assuming that .NET 6 can write long paths even if the registry is key is not set, then I would suggest just removing the registry check and always proceeding as if long path support is enabled.

It may be the case and perhaps I'm a bit lost here. The project I'm fiddling with [says it is `netstandard2.0`](https://github.com/icsharpcode/ILSpy/blob/e2d79774cd3eb98b2b14f2a1fbeda1d9d0089d60/ICSharpCode.Decompiler/ICSharpCode.Decompiler.csproj#L5),

avengerx:
Current state already passes all unit tests. Making validation specific unit tests. It's a bit tricky to go multi-platform and all... :P

dgrunwald:
I think this PR is way too complex. We should be able to simplify the code thanks to .NET 6. This PR is the opposite of that. We're not going to accept this approach.

PR #2726, opened 2022-06-26
Add JetBrain's IDE-specific config files

clin1234:
I presume that non-Visual Studio IDE configuration files are allowed uptream.


christophwille:
Please ask before providing such a PR - we usually do not accept vendor-specific files especially if no one on the team uses those tools.

PR #2729, opened 2022-07-01
Use SRM metadata to check if a type is nested.

ElektroKill:
Link to issue(s) this covers
N/A

### Problem
The implementation of `MetadataTypeDefinition` relied on the type name in order to check whether a type was nested or not. This implementation works fine for normal files. A problem occurs if someone deliberately modifies the type names in order to break ILSpy. This can be done by inserting a malicious nested type delimiter into the type name which could result in incorrect behavior during decompilation.

### Solution
The solution is rather quite simple. The information on whether a type is nested can be read straight from metadata instead of relying on type names. In this PR, I changed the code to use the `TypeDefinition.IsNested` property provided by SRM.

### DIFF
--- a/ICSharpCode.Decompiler/TypeSystem/Implementation/MetadataTypeDefinition.cs
+++ b/ICSharpCode.Decompiler/TypeSystem/Implementation/MetadataTypeDefinition.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) 2018 Daniel Grunwald
+// Copyright (c) 2018 Daniel Grunwald
 // 
 // Permission is hereby granted, free of charge, to any person obtaining a copy of this
 // software and associated documentation files (the "Software"), to deal in the Software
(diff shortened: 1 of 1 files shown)


dgrunwald:
What problem concretely do you think occurs with the code as is?

I think `fullTypeName.IsNested`  doesn't actually depend on the type name in the metadata; it just depends on whether the type has a parent type `if ((declaringTypeHandle = td.GetDeclaringType()).IsNil)`.
I think obfuscated assemblies can set `td.IsNested` independently from `DeclaringType`. ILSpy expects every nested type to have a declaring type (otherwise there might be NullReferenceExceptions).

ElektroKill:
> I think `fullTypeName.IsNested` doesn't actually depend on the type name in the metadata; it just depends on whether the type has a parent type `if ((declaringTypeHandle = td.GetDeclaringType()).IsNil)`. I think obfuscated assemblies can set `td.IsNested` independently from `DeclaringType`. ILSpy expects every nested type to have a declaring type (otherwise there might be NullReferenceExceptions).

This is correct. I made a mistake and did not look at the implementation behind the `GetFullTypeName` extension method and just assumed it relied on type names 🙃The current implementation does not actually break like I assumed it would. The PR can be closed. Sorry for the inconvenience 😅

PR #2235, opened 2020-11-29
Remove Humanizer dependency

siegfriedpammer:
Fixes #2232: We only need a small percentage of features provided by Humanizer. Therefore we copy the relevant parts and remove the dependency in order to avoid such problems in the future.

PR #2755, opened 2022-07-17
Update Vocabulary.cs

ArFlre:
More shorter condition.


christophwille:
As the folder implies, this is a copy-in-time of https://github.com/Humanizr/Humanizer/, for reason please see https://github.com/icsharpcode/ILSpy/pull/2235.

christophwille:
We are only going to copy a newer version (if ever), but not deviate from the original. Please consider providing your fix upstream with Humanizer.

PR #2760, opened 2022-08-01
Try to resolve overspecified method references

hexafluoride:
Link to issue(s) this covers

N/A

### Problem
When certain method calls* are compiled with netstandard1.x, the emitted `callvirt` instruction references a member that does not exist. At runtime, these accesses are resolved to members of base types, however ILSpy fails to do so, and fake members are generated instead.

\* = calls to methods that belong to a base type of a class. Example: [TypeInfo.Assembly](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.typeinfo?view=net-6.0#properties), which is an inherited member from Type. When an expression like `typeof(object).GetTypeInfo().Assembly` is compiled with a netstandard1.x target, the emitted instruction is:

    callvirt instance class [System.Reflection]System.Reflection.Assembly [System.Reflection]System.Reflection.TypeInfo::get_Assembly()

whereas on other targets (ex. net6.0), this is emitted:

    callvirt instance class [System.Runtime]System.Reflection.Assembly [System.Runtime]System.Type::get_Assembly()

At runtime, the former member reference is correctly resolved to Type::get_Assembly, but ILSpy's type system fails to do so. See the following screenshots for how this impacts the generated AST:

pre-patch:

incorrectly_resolved

post-patch:

correctly_resolved

This problem was encountered while decompiling an Autofac build that targets netstandard1.1 ([link](https://github.com/autofac/Autofac/releases/download/v4.8.1/autofac.4.8.1.nupkg)), so there definitely are examples of this weird IL out in the wild.

### Solution

This solution may slightly degrade method resolution performance for references that can't be resolved, although this is an operation performed once per method reference. I do not anticipate a regression for valid/runnable IL, however it is possible that previously invalid method calls (perhaps due to access modifiers) may be resolved to inaccessible members in the target type hierarchy. This would require incorrect IL to be provided, though.

These method references are found only in callvirt instructions (as call would fail with this overspecified reference), but ResolveMethodReference is oblivious to what IL instruction contains the reference. A cleaner solution would only execute this logic when resolving callvirt references, though I'm not sure that this is possible to begin with, as member reference resolution is deduplicated across the entire module from what I can tell.

Further testing to ensure no regressions.

The existing TypeSystem tests cover ResolveMethodReference. I wanted to add a test specifically for this IL, but I couldn't figure out where exactly to place it in the test hierarchy, as this is a pretty vertical testcase where handmade IL is provided and later the AST must be analyzed.

Looking forward to any feedback.

### DIFF
--- a/ICSharpCode.Decompiler/TypeSystem/MetadataModule.cs
+++ b/ICSharpCode.Decompiler/TypeSystem/MetadataModule.cs
@@ -491,6 +491,11 @@ namespace ICSharpCode.Decompiler.TypeSystem
 				var statedDeclaringType = ResolveDeclaringType(memberRef.Parent, context);
 				var statedDeclaringTypeDefinition = statedDeclaringType.GetDefinition();
 
+				// Note: statedDeclaringType might be parameterized, but the signature is for the original method definition.
+				// We'll have to search the member directly on statedDeclaringTypeDefinition.
+				signature = memberRef.DecodeMethodSignature(TypeProvider,
+					new GenericContext(statedDeclaringTypeDefinition?.TypeParameters));
+
 				if (statedDeclaringTypeDefinition != null)
 				{
 					var candidateDeclaringTypes = statedDeclaringType.GetAllBaseTypes().ToList();
@@ -506,10 +511,10 @@ namespace ICSharpCode.Decompiler.TypeSystem
 						{
 							classTypeArguments = declaringType.TypeArguments;
 						}
-						// Note: declaringType might be parameterized, but the signature is for the original method definition.
-						// We'll have to search the member directly on declaringTypeDefinition.
+
 						signature = memberRef.DecodeMethodSignature(TypeProvider,
 							new GenericContext(declaringTypeDefinition?.TypeParameters));
+
 						if (declaringTypeDefinition != null)
 						{
 							// Find the set of overloads to search:
--- a/ICSharpCode.Decompiler/TypeSystem/MetadataModule.cs
+++ b/ICSharpCode.Decompiler/TypeSystem/MetadataModule.cs
@@ -491,75 +491,70 @@ namespace ICSharpCode.Decompiler.TypeSystem
 				var statedDeclaringType = ResolveDeclaringType(memberRef.Parent, context);
 				var statedDeclaringTypeDefinition = statedDeclaringType.GetDefinition();
 
-				// Note: statedDeclaringType might be parameterized, but the signature is for the original method definition.
-				// We'll have to search the member directly on statedDeclaringTypeDefinition.
-				signature = memberRef.DecodeMethodSignature(TypeProvider,
-					new GenericContext(statedDeclaringTypeDefinition?.TypeParameters));
-
-				if (statedDeclaringTypeDefinition != null)
+				// GetAllBaseTypes() returns types ordered from base -> specific. Traverse the list in reverse
+				// to check in the opposite direction (most specific to least specific).
+				var candidateDeclaringTypes = statedDeclaringType.GetAllBaseTypes().ToList();
+				for (int i = candidateDeclaringTypes.Count - 1; i >= 0; i--)
 				{
-					var candidateDeclaringTypes = statedDeclaringType.GetAllBaseTypes().ToList();
+					var declaringType = candidateDeclaringTypes[i];
+					var declaringTypeDefinition = declaringType.GetDefinition();
 
-					// GetAllBaseTypes() returns types ordered from base -> specific. Traverse the list in reverse
-					// to check in the opposite direction (most specific to least specific).
-					for (int i = candidateDeclaringTypes.Count - 1; i >= 0; i--)
+					if (declaringType.TypeArguments.Count > 0)
 					{
-						var declaringType = candidateDeclaringTypes[i];
-						var declaringTypeDefinition = declaringType.GetDefinition();
+						classTypeArguments = declaringType.TypeArguments;
+					}
 
-						if (declaringType.TypeArguments.Count > 0)
+					// Note: declaringType might be parameterized, but the signature is for the original method definition.
+					// We'll have to search the member directly on declaringTypeDefinition.
+					signature = memberRef.DecodeMethodSignature(TypeProvider,
+						new GenericContext(declaringTypeDefinition?.TypeParameters));
+
+					if (declaringTypeDefinition != null)
+					{
+						// Find the set of overloads to search:
+						IEnumerable<IMethod> methods;
+						if (name == ".ctor")
 						{
-							classTypeArguments = declaringType.TypeArguments;
+							methods = declaringTypeDefinition.GetConstructors();
+						}
+						else if (name == ".cctor")
+						{
+							methods = declaringTypeDefinition.Methods.Where(m => m.IsConstructor && m.IsStatic);
+						}
+						else
+						{
+							methods = declaringTypeDefinition.GetMethods(m => m.Name == name, GetMemberOptions.IgnoreInheritedMembers)
+								.Concat(declaringTypeDefinition.GetAccessors(m => m.Name == name, GetMemberOptions.IgnoreInheritedMembers));
+						}
+						// Determine the expected parameters from the signature:
+						ImmutableArray<IType> parameterTypes;
+						if (signature.Header.CallingConvention == SignatureCallingConvention.VarArgs)
+						{
+							parameterTypes = signature.ParameterTypes
+								.Take(signature.RequiredParameterCount)
+								.Concat(new[] { SpecialType.ArgList })
+								.ToImmutableArray();
+						}
+						else
+						{
+							parameterTypes = signature.ParameterTypes;
+						}
+						// Search for the matching method:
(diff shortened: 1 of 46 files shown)


siegfriedpammer:
> When certain method calls* are compiled with netstandard1.x, the emitted callvirt instruction references a member that does not exist. At runtime, these accesses are resolved to members of base types, however ILSpy fails to do so, and fake members are generated instead.

It seems this is also the case for Portable Class Libraries, am I correct? At least we were seeing the same weird method references when trying to decompile a PCL. We had to disable our PCL roundtrip test of NSJ (see https://github.com/icsharpcode/ILSpy/blob/master/ICSharpCode.Decompiler.Tests/RoundtripAssembly.cs#L65).

hexafluoride:
Yes, that is correct. Check out the following C#/IL from the PCL build of Newtonsoft.Json:

    throw Newtonsoft.Json.JsonSerializationException.Create(reader, Newtonsoft.Json.Utilities.StringUtils.FormatWith("Could not find member '{0}' on object of type '{1}'", System.Globalization.CultureInfo.InvariantCulture, text, contract.UnderlyingType.get_Name()));

with the IL:

    IL_0180: callvirt instance string [System.Runtime]System.Type::get_Name()

Type::Name is actually MemberInfo::Name, so this patch fixes the member resolution and ensures that `.Name` is emitted for the C# instead of `.get_Name()`. I can confirm that the PCL Newtonsoft.Json binary found in ILSpy-tests is decompiled to compilable C# with this patch applied.

ICSharpCode.Decompiler/TypeSystem/MetadataModule.cs
```diff
@@ -477,77 +477,95 @@ IMethod ResolveMethodReference(MemberReferenceHandle memberRefHandle, GenericCon
...
-					{
-						parameterTypes = signature.ParameterTypes
-							.Take(signature.RequiredParameterCount)
-							.Concat(new[] { SpecialType.ArgList })
-							.ToImmutableArray();
-					}
-					else
+					var candidateDeclaringTypes = new List<IType>() { statedDeclaringType };
+
+					// Search through type hierarchy of declaring type, traversing up the chain.
```
siegfriedpammer, line 498: I think you could use `GetAllBaseTypes()` or `GetAllBaseTypeDefinitions()` found in `TypeSystemExtensions` to implement this.

ICSharpCode.Decompiler/TypeSystem/MetadataModule.cs
```diff
@@ -477,77 +477,95 @@ IMethod ResolveMethodReference(MemberReferenceHandle memberRefHandle, GenericCon
...
-					method = null;
-					foreach (var m in methods)
+
+					for (int i = 0; i < candidateDeclaringTypes.Count; i++)
 					{
-						if (m.TypeParameters.Count != signature.GenericParameterCount)
-							continue;
-						if (CompareSignatures(m.Parameters, parameterTypes) && CompareTypes(m.ReturnType, signature.ReturnType))
+						var declaringType = candidateDeclaringTypes[i];
+						var declaringTypeDefinition = i == 0 ? statedDeclaringTypeDefinition : candidateDeclaringTypes[i].GetDefinition();
```
siegfriedpammer, line 507: For `ITypeDefinition`s `GetDefinition()` returns `this`, so this should not be necessary.

ICSharpCode.Decompiler/TypeSystem/MetadataModule.cs
```diff
@@ -477,77 +477,95 @@ IMethod ResolveMethodReference(MemberReferenceHandle memberRefHandle, GenericCon
...
-							.ToImmutableArray();
-					}
-					else
+					var candidateDeclaringTypes = new List<IType>() { statedDeclaringType };
+
+					// Search through type hierarchy of declaring type, traversing up the chain.
+					if (statedDeclaringTypeDefinition.DirectBaseTypes?.Any() == true)
 					{
-						parameterTypes = signature.ParameterTypes;
+						candidateDeclaringTypes.AddRange(statedDeclaringTypeDefinition.DirectBaseTypes);
```
siegfriedpammer, line 501: This seems strange... why are we only interested in "one" level of base types, instead of the whole hierarchy? The comment above mentions the whole hierarchy.

ICSharpCode.Decompiler/TypeSystem/MetadataModule.cs
```diff
@@ -477,77 +477,95 @@ IMethod ResolveMethodReference(MemberReferenceHandle memberRefHandle, GenericCon
...
+
+					for (int i = 0; i < candidateDeclaringTypes.Count; i++)
 					{
-						if (m.TypeParameters.Count != signature.GenericParameterCount)
-							continue;
-						if (CompareSignatures(m.Parameters, parameterTypes) && CompareTypes(m.ReturnType, signature.ReturnType))
+						var declaringType = candidateDeclaringTypes[i];
+						var declaringTypeDefinition = i == 0 ? statedDeclaringTypeDefinition : candidateDeclaringTypes[i].GetDefinition();
+
+						if (declaringType.TypeArguments.Count > 0)
```
siegfriedpammer, line 509: The code that works with one declaring type could be extracted into a separate method/local function to increase readability?

siegfriedpammer (commented):
Thank you for the contribution! I hope, @dgrunwald will take another closer look at the details, once he has time.

Can you please `git pull --rebase` from icsharpcode:master so that we can get the tests to pass? Thanks!

ICSharpCode.Decompiler/TypeSystem/MetadataModule.cs
hexafluoride: With the latest change, I'm not sure that this is necessary. Can you check again?

hexafluoride:
Looking at the failing tests right now. Apologies, they just take like half an hour to run on my dev machine so I can't screen them all before I push 😅

ICSharpCode.Decompiler/TypeSystem/MetadataModule.cs
siegfriedpammer: Yes, I think it's fine.

ICSharpCode.Decompiler/TypeSystem/MetadataModule.cs
```diff
@@ -491,75 +491,70 @@ IMethod ResolveMethodReference(MemberReferenceHandle memberRefHandle, GenericCon
...
 
-				// Note: statedDeclaringType might be parameterized, but the signature is for the original method definition.
-				// We'll have to search the member directly on statedDeclaringTypeDefinition.
-				signature = memberRef.DecodeMethodSignature(TypeProvider,
-					new GenericContext(statedDeclaringTypeDefinition?.TypeParameters));
-
-				if (statedDeclaringTypeDefinition != null)
+				// GetAllBaseTypes() returns types ordered from base -> specific. Traverse the list in reverse
+				// to check in the opposite direction (most specific to least specific).
+				var candidateDeclaringTypes = statedDeclaringType.GetAllBaseTypes().ToList();
```
dgrunwald, line 496: Is it correct to search all base types? I would assume the runtime only searches base classes, not interfaces.

ICSharpCode.Decompiler/TypeSystem/MetadataModule.cs
```diff
@@ -477,77 +477,91 @@ IMethod ResolveMethodReference(MemberReferenceHandle memberRefHandle, GenericCon
...
+						classTypeArguments = declaringType.TypeArguments;
 					}
-					// Search for the matching method:
-					method = null;
-					foreach (var m in methods)
+
+					// Note: declaringType might be parameterized, but the signature is for the original method definition.
+					// We'll have to search the member directly on declaringTypeDefinition.
+					signature = memberRef.DecodeMethodSignature(TypeProvider,
+						new GenericContext(declaringTypeDefinition?.TypeParameters));
```
dgrunwald, line 510: I don't think this deal with `GenericContext` is correct once inheritance comes into play -- the base class may have different type parameters than the derived class.
Consider a class hierarchy like:
```
class B<T> { void Add(T); }
class C<U> : B<List<U>> { }
```
Now I believe an Add() call on `C<int>` can be encoded as a call to `C<U=int>.Add(List<U>)`. This overload won't be found in the declarationTypeDefinition of the base class. In edge cases you might end up returning the wrong overload.

dgrunwald: Really specialization needs to happen in two steps here:
We need to use the partial specialization `B<List<U>>` for comparing the signature of `Add`, and only then substitute `U=int`.
So you need to go from declaringType to declaringTypeDefinition just once for the declaringType mentioned in the IL, not for each base class separately.

siegfriedpammer:
@hexafluoride any update?

hexafluoride:
> @hexafluoride any update?

Apologies, life has gotten in the way. I will rebase and update the PR by next week.

siegfriedpammer:
@hexafluoride any update?

dgrunwald:
Feel free to re-open when you get around to updating this PR.

PR #2761, opened 2022-08-08
Set LangVersion of all net6 csprojs to default, and bump versions of a few packages.

clin1234:
Perhaps we need a GH Action to automate updates of trivial packages...

### DIFF
--- a/packages.props
+++ b/packages.props
@@ -6,7 +6,7 @@
 	-->
 	<PropertyGroup>
 		<SystemCollectionsImmutableVersion>6.0.0</SystemCollectionsImmutableVersion>
-		<SystemReflectionMetadataVersion>6.0.0</SystemReflectionMetadataVersion>
+		<SystemReflectionMetadataVersion>6.0.1</SystemReflectionMetadataVersion>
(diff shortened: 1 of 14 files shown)


christophwille:
We very much intentionally do not have such an update task.

packages.props
```diff
@@ -6,7 +6,7 @@
 	-->
 	<PropertyGroup>
 		<SystemCollectionsImmutableVersion>6.0.0</SystemCollectionsImmutableVersion>
-		<SystemReflectionMetadataVersion>6.0.0</SystemReflectionMetadataVersion>
+		<SystemReflectionMetadataVersion>6.0.1</SystemReflectionMetadataVersion>
```
christophwille, line 9: @siegfriedpammer I take we are intentionally not bumping this beyond 6.0.0 because ICS.D is using it (and thus Roslyn).

siegfriedpammer: Yes, please revert this. Thanks!

christophwille: With the exception of ICSharpCode.Decompiler.csproj - what is the point of the exercise anyways?

PR #2762, opened 2022-08-08
Add option to always fully qualify type names with global::

hexafluoride:
Link to issue(s) this covers
N/A

### Problem
There is no option to always fully qualify type names to the strongest extent: even AlwaysQualifyMemberReferences | !UsingDeclarations omits the use of global::. This can be a problem when dealing with name confusion such as in [this SO answer](https://stackoverflow.com/a/72318963). I haven't been able to quickly generate a repro case, but anecdotally, using this option has resolved a number of cases in which the ILSpy generated code will not compile. In general, when compilability of generated code is a goal, this option seems like it would be handy to have.

Of note is that StringInterpolation will probably need to be set to false when this option is enabled, as the C# compiler confuses the colon in "global::" for a formatting specifier. Perhaps it's possible to also emit surrounding braces when global:: is used inside of a string interpolation argument, which is a consideration regardless of this option.


siegfriedpammer:

PR #2767, opened 2022-08-24
Use latest VS SDK

christophwille:
Update https://www.nuget.org/packages/Microsoft.VSSDK.BuildTools to 7.3.2093 to match https://github.com/actions/runner-images/blob/af3dd14fe039b539ac16336e571f761173676b3e/images/win/Windows2022-Readme.md 

Matching VS Build number to version: https://docs.microsoft.com/en-us/visualstudio/install/visual-studio-build-numbers-and-release-dates?preserve-view=true&view=vs-2022

Now both projects need latest VSSDK.


siegfriedpammer (approved):
LGTM

PR #2774, opened 2022-09-03
Refresh ready to run for .NET 7

cshung:
This change updates the `ILCompiler.Reflection.ReadyToRun.Experimental` package to a newer version to be compatible with .NET 7.

### DIFF
--- a/ILSpy.ReadyToRun/ILSpy.ReadyToRun.csproj
+++ b/ILSpy.ReadyToRun/ILSpy.ReadyToRun.csproj
@@ -38,7 +38,7 @@
 
   <ItemGroup>
     <PackageReference Include="Iced" Version="1.13.0" />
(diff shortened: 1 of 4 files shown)


christophwille:
@cshung please test master after merge - we had to introduce #2776 so the SRM upgrade in your PR doesn't interfere with ICS.D version-locked dependencies - just to make sure that the 6.0.1 SRM works for your addin properly.

cshung:
> @cshung please test master after merge - we had to introduce #2776 so the SRM upgrade in your PR doesn't interfere with ICS.D version-locked dependencies - just to make sure that the 6.0.1 SRM works for your addin properly.

I tested ready to run on master (0fbbb6d95e2ab8af48fd619cccef0f97dddfd268) and I confirm it works.

PR #2785, opened 2022-09-15
Fix #2763: Improve enum

### LINKED ISSUE #2763: Improve enum
Original : 
```csharp
public static void Test(int value)
{
    var color = (KnownColor)value;
    switch (color)
    {
        case KnownColor.DarkBlue:
            Console.WriteLine();
            break;
        case KnownColor.DarkCyan:
            Console.WriteLine();
            break;
        case KnownColor.DarkGoldenrod:
            Console.WriteLine();
            break;
        case KnownColor.DarkGray:
            Console.WriteLine();
            break;
        case KnownColor.DarkGreen:
            Console.WriteLine();
            break;
        case KnownColor.DarkKhaki:
            Console.WriteLine();
            break;
    }
}
```

ILSpy : 
```csharp
public static void Test(int value)
{
	switch (value)
	{
	case 49:
		Console.WriteLine();
		break;
	case 50:
		Console.WriteLine();
		break;
	case 51:
		Console.WriteLine();
		break;
	case 52:
		Console.WriteLine();
		break;
	case 53:
		Console.WriteLine();
		break;
	case 54:
		Console.WriteLine();
		break;
	}
}
```

mpMelnikov:
Enum in switch was replaced with the int incorrectly in some cases.

#2763

### Problem
The issue is related to the code that inlines multiple ldarg-stloc pairs in IL code into one variable (the classes ILInlining and CopyPropagation).
The following code samples were used to reproduce the issue
```csharp
public class TestClass
{
    public enum KnownColor
    {
        DarkBlue = 749,
        DarkCyan = 750,
    }

    public static void EnumReplacedWithIntIncorrectly(int intValue)
    {
        var color = (KnownColor)intValue;
        switch (color)
        {
            case KnownColor.DarkBlue:
                Console.WriteLine("DarkBlue");
                break;
            case KnownColor.DarkCyan:
                Console.WriteLine("DarkCyan");
                break;
        }
    }

    public static void EnumReplacedWithIntIncorrectly_AndIncorrectConditionGenerated_AndUnusedVariables(int intValue)
    {
        var color = (KnownColor)intValue;
        var color2 = (KnownColor)intValue;
        var color3 = (KnownColor)intValue;
        if (color == KnownColor.DarkBlue)
        {

        }

        if (color2 == KnownColor.DarkBlue)
        {

        }
    }

    public static void WorkedCorrectlyBefore(int intValue)
    {
        var color = (KnownColor)intValue;
        switch (color)
        {
            case KnownColor.DarkBlue:
                Console.WriteLine("DarkBlue: " + color);
                break;
            case KnownColor.DarkCyan:
                Console.WriteLine("DarkCyan: " + color);
                break;
        }
    }

    public static void WorkedCorrectlyBefore2(KnownColor color)
    {
        switch (color)
        {
            case KnownColor.DarkBlue:
                Console.WriteLine("DarkBlue");
                break;
            case KnownColor.DarkCyan:
                Console.WriteLine("DarkCyan");
                break;
        }
    }
}
```

### Solution
* Added a method `IsCast` that detects when the casting of types happens instead of just assignment.
* The `IsCast` method is called in ILInlining and CopyPropagation classes.
* Added `SwitchWithEnum` to Switch.cs (doesn't fail even before the changes, so need a piece of advice here).

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Correctness/Switch.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Correctness/Switch.cs
@@ -37,6 +37,7 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Correctness
 			SwitchWithGoto(2);
 			SwitchWithGoto(3);
 			SwitchWithGoto(4);
+			SwitchWithEnum(5);
 		}
 
 		static void TestCase<T>(Func<T, string> target, params T[] args)
@@ -244,6 +245,30 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Correctness
 					Console.WriteLine("default");
 					break;
 			}
+		}
+
+		public static void SwitchWithEnum(int i)
+		{
+			Console.WriteLine("SwitchWithEnum: " + i);
+			var color = (KnownColor)i;
+			switch (color)
+			{
+				case KnownColor.DarkBlue:
+					Console.WriteLine("one");
+					break;
+				case KnownColor.DarkCyan:
+					Console.WriteLine("two");
+					break;
+				default:
+					Console.WriteLine("default");
+					break;
+			}
+		}
+		
+		public enum KnownColor
+		{
+			DarkBlue = 749,
+			DarkCyan = 750,
 		}
 	}
 }
--- a/ICSharpCode.Decompiler/IL/ILVariable.cs
+++ b/ICSharpCode.Decompiler/IL/ILVariable.cs
@@ -199,6 +199,28 @@ namespace ICSharpCode.Decompiler.IL
 		public bool HasGeneratedName { get; set; }
 
 		/// <summary>
+		/// Is the variable declaration the casting of one type to another.
+		/// </summary>
+		public bool IsCast
+		{
+			get
+			{
+				if (StoreInstructions.Count != 1)
+					return false;
+				if (LoadInstructions.Count != 1)
+					return false;
+				var stLoc = StoreInstructions[0] as StLoc;
+				if (stLoc == null)
+					return false;
+				var ldLoc = stLoc.Value as LdLoc;
+				if (ldLoc == null)
+					return false;
+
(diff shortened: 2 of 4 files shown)


ICSharpCode.Decompiler/IL/ILVariable.cs
```diff
@@ -198,6 +198,28 @@ internal void CheckInvariant()
 
 		public bool HasGeneratedName { get; set; }
 
+		/// <summary>
+		/// Is the variable declaration the casting of one type to another.
+		/// </summary>
+		public bool IsCast
```
siegfriedpammer, line 204: This does not belong here. Please create an extension method or static helper.

ICSharpCode.Decompiler.Tests/TestCases/Correctness/Switch.cs
```diff
@@ -37,6 +37,7 @@ public static void Main()
 			SwitchWithGoto(2);
 			SwitchWithGoto(3);
 			SwitchWithGoto(4);
+			SwitchWithEnum(5);
```
siegfriedpammer, line 40: You added the test case to the correctness tests. These tests are for checking semantic equivalence and are executed. 

This test case should be added to the pretty tests: https://github.com/icsharpcode/ILSpy/blob/master/ICSharpCode.Decompiler.Tests/TestCases/Pretty/Switch.cs

There the generated code is compared.

siegfriedpammer: For the runtime, enum values are numeric constants anyway, so the behavior of the code is the same.

mpMelnikov: During adding a new test to [Pretty/Switch.cs](https://github.com/icsharpcode/ILSpy/blob/master/ICSharpCode.Decompiler.Tests/TestCases/Pretty/Switch.cs), I found out that my solution worked only when there is a PDB file next to the DLL. It happens because I use `HasGeneratedName` property inside my check. I will try to find a working solution.

PR #2792, opened 2022-09-29
Fix initialization of FakeProperty with setter 

exyi:
ILSpy crashed with NullReferenceException when it couldn't resolve some dependency (Newtonsoft.Json in my case, the problem was around `JObject.Item[string]` indexer)

Seems like a typo when FakeProperty is initialized - the setter method was assigned to the Getter property
### Solution

First commit contains the actual fix.

In the second commit I included some asserts and a ToString implementation I used to debug the problem. I'm guessing that more asserts would be welcome, but if you don't like it, just merge only the first commit.

  - Sorry, but I couldn't make the tests run on Linux, is there any way?
  - I also don't know how to replicate this problem in unit tests - I'd have to compile a dll against some library which the decompiler won't see at runtime.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/ILPretty/GuessAccessors.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/ILPretty/GuessAccessors.cs
@@ -15,9 +15,8 @@ namespace ClassLibrary1
 			//IL_0007: Expected O, but got Unknown
 			UnknownClass val = new UnknownClass();
 			int? unknownProperty = val.UnknownProperty;
-			int? num = unknownProperty.GetValueOrDefault();
-			val.UnknownProperty = num;
-			int? num2 = num;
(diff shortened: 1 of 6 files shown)


siegfriedpammer:
> * I also don't know how to replicate this problem in unit tests - I'd have to compile a dll against some library which the decompiler won't see at runtime.

You could create an IL-Pretty-Test like this one: https://github.com/icsharpcode/ILSpy/blob/master/ICSharpCode.Decompiler.Tests/TestCases/ILPretty/UnknownTypes.il

ILAsm will happily assemble a file that contains references to unknown types/members.

exyi:
I added a test case which I believe covers the changed code.

I also needed to slightly change output of one the existing test. It does not change the meaning of the code, but it's now eager to merge assignments into `var v = (x.Prop=c.Prop)` expression and then it does not know how to name the variable `v`.

siegfriedpammer:

Issue #2204, opened 2020-11-04
Easy to switch int format (dec/hex/bin/oct)


ahdung:
I hope provide a easy way to switch int format, like toolbar/context menu. Thanks.

siegfriedpammer:
Are you requesting this for the decompiler output or for the UI?

ahdung:
UI

ahdung:
I might answer wrong, I mean this:
<img width="818" height="543" alt="Image" src="https://github.com/user-attachments/assets/9fd64593-1435-4730-b6da-bdf28b4ce17c" />

PR #2798, opened 2022-10-05
Add option to display row numbers in hex

davidwengier:
Link to issue(s) this covers
https://github.com/icsharpcode/ILSpy/issues/2204 maybe? Not exactly clear what that issue is asking for.

### Problem
I'm preparing a talk about metadata tables, and showing lots of screenshots of ILSpy, and it struck me as odd that I could point to a metadata token of `0600000C` but when looking at the Method table, there was no row that had a RID of `C`.

### Solution
Looks like this:
ILSpyHexRowNumbers

Couldn't see any relevant tests for this area, but let me know if I missed something.

### DIFF
--- a/ILSpy/Metadata/Helpers.cs
+++ b/ILSpy/Metadata/Helpers.cs
@@ -94,11 +94,22 @@ namespace ICSharpCode.ILSpy.Metadata
 
 		internal static void View_AutoGeneratingColumn(object sender, DataGridAutoGeneratingColumnEventArgs e)
 		{
+			var displaySettings = MainWindow.Instance.CurrentDisplaySettings;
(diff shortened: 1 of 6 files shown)


dgrunwald:
The `Token` column already shows the row number in hex (after the prefix with the table number).

davidwengier:
🤦‍♂️

You're right of course.

However, I was running into the confusion when looking at PDB metadata (eg, look up MethodDebugInformation to see the Document token, then go find the Document row) and it seems like PDB metadata has no Token columns at all. Perhaps that is the better fix?

davidwengier:
If you've added that then I guess there is no need for this after all. Thanks for the chat 😬

PR #2799, opened 2022-10-07
CustomDebugInformationTableTreeNode: split Kind, fix Offset, add EncStateMachineStateMap, 

fowl2:
Improvements for CustomDebugInformationTableTreeNode:
* add kind EncStateMachineStateMap as "Edit And Continue State Machine State Map (C# / VB)"
* split Kind column into Kind, KindGuid, and KindGuidHeapOffset columns
* Show offset for embedded
* add Parent kind to tooltip
* Format value blob heap offset to 8 digits

**Before**:
image
Parent Tooltip:
image

**After**:
image

Parent Tooltip:
image

### DIFF
--- a/ICSharpCode.Decompiler/DebugInfo/KnownGuids.cs
+++ b/ICSharpCode.Decompiler/DebugInfo/KnownGuids.cs
@@ -10,11 +10,13 @@ namespace ICSharpCode.Decompiler.DebugInfo
 		public static readonly Guid VBLanguageGuid = new Guid("3a12d0b8-c26c-11d0-b442-00a0244a1dd2");
 		public static readonly Guid FSharpLanguageGuid = new Guid("ab4f38c9-b6e6-43ba-be3b-58080b2ccce3");
 
+		// https://github.com/dotnet/roslyn/blob/main/src/Dependencies/CodeAnalysis.Debugging/PortableCustomDebugInfoKinds.cs
 		public static readonly Guid StateMachineHoistedLocalScopes = new Guid("6DA9A61E-F8C7-4874-BE62-68BC5630DF71");
 		public static readonly Guid DynamicLocalVariables = new Guid("83C563C4-B4F3-47D5-B824-BA5441477EA8");
 		public static readonly Guid DefaultNamespaces = new Guid("58b2eab6-209f-4e4e-a22c-b2d0f910c782");
 		public static readonly Guid EditAndContinueLocalSlotMap = new Guid("755F52A8-91C5-45BE-B4B8-209571E552BD");
 		public static readonly Guid EditAndContinueLambdaAndClosureMap = new Guid("A643004C-0240-496F-A783-30D64F4979DE");
+		public static readonly Guid EncStateMachineStateMap = new Guid("8B78CD68-2EDE-420B-980B-E15884B8AAA3");
 		public static readonly Guid EmbeddedSource = new Guid("0e8a571b-6926-466e-b4ad-8ab04611f5fe");
 		public static readonly Guid SourceLink = new Guid("CC110556-A091-4D38-9FEC-25AB9A351A6A");
 		public static readonly Guid MethodSteppingInformation = new Guid("54FD2AC5-E925-401A-9C2A-F94F171072F8");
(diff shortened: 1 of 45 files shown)


siegfriedpammer:
Thanks for adding the missing custom debug info kind.

Would you mind explaining why you think the other changes are needed? Thanks!

fowl2:
> split Kind column into Kind, KindGuid, and KindGuidHeapOffset columns

Because previously this was just all combined together, unlabelled in a single column. 

> Show offset for embedded

Because the column was blank before. 

> add Parent kind to tooltip

Because otherwise "@30000001" is pretty opaque. I'm working on linking this up. 

> Format value blob heap offset to 8 digits

I think this was to be consistent with other view. 

Hope this helps

siegfriedpammer:
> > split Kind column into Kind, KindGuid, and KindGuidHeapOffset columns
>
> Because previously this was just all combined together, unlabelled in a single column.

That was a deliberate design decision. I only wanted to show columns that actually exist in the physical metadata format (except for RID, Token and Offset, which are kind of necessary to make the table usable). Additional info / text representation is usually added in tooltips. The reason I added the text representation in that column (instead of just showing the raw GUID offset), is because I wanted to be able to filter the debug info kinds by name.

I think adding two new columns to that table for one piece of information puts too much emphasis on that piece of information. Especially `KindGuidHeapOffset` seems to be quite useless. How about adding a tooltip that contains the information? I would keep the "human readable" description of the kind directly in the table cell (to allow easy filtering) and move the GUID representation to the tooltip. And probably get rid of the GUID heap offset altogether. What do you think?

> > Show offset for embedded
> 
> Because the column was blank before.

Yeah, I fixed that in my latest commit. It now shows `n/a` like in all the other tables. The  offset is generally not useful, when the debug metadata is embedded, because the stream containing the data is compressed and there is no physical offset in the image, where the offset could refer to.

> > add Parent kind to tooltip
> 
> Because otherwise `@30000001` is pretty opaque. I'm working on linking this up.

I am not against changing that in general, however, I think we could use the format: `<token> (<table name>)`. What do you think?
 
> > Format value blob heap offset to 8 digits
> 
> I think this was to be consistent with other view. 

I think, I deliberately left it like that to distinguish heap offsets from tokens.

I hope you don't mind discussing this with me further, so we can come to a conclusion...

fowl2:
> I think adding two new columns to that table for one piece of information puts too much emphasis on that piece of information. Especially `KindGuidHeapOffset` seems to be quite useless. How about adding a tooltip that contains the information? I would keep the "human readable" description of the kind directly in the table cell (to allow easy filtering) and move the GUID representation to the tooltip. And probably get rid of the GUID heap offset altogether. What do you think?

As someone not that familiar with the format I couldn't identify the various parts without the column headings and I just didn't want to remove some functionality in case some one was [relying on them](https://xkcd.com/1172/) for some reason. It absolutely makes sense they don't need to be in as much focus.

> Yeah, I fixed that in my latest commit. It now shows `n/a` like in all the other tables. The offset is generally not useful, when the debug metadata is embedded, because the stream containing the data is compressed and there is no physical offset in the image, where the offset could refer to.

Ahh embedded implies compressed - that I wasn't aware of. I guess it's completely hidden from the object model. How about a tool tip explaining that because the stream is compressed the offset might not be as meaningful?

> I am not against changing that in general, however, I think we could use the format: `<token> (<table name>)`. What do you think?

Sure. Would that be with the expanded version on the second line? (eg. method names get reaaaalllly long)

> I think, I deliberately left it like that to distinguish heap offsets from tokens.

Makes sense. Could we use additionally use different column heading to make it clearer?

> I hope you don't mind discussing this with me further, so we can come to a conclusion...

Of course - thanks for the project and engaging with me so readily!

siegfriedpammer:
> Sure. Would that be with the expanded version on the second line? (eg. method names get reaaaalllly long)

I would only add that for items that previously had no specialized tooltip and were just showing `@XXXXXXXX`.

> Ahh embedded implies compressed - that I wasn't aware of. I guess it's completely hidden from the object model. How about a tool tip explaining that because the stream is compressed the offset might not be as meaningful?

So you would still want to display the offset instead of "n/a", or just add an explanatory tooltip, as to why it shows "n/a"?

> Makes sense. Could we use additionally use different column heading to make it clearer?

Yes, any suggestions? :)

siegfriedpammer:
I have made some adjustments to your code following our previous discussion. Thank you for contributing and discussing things with me. If there is anything else you would like to see changed, please feel free to open an issue.

Issue #2801, opened 2022-10-12
Proposal: Better progress reporting and cancellation for PortablePdbWriter


andrewcrawley:
I'm continuing to work on improving the integration between the decompiler and the Visual Studio debugger.  We've found that a surprisingly high number of customers who try the decompiler end up attempting to cancel the decompilation operation, and our current hypothesis is that the lack of progress feedback leads them to think that the operation is "hung", or otherwise unresponsive.  Since the decompiler doesn't really support cancellation, this leads to an undesirable situation where the user attempts to cancel the operation in VS, and we're forced to wait for it to complete (probably successfully!) before discarding the results and informing the user that the "cancellation" succeeded.

To improve things, I'm looking at implementing better support for cancellation and progress reporting in `PortablePdbWriter`, the entry point used by the VS debugger.

I've prototyped the following API, which I believe addresses both of the pain points I mentioned above - better progress reporting so users hopefully won't cancel as often, and better cancellation so they won't have to wait as long when they do cancel.

```C#
public static void WritePdb(
    PEFile file,
    CSharpDecompiler decompiler,
    DecompilerSettings settings,
    Stream targetStream,
    bool noLogo = false,
    BlobContentId? pdbId = null,
    CancellationToken cancellationToken = default(CancellationToken), // New!
    IProgress<WritePortablePdbProgress> progress = null)              // New!

public struct WritePortablePdbProgress
{
    int TotalTypes { get; }
    int TypesWritten { get; }
}
```

The implementation merely updates the loop over types in `WritePdb` to check the CancellationToken and send a progress update per iteration.

If you're satisfied with this design, I'd be happy to clean up my prototype and send a PR.  We're hoping to be able to get this into the 8.0 release.

dgrunwald:
Cancellation is already supported via the `CSharpDecompiler.CancellationToken` property.
Adding support for `IProgress` makes sense.

siegfriedpammer:
For now, I will only add this to WholeProjectDecompiler / PortablePdbWriter on a per-file basis...

andrewcrawley:
@dgrunwald - Thanks for the info on cancellation, I just submitted changes to the debugger to get that hooked up.

I'll follow up with a PR for the progress reporting, thanks!

PR #2802, opened 2022-10-14
Add progress reporting to PortablePdbWriter

andrewcrawley:
### Problem
As discussed in #2801, the Visual Studio debugger would like to be able to provide better progress feedback to users while a decompilation operation is in progress.

### Solution
This commit adds a new parameter to `PortablePdbWriter.WritePdb` that allows the caller to provide an implementation of `IProgress<>` to receive status updates on the pdb generation process.  Currently, the progress reports include the number of files generated so far and the total number of files.

#2801 mentions adding progress reporting to `WholeProjectDecompiler` as well, but that appears to already be present - see: https://github.com/icsharpcode/ILSpy/blob/master/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs#L256

### DIFF
--- a/ICSharpCode.Decompiler.PowerShell/GetDecompiledProjectCmdlet.cs
+++ b/ICSharpCode.Decompiler.PowerShell/GetDecompiledProjectCmdlet.cs
@@ -33,8 +33,8 @@ namespace ICSharpCode.Decompiler.PowerShell
 			lock (syncObject)
 			{
 				completed++;
-				progress = new ProgressRecord(1, "Decompiling " + fileName, $"Completed {completed} of {value.TotalNumberOfFiles}: {value.Status}") {
-					PercentComplete = (int)(completed * 100.0 / value.TotalNumberOfFiles)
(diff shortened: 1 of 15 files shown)


siegfriedpammer:
@andrewcrawley thank you for providing the implementation. I have generalized and extended progress reporting in our code base... I hope that you are OK with the changes.

This PR now also implements cancellation of PDB generation in ILSpy and some visual changes:

The size of the progress bar was increased:

image

Progress now shows more information when exporting a project:

image

Progress reporting for PDB generation:

image

andrewcrawley:
The changes look great, thank you!  I was hesitant to wander too far from my original proposal : )

Is there a roadmap for the final release of 8.0, or any info you could share on when that might be?

christophwille:
No, there is no roadmap and/or timeline for 8.0. Most likely we will have a preview3 late this year, followed by again most likely a couple of RCs. Guesstimate: late winter/early spring next year.

PR #2810, opened 2022-10-21
Include assembly name in decompiled code

miloush:
### Problem

When exploring decompiled methods and other members and jumping all around through analysis tree, inheritance etc., it is difficult to figure out to which assembly the currently displayed code belongs. On has to scroll the treeview to an exact narrow position, or just keep collapsing it manually.

### Solution

The proposed solution is to include the assembly name as a comment above the existing comment containing full type name.

Another approach would be to include it with the type comment in `CSharpAmbience.ConvertSymbol`, possibly using `ConversionFlags.UseAssemblyName`, but that seemed unnecessarily complex.

Namespaces do not show assembly because it is not readily available around.

This does not come with tests, happy to add some if deemed necessary and given some pointers whereabouts are similar tests e.g. for the type comment. `assembly` cannot be null because `CreateDecompiler` already dereferences it.


siegfriedpammer:

Issue #2807, opened 2022-10-20
Permit source generation for the `<Module>` from the project decompile


AaronRobinsonMSFT:
**Is your feature request related to a problem? Please describe.**

I am looking at generating C# from a C++/CLI binary compiled with `/clr:pure`. A lot of code is pinned on `<Module>` and not emitting this is annoying because now I must go to each node in the module. Would it be possible to permit this? I can submit a PR if this seems helpful.

**Describe the solution you'd like**
I've removed the `"<Module>"` check in the below path and it works as expected.

https://github.com/icsharpcode/ILSpy/blob/6e70db7dd3f02aa230be8d6c938b2f76005a14eb/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs#L172-L173

dgrunwald:
We don't want to output `<Module>` when decompiling a C# assembly; so you'd somehow need to detect C++/CLI.

AaronRobinsonMSFT:
@dgrunwald Thoughts on making it an opt-in setting?

AaronRobinsonMSFT:
Perhaps adding a new checkbox under **Other**, off by default, to emit `<Module>`?

![image](https://user-images.githubusercontent.com/30635565/197062005-7653cf79-564d-4fcd-844b-bc8d1ff97f71.png)

siegfriedpammer:
> We don't want to output `<Module>` when decompiling a C# assembly; so you'd somehow need to detect C++/CLI.

Couldn't we detect whether code from <Module> is referenced elsewhere and include it in those cases?

AaronRobinsonMSFT:
> Couldn't we detect whether code from <Module> is referenced elsewhere and include it in those cases?

That would also solve it. I've figured out how to add a new option and plumb that through I can submit a PR for that approach. Detecting a reference is beyond my level of understanding in this system so I'd need some pointers if that is the preferred option.

AaronRobinsonMSFT:
I've submitted a possible solution for this issue - https://github.com/icsharpcode/ILSpy/pull/2815.

siegfriedpammer:
I have implemented my proposed solution, i.e., automatic detection of missing types... https://github.com/icsharpcode/ILSpy/tree/fix/2807

What do you think? @AaronRobinsonMSFT

AaronRobinsonMSFT:
> What do you think? @AaronRobinsonMSFT

[...]

PR #2815, opened 2022-10-25
Provide an option for emitting `<Module>`.

AaronRobinsonMSFT:
This is set to "false" by default, so no behavioral change for those not interested in the `<Module>`.

Fixes https://github.com/icsharpcode/ILSpy/issues/2807

### Problem
I would like to be able to optionally emit the `<Module>` to disk.

### Solution
Added a new **Other** option that permits the user to emit the `<Module>` if desired.

image


siegfriedpammer:
We have discussed your proposed solution internally and we came to the conclusion that we don't really like having too many possible options/combinations as it makes everything more complicated. I have created my/our proposed solution in https://github.com/icsharpcode/ILSpy/tree/fix/2807

AaronRobinsonMSFT:
Closing in deference to https://github.com/icsharpcode/ILSpy/tree/fix/2807

PR #2833, opened 2022-11-10
C# 11: Required members

dgrunwald:
Implements support for C# 11 required members.
Also adds another test for static abstract interface members.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/Structs.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/Structs.cs
@@ -51,4 +51,28 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 		}
 	}
 #endif
+
+#if CS110
+	public struct StructWithRequiredMembers
+	{
+		public required string FirstName;
+		public required string LastName { get; set; }
+	}
+#endif
+}
+
+namespace System.Runtime.CompilerServices
+{
+	[AttributeUsage(AttributeTargets.All, AllowMultiple = true, Inherited = false)]
+	internal sealed class CompilerFeatureRequiredAttribute : Attribute
+	{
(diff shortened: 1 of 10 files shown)


ICSharpCode.Decompiler.Tests/TestCases/Pretty/Structs.cs
```diff
@@ -51,4 +51,28 @@ public StructWithDefaultCtor()
...
+		public required string FirstName;
+		public required string LastName { get; set; }
+	}
+#endif
+}
+
+namespace System.Runtime.CompilerServices
+{
+	[AttributeUsage(AttributeTargets.All, AllowMultiple = true, Inherited = false)]
+	internal sealed class CompilerFeatureRequiredAttribute : Attribute
```
siegfriedpammer, line 67: I think we should add a transform that removes these, if the target framework is net7.0.

Also note that https://sharplab.io/#v2:EYLgtghglgdgNAExAagD4AEBMBGAsAKHQGYACLEgYRIG8SCSGzSAnAUwEcBXKNhE2AC4kAgjRIBzVgIDcJAM5TZAX3qNiJNlx6s+gkgDFpBJUA== fully qualifies the `RefSafetyRules` module attribute. This could be because it is defined in two places, i.e., the preview framework used by sharplab as well as the local definition emitted by the compiler.

siegfriedpammer (approved):

ICSharpCode.Decompiler.Tests/TestCases/Pretty/Structs.cs
dgrunwald: These attributes were necessary for the test case to compile on older dotnet versions.
I think we should output the attribute if it is present in the assembly. It doesn't get embedded automatically (unlike some other attributes).

Issue #2834, opened 2022-11-12
Enhance ReflectionDisassembler to support filtering/sorting


tom-englert:
**Is your feature request related to a problem? Please describe.**
When using IL disassembling to verify the outcome of some IL modification, it frequently happens that tests break just because the IL layout has slightly changed, but only by the ordering of classes or members, while the IL is still the same.
Another use case is to validate IL changes by before/after comparison, which is impossible when not only a few IL statements, but the complete layout has changed.

**Describe the solution you'd like**
ReflectionDisassembler should support an extension point to enable sorting or filtering. By sorting classes or members, the generated IL would be identical if only the ordering of members has changed.
Filtering could be helpful to reduce the output just to the points of interest.

**Additional context**
See implementation and tests in #2835

siegfriedpammer:
Fixed by https://github.com/icsharpcode/ILSpy/pull/2835

PR #2835, opened 2022-11-12
Add filtering/sorting capabilities to ReflectionDisassembler

tom-englert:
Link to issue(s) this covers

### Problem
#2834

### Solution
* Wrap type or member handles in adapter classes so they can be easily exposed to a filter method


ICSharpCode.Decompiler/Disassembler/Filter.cs
```diff
@@ -0,0 +1,38 @@
...
+// DEALINGS IN THE SOFTWARE.
+
+#nullable enable
+
+using System.Collections.Generic;
+using System.Linq;
+
+namespace ICSharpCode.Decompiler.Disassembler
+{
+	public interface IFilter
```
siegfriedpammer, line 26: I feel that the name `IFilter` does not quite show the intent of the interface... Maybe call it `IEntityProcessor`?

ICSharpCode.Decompiler/Disassembler/Adapters.cs
```diff
@@ -0,0 +1,218 @@
...
+
+using ICSharpCode.Decompiler.Metadata;
+
+using System.Reflection.Metadata;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace ICSharpCode.Decompiler.Disassembler
+{
+	public abstract class Adapter
```
siegfriedpammer, line 31: I feel that a whole class hierarchy is a bit over-engineered for something that could just be two parameters, `MetadataReader` and `EntityHandle`.

Previously, when doing the big "SRM rewrite" of the decompiler engine, I added a lot of wrappers for all the different handle kinds, however almost all of them have disappeared because they do not provide much value over directly working with `MetadataReader` and a handle.

Would it be possible for you to describe your use-case for these classes? If it's about performance or allocations, I'd like to point out that we're calling `MetadataReader.GetString` all the time and it has not been a bottleneck so far. If you did some profiling, would it be possible for you to share the results? Thanks!

ICSharpCode.Decompiler/Disassembler/Filter.cs
tom-englert: OK, names are updated

ICSharpCode.Decompiler/Disassembler/Adapters.cs
tom-englert: I've added these adapters to make the exposed new interface as lean as possible; using an interface should not require to understand all the implementation details if not needed.
The adapters are as lean as possible, and I tried to touch the existing code as few as possible.

Every other approach ended up in a clumsy interface and a very complex implementation.

Performance: For methods it's not only `MetadataReader.GetString`, and using this in sorting will call it up to nearly N² times for N methods,  so ignoring this and waiting for the community to report performance problems sounds dangerous to me.

dgrunwald: You're adding 11 new classes to the public API, just for sorting a few lists. I think that's an absolute no-go.
If you think this is truly necessary, you will need to come up with some convincing arguments. If it's about perf, include actual benchmarks.

dgrunwald: Sorting isn't quadratic. I strongly suspect that when you run your benchmark you will find that the sorting has no influence on the overall disassembly performance.

tom-englert: e.g. QSort does up to O(N²) comparisons, so you don't want to calculate the two comparands every time.

dgrunwald: But .NET's `Sort` isn't a stupid quadratic sort.
With any decent sort, there will be O(N lg N) string comparisons. Whether we perform O(N) or O(N lg N) `GetString` calls doesn't change the overall runtime -- that remains O(N lg N).
You're only improving performance by a small constant factor. But if the sorting algorithm only takes a negligible portion of the disassembly time to begin with, your performance improvement won't be big enough to be measurable.

tom-englert: @dgrunwald The EnumerableSorter reads the keys only once, that's what makes the difference.

tom-englert:
Release build terminated with timeout. Can s.o. restart it?

tom-englert:
@siegfriedpammer any objections on the latest changes?

siegfriedpammer:
No... Looks good now... Will do some clean up and merge soon... Sorry for the delay...

tom-englert:
@siegfriedpammer using `InstructionOutputExtensions.WriteTo(...` to get a methods sort key now includes the method return type, which I explicitly excluded in my implementation.

tom-englert:
Any objections to revert that change?

siegfriedpammer:
> Any objections to revert that change?

No, your changes are fine. Thanks!

PR #2840, opened 2022-11-18
Fix dynamic type index calculation in `ApplyAttributeTypeVisitor`

ElektroKill:
Link to issue(s) this covers:
N/A

### Problem
ILSpy did not correctly decompile `volatile dynamic` fields.

Example:
```csharp
static volatile dynamic volatileField;
```
would decompile to:
```csharp
static volatile object volatileField;
```
The issue lies in the implementation of `ApplyAttributeTypeVisitor`. This class did not properly handle modifier signatures (`CModOpt` and `CModReqd`) which are used to indicate that a field is volatile.

Roslyn implementation: https://github.com/dotnet/roslyn/blob/29cdbf9a70ae83e6b7b914182fc78a7ed122a0a0/src/Compilers/CSharp/Portable/Symbols/Compilation_WellKnownMembers.cs#L850

### Solution
Modifier signatures are now correctly handled. The `dynamicTypeIndex` is incremented for every modifier type to match Roslyn's behavior when it generates the `DynamicAttribute` attributes. The "crazy special case" for readonly return types has been removed as those include a modifier signature of `InAttribute`.

The `dynamic` "pretty" test was extended to include a `dynamic volatile` field.

### DIFF
--- a/ICSharpCode.Decompiler/TypeSystem/Implementation/MetadataField.cs
+++ b/ICSharpCode.Decompiler/TypeSystem/Implementation/MetadataField.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) 2018 Daniel Grunwald
+// Copyright (c) 2018 Daniel Grunwald
 // 
 // Permission is hereby granted, free of charge, to any person obtaining a copy of this
 // software and associated documentation files (the "Software"), to deal in the Software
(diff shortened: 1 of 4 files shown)


ICSharpCode.Decompiler/TypeSystem/Implementation/MetadataField.cs
```diff
@@ -225,7 +225,6 @@ private IType DecodeTypeAndVolatileFlag()
 				if (ty is ModifiedType mod && mod.Modifier.Name == "IsVolatile" && mod.Modifier.Namespace == "System.Runtime.CompilerServices")
 				{
 					Volatile.Write(ref this.isVolatile, true);
-					ty = mod.ElementType;
```
dgrunwald, line 228: We need to get rid of the `ModifiedType`, otherwise it might confuse other parts of the type system that don't expect it to be present (e.g. in overload resolution).
But I think you can re-order the `ApplyAttributesToType` call to come before this if.

ElektroKill: Doesn't `ApplyAttributesToType` already remove modifier types? Or is that option not enabled by default?

dgrunwald: Ah right, I forgot that `ApplyAttributesToType` usually removed modopts.
Pretty much only the ILSpy UI uses `TypeSystemOptions.KeepModifiers` to keep them around, and in that case it seems harmless to also keep `IsVolatile` around.

PR #2852, opened 2022-11-26
Enable tests which no longer fail

ElektroKill:
Link to issue(s) this covers:
N/A

### Problem
The ICSharpCode.Decompiler.Tests project contained a few tests that were marked to be ignored due to them being broken when in reality they pass without any issues!

### Solution
* Re-enable said tests


siegfriedpammer:

PR #2853, opened 2022-11-26
Add support for Visual Basic Async Await state machine decompilation

ElektroKill:
Link to issue(s) this covers:
N/A

### Problem
ILSpy did not support async await state machines generated by both the legacy Visual Basic compiler as well as Roslyn Visual Basic compiler.

The code generation differed for the state machine creation code, the placement of the `doFinallyBodies` initial assignment, additional calls to VB-specific methods in the catch handler, additional stack slots, and different order of field assignments in the resume blocks.

### Solution
* Adjust the async await decompilation logic to handle the different code generation properly
* Enable, fix and extend the VBPretty test for Async Await.


siegfriedpammer:

LGTM

Issue #2620, opened 2022-02-03
Pointer increments and compund assignements not always decompiled in a "pretty" way.


ElektroKill:
### Input code
```csharp
unsafe class UIntCases {
	static void Sample1() {
		uint local = 0;
		uint* r = &local;
		uint g = (*r++) * (*r++);
		Console.WriteLine(g);
	}
	static void Sample1(uint param) {
		uint* r = &param;
		uint g = (*r++) * (*r++);
		Console.WriteLine(g);
	}
	static void Sample5() {
		uint local = 0;
		uint* r = &local;
		uint g = (*r += 5) * (*r += 5);
		Console.WriteLine(g);
	}
	static void Sample5(uint param) {
		uint* r = &param;
		uint g = (*r += 5) * (*r += 5);
		Console.WriteLine(g);
	}
	static void Sample52() {
		uint local = 0;
		uint* r = &local;
		uint g = (*(r += 5)) * (*(r += 5));
		Console.WriteLine(g);
	}
	static void Sample52(uint param) {
		uint* r = &param;
		uint g = (*(r += 5)) * (*(r += 5));
		Console.WriteLine(g);
	}
}
unsafe class ByteCases {
	static void Sample1() {
		byte local = 0;
		byte* r = &local;
		int g = (*r++) * (*r++);
		Console.WriteLine(g);
	}
	static void Sample1(byte param) {
		byte* r = &param;
		int g = (*r++) * (*r++);
		Console.WriteLine(g);
	}
	static void Sample5() {
		byte local = 0;
		byte* r = &local;
		int g = (*r += 5) * (*r += 5);
		Console.WriteLine(g);
	}
	static void Sample5(byte param) {
		byte* r = &param;
		int g = (*r += 5) * (*r += 5);
		Console.WriteLine(g);
	}
	static void Sample52() {
		byte local = 0;
		byte* r = &local;
		int g = (*(r += 5)) * (*(r += 5));
		Console.WriteLine(g);
	}
	static void Sample52(byte param) {
		byte* r = &param;
		int g = (*(r += 5)) * (*(r += 5));
		Console.WriteLine(g);
	}
}
```

### Erroneous output
```c#
// ILSpyPointerIncrement.UIntCases
using System;

internal class UIntCases
{
	private unsafe static void Sample1()
	{
		uint local = 0u;
		uint* num = &local;
		uint* r = num + 1;
		uint num2 = *num;
		uint* num3 = r;
		r = num3 + 1;
[...]

PR #2856, opened 2022-12-03
Improve support for post-increment/decrement on pointers

ElektroKill:
Link to issue(s) this covers:
https://github.com/icsharpcode/ILSpy/issues/2620 (This PR only partially resolves this issue)

### Problem
ILSpy did not properly handle post-increment and post-decrement on pointers for other pointer types than `byte*`

### Solution
* Extend pattern-matching logic to support more post-increment and post-decrement cases
* Re-enabled a previously disabled test in `CompoundAssignmentTest` pretty test which tests this newly added behavior.

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) 2014-2020 Daniel Grunwald
+// Copyright (c) 2014-2020 Daniel Grunwald
 // 
 // Permission is hereby granted, free of charge, to any person obtaining a copy of this
 // software and associated documentation files (the "Software"), to deal in the Software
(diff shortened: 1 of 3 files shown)


ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -1901,7 +1901,16 @@ TranslatedExpression HandleCompoundAssignment(NumericCompoundAssign inst, Assign
 			if (inst.EvalMode == CompoundEvalMode.EvaluatesToOldValue)
 			{
 				Debug.Assert(op == AssignmentOperatorType.Add || op == AssignmentOperatorType.Subtract);
-				Debug.Assert(inst.Value.MatchLdcI(1) || inst.Value.MatchLdcF4(1) || inst.Value.MatchLdcF8(1));
+#if DEBUG
+				if (inst.Type is PointerType ptrType)
```
dgrunwald, line 1905:
```suggestion
				if (target.Type is PointerType ptrType)
```
`target.Type` is the actual C# type of the expression being incremented, so that's where it's important that the pointer element size matches.
I think `target.Type` and `inst.Type` will be equivalent in all the cases allowed by the compound assignment transform, so the changed assertion should still succeed.

PR #2863, opened 2022-12-07
Add MethodSignatureProvider and PEFileExtensions

### LINKED ISSUE #2862: Provide method signature to distinguish overloaded methods
When trying to disassemble a single method, it's not easily possible to find overloaded methods, since the parameter types are not available in the first place, method definition only provides parameter names, which are not useful here.

Provide a method signature that lists the individual parameter types in a C# style notation, so overloaded methods can be easily distinguished by their parameter types.

**Additional context**
See PR

tom-englert:
Fixes #2862: Provide method signature to distinguish overloaded methods 

### Problem
overloaded methods are not distinguishable, as their parameter types are not available in the first place.

### Solution
  - the new code is mainly standalone
  - MethodSignatureProvider


dgrunwald:
We already have a type system (`DecompilerTypeSystem`); why are you trying to introduce yet another?

PR #2874, opened 2022-12-23
Add support for Visual Basic Yield Return state machine decompilation

ElektroKill:
Link to issue(s) this covers
N/A

### Problem
ILSpy did not support yield return state machines generated by both the legacy Visual Basic compiler as well as the Roslyn Visual Basic compiler.

The following changes were made:
* Adjust detection of methods such as `MoveNext` or `Dispose` to support the interface implementations generated by the VB.NET compiler.
* Detect VB.NET `doFinallyBodies` variable used by the Legacy Visual Basic compiler.
* Add necessary code to cleanup finally blocks by removing dead conditions.
* Extend state machine detection logic for VB.NET generated code. I'm still not entirely happy with my way of detecting whether a state machine was compiled by Legacy or Roslyn Visual Basic compilers, but it works.
* Execute additional cleanup steps before analyzing `MoveNext` to accommodate for dead code generated by the VB compiler.
* Use `-1` as the initial state when decompiling code generated by the Legacy VB.NET compiler.
* Add additional logic to disregard conditions testing on `disposeField` when converting the body.
* Extract the real body out of the try-catch handler covering all the code emitted by the Legacy Visual Basic compiler.

### Solution
* Extend the yield return decompilation logic to handle the different code generation properly.
* Add a VBPretty test for the newly added behavior.

Feedback is greatly appreciated, this is probably my most complicated PR to the decompiler yet and I'm not sure whether I implemented everything up to standard :P

EDIT: Would using an enum for the compiler used for state machines be beneficial for code quality? It would help clean up all the boolean fields and we could even use bitwise operations and masks to implement the two visual basic fields in a prettier way.

### DIFF
--- a/ICSharpCode.Decompiler/IL/ControlFlow/SymbolicExecution.cs
+++ b/ICSharpCode.Decompiler/IL/ControlFlow/SymbolicExecution.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) 2011 AlphaSierraPapa for the SharpDevelop Team
+// Copyright (c) 2011 AlphaSierraPapa for the SharpDevelop Team
 // 
 // Permission is hereby granted, free of charge, to any person obtaining a copy of this
 // software and associated documentation files (the "Software"), to deal in the Software
@@ -118,7 +118,7 @@ namespace ICSharpCode.Decompiler.IL.ControlFlow
 
 		public SymbolicValue Eval(ILInstruction inst)
 		{
-			if (inst is BinaryNumericInstruction bni && bni.Operator == BinaryNumericOperator.Sub && !bni.CheckForOverflow)
+			if (inst is BinaryNumericInstruction bni && bni.Operator == BinaryNumericOperator.Sub)
 			{
 				var left = Eval(bni.Left);
 				var right = Eval(bni.Right);
--- a/ICSharpCode.Decompiler/IL/ControlFlow/SymbolicExecution.cs
+++ b/ICSharpCode.Decompiler/IL/ControlFlow/SymbolicExecution.cs
@@ -99,10 +99,12 @@ namespace ICSharpCode.Decompiler.IL.ControlFlow
 	class SymbolicEvaluationContext
 	{
 		readonly IField stateField;
+		readonly bool legacyVisualBasic;
 		readonly List<ILVariable> stateVariables = new List<ILVariable>();
 
-		public SymbolicEvaluationContext(IField stateField)
+		public SymbolicEvaluationContext(IField stateField, bool legacyVisualBasic = false)
 		{
+			this.legacyVisualBasic = legacyVisualBasic;
 			this.stateField = stateField;
 		}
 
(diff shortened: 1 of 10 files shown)


ICSharpCode.Decompiler/IL/ControlFlow/SymbolicExecution.cs
```diff
@@ -118,7 +118,7 @@ public void AddStateVariable(ILVariable v)
 
 		public SymbolicValue Eval(ILInstruction inst)
 		{
-			if (inst is BinaryNumericInstruction bni && bni.Operator == BinaryNumericOperator.Sub && !bni.CheckForOverflow)
+			if (inst is BinaryNumericInstruction bni && bni.Operator == BinaryNumericOperator.Sub)
```
siegfriedpammer, line 121: I don't think this change is safe. Maybe we need an extra flag switching symbolic evaluation to a "VB mode"?

ICSharpCode.Decompiler/IL/Instructions/ILFunction.cs
```diff
@@ -98,6 +98,11 @@ partial class ILFunction
 		/// </summary>
 		public bool StateMachineCompiledWithMono;
 
+		/// <summary>
+		/// Gets whether the YieldReturnDecompiler determined that the Legacy VB compiler was used to compile this function.
+		/// </summary>
+		public bool StateMachineCompiledWithLegacyVisualBasic;
```
siegfriedpammer, line 104: As you suggested in your PR intro, maybe switching this to a flags enum is a good idea?

siegfriedpammer:
Otherwise LGTM. Thanks a lot for the contribution!

ICSharpCode.Decompiler/IL/Instructions/ILFunction.cs
ElektroKill: Hi, I came up with the following enum and additional extension methods. What do you think of this?
```csharp
[Flags]
public enum CompilerFlags : ushort
{
	// Lower 8 bits are used for compiler flags, allowing for 256 possible languages
	Language_None = 0,
	Language_CSharp = 0x1,
	Language_VisualBasic = 0x2,
	LanguageMask = 0xFF,

	// Upper 8 bits are used for compiler kind, allowing for 256 possible compiler kinds
	CompilerKind_None = 0,
	CompilerKind_Mono = 0x100,
	CompilerKind_Legacy = 0x200,
	CompilerKindMask = 0xFF00,

	CSharp = Language_CSharp | CompilerKind_None,
	CSharpLegacy = Language_CSharp | CompilerKind_Legacy,
	CSharpMono = Language_CSharp | CompilerKind_Mono,

	VisualBasic = Language_VisualBasic | CompilerKind_None,
	VisualBasicLegacy = Language_VisualBasic | CompilerKind_Legacy,
	VisualBasicMono = Language_VisualBasic | CompilerKind_Mono,
}

public static class CompilerFlagsExtensions
{
	public static bool IsCSharp(this CompilerFlags flags) => (flags & CompilerFlags.LanguageMask) == CompilerFlags.Language_CSharp;

	public static bool IsVisualBasic(this CompilerFlags flags) => (flags & CompilerFlags.LanguageMask) == CompilerFlags.Language_VisualBasic;

	public static bool IsLegacy(this CompilerFlags flags) => (flags & CompilerFlags.CompilerKindMask) == CompilerFlags.CompilerKind_Legacy;

	public static bool IsMono(this CompilerFlags flags) => (flags & CompilerFlags.CompilerKindMask) == CompilerFlags.CompilerKind_Mono;
}
```
The rationale behind splitting up compiler kind and language is to allow easy future expansion in the case of entries for other languages and compilers being necessary in the future.

siegfriedpammer: I think we should do the refactoring in a later PR... the above proposal looks fine... maybe a little bit overengineered... I doubt there will be need for up to 256 languages or compiler kinds :)

I think that C# and None should share the same enum value... but the details are still up for discussion!

PR #2895, opened 2023-02-08
Fix `ResXResourceWriter` support for `MemoryStream` resource element

ElektroKill:
Link to issue(s) this covers:
N/A

### Problem
In newer runtime versions, MemoryStream is no longer a serializable type like it was in .NET Framework. This leads to an exception when exporting an assembly with resources using a resource element of the `Stream` type code.

### Solution
* Implement explicit support for the `MemoryStream` class instantiated by `ResourcesFile`.


greenozon:
I've recently hit such an assembly that  uses MemoryStream
any plans to merge this PR please?

PR #2903, opened 2023-02-19
Use `.interfaceimpl type` syntax

ltrzesniewski:
### Problem

I noticed the following output when decompiling an assembly:

image

That `implements .custom` part seemed wrong, and indeed, `ildasm` uses another syntax instead: `.interfaceimpl type` ([see here](https://github.com/dotnet/runtime/blob/7908e8ef62c7d5ebdcd6459faccd755b9b10d369/src/coreclr/ildasm/dasm.cpp#L4847-L4870)).

Here is the result:

image

### Solution

  This PR moves the custom attributes on interface implementations inside the type body and adds one `.interfaceimpl type` declaration for each interface implementation which has a custom attribute.

  While implementing this, I noticed that `ilasm` is inconsistent with `ildasm` when multiple attributes are applied to an interface implementation. I believe `ilasm` is wrong here, and reported the following issue:  
  https://github.com/dotnet/runtime/issues/82373

  Because of this, the unit test only applies a single attribute on the interface implementation. I didn't want to add an additional `.expected.il` file just because of this inconsistency.

  This PR implements the same behavior as in `ildasm`.

  As for the UI, I wasn't sure if I should highlight the `type` keyword in green (which looked wrong) or in blue (which could lead to many false positives), so I left it as-is.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ _ReSharper*/
 *.ReSharper
 *.patch
 .vs/
+.idea/
 /ILSpy.AddIn*/Packages/*
 /ILSpy.AddIn*/source.extension.vsixmanifest
 /ICSharpCode.Decompiler.Tests/TestCases/Disassembler/Pretty/*.dll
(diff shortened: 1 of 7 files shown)


ltrzesniewski:
BTW this is how dotPeek displays an interface implementation with two custom attributes:

image

siegfriedpammer:
> * As for the UI, I wasn't sure if I should highlight the `type` keyword in green (which looked wrong) or in blue (which could lead to many false positives), so I left it as-is.

As `ilasm` treats `type` as a keyword and fails to assemble a method that is named `type`:  `error : syntax error at token 'type'`. I would say that you can safely highlight it, there would be no confusion in existing code.

ltrzesniewski:
Right, I should have tested it. I added the `type` keyword.

But upon checking, I noticed that single-quoted identifiers are still highlighted as kewords. This also applies to other keywords such as `instance`:

image

So I also fixed that in this PR:

image

PR #2936, opened 2023-03-26
Fix 2933

christophwille:
@sharwell would that be the intended fix?

### DIFF
--- a/ICSharpCode.Decompiler/Documentation/XmlDocLoader.cs
+++ b/ICSharpCode.Decompiler/Documentation/XmlDocLoader.cs
@@ -18,6 +18,7 @@
 
 using System;
 using System.Diagnostics;
+using System.Globalization;
 using System.IO;
 using System.Runtime.CompilerServices;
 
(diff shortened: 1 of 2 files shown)


ICSharpCode.Decompiler/Documentation/XmlDocLoader.cs
```diff
@@ -135,7 +147,7 @@ static string LookupLocalizedXmlDoc(string fileName)
 			return null;
 		}
 
-		static string GetLocalizedName(string fileName, string language)
+		private static string GetLocalizedName(string fileName, string language)
 		{
 			string localizedXmlDocFile = Path.GetDirectoryName(fileName);
 			localizedXmlDocFile = Path.Combine(localizedXmlDocFile, language);
```
siegfriedpammer, line 153: Can we merge these two calls to Path.Combine?

christophwille: I intentionally kept everything "as is" except for the lookup logic

siegfriedpammer: Please?

christophwille: Done

PR #2938, opened 2023-03-26
Improve selected text highlighting

Konctantin:
Link to issue(s) this covers
There is no issue here

### Problem
Just improve selected text highlighting, like that

Before:
image

After:
image


christophwille:
Thanks!

PR #2939, opened 2023-03-27
How could "Explain this code" look like in ILSpy

christophwille:
Based on the experiment from https://github.com/christophwille/ExplainThisCode/blob/main/OpenAi.TestConsole/Program.cs

image

To be determined: is this useful at all?


christophwille:
https://chat.openai.com/chat offers a different, more concise, output:

image

christophwille:
An interactive session in ChatGPT (or Microsoft Security Copilot) with the ability to adapt the prompt and ask deeper questions will be more useful than this feature.
image

PR #2944, opened 2023-04-03
Support disassembling ReadyToRun binaries compiled using composite mode

cshung:
This late PR added the support for disassembling ReadyToRun binaries compiled using composite mode introduced officially in .NET 6.

By virtue of using the latest version, this version also supports disassembling binaries compiled using the experimental hot-cold-splitting mode introduced [here](https://github.com/dotnet/runtime/pull/74963).

Thanks @eduardo-vp for introducing the `OwnerCompositeExecutable` to make this work possible.

FYI @amanasifkhalid, @EugenioPena, this will fix https://github.com/dotnet/runtimelab/issues/1910.


christophwille:
Thanks, this is the last item that will make it into preview 4!

PR #2947, opened 2023-04-04
feat: Optimized the dark mode, now the title bar will also be dark

SlimeNull:
Optimized the dark mode, now the title bar will also be dark
优化了暗色模式, 现在标题栏也会有暗色

> Based on `DwmSetWindowAttribute`

### DIFF
--- a/ILSpy/NativeMethods.cs
+++ b/ILSpy/NativeMethods.cs
@@ -193,6 +193,16 @@ namespace ICSharpCode.ILSpy
 				return null;
 			}
 		}
+
+		[DllImport("dwmapi.dll", PreserveSig = true)]
+		public static extern int DwmSetWindowAttribute(IntPtr hwnd, DwmWindowAttribute attr, ref int attrValue, int attrSize);
+
+		public static bool EnableDarkModeForWindow(IntPtr hWnd, bool enable)
+		{
+			int darkMode = enable ? 1 : 0;
+			int hr = DwmSetWindowAttribute(hWnd, DwmWindowAttribute.UseImmersiveDarkMode, ref darkMode, sizeof(int));
+			return hr >= 0;
+		}
 	}
 
 	[return: MarshalAs(UnmanagedType.Bool)]
@@ -211,5 +221,34 @@ namespace ICSharpCode.ILSpy
 			this.Size = size;
 			this.Buffer = buffer;
 		}
+	}
+
+	public enum DwmWindowAttribute : uint
+	{
+		NCRenderingEnabled = 1,
+		NCRenderingPolicy,
+		TransitionsForceDisabled,
+		AllowNCPaint,
+		CaptionButtonBounds,
+		NonClientRtlLayout,
+		ForceIconicRepresentation,
+		Flip3DPolicy,
+		ExtendedFrameBounds,
+		HasIconicBitmap,
+		DisallowPeek,
+		ExcludedFromPeek,
+		Cloak,
+		Cloaked,
+		FreezeRepresentation,
+		PassiveUpdateMode,
+		UseHostBackdropBrush,
+		UseImmersiveDarkMode = 20,
+		WindowCornerPreference = 33,
+		BorderColor,
+		CaptionColor,
+		TextColor,
+		VisibleFrameBorderThickness,
+		SystemBackdropType,
+		Last
 	}
 }
--- a/ILSpy/Themes/ThemeManager.cs
+++ b/ILSpy/Themes/ThemeManager.cs
@@ -24,6 +24,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Windows;
 using System.Windows.Controls;
+using System.Windows.Interop;
 
 using ICSharpCode.AvalonEdit.Highlighting;
 
@@ -58,6 +59,8 @@ namespace ICSharpCode.ILSpy.Themes
 			set => UpdateTheme(value);
 		}
 
+		public bool ThemeIsDark => IsDarkTheme(Theme);
+
 		public Button CreateButton()
 		{
 			return new Button {
@@ -91,6 +94,9 @@ namespace ICSharpCode.ILSpy.Themes
 			}
 		}
 
+		private bool IsDarkTheme(string? themeName) =>
+			themeName?.Contains("Dark", StringComparison.CurrentCultureIgnoreCase) ?? false;
+
 		private void UpdateTheme(string? themeName)
 		{
 			_theme = themeName ?? DefaultTheme;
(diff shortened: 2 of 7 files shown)


christophwille:
Please see #2412 - styling the title bar is something you have to turn on (off by default)

image

SlimeNull:
> Please see #2412 - styling the title bar is something you have to turn on (off by default)
> 
> image

But I found that if I use that option, I can't use the layout options that appear when hovering over the maximize window with the mouse.

christophwille:
I don't follow what you mean - could you please provide screenshots highlighting what you meant?

ltrzesniewski:
Looks like this thing doesn't work when titlebar styling is enabled:

image

siegfriedpammer:
This thing is from Windows 11?

ltrzesniewski:
Yes

christophwille:
Interesting. Been on Win11 for nearly half a year and never seen this (I'm the Alt+Spacebar+X type)

christophwille:
@tom-englert Did that come up for your library yet?

tom-englert:
@christophwille my library only deals with custom drawing of the title bar, which is disabled by default, because of these side effects in newer windows versions.
This PR deals with changing the color of the windows default title bar. Both could co-exist, but mutual exclusive.

ILSpy/Views/CreateListDialog.xaml.cs
```diff
@@ -31,5 +34,11 @@ private void OKButton_Click(object sender, RoutedEventArgs e)
 			get => ListNameBox.Text;
 			set => ListNameBox.Text = value;
 		}
+
+		protected override void OnSourceInitialized(EventArgs e)
+		{
+			base.OnSourceInitialized(e);
+			ThemeManager.Current.ApplyThemeForWindow(this);
```
tom-englert, line 41: Code duplication: We don't need to repeat this for every window, (SPOM), it can be done once in the `WindowStyleManagerBehavior`; also other window related code better goes there.

tom-englert (changes requested):

ILSpy/Views/CreateListDialog.xaml.cs
SlimeNull: If the main window has a dark color, will the child window also have it? I have not tested this.

christophwille:
> However maybe the custom drawing of the title bar should rather be discontinued in favor of this?

Yes, I am starting to lean on this solution based on your analysis.

ILSpy/Views/CreateListDialog.xaml.cs
tom-englert: `WindowStyleManagerBehavior` is attached to every window by style

ILSpy/NativeMethods.cs
```diff
@@ -193,6 +193,16 @@ public unsafe static string GetProcessNameFromWindow(IntPtr hWnd)
 				return null;
 			}
 		}
+
+		[DllImport("dwmapi.dll", PreserveSig = true)]
```
christophwille, line 197: Just a comment - too bad we can't yet use https://learn.microsoft.com/en-us/dotnet/standard/native-interop/pinvoke-source-generation

ILSpy/Themes/ThemeManager.cs
```diff
@@ -58,6 +59,8 @@ private ThemeManager()
 			set => UpdateTheme(value);
 		}
 
+		public bool ThemeIsDark => IsDarkTheme(Theme);
```
christophwille, line 62: What is this property used for?

ILSpy/Views/CreateListDialog.xaml.cs
christophwille: @tom-englert what should we remove from #2412 if anything?

tom-englert:
@christophwille 
> Yes, I am starting to lean towards this solution based on your analysis.

- Should it then be always active, or configurable?
- is just light/dark OK, or individually configurable per theme?

christophwille:
@tom-englert 

> > Yes, I am starting to lean towards this solution based on your analysis.
> 
>     * Should it then be always active, or configurable?
>     * is just light/dark OK, or individually configurable per theme?

Very good questions indeed. Currently, we only have three light/dark "combos", and nothing that is like say a blue theme or the like. Given that the API is all about immersive dark mode... maybe "Add Dark to the theme name to get immersive dark mode for the title bar too" is sufficient for now (could later be expanded to actually look at a theme property that says "I want dark mode for the title bar"

ltrzesniewski:
> (could later be expanded to actually look at a theme property that says "I want dark mode for the title bar")

That's easily doable on the same principle as this:

https://github.com/icsharpcode/ILSpy/blob/8b0c7fbe85a53779dd437a1e40f0cf2fc1ff6d7d/ILSpy/Themes/Base.Dark.xaml#L67

The code could try to get a ressource with a known key such as `ResourceKeys.IsDarkTheme` from `_themeDictionaryContainer`. That ressource would be defined in `Base.Dark.xaml`.

(or just use the presence of `ResourceKeys.ThemeAwareButtonEffect` as a flag, but that's not pretty ^^)

tom-englert:
@ltrzesniewski we can easily find out by comparing foreground & background color, we don't need to define some extra key: c0d4075
We could also have a resource dedicated for the caption color, instead of just dark or light, by using this instead:
```
DWMWA_CAPTION_COLOR
Use with [DwmSetWindowAttribute](https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/nf-dwmapi-dwmsetwindowattribute). Specifies the color of the caption. The pvAttribute parameter points to a value of type [COLORREF](https://learn.microsoft.com/en-us/windows/win32/gdi/colorref).```

**Comment by tom-englert** [comment]
BTW this would all work only on Win11

**Comment by christophwille** [comment]
"Work only on Win11" - what would happen on older operating systems? (eg I am the only one from the core team that is on 11)

**Comment by tom-englert** [comment]
It just has no effect

**Comment by tom-englert** [comment]
`DWMWA_USE_IMMERSIVE_DARK_MODE`
Use with DwmSetWindowAttribute. Allows the window frame for this window to be drawn in dark mode colors **when the dark mode system setting is enabled.**

Does this really do what is intended?
Did s.o. test this on Win11?

**Comment by christophwille** [comment]
I did not yet test this PR

File: ILSpy/Themes/ThemeManager.cs (hunk 3 above)
Comment (SlimeNull) [reply ILSpy/Themes/ThemeManager.cs:62]: I just created this property, so if anyone needs it, they can use it directly.

Comment (SlimeNull) [reply ILSpy/Themes/ThemeManager.cs:62]: Of course, if you think it's not necessary, you can also delete it.

**Comment by SlimeNull** [comment]
DwmSetWindowAttribute is added to Windows since Windows Vista, and "DWMWA_USE_IMMERSIVE_DARK_MODE" is supported starting with Windows 11 Build 22000.

> [learn.microsoft.com/en-us/windows/win32/api/dwmapi/nf-dwmapi-dwmsetwindowattribute](https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/nf-dwmapi-dwmsetwindowattribute)

image

When we use this API, the title bar of the window will have a dark Mica background.

**Comment by SlimeNull** [comment]
Many people are not aware that ILSpy has an option to style the title bar, including myself. But now, we can make the Windows-based dark title bar coexist with the WPF-styled dark title bar. This way, if the user is a Windows 11 user, they can use the dark title bar provided by Windows 11 even if they don't enable the option. If they prefer a solid dark title bar background that matches the content, they can turn on that option.

**Comment by tom-englert** [comment]
> When we use this API, the title bar of the window will have a dark Mica background

Even when Windows itself is in light mode?

**Comment by SlimeNull** [comment]
> > When we use this API, the title bar of the window will have a dark Mica background
> 
> Even when Windows itself is in light mode?

No, there is an internal check for the current theme being used to determine if it is a dark theme.

File: ILSpy/Themes/ThemeManager.cs [hunk 4]
```diff
@@ -119,6 +125,36 @@ void ProcessDictionary(ResourceDictionary resourceDictionary)
...
+				ApplyThemeForWindow(mainWindow);
+		}
+
+		public void ApplyThemeForWindow(Window window)
+		{
+			if (window is null)
+				throw new ArgumentNullException(nameof(window));
+
+			IntPtr hwnd = new WindowInteropHelper(window).Handle;
+			bool enable = IsDarkTheme(Theme);
```
SlimeNull, line 139: Check whether the current theme is dark theme

tom-englert:
I have created an issue #2950 for this, since discussing all this stuff in the PR is not the right place.
Here we should only review the code, but not discuss the general feature, so please add comment about the feature in #2950

christophwille:
Closing in favor of #2948

PR #2953, opened 2023-04-06
Consider constructor type when lifting decimal constants

### LINKED ISSUE #2949: decompilation of large uint constants cast to decimal is incorrect
### Steps to reproduce
1. Compile `(decimal)uint.MaxValue` under .NET Core 6 or later (or otherwise produce the following IL)
2. This should result in IL containing `ldc.i4.m1; newobj System.Decimal::.ctor(uint32)`
3. This incorrectly decompiles to `-1m` instead of something like `4294967295m`

Note, this affects any uint constant > int.MaxValue. E.g.
1. Compile `(decimal)(uint.MaxValue-1000)`
2. This results in `ldc.i4 -1001; newobj System.Decimal::.ctor(uint32)`
3. This incorrectly decompiles to `-1001m`

I haven't tested the behavior of large ulong constants cast to decimal, but it'd be worth looking at.

### Screenshot
image

### Details
* Product in use: ILSpy
* Version in use: 7.2.1.6856

ElektroKill:
Link to issue(s) this covers:
Fixes #2949 

### Problem
ILSpy did not use the proper `decimal` constructor when transforming the `LdcI4` instruction into a `LdcDecimal` instruction. Furthermore, the constructors with parameter type `long` and `ulong` were not supported at all as the pattern did not account for `LdcI8`.

Code to reproduce the missing constructor issue:
```csharp
public static decimal LongToDecimal() {
	return (decimal)long.MaxValue;
}

public static decimal ULongToDecimal() {
	return (decimal)ulong.MaxValue;
}
```

### Solution
* Ensure the correct decimal constructor is used when creating `LdcDecimal` instructions.
* Add support for the two aforementioned previously unsupported constructors.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Correctness/DecimalFields.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Correctness/DecimalFields.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) 2014 AlphaSierraPapa for the SharpDevelop Team
+// Copyright (c) 2014 AlphaSierraPapa for the SharpDevelop Team
 // 
 // Permission is hereby granted, free of charge, to any person obtaining a copy of this
 // software and associated documentation files (the "Software"), to deal in the Software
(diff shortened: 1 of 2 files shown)


siegfriedpammer:
Would be nice if you could add the example code as test cases somewhere... Thanks!

ElektroKill:
> Would be nice if you could add the example code as test cases somewhere... Thanks!

I went with extending the correctness test for `decimal` since ILSpy will always decompile decimal constants using the raw value and will not introduce special constants like `uint.MaxValue` so a pretty test would not have been ideal.

siegfriedpammer:

PR #2966, opened 2023-04-19
Debugger

modz2014:
Adding Debugger


christophwille:
Is that the old ILSpy debugger?

dgrunwald:
We used to have a debugger in ILSpy, but that was removed because it ended up being unmaintained.

We now consider a debugger as out-of-scope for ILSpy.

modz2014:
why was it removed it a good thing to keep in i can maintain it if it was uploaded to it could you consider of upload the code for it please

Issue #2967, opened 2023-04-20
Searching resource tables


miloush:
Trying to look for particular resources in a string/object table of .resources files. Ctrl+F in the decompiled .resources file does not work, as it is an embedded user control.

A workaround is to select all rows in the listview, copy them, paste into notepad and search there.

I would like to have the ability to search either by resource key or value in a decompiled .resources file. I am not expecting a search provider to look through these.

christophwille:
PR merged, closing (wasn't linked to the PR so it slipped through).

PR #2969, opened 2023-04-20
Search box for resource tables

miloush:
Implements #2967 

### Problem
Unable to search for resource keys or values inside decompiled .resource files.

### Solution
Allow filtering of the string and object tables:

String tables: filters key and value
Object tables: filters key and value string, not type

Filtering is by simple case-insensitive substring. There are no tests covering the user control as far as I can tell, let me know if I missed them. Object table for testing purposes can be found e.g. in System.Windows.Forms.

There is no watermark in the search box, it feels obvious, but happy to add one if needed.


miloush:
Not sure what to do with the whitespace issues, it seems to be complaininig about lines I have not touched.

Issue #2924, opened 2023-03-14
ILSpy produces code with a redundant return statement not present in the CIL code


ElektroKill:
### Input code
```c#
private static Hashtable DuplicateReturnTest(string[] arg)
{
	var results = new Hashtable();

	using (MemoryStream memoryStream = new MemoryStream())
	{
		if (memoryStream != null)
		{
			foreach (string name in arg)
			{

			}
		}
	}

	return results;
}
```

IL produced by compiler:
```il
.method private hidebysig static 
		class [mscorlib]System.Collections.Hashtable DuplicateReturnTest (
			string[] arg
		) cil managed 
	{
		// Method begins at RVA 0x2050
		// Header size: 12
		// Code size: 49 (0x31)
		.maxstack 2
		.locals init (
			[0] class [mscorlib]System.Collections.Hashtable results,
			[1] class [mscorlib]System.IO.MemoryStream memoryStream,
			[2] string[],
			[3] int32
		)

		IL_0000: newobj instance void [mscorlib]System.Collections.Hashtable::.ctor()
		IL_0005: stloc.0
		IL_0006: newobj instance void [mscorlib]System.IO.MemoryStream::.ctor()
		IL_000b: stloc.1
		.try
		{
			IL_000c: ldloc.1
			IL_000d: brfalse.s IL_0023

			IL_000f: ldarg.0
			IL_0010: stloc.2
			IL_0011: ldc.i4.0
			IL_0012: stloc.3
			IL_0013: br.s IL_001d
			// loop start (head: IL_001d)
				IL_0015: ldloc.2
				IL_0016: ldloc.3
				IL_0017: ldelem.ref
				IL_0018: pop
				IL_0019: ldloc.3
				IL_001a: ldc.i4.1
				IL_001b: add
				IL_001c: stloc.3

				IL_001d: ldloc.3
				IL_001e: ldloc.2
				IL_001f: ldlen
				IL_0020: conv.i4
				IL_0021: blt.s IL_0015
			// end loop

			IL_0023: leave.s IL_002f
		} // end .try
		finally
		{
			IL_0025: ldloc.1
			IL_0026: brfalse.s IL_002e

			IL_0028: ldloc.1
			IL_0029: callvirt instance void [mscorlib]System.IDisposable::Dispose()

			IL_002e: endfinally
		} // end handler

		IL_002f: ldloc.0
		IL_0030: ret
	} // end of method DuplicatedReturn::DuplicateReturnTest
```

[...]

PR #2972, opened 2023-04-22
Make return duplication in `ControlFlowSimplification` less aggressive

ElektroKill:
Link to issue(s) this covers:
Fixes https://github.com/icsharpcode/ILSpy/issues/2924

### Problem
ILSpy was a bit too aggressive when it came to duplicating return blocks which resulted in unoptimal and confusing code with tons of unnecessary return statements which differed from the original source code used to produce the binary. For examples see https://github.com/icsharpcode/ILSpy/issues/2924#issue-1624058543 and https://github.com/icsharpcode/ILSpy/issues/2924#issuecomment-1502212321.

### Solution
* Make the return block duplication step in `ControlFlowSimplification` less aggressive by effectively making it only perform the duplication if the incoming edge count was 1. This effectively turned it into a simple movement of a block from one container to another.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/ReduceNesting.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/ReduceNesting.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 
 namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 {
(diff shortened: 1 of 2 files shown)


dgrunwald:
Can you add a pretty test that demonstrates this improvement?

siegfriedpammer:
Thank you for the contribution. Sorry for taking so long to get to this!

Issue #2973, opened 2023-04-22
Can't create pdb from assembly, doesn't contain PE Debug Directory Entry


Meister1593:
Not sure if i understand what this means, or why it wouldn't create it, maybe i'm not understanding something obvious?
In general, my overall idea is to attach debugger (rider in my case) and have a debug symbols for the attached application assemblies to see code. The first part of attaching game i already completed, but without symbols it's not very useful, and rider doesn't generate them by itself. I tried using dotpeek, but it doesn't really want to run under wine, so while looking at solutions, ilspy seemed most promising for that purpose.

### Steps to reproduce
It seems quite random, but it happens for some Dll's, but not for others. In my case, i tried to generate pdb for game files (NeosVR)
and some assemblies works and they generate pdb files (CodeX.dll from Managed folder), but some don't (FrooxEngine.dll for instance)
Since this is a someone's product, i probably can't really redistribute dlls, but they are stored with it and it's free. There is also a link for a standalone version with same links outside steam, but not sure if it would be needed.

### Error message shown
![image](https://user-images.githubusercontent.com/7141787/233800214-ac29e89a-1272-4165-922b-3e0d757cdd5d.png)

### Details
* Product in use: e.g. ILSpy under wine, AvaloniaILSpy native on linux (same errors)
* Version in use: ILSpy version 7.2.1.6856 .NET version 4.7.3062.0 under wine and ILSpy version 7.2.0.0.rc (AvaloniaILSpy on linux native)

trivalik:
The pdb create is blocked, the question is why: see https://github.com/icsharpcode/ILSpy/blob/fa900be124cf2997e9d60d3bf0a5735cad572625/ILSpy/Commands/GeneratePdbContextMenuEntry.cs#L62

trivalik:
As we found out in the PR that for assemblies built without an pdb it is not possible to load a pdb in Visual Studio. This makes this bug report no sense and should be closed.

PR #2974, opened 2023-04-24
remove CodeView check on pdb create

trivalik:
try to fix https://github.com/icsharpcode/ILSpy/issues/2973

### Problem
No CodeView blocks pdb generation.

### Solution
Generate a guid with zeroed timestamp in case no CodeView exists.

### DIFF
--- a/ICSharpCode.Decompiler/DebugInfo/PortablePdbWriter.cs
+++ b/ICSharpCode.Decompiler/DebugInfo/PortablePdbWriter.cs
@@ -26,10 +26,8 @@ using System.Linq;
 using System.Reflection.Metadata;
 using System.Reflection.Metadata.Ecma335;
 using System.Reflection.PortableExecutable;
-using System.Runtime;
 using System.Security.Cryptography;
 using System.Text;
-using System.Threading;
 
 using ICSharpCode.Decompiler.CSharp;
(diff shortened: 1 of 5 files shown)


christophwille:
Can you provide evidence (eg some docs) why that is a proper fix?

trivalik:
No, but generation works.

siegfriedpammer:
"generation works", yes, but debuggers might reject the PDB because it does not match the timestamp/hash stored in the DLL. Some debuggers will refuse to load any PDB at all, if the DebugView entry is missing.

Someone from the Visual Studio Debugger team even contributed a way for the Visual Studio Debugger to pass an ID to the PDB Generator that is used/accepted by the VS Debugger: https://github.com/icsharpcode/ILSpy/pull/2678... there are a couple of pointers in the PR description.

trivalik:
I see, after further testing. If you build a project without a pdb. Then it is impossible to use the created pdb in Visual Studio. Maybe a improve of the error text would make sense then, that builded programs without debug information the use of a generated is not possible. Only ILSpy itself could load the pdb right now, but makes not really sense, as long as obfuscated names could not be overridden via pdb.

christophwille:
I am closing this because aside from generating a pdb, it doesn't create a working pdb.

Issue #2997, opened 2023-06-07
Adding Auto-Update Functionality to ILSpy


christophwille:
# Status Quo

ILSpy has been shipping with update check & notification functionality for a long time. It is located in the About page:

![image](https://github.com/icsharpcode/ILSpy/assets/344208/5cb8c8c1-44a5-456c-ada0-08e684b01842)

The result of a check that shows a new version (determined via https://ilspy.net/updates.xml)

![image](https://github.com/icsharpcode/ILSpy/assets/344208/611509ce-8d55-4ac9-bafb-4f33586e9fc9)

Hitting the Download button will open a Web browser showing the respective Github Releases page - and the user choosing the install option they prefer.

Note that the MSIX was built to exclude this update check.

# Motivation

A user-install auto-update mechanism because we had that with the Microsoft Store MSIX project. Note that (semi-)auto-updates via package managers is available today, eg https://community.chocolatey.org/packages/ilspy (`choco upgrade all -y` is my daily driver) - but that usually means admin-level permissions are necessary.

# The Matrix of Installation Options

We ship many different options to "install" ILSpy
* zip - our oldest and most-used distribution mechanism. Unpack, run.
* vsix - available via GH Releases and on RTM via the VS Marketplace.
* msi - available via GH Release and on RTM in the Microsoft Store - both are user installers (non-admin), but manual update

When using the current update check it is up to the user to pick the correct deliverable - eg it is perfectly fine to use pre-release vsix when manually downloading and updating it outside the VS Marketplace. Or having multiple versions sitting side by side with the zip download.

This leads to the problem which installation option should offer an auto-update
* vsix makes no sense at all because its version of ILSpy is squirreled away by VS in some custom location. And auto-updating there might get us into serious trouble (do we update ILSpy or the extension?).
* zip doing an in-place upgrade in the extracted folder also most likely annoys people that have master or specific branch versions installed. (ILSpy settings are globally shared today re:update check)
* msi would be the most likely candidate for an auto-update where such functionality is the expected and a welcomed experience by end-users.

# Possible Way of Doing It

* Keep the existing check functionality for all distributions except msi
* Allow the msi variant to check for updates & install silently over the existing msi installation

[...]

PR #2998, opened 2023-06-07
PoC Auto-Update with Netsparkle Updater

christophwille:
Motivation - see #2997. This is a PoC only! No intention to merge.

Reason for choosing Netsparkle over others: signature validation of update files (SSL these days doesn't mean a thing, and [evilginx2](https://github.com/kgretzky/evilginx2) and friends are way too easy to use to fake update files)

## Notes for using Netsparkle

Install dotnet tool for Netsparkle

`dotnet tool install --global NetSparkleUpdater.Tools.AppCastGenerator`

Create keys

`netsparkle-generate-appcast --generate-keys --key-path ./signingkeys`

Create appcast 
`
netsparkle-generate-appcast -n "ILSpy" -a ./output -e msi -b ./ -o windows -f true -u https://github.com/icsharpcode/ILSpy/releases/tag/v8.0 --description-tag ".NET Decompiler" --key-path ./signingkeys --change-log-path ./`

This assumes `ILSpy_Installer_8.0.0.7400.msi` and `8.0.0.7400.md` (both files entirely fake) in the root folder where the command is run.

## Notes for deploying the appcast

When you add release notes to the appcast, it will contain newlines, eg https://github.com/icsharpcode/ILSpy/blob/11a5bb50682fdf31c08a01e0937673ab6faa29a7/appcast.xml. When pushing from Windows and consuming as is the case here on Linux, you need `appcast.xml eol=crlf` in https://github.com/icsharpcode/ILSpy/blob/11a5bb50682fdf31c08a01e0937673ab6faa29a7/.gitattributes

## Implementation notes

Comments included address a few of the edge cases found (and some not yet resolved). In general, if going down the route of #2997 we should have one update strategy for checking only, and one for auto-updating (think interface and two implementations that can be used depending on deployment variant)

### DIFF
--- a/ILSpy/AboutPage.cs
+++ b/ILSpy/AboutPage.cs
@@ -218,74 +218,6 @@ namespace ICSharpCode.ILSpy
 			return latestAvailableVersion;
 		}
 
-		sealed class AvailableVersionInfo
-		{
-			public Version Version;
-			public string DownloadUrl;
-		}
-
-		sealed class UpdateSettings : INotifyPropertyChanged
-		{
-			public UpdateSettings(ILSpySettings spySettings)
-			{
-				XElement s = spySettings["UpdateSettings"];
-				this.automaticUpdateCheckEnabled = (bool?)s.Element("AutomaticUpdateCheckEnabled") ?? true;
-				try
-				{
-					this.lastSuccessfulUpdateCheck = (DateTime?)s.Element("LastSuccessfulUpdateCheck");
-				}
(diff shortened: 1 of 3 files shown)


nnfhhvbb (approved):

christophwille:
Given that the number of updates per year is low, and that we'd need multiple channels (eg release, preview) plus the problem of disabling it inside eg VSIX makes this more complicated than it is worth. Auto-update won't be considered for the near term future.

Issue #2999, opened 2023-06-07
VSIX not arm64-aware


christophwille:
VSIX is picking up binaries from default output folder https://github.com/icsharpcode/ILSpy/blob/f12a3d35727ac6ff0a38128d5543a8c9fe232bed/ILSpy.AddIn.VS2022/ILSpy.AddIn.VS2022.csproj#L92 - this is not an issue inside VS, but when ilspy.exe is being launched (then it will be by default x64, see thread starting at https://github.com/icsharpcode/ILSpy/issues/2984#issuecomment-1565260509).

Ideally, that would be using the publish folders now being used by the Installer. Maybe transfer the VSIX projects to the Installer sln if no other build time references are really necessary for VSIX. @Rpinski what constraints do you remember? Or would a pre-built publish folder be a-ok?

Note: haven't looked into publishing x64/arm64 specific VS extensions...

Rpinski:
Sorry for the late reply. We can try moving it to the Installer solution, of course. I don't remember any dependencies on the projects' code. We link some .cs files from Decompiler directly, but that shouldn't be an issue.

Targeting x64+arm64 in same extension is possible, basic work is described here:
https://learn.microsoft.com/en-us/visualstudio/extensibility/arm64/target-arm64-visual-studio-extension?view=vs-2022

Should we then ship both ILSpy variants inside of VSIX package and somehow decide which one to start?

christophwille:
The extension itself is net472 without a CPU preference at the moment; maybe the simplest solution really would be to package both versions (x64 and arm64 fw-dependent) and then decide at runtime which one to call (I don't really want to figure what packaging & deploying two targets means).

PR #3009, opened 2023-06-14
Split out VS Extensions into separate sln & add arm64 binaries to 2022 VSIX

christophwille:
See #2999 

* VS 2022 Extension has both \x64 and \arm64 binaries of published ILSpy included
* Architecture decides which one is launched


christophwille:
Screenshot on Arm64 VS 17.6.3
image

PR #3024, opened 2023-06-24
Copy key/value/names ContextMenu for resource tables

### LINKED ISSUE #3023: Resource tables do not have their context menus
### Steps to reproduce
1. Navigate to e.g. System.Windows.Forms > Resources
2. Right click on a row in the String Table

Actual: Copy is disabled, Select All selects the decompiled text (i.e. the decompilation context menu is shown)

### Details
* Product in use: ILSpy
* Version in use: 8.0.0.7417

miloush:
Fixes #3023 

### Problem
The resource tables show document context menu that does not allow copying items.

### Solution
This PR adds a context menu with a few copy options.

It also fixes the issue that the normal copy uses simple ToString() which
* for object table is not overloaded and just copies the internal ILSpy type name 
* for string table it used [Key, Value] syntax

Now it separates the columns by tabs so that they can be pasted into Excel etc.

Before:
image

After:
image

### DIFF
--- a/ILSpy/Controls/ResourceObjectTable.xaml
+++ b/ILSpy/Controls/ResourceObjectTable.xaml
@@ -26,9 +26,9 @@
 					<Setter.Value>
 						<ContextMenu>
 							<MenuItem Header="_Copy" Command="ApplicationCommands.Copy" />
-							<MenuItem Header="Copy _name" Command="ApplicationCommands.Copy" CommandParameter="Key" />
-							<MenuItem Header="Copy _value" Command="ApplicationCommands.Copy" CommandParameter="Value" />
-							<MenuItem Header="Copy _type" Command="ApplicationCommands.Copy" CommandParameter="Type" />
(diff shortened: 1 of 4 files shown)


siegfriedpammer:
Small nit: having the same shortcut for three different menu commands seems a bit strange to me.

christophwille:
Yes, only one Copy item should have the Ctrl+C shortcut.

miloush:
Yeah only this first one has the shortcut actually. If you have a smart idea how to not show it for the others, let me know. Otherwise this needs to be much more complicated PR with new commands or templated MenuItem and it is not exactly the most prominent menu.

Unless you are happy with something like InputGestureText=” “ that might work…?

siegfriedpammer:
> Unless you are happy with something like InputGestureText=” “ that might work…?

That is the solution I was going to suggest. No need to over-complicate things :)

miloush:
That did the trick indeed.

siegfriedpammer:

PR #3030, opened 2023-07-01
Fix #3025: Annotate nested types

### LINKED ISSUE #3025: The outer class of a nested class is not clickable
As the title said, the outer class is not clickable and no specific colour just like namespace in all built-in theme, plain black and white depends on light/dark theme. The hover tooltip is coloured as expected.
image

```csharp
namespace Namespace;
public class Layer1
{
    public class Layer2
    {
        public class Layer3
        {
            public class Layer4
            {
                public static void Method()
                {
                }
            }

            public static void Method()
            {
            }
        }

        public static void Method()
        {
        }
    }

    public static void Method()
    {
        Layer2.Method();
        Layer2.Layer3.Method();
        Layer2.Layer3.Layer4.Method();
    }
}
```
Perhaps this should be posted to [AvalonEdit](https://github.com/icsharpcode/AvalonEdit/)?

ltrzesniewski:
Fixes #3025

### Problem

- #3025

### Solution

Normally, `AddTypeAnnotation` is called right after `ConvertTypeHelper` in the `ConvertType` method:

https://github.com/icsharpcode/ILSpy/blob/bf0d74d0c713a0ae231a7f650cc19d06a56c784c/ICSharpCode.Decompiler/CSharp/Syntax/TypeSystemAstBuilder.cs#L244-L245

But in the case of nested types, `ConvertTypeHelper` calls itself recursively for declaring types, and therefore `AddTypeAnnotation` was never called for these nodes.

This is a very simple change.

I couldn't find any existing unit tests which check for type annotations (`ResolveResult`), so I wasn't sure where to put this. If you'd like me to add a unit test, please tell me in which way you'd like for type annotations to be tested.

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/Syntax/TypeSystemAstBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/Syntax/TypeSystemAstBuilder.cs
@@ -538,6 +538,7 @@ namespace ICSharpCode.Decompiler.CSharp.Syntax
 			{
 				// Handle nested types
 				result.Target = ConvertTypeHelper(genericType.DeclaringType, typeArguments);
+				AddTypeAnnotation(result.Target, genericType.DeclaringType);
 			}
 			else
 			{


siegfriedpammer:

PR #3032, opened 2023-07-02
Fix #3031: emit sequence points for expression-bodied properties and indexers

### LINKED ISSUE #3031: Sequence points not emitted for expression-bodied members
Decompile this program:

```csharp
class C
{
    int Property => 42;

    C this[int index]
    {
        get => null;
    }

    static void Main()
    {
        var c = new C();

        _ = c.Property;
        _ = c[0];
    }
}
```

When debugging, it should be possible to step in the expression body of the property and indexer, and set breakpoints, etc. However the sequence points for these members are not emitted. 

See discussion starting at comment https://github.com/icsharpcode/ILSpy/issues/1901#issuecomment-1605829503

KirillOsenkov:
Need to visit properties and indexers in SequencePointBuilder to account for ExpressionBody.

Adds a failing test for all kinds of members. Currently failing with an unrelated duplicate sequence points for a local function body.

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/Transforms/NormalizeBlockStatements.cs
+++ b/ICSharpCode.Decompiler/CSharp/Transforms/NormalizeBlockStatements.cs
@@ -218,6 +218,7 @@ namespace ICSharpCode.Decompiler.CSharp.Transforms
 				return;
 			propertyDeclaration.Modifiers |= propertyDeclaration.Getter.Modifiers;
 			propertyDeclaration.ExpressionBody = m.Get<Expression>("expression").Single().Detach();
+			propertyDeclaration.CopyAnnotationsFrom(propertyDeclaration.Getter);
 			propertyDeclaration.Getter.Remove();
 		}
 
@@ -230,6 +231,7 @@ namespace ICSharpCode.Decompiler.CSharp.Transforms
 				return;
 			indexerDeclaration.Modifiers |= indexerDeclaration.Getter.Modifiers;
 			indexerDeclaration.ExpressionBody = m.Get<Expression>("expression").Single().Detach();
+			indexerDeclaration.CopyAnnotationsFrom(indexerDeclaration.Getter);
 			indexerDeclaration.Getter.Remove();
 		}
 	}
--- a/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
+++ b/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
@@ -325,6 +325,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <Content Include="TestCases\PdbGen\Members.xml" />
     <Content Include="TestCases\PdbGen\ProgressReporting.xml" />
     <Content Include="TestCases\PdbGen\ForLoopTests.xml" />
     <Content Include="TestCases\PdbGen\CustomPdbId.xml" />
--- a/ICSharpCode.Decompiler.Tests/PdbGenerationTestRunner.cs
+++ b/ICSharpCode.Decompiler.Tests/PdbGenerationTestRunner.cs
@@ -44,6 +44,13 @@ namespace ICSharpCode.Decompiler.Tests
 		}
 
 		[Test]
+		[Ignore("Duplicate sequence points for local function")]
+		public void Members()
+		{
+			TestGeneratePdb();
+		}
+
+		[Test]
 		public void CustomPdbId()
 		{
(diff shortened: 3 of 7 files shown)


siegfriedpammer:
Thank you for the proposed fix. If I remember correctly, `SequencePointBuilder` should only handle statements and expressions, so everything inside a method, ctor, property, indexer, etc. The correct place to fix this would be `DebugInfoGenerator` which maps the generated sequence points to MethodDefinitions in the assembly. See my proposed fix in https://github.com/icsharpcode/ILSpy/tree/pdbgen-fixes...

There is one more nut I need to crack regarding local functions, but I might be able to get to that this week.

Note that we intentionally do not use expression-bodied members for single accessors, but only for getter-only properties/indexers. Also note that my branch mentioned above fixes another issue in the getter-only transform.

Thanks for creating a test case. I hope we'll be able to solve the PDB gen unit testing problem soon!

siegfriedpammer:
I think I will be basing the fix on the changes in my branch and use as much as possible from this PR. There is no need for you to invest any more time. However, I would very grateful, if you could take my final solution for a test ride. Thanks you very much!

siegfriedpammer:
> Thank you for the proposed fix. If I remember correctly, `SequencePointBuilder` should only handle statements and expressions, so everything inside a method, ctor, property, indexer, etc. The correct place to fix this would be `DebugInfoGenerator` which maps the generated sequence points to MethodDefinitions in the assembly. See my proposed fix in [`pdbgen-fixes`](https://github.com/icsharpcode/ILSpy/tree/pdbgen-fixes?rgh-link-date=2023-07-02T06%3A49%3A24Z)...

I have to retract this previous statement: the changes to `SequencePointBuilder` are equally necessary. Sorry for the confusion on my end. :) It's been a long time since I touched that part of the decompiler...

siegfriedpammer:
@KirillOsenkov I have pushed my changes. I have cleaned up the PR a bit. Could you try and test the generated PDB on your end? It seems to work fine with VS 17.6.4. Thanks a lot!

KirillOsenkov:
Sorry, I should have clarified, I don't work on the Visual Studio debugger team and I am not using the pdb generation logic at all. I'm not using the DebugInfoGenerator for my purposes, I only use the SequencePointBuilder to get the raw sequence points. The changes in my PR are sufficient for my needs, however the VS debugger team is likely to need both your and my changes (I'm guessing).

ICSharpCode.Decompiler/CSharp/Transforms/NormalizeBlockStatements.cs
```diff
@@ -218,6 +218,7 @@ void SimplifyPropertyDeclaration(PropertyDeclaration propertyDeclaration)
 				return;
 			propertyDeclaration.Modifiers |= propertyDeclaration.Getter.Modifiers;
 			propertyDeclaration.ExpressionBody = m.Get<Expression>("expression").Single().Detach();
+			propertyDeclaration.CopyAnnotationsFrom(propertyDeclaration.Getter);
```
KirillOsenkov, line 221: I left a comment on your old branch where I think these are redundant. After we copy annotations, we end up with a duplicate ILFunction annotation. Have you checked that this works as expected?

KirillOsenkov:
Other than the potential duplicate ILFunction annotations in CopyAnnotationsFrom, it looks good as far as I can see. Thanks!

ICSharpCode.Decompiler/CSharp/Transforms/NormalizeBlockStatements.cs
siegfriedpammer: The property declaration AST node has no ILFunction annotation at all, which makes sense because there is no 1:1 mapping between IL code and metadata, if there are multiple accessors, so we store the ILFunction on the accessors. When removing the accessor syntax from the output, we need to copy all annotations from the single accessor to the other declaration to ensure that the ILFunction is still available to others. This was a small oversight on my end, when I originally implemented that transform.

The problem is that somehow the accessor declaration already has two duplicate ILFunction annotations. This can easily be verified by stepping through CSharpDecompiler.DoDecompile(IProperty, ...) and inspecting the Annotations of the property declaration and accessor declarations step by step.

I will investigate further and hopefully be able to fix the root cause, but in general I don't think that it's a problem having two references to the same ILFunction in the collection. Usually our code uses `FirstOrDefault` semantics when working with annotations.

KirillOsenkov (commented):
Looks great as far as I can see. Thanks so much! 👍🏻

siegfriedpammer:
Thank you for your help with this!

PR #3035, opened 2023-07-12
Update Check for ilspycmd dotnet tool

christophwille:
* Update McMaster.Extensions.CommandLineUtils to 4.0.1 (actually switch to McMaster.Extensions.Hosting.CommandLine for generic host support)
* Fix [DirectoryExists] for ReferencePaths due to subtle change in 3.1.0 -> 4.0.1 upgrade
* Add DotNetToolUpdateChecker.CheckForPackageUpdateAsync inspired by efcpt dotnet tool

### DIFF
--- a/ICSharpCode.ILSpyCmd/DotNetToolUpdateChecker.cs
+++ b/ICSharpCode.ILSpyCmd/DotNetToolUpdateChecker.cs
@@ -1,0 +1,51 @@
+﻿using System;
+using System.Linq;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+
+using NuGet.Common;
+using NuGet.Protocol;
+using NuGet.Protocol.Core.Types;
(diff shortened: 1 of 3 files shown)


ICSharpCode.ILSpyCmd/DotNetToolUpdateChecker.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System;
```
siegfriedpammer, line 1: License?

siegfriedpammer (approved):

ICSharpCode.ILSpyCmd/DotNetToolUpdateChecker.cs
christophwille: https://github.com/ErikEJ/EFCorePowerTools/blob/master/LICENSE (MIT)

PR #3045, opened 2023-07-27
Copy to clipboard with full color

### LINKED ISSUE #3039: Feature request: copy to clipboard with formatting
My request is to maintain code-coloring and other formatting when code is copied to the clipboard.

We make heavy use of ILSpy against crash dump output to help developers understand bugs in their code. Way back in version 2.x, copy-paste from ILSpy retained all the code-coloring and other formatting. Somewhere along the way this has been lost. We actually use a 2.x version when we do writeups because the code formatting makes complex code walk-throughs easier to read and understand, but because that version pre-dated generics and other modern language features, that's no longer a reasonable option much of the time.

ltrzesniewski:
I was bored so I wrote this 😅

Fixes #3039

### Problem

- #3039 

  The text copied to the clipboard only had the basic syntax highlighting as defined in the xshd file. Additional highlighting provided by `CSharpHighlightingTokenWriter` wasn't present.

### Solution

  - I took the HTML copy code from AvalonEdit's `Selection.CreateHtmlFragment` as a baseline
  - The new code creates a `HighlightedLine` from the active `RichTextModel`
  - It then merges it into the one provided by the `IHighlighter`

  The code is straightforward enough. I only wondered if this situation (having multiple `DocumentColorizingTransformer` instances) is common enough to warrant being handled by AvalonEdit itself, but that would require a rather large refactoring.

  No test; only the UI part changed.

### DIFF
--- a/ILSpy/TextView/DecompilerTextView.cs
+++ b/ILSpy/TextView/DecompilerTextView.cs
@@ -25,6 +25,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection.Metadata;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Windows;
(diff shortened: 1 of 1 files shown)


siegfriedpammer:
Thank you very much! :)

PR #3052, opened 2023-08-08
Add the Entry Size Info

startewho:
### Problem
When use a single file fomart, can not found the detal size of assmessly

### Solution
* Add the size info on the entry output 
image


siegfriedpammer:

Issue #1885, opened 2020-01-02
[ILSpy.ReadyToRun] Decorate the disassembly with GCInfo


cshung:
The garbage collector of .NET is compacting. In order to move objects, it must be able to update all references to the object. To update all references to the object, it must know where the references are, in particular, it must know all the references that is currently living on the stack. It must also be precise. The GC should not be modifying a pointer-like integer on the stack.

To make this possible, the methods are annotated with GCInfo. The GCInfo describes where the GC references are (and the type of them) on a method-by-method basis. Depending on where exactly the code is executing, the compiler could decide where to put the GC references, and therefore we have a time-varying set of locations where GC references are stored. This information is compactly stored with a differential encoding, for example, we could add/remove a register or a stack slot from the set of GC references after executing a certain instruction.

Because the encoding is associated with point in the instruction execution, it would be nice to display the GC info weaved with the disassembly. A sample implementation of how to weave it with the disassembly is available in [`TextDumper.DumpDisasm`](https://github.com/dotnet/runtime/blob/e440e7081f85efa5a3492e418b3aca4a83ab7040/src/coreclr/src/tools/r2rdump/TextDumper.cs#L215-L221)

PR #3076, opened 2023-09-01
Decorate the disassembly with GCInfo

eduardo-vp:
Fixes https://github.com/icsharpcode/ILSpy/issues/1885

Shows GC Info at the beginning and the GC transitions as well. Also it stops iterating through the whole `runtimeFunction.DebugInfo.BoundsList` for each instruction for a slight performance improvement.

image

### DIFF
--- a/ILSpy.ReadyToRun/Properties/Resources.zh-Hans.resx
+++ b/ILSpy.ReadyToRun/Properties/Resources.zh-Hans.resx
@@ -129,4 +129,7 @@
   <data name="ShowUnwindInfo" xml:space="preserve">
     <value>显示展开信息</value>
   </data>
+  <data name="ShowGCInfo" xml:space="preserve">
+    <value>?</value>
+  </data>
 </root>
--- a/ILSpy.ReadyToRun/ReadyToRunDisassembler.cs
+++ b/ILSpy.ReadyToRun/ReadyToRunDisassembler.cs
@@ -129,6 +129,7 @@ namespace ICSharpCode.ILSpy.ReadyToRun
 						}
 					}
 				}
+				DecorateGCInfo(instr, baseInstrIP, readyToRunMethod.GcInfo);
 				formatter.Format(instr, tempOutput);
 				output.Write(instr.IP.ToString("X16"));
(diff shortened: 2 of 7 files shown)


ILSpy.ReadyToRun/Properties/Resources.zh-Hans.resx
```diff
@@ -129,4 +129,7 @@
   <data name="ShowUnwindInfo" xml:space="preserve">
     <value>显示展开信息</value>
   </data>
+  <data name="ShowGCInfo" xml:space="preserve">
+    <value>?</value>
```
eduardo-vp, line 133: Don't know how to get a proper value here

ILSpy.ReadyToRun/ReadyToRunDisassembler.cs
```diff
@@ -47,10 +47,19 @@ public ReadyToRunDisassembler(ITextOutput output, ReadyToRunReader reader, Runti
 
 		public void Disassemble(PEFile currentFile, int bitness, ulong address, bool showMetadataTokens, bool showMetadataTokensInBase10)
 		{
-			// TODO: Decorate the disassembly with GCInfo
 			ReadyToRunMethod readyToRunMethod = runtimeFunction.Method;
 			WriteCommentLine(readyToRunMethod.SignatureString);
 
+			if (ReadyToRunOptions.GetIsShowGCInfo(null) && readyToRunMethod.GcInfo != null)
```
eduardo-vp, line 53: Also wondering if it's possible that `readyToRunMethod.GcInfo` returns null, in that case should we show a comment that the GC Info was not found (if the checkbox was enabled)?

ILSpy.ReadyToRun/Properties/Resources.zh-Hans.resx
dgrunwald: Initially you can leave the chinese version unmodified, it'll fall back to the english resources for missing strings.

cshung (approved):
LGTM

siegfriedpammer:

PR #3081, opened 2023-09-09
Decompile base & derived types in place

miloush:
### Problem
The list of base and derived types in ILSpy does not show anything useful in the output. At least assembly name and list of base types would be helpful, but I don't see any reason why the whole type, which already provides this information, cannot be decompiled in place - it is a simple change in the code and users do not have to jump back and forth when going through and comparing the implementations.

### Solution
* This PR makes the base and derived type nodes decompile the respective types.
* The parent nodes (`BaseTypesTreeNode` and `DerivedTypesTreeNode`) were calling all children to decompile. However, with increasing information in child nodes, it is not sustainable. They now explicitly list the children type names only, like they did before.
* While trivial in base types, the derived types are loaded asynchronously and children are not available on first decompilation. I don't see any easy way to await/add continuation to the loading, but I did not deem the output of _Derived Types_ node to be critical enough, so the output will be empty the first time. Open to ideas if that is not satisfactory.

### DIFF
--- a/ILSpy/TreeNodes/DerivedTypesTreeNode.cs
+++ b/ILSpy/TreeNodes/DerivedTypesTreeNode.cs
@@ -16,8 +16,10 @@
 // OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
+using System;
 using System.Collections.Generic;
 using System.Threading;
+using System.Windows.Threading;
 
 using ICSharpCode.Decompiler;
 using ICSharpCode.Decompiler.TypeSystem;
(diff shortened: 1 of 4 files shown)


ILSpy/TreeNodes/DerivedTypesTreeNode.cs
```diff
@@ -101,7 +103,13 @@ IEnumerable<ILSpyTreeNode> FetchChildren(CancellationToken cancellationToken)
 
 		public override void Decompile(Language language, ITextOutput output, DecompilationOptions options)
 		{
-			threading.Decompile(language, output, options, EnsureLazyChildren);
+			App.Current.Dispatcher.Invoke(DispatcherPriority.Normal, new Action(EnsureLazyChildren));
+			for (int i = 0; i < Children.Count; i++)
+			{
+				// LazyChildren are async and will not be ready on first call, avoid foreach.
```
dgrunwald, line 109: I don't see how avoiding `foreach` helps here, you still have a race condition.

siegfriedpammer:
Closing, because it's stale. If you or anyone else is still interested in having such a feature in ILSpy, please open an issue.

miloush:
I am keeping this commit in my branch if anyone is interested, and with the updates in the past year the async problem seems to be void now. https://github.com/miloush/ILSpy/commit/2004fcbcde6b8c4cf0620e075db5b55cddc3b742

PR #3082, opened 2023-09-09
Show base members

miloush:
### Problem
There isn't any way to show all members, including inherited, available on a type. This, for example, makes it slightly challenging to find out where a member that user knows about comes from.

### Solution
* Added _View > Show base members_ setting stored in filter settings.
* When the setting is on, the node concatenates `NestedTypes`, `Fields`, `Properties`, `Events` and `Methods` from the list of non-interface base types on the `ITypeDefinition`.
* Searching is not affected, i.e. searching for "button.fontstyle" will not yield any results even when the setting is on. This could be added in the future if needed.
* If expanded, a type node will reload children when the setting changes.

An alternative approach would be for the inheritance check to be part of the standard filtering. That might, however, slow down enumareting children considerably, as all inherited members would have to be checked each time.

Having a global setting is not necessary. If preferrable, we could only have explicit _Show base members_ in context menu on individual type nodes that would only affect them on by one. The current approach felt simpler (and possibly allowing for the search to take the setting into account in the future).

### DIFF
--- a/ILSpy/FilterSettings.cs
+++ b/ILSpy/FilterSettings.cs
@@ -47,6 +47,7 @@ namespace ICSharpCode.ILSpy
 		public FilterSettings(XElement element)
 		{
 			this.ShowApiLevel = (ApiVisibility?)(int?)element.Element("ShowAPILevel") ?? ApiVisibility.PublicAndInternal;
+			this.ShowApiInherited = (bool?)element.Element("ShowAPIInherited") ?? false;
 			this.Language = Languages.GetLanguage((string)element.Element("Language"));
 			this.LanguageVersion = Language.LanguageVersions.FirstOrDefault(v => v.Version == (string)element.Element("LanguageVersion"));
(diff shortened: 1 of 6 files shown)


miloush:
> the method simply returns the `Fields` property in that case.

That turns out to be the case for fields, properties and events. However, `GetMethods` filters out constructors, will have to look into that.

miloush:
Not only `GetMethods` and `GetMembers` do not return constructors, but also `GetConstructors` does not return static ones. Decided for using the properties for consistency and concatenate them from base classes. I am now done with this PR until asked for changes.

siegfriedpammer:
The team is not interested in adding such a feature. We strongly feel that there should be a 1 : 1 relationship between the metadata information and the tree view. Also, we'd like the normal decompilation and project decompilation to behave as similar as possible.

miloush:
No problem, thank you for your time checking this!

Issue #2690, opened 2022-05-12
Used by for Assemblies


christophwille:
Today, I can go into Metadata and somewhat "easily" see what is used by an assembly (in this screenshot: ILSpy's usage of ILSpyX, which is easy because it is one namespace, not all assemblies have one namespace):
![image](https://user-images.githubusercontent.com/344208/168080155-83c0ddd3-0178-4ce2-b7e7-308d017c3f10.png)
Goal: Have a context menu item on the References/&lt;assembly&gt; nodes called "Used by" that lists for each member the call sites (made up sample of a tree view):

* AssemblyList
  * Load
     * ClassName1.Method1
     * ClassName2.Method5
* BlahClassFromReference
  * get_Name
     * CallSiteX

Issue #3068, opened 2023-08-29
Support reading raw metadata blobs and Portable PDBs


jkoritzinsky:
**Is your feature request related to a problem? Please describe.**

I have been working on a new reader/writer for ECMA-335 metadata blobs. I've regularly used ILSpy for decompiling .NET assemblies, and I've had very positive experiences using the Metadata tab to explore the metadata tables. However, I've discovered that ILSpy does not support loading raw ECMA-335 blobs or Portable PDBs to explore the metadata tables.

**Describe the solution you'd like**

I'd like to be able to open a raw ECMA-335 metadata blob or a Portable PDB in ILSpy and explore the metadata tables using the Metadata Tab. I don't care if any other features of ILSpy are supported (especially as many of them don't make sense when all you have is a PDB or raw metadata).

**Additional context**

I'd be willing to help with the contribution of this feature.

siegfriedpammer:
Do you have examples of raw ECMA-335 blobs? Are they different from portable PDBs? Thanks!

jkoritzinsky:
I'll pull some examples next week for raw blobs for you. Portable PDBs are full PE files with metadata.

jkoritzinsky:
Here's an example of a raw ECMA-335 blob (wrapped in a ZIP file to make uploading to GitHub easy):
[System.Reflection.Metadata.ApplyUpdate.Test.AddStaticLambda.dll.composite.zip](https://github.com/icsharpcode/ILSpy/files/12590160/System.Reflection.Metadata.ApplyUpdate.Test.AddStaticLambda.dll.composite.zip)

siegfriedpammer:
@jkoritzinsky I have trouble reading the dmeta file you sent using SRM... is there something I am missing, i.e. do I need to implement a custom loader?

I trying to load it using:

	using var fileStream = new FileStream(fileName, FileMode.Open, FileAccess.Read));
	var provider = MetadataReaderProvider.FromMetadataStream(fileStream, MetadataStreamOptions.PrefetchMetadata);
	var metadata = provider.GetMetadataReader();

leads to the following exception:

	System.BadImageFormatException
	  HResult=0x8007000B
	  Message=Not enough space for stream header name.
	  Source=System.Reflection.Metadata
	  StackTrace:
	   at System.Reflection.Metadata.MetadataReader.ReadStreamHeaders(BlobReader& memReader)
[...]

PR #3092, opened 2023-10-05
add "Referenced Types" under References

fowl2:
### Problem
Seeing the types/members referenced in an assembly reference is cumbersome.

### Solution
Add a "Referenced Types" child node to each reference which contains the referenced types, including exported/forwarded types.

Future enhancements:
* C#-ify the names
* other children eg. ManifestResource

image
screenshot including forwarded/exported types

partially fixes https://github.com/icsharpcode/ILSpy/issues/2690. It looked like trying to add "analysis" for anything that's not `IEntity` was going to be a bad time.

### DIFF
--- a/ICSharpCode.Decompiler/ICSharpCode.Decompiler.csproj
+++ b/ICSharpCode.Decompiler/ICSharpCode.Decompiler.csproj
@@ -98,6 +98,7 @@
     <Compile Include="Disassembler\IEntityProcessor.cs" />
     <Compile Include="Disassembler\SortByNameProcessor.cs" />
     <Compile Include="IL\ApplyPdbLocalTypeInfoTypeVisitor.cs" />
+    <Compile Include="Metadata\ModuleReferenceMetadata.cs" />
     <Compile Include="NRTAttributes.cs" />
     <Compile Include="PartialTypeInfo.cs" />
     <Compile Include="CSharp\ProjectDecompiler\IProjectFileWriter.cs" />
--- a/ICSharpCode.Decompiler/ICSharpCode.Decompiler.csproj
+++ b/ICSharpCode.Decompiler/ICSharpCode.Decompiler.csproj
@@ -362,6 +362,7 @@
     <Compile Include="IL\Transforms\TransformDisplayClassUsage.cs" />
     <Compile Include="IL\Transforms\UserDefinedLogicTransform.cs" />
     <Compile Include="Metadata\AssemblyReferences.cs" />
+    <Compile Include="Metadata\ExportedTypeMetadata.cs" />
     <Compile Include="Metadata\MemberReferenceMetadata.cs" />
     <Compile Include="Metadata\TypeReferenceMetadata.cs" />
     <Compile Include="Metadata\CodeMappingInfo.cs" />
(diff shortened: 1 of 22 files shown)


siegfriedpammer:
Thank you for proposing this feature! It definitely looks interesting, some notes/comments from my side:

* I am currently working on #3068 which will most likely require some refactoring in ILSpy's tree nodes, which will conflict with this feature. Until the refactoring is done, I won't be able to merge this.
* The icon for exported types looks really strange, maybe we can come up with some better symbol.
* I don't think any tree nodes should print their signature info in the tooltip, I'd prefer to do the same as we do for all other tree nodes of type definitions and members.
* The tree node label (i.e., Text property) should use the appropriate methods of the currently selected Language implementation.

fowl2:
> Thank you for proposing this feature! It definitely looks interesting, some notes/comments from my side:

:)

>     * I am currently working on [Support reading raw metadata blobs and Portable PDBs #3068](https://github.com/icsharpcode/ILSpy/issues/3068) which will most likely require some refactoring in ILSpy's tree nodes, which will conflict with this feature. Until the refactoring is done, I won't be able to merge this.

Sure.

>     * The icon for exported types looks really strange, maybe we can come up with some better symbol.

Yeah. Maybe something like the "Type" icon with a "Shortcut" overlay? (not that I know how to implement that)

>     * I don't think any tree nodes should print their signature info in the tooltip, I'd prefer to do the same as we do for all other tree nodes of type definitions and members.

I'm not 100% sure what you mean here, but I'll try to figure it out the next time I have a chance.

>     * The tree node label (i.e., Text property) should use the appropriate methods of the currently selected Language implementation.

Do you mean completing the TODO inside `CSharpLanguage::GetEntityName`, or using a different method?

siegfriedpammer:
@fowl2 if you don't mind, I will be taking your branch and add the missing things, clean it up, etc. and push my changes here for further discussion...

fowl2:
@siegfriedpammer sure, any way I can be help.

Have been looking into wether it would make sense to have an implementation of `IMember`, possibly unified with `FakeMember` etc. to get the correct language specific formatting and possibly the analysis features working, but I'm still thinking through any "fuzziness" with mismatches, etc.

ICSharpCode.Decompiler/Metadata/ExportedType.cs
```diff
@@ -0,0 +1,100 @@
...
+using System;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Reflection;
+using System.Reflection.Metadata;
+
+namespace ICSharpCode.Decompiler.Metadata
+{
+#if !VSADDIN
+	public sealed class ExportedType
```
siegfriedpammer, line 30: Instead of sprinkling `using SRM = System.Reflection.Metadata;` everywhere I would suggest using different names for the new types.

Also, I am not sure, if we really need them. I think that the object model for the "Referenced Types" feature should be part of the UI only.

siegfriedpammer:
Note: looking at the code, it seems it does not support showing nested type references.

siegfriedpammer:
One problem with getting nice member signatures in the tree view is that we probably want to show all members even those from assemblies that are not resolvable. So probably `Language.GetEntityName` based on SRM is the best we can do, without creating yet another layer doing similar things.

siegfriedpammer (approved):

siegfriedpammer:
Thank you for proposing and implementing this feature. Sorry that it took so long to get this merged.

PR #3106, opened 2023-10-18
Reduce "unsafe" by replacing byte* pointer usage with  ReadOnlySpan<byte>

fowl2:
This replaces all the `unsafe` code involved in metadata parsing in the ILSpy project with the equivalent using `ReadOnlySpan<byte>`. There's a little bit left in ICSharpCode.Decompiler, mostly around resource handling.

Not sure if it's worth it, especially given possible performance regression when running on netfx, but since I already did it as an experiment, it might be useful to the project.

side note: I think all ECMA 335 metadata is little endian, but the code doesn't really seem to deal with that. I don't think .net runs on any big endian machines anymore anyway.

### DIFF
--- a/ILSpy/Metadata/Helpers.cs
+++ b/ILSpy/Metadata/Helpers.cs
@@ -212,19 +212,18 @@ namespace ICSharpCode.ILSpy.Metadata
 			}
 		}
 
-		[Obsolete("Use ReadOnlySpan<byte> overload")]
+		[Obsolete("Use safe GetValueLittleEndian(ReadOnlySpan<byte>) or appropriate BinaryPrimitives.Read* method")]
 		[MethodImpl(MethodImplOptions.AggressiveInlining)]
 		public static unsafe int GetValue(byte* ptr, int size)
-			=> GetValue(new ReadOnlySpan<byte>(ptr, size));
+			=> GetValueLittleEndian(new ReadOnlySpan<byte>(ptr, size));
 
 		[MethodImpl(MethodImplOptions.AggressiveInlining)]
-		public static int GetValue(ReadOnlySpan<byte> ptr, int size)
-			=> GetValue(ptr.Slice(0, size));
+		public static int GetValueLittleEndian(ReadOnlySpan<byte> ptr, int size)
+			=> GetValueLittleEndian(ptr.Slice(0, size));
 
 		[MethodImpl(MethodImplOptions.AggressiveInlining)]
-		public static int GetValue(ReadOnlySpan<byte> ptr)
+		public static int GetValueLittleEndian(ReadOnlySpan<byte> ptr)
 		{
-			// endianess?
 			int result = 0;
 			for (int i = 0; i < ptr.Length; i += 2)
 			{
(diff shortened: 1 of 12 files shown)


ILSpy/Metadata/Helpers.cs
```diff
@@ -212,11 +212,21 @@ static void ApplyAttributes(PropertyDescriptor descriptor, Binding binding, Data
...
+			=> GetValue(new ReadOnlySpan<byte>(ptr, size));
+
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
+		public static int GetValue(ReadOnlySpan<byte> ptr, int size)
+			=> GetValue(ptr.Slice(0, size));
+
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
+		public static int GetValue(ReadOnlySpan<byte> ptr)
 		{
+			// endianess?
```
siegfriedpammer, line 227: If I am not mistaken, the code below assumes little endian byte-order and ECMA335 states that big endian byte-order is only used in signatures. Maybe we should call this method "GetValueLittleEndian"?

siegfriedpammer: When decoding signatures, we should not need this method, as we already have access to the raw bytes and SRM `BlobReader` already exposes helpers that handle compressed integers.

fowl2: 1. renamed `GetValueLittleEndian`
2. unless I'm misunderstanding these are not "compressed", they are just either "small" (2 byte) or "large" (4 bytes) depending on the size of the table they're indexing. Maybe this could be a `ReadTableIndex` or something, but I'm not sure how that would work.

siegfriedpammer: I don't follow...

fowl2: I'm sorry, what is unclear? Which method on BlobReader could we use instead of GetValueLittleEndian?

siegfriedpammer: OK... now I understand... sorry.... apparently my second comment yesterday was the confusing one: It was just meant as additional information regarding LE vs BE encoding used in ECMA335. BE is only used in signature blobs, in which case we don't need GetValue (now called "GetValueLittleEndian"), because SRMs BlobReader already supports all the different primitive types used in signatures.

I did not mean to suggest using BlobReader instead of GetValue/GetValueLittleEndian. However, now that I think more about it, maybe we could/should use the methods found in BinaryPrimitives instead? Not sure whether that's a good idea for all the different uses of the method.

fowl2: You mean reimplementing `GetValueLittleEndian` in terms of `BinaryPrimitives`, or in each of the callers?

Also, it is public - do we have to go though a obsoletion/deprecation process or can it just be changed/removedv?

siegfriedpammer: I don't expect anyone to actually use the method outside of the decompiler engine or ILSpy, however, wouldn't do any harm to deprecate the original overload (the one that uses pointers) and keep the name "GetValue" so we don't break anyone, right?

I think all the uses of GetValue(LittleEndian) could be directly replaced by BinaryPrimitives?

siegfriedpammer (approved):

siegfriedpammer:

PR #3111, opened 2023-11-06
Fix display of `DynamicLocalVariables` and `DefaultNamespaces` CDIs

ElektroKill:
Link to issue(s) this covers:
N/A

### Problem
The ILSpy UI incorrectly displayed `DynamicLocalVariables` and  `DefaultNamespaces`  custom debug info due to faulty return values.

### Solution
* Corrected the return values


siegfriedpammer:
Thanks for the fix!

Issue #3112, opened 2023-11-06
Use tuple element name and dynamic type information for local from PDB


ElektroKill:
Currently, the decompiler will emit a tuple type with no element names for local variables. Some which where of type `dynamic` in source will also appear as `object` variables due to the decompiling not being able to establish that the variable was in fact `dynamic`.

Example:
```csharp
public static void Main(string[] args) {
	dynamic obj = GetDynamicAnonymousObject();
	Console.WriteLine((object)obj);
	UseDynamic((object)obj);

	(int d, int e) tuple = GetTuple();
	Console.WriteLine(tuple);
	UseTuple(tuple);

	(int, int) tuple2 = GetTuple();
	Console.WriteLine(tuple2);
	UseTuple(tuple2);

	(int, int g) tuple3 = GetTuple();
	Console.WriteLine(tuple3);
	UseTuple(tuple3);
}

public static dynamic GetDynamicAnonymousObject() {
	return new { Name = "John", Age = 30 };
}

public static (int a, int b) GetTuple() {
	return (1, 2);
}

public static void UseDynamic(dynamic dynamic) {
	var obj = ((object)dynamic).ToString();
}

public static void UseTuple((int a, int b) tuple) {
	var obj = tuple.ToString();
}
```
Decompiled code:
```csharp
public static void Main(string[] args)
{
	object obj = GetDynamicAnonymousObject();
	Console.WriteLine(obj);
	UseDynamic(obj);
	(int, int) tuple = GetTuple();
	Console.WriteLine(tuple);
	UseTuple(tuple);
	(int, int) tuple2 = GetTuple();
	Console.WriteLine(tuple2);
	UseTuple(tuple2);
	(int, int) tuple3 = GetTuple();
	Console.WriteLine(tuple3);
	UseTuple(tuple3);
}

public static dynamic GetDynamicAnonymousObject()
{
	return new
	{
		Name = "John",
		Age = 30
	};
}

public static (int a, int b) GetTuple()
{
	return (1, 2);
}

public static void UseDynamic(dynamic dynamic)
{
	((object)dynamic).ToString();
}

public static void UseTuple((int a, int b) tuple)
{
	tuple.ToString();
}
```

The decompiled code loses the `dynamic` type and the tuple element names as they are not present in the compiled assembly. However, they are present in the PDB (both PortablePdb and WindowsPdb support this extra information)
[...]

PR #3114, opened 2023-11-07
Read and use tuple element names and dynamic type information from PDBs

ElektroKill:
Link to issue(s) this covers:
fixes https://github.com/icsharpcode/ILSpy/issues/3112

### Problem
ILSpy did not utilize helpful type information from PDB files when determining the type of local variables.

### Solution
* Implementation as mentioned in the original issue.
* Added additional code to extract the necessary information from the PDB file and added a new `TypeVisitor` to apply this information to the type system.

### DIFF
--- a/ICSharpCode.Decompiler/DebugInfo/IDebugInfoProvider.cs
+++ b/ICSharpCode.Decompiler/DebugInfo/IDebugInfoProvider.cs
@@ -23,6 +23,7 @@ namespace ICSharpCode.Decompiler.DebugInfo
 		IList<SequencePoint> GetSequencePoints(MethodDefinitionHandle method);
 		IList<Variable> GetVariables(MethodDefinitionHandle method);
 		bool TryGetName(MethodDefinitionHandle method, int index, out string name);
+		bool TryGetExtraTypeInfo(MethodDefinitionHandle method, int index, out string[] tupleElementNames, out bool[] dynamicFlags);
 		string SourceFileName { get; }
 	}
 }
--- a/ICSharpCode.Decompiler/DebugInfo/IDebugInfoProvider.cs
+++ b/ICSharpCode.Decompiler/DebugInfo/IDebugInfoProvider.cs
@@ -17,13 +17,19 @@ namespace ICSharpCode.Decompiler.DebugInfo
 		public string Name { get; }
 	}
 
+	public struct PdbExtraTypeInfo
+	{
+		public string[] TupleElementNames;
+		public bool[] DynamicFlags;
+	}
+
 	public interface IDebugInfoProvider
 	{
 		string Description { get; }
 		IList<SequencePoint> GetSequencePoints(MethodDefinitionHandle method);
 		IList<Variable> GetVariables(MethodDefinitionHandle method);
 		bool TryGetName(MethodDefinitionHandle method, int index, out string name);
-		bool TryGetExtraTypeInfo(MethodDefinitionHandle method, int index, out string[] tupleElementNames, out bool[] dynamicFlags);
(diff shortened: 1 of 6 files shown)


ICSharpCode.Decompiler/DebugInfo/IDebugInfoProvider.cs
```diff
@@ -23,6 +23,7 @@ public interface IDebugInfoProvider
 		IList<SequencePoint> GetSequencePoints(MethodDefinitionHandle method);
 		IList<Variable> GetVariables(MethodDefinitionHandle method);
 		bool TryGetName(MethodDefinitionHandle method, int index, out string name);
+		bool TryGetExtraTypeInfo(MethodDefinitionHandle method, int index, out string[] tupleElementNames, out bool[] dynamicFlags);
```
siegfriedpammer, line 26: In general I am fine with out parameters, but it might be beneficial to use a class/struct instead to avoid binary breaking changes in interface methods. So if there is a new language feature we want to add support for, we don't have to add extra overloads or do a new major version. What do you think?

siegfriedpammer:
In general LGTM, however, we should discuss the method signature of `TryGetExtraTypeInfo`.

ICSharpCode.Decompiler/DebugInfo/IDebugInfoProvider.cs
ElektroKill: I'd agree with introducing a struct for the out parameters to make way for future information that could find its way into PDB files. 

I think something like this would be fine:
```csharp
bool TryGetExtraTypeInfo(MethodDefinitionHandle method, int index, out PdbExtraTypeInfo extraTypeInfo);
```

```csharp
struct PdbExtraTypeInfo {
	public string[] TupleElementNames;
	public bool[] DynamicFlags;
}
```

siegfriedpammer: Yeah, I think that's fine.

ICSharpCode.ILSpyX/PdbProvider/PortableDebugInfoProvider.cs
```diff
@@ -161,5 +161,75 @@ public bool TryGetName(MethodDefinitionHandle method, int index, [NotNullWhen(tr
...
+				var kind = metadata.GetGuid(cdi.Kind);
+				if (kind == KnownGuids.TupleElementNames && extraTypeInfo.TupleElementNames is null)
+				{
+					var reader = metadata.GetBlobReader(cdi.Value);
+					var list = new List<string?>();
+					while (reader.RemainingBytes > 0)
+					{
+						int length = reader.IndexOf(0);
+						string s = reader.ReadUTF8(length);
+						reader.ReadByte();
```
siegfriedpammer, line 209: Should probably use `ReadUTF8StringNullTerminated` instead or add a comment why the extra `ReadByte` is necessary.

ElektroKill: `ReadUTF8StringNullTerminated` is not available in `ILSpyX` project as it's defined in `ILSpy`. I will go ahead and add a comment stating that this extra `ReadByte` is responsible for reading the null terminator.

siegfriedpammer: Ah, sorry, that method only exists in ILSpy.

ElektroKill:
The build seems to be failing after the recent upgrade of the upload artifact GH action. Looking at the error it looks like a limit for the amount of artifacts was imposed.

riQQ:
See https://github.com/actions/upload-artifact/blob/v4.0.0/README.md#breaking-changes

PR #3132, opened 2023-11-28
Migrate to NUnit 4

christophwille:


ICSharpCode.Decompiler.Tests/CorrectnessTestRunner.cs
```diff
@@ -486,7 +486,7 @@ async Task RunIL(string testFileName, CompilerOptions options = CompilerOptions.
 			bool optionsForce32Bit = options.HasFlag(CompilerOptions.Force32Bit);
 			bool asmOptionsForce32Bit = asmOptions.HasFlag(AssemblerOptions.Force32Bit);
 
-			Assert.AreEqual(optionsForce32Bit, asmOptionsForce32Bit, "Inconsistent architecture.");
+			Assert.That(asmOptionsForce32Bit, Is.EqualTo(optionsForce32Bit), "Inconsistent architecture.");
```
siegfriedpammer, line 489: I really don't like that overly verbose syntax.... what's wrong with `AreEqual`?

siegfriedpammer: https://docs.nunit.org/articles/nunit/writing-tests/assertions/assertions.html... why not keep the classic model for all the old tests?

christophwille: Guess what will happen if we keep classic. It will be copy pasted forever.

siegfriedpammer: I see no problem with that... :D the short forms are more intuitive...

OsirisTerje:
Version 4.0.1 is out on nuget,  https://docs.nunit.org/articles/nunit/release-notes/framework.html#nunit-401---december-2-2023

PR #3147, opened 2023-12-29
feat: OpenWith

workgroupengineering:
Add to OpenWith Windows ContextMenu `Open With ILSpy`

### Problem
Allow quick inspect .dll, .nupkg and .vsix file.

### Solution

Add entry `Open With ILSpy` to  OpenWith Windows ContextMenu

### DIFF
--- a/ILSpy/App.xaml.cs
+++ b/ILSpy/App.xaml.cs
@@ -86,6 +86,7 @@ namespace ICSharpCode.ILSpy
 											  Hyperlink.RequestNavigateEvent,
 											  new RequestNavigateEventHandler(Window_RequestNavigate));
 			ILSpyTraceListener.Install();
+			RegisterFilesAssociation();
 		}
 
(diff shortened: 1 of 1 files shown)


siegfriedpammer:
Thank you for your contibution. However, ILSpy currently cannot open VSIX files.

How is this different from the existing context menu integration? IMO this **should not** happen without the user's approval, i.e., if we add this, it needs to go into the misc options dialog like the existing integration...
But I think, we could simply reuse the existing integration and add it for nupkg files.

workgroupengineering:
> Thank you for your contibution. However, ILSpy currently cannot open VSIX files.

Works 
immagine


> 
> How is this different from the existing context menu integration? IMO this **should not** happen without the user's approval, i.e., if we add this, it needs to go into the misc options dialog like the existing integration... But I think, we could simply reuse the existing integration and add it for nupkg files.

I didn't know its existence. Thank you

siegfriedpammer:
So then I discovered a bug, while testing this... thanks for pointing that out!

PR #3191, opened 2024-03-31
Add IFileLoader API

siegfriedpammer:


ICSharpCode.ILSpyX/FileLoaders/LoadResult.cs
```diff
@@ -0,0 +1,39 @@
...
+		public MetadataFile? MetadataFile { get; init; }
+		public Exception? FileLoadException { get; init; }
+		public LoadedPackage? Package { get; init; }
+	}
+
+	public record FileLoadSettings(bool ApplyWinRTProjections);
+
+	public interface IFileLoader
+	{
+		LoadResult? Load(string fileName, Stream stream, FileLoadSettings settings);
```
dgrunwald, line 37: Maybe this should be async? We're already calling it within an async method; so this would give loaders some additional implementation options.

PR #3195, opened 2024-04-09
Fix NETCORE to NET8(or newer) directive

christophwille:


ICSharpCode.Decompiler/Util/KeyComparer.cs
```diff
@@ -81,7 +81,8 @@ public bool Equals(TElement? x, TElement? y)
 
 		public int GetHashCode(TElement obj)
 		{
-			return keyEqualityComparer.GetHashCode(keySelector(obj));
+			var key = keySelector(obj)!;
```
siegfriedpammer, line 84: @dgrunwald anything wrong with this?

PR #3196, opened 2024-04-15
Natural Sort without interop

christophwille:
Using NaturalSort.Extensions 4.3 that was updated for Unicode as per https://github.com/tompazourek/NaturalSort.Extension/issues/74

Benchmark https://github.com/christophwille/poc-oh/blob/main/src/NaturalSortBenchmark/README.md

### DIFF
--- a/ILSpy/TreeNodes/NaturalStringComparer.cs
+++ b/ILSpy/TreeNodes/NaturalStringComparer.cs
@@ -17,7 +17,6 @@
 // DEALINGS IN THE SOFTWARE.
 
 using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.Runtime.InteropServices;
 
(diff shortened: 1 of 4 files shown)


ILSpy/TreeNodes/NaturalStringComparer.cs
```diff
@@ -16,22 +16,37 @@
...
+		// public static readonly NaturalStringComparer Instance = new NaturalStringComparer();
+	}
+
+#if WINDOWS
+	/// <summary>
+	/// Native-Interop natural string comparion using StrCmpLogicalW from shlwapi.dll
+	/// </summary>
 	[System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1060:MovePInvokesToNativeMethodsClass")]
-	public sealed class NaturalStringComparer : IComparer<string>
+	public sealed class NativeNaturalStringComparer : IComparer<string>
```
siegfriedpammer, line 41: Why do we keep the old one around?

christophwille: That is just for optionality reasons - eg if we ever wanted to compare for a bug report on sorting.

PR #3205, opened 2024-05-03
`@file` support with breaking changes to command line options

christophwille:
See the amended "Command Line.txt".

Notes:
* `singleInstance` is removed. Reason: if the user has a configuration for allowing multiple instances of ILSpy and there are already multiple instances actually running, we cannot enforce that parameter anyways.
* `separate` has been replaced with `newinstance`. This will always open a new instance of ILSpy irrespective of what the user has configured.
* we no longer use `/` but standard `--` for options
* we now support `@file` for passing command line parameters
* VS Addin uses the new `@file` approach

### DIFF
--- a/ILSpy.AddIn.Shared/ILSpyInstance.cs
+++ b/ILSpy.AddIn.Shared/ILSpyInstance.cs
@@ -1,11 +1,8 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
-using System.Text;
-using System.Threading.Tasks;
 
 namespace ICSharpCode.ILSpy.AddIn
 {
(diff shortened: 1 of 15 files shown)


siegfriedpammer:
```ilspy --help``` does nothing and prints nothing

File: ILSpy.AddIn.Shared/ILSpyInstance.cs [hunk 1]
```diff
@@ -47,85 +44,28 @@ public void Start()
...
-						}, IntPtr.Zero);
+				string assemblyArguments = string.Join(" ", commandLineArguments);
 
-					// Wait some time before next attempt
-					await Task.Delay(500);
-					remainingAttempts--;
-				} while (!success && (remainingAttempts > 0));
-			});
-		}
+				string filepath = Path.GetTempFileName();
```
siegfriedpammer, line 55: who is going to delete the file?

christophwille: Resolved via https://github.com/icsharpcode/ILSpy/pull/3205/commits/d2221783ba211be78072701876a1ed490248230c

christophwille:
> `ilspy --help` does nothing and prints nothing

done - with a dialog that pops up on --help.

PR #3215, opened 2024-05-30
fixes: Decompile Package (.nuget;.vsix; ecc...)

workgroupengineering:
- Avoid exception on ArchiveFileLoader, BundleFileLoader and WebCilFileLoader
- Try Resolve .exe file

Link to issue(s) this covers

### Problem
in the last master d7771486d20206d94481f0a0f49275a101dc60c3, if you try to load the nuget or vsix package, when you select a library in the decompiler tab a FileLoadException is reported

### Solution


siegfriedpammer:
Thank you for the proposed fix! For a long time I had a fix for the bug on my computer, but had no time to actually commit/push it, which I now did in 77b237d5a4dc3bd9ed1725f4fc569bff9e94e96a. It avoids `File.Exists` by tracking, if we're in a bundle/package context.

PR #3219, opened 2024-06-07
Update README.md

nortages:
Fix help option

### Problem
Incorrect syntax of the help command option.


christophwille:
Thanks for pointing this out, however, I found that there is actually more off (old, outdated) than just this and fixed it accordingly in https://github.com/icsharpcode/ILSpy/commit/dd7d7b806f5cd0122c88425307b426e91c6c9117

PR #3234, opened 2024-07-15
Optimize the scrolling experience

SlimeNull:
1. Makes scrolling smoother when using the mouse
2. Makes scrolling distances more reasonable when scrolling with the touchpad

ilspy-scrolling2


christophwille:
We are currently not going to add new WPF package dependencies. Do you see a way to change our existing WPF code to incorporate some or all of your improvements?

SlimeNull:
That's ok, I'll go back and copy the code.

christophwille:
@ltrzesniewski could you please join the review? Wrt to does it actually do what it says on the tin and whether it could be further simplified?

SlimeNull:
The current approach is to put the scrolling optimization part into a separate project. Since both ICSharpCode.TreeView and ILSpy projects need to use this logic, if we don't put it in a separate project, it means that we need to copy this code twice.

Is there a better way to do it? After all, scroll optimization is just a rewrite of ScrollViewer's OnMouseWheel

ltrzesniewski:
> @ltrzesniewski could you please join the review? Wrt to does it actually do what it says on the tin and whether it could be further simplified?

Sure, I can help, but not before next week - I'm currently on vacation and I don't have access to my dev tools.

In the meantime, let's ping @tom-englert: would you be willing to add a smooth scrolling feature to TomsToolbox.Wpf.Styles if this PR could be further simplified? IMO that would be a  better place for such code than a separate project here.

tom-englert:
@ltrzesniewski generally yes, however the actual implementation looks like a huge overkill for "just a rewrite of ScrollViewer's OnMouseWheel"

tom-englert:
Maybe @SlimeNull can point out where the core functionality is located?

SlimeNull:
The ScrollViewer class in project “EleCho.WpfSuite” is the core logic of scrolling optimization

SlimeNull:
In addition, I also have a scrolling adaptation solution for the stylus device here, I wonder if ILSpy needs this? The core logic is to simulate a touch device to achieve scrolling without affecting existing mouse operations.

About touch simulation：[WPF Suite - Touch Simulation](https://wpfsuite.elecho.dev/en/enhanced-controls#touch-simulation)

tom-englert:
If it's all about to capture `OnMouseWheel`, maybe a Behavior that finds the `ScrollViewer` as a child of the control and captures the `PreviewMouseWheel` event will do the job?

SlimeNull:
It should be a little bit tricky, because for virtualized panels, we also need to calculate the real scroll distance based on the control's template child, PART_ScrollContentPresenter. GetTemplatedChild is protected. Of course we can use VisualTree to find it.

tom-englert:
Another interesting approach: https://www.wpf-controls.com/wpf-smooth-scroll-viewer/

tom-englert:
@SlimeNull here's the scaffold for the behavior:

https://github.com/tom-englert/ILSpy/blob/dev/smoothscrolling/EleCho.WpfSuite/Controls/SmoothScrollingBehavior.cs

Actually it just blocks mouse wheel; you can just add your scrolling logic.

SlimeNull:
Wait, one more question. There is a ScrollViewer in ILSpy that responds to zooming, and it does so by overriding the OnMouseWheel method. If we set IsHandled to true in the PreviewMouseWheel handler, we're not going to be able to get rid of that scaling logic, are we?

tom-englert:
Of course you have to implement it correctly and only set it to true if it was a scroll

tom-englert:
PS. I have tried this with the mouse wheel, and the improvement is marginal, if even noticeable, is it really worth spending time on this?

SlimeNull:
For now, the cost of change is just to change the base class of ScrollViewer, ListBox, ListView, but if you want to abstract it into behavior, it will take more time

SlimeNull:
I don't really want to abstract it into a behavior, which means you may need to write scaling logic in the behavior. It's not very good. Because not every ScrollViewer has a zoom feature.

tom-englert:
> For now, the cost of change is just to change the base class of ScrollViewer, ListBox, ListView,

and DataGrid, TextBox, ... and whatever control uses a scroll viewer. 

Also you not only override the code, you also need to maintain a full copy of all templates of all affected controls - which in your current implementation even messes up the design of the affected controls.

And the problem with deriving instead of decorating is that you never can combine any two behaviors if you are not in full control of the source code.

SlimeNull:
I have now moved smooth movement to the Behavior implementation. However, once I set Handled to true in the Preview event, the ScrollViewer's OnMouseWheel method is not executed.

What should I do? Put the ZoomScrollViewer's zoom-related logic into PreviewMouseWheel as well? I don't want to write zoom-related logic in the Behavior.

SlimeNull:
I've switched to the Behavior implementation, but I've also changed the ZoomScrollViewer's OnMouseWheel to OnPreviewMouseWheel. They work fine now. Are there any other changes you'd like to make?

ILSpy/App.xaml
```diff
@@ -6,19 +6,19 @@
...
-			<Setter Property="TextOptions.TextFormattingMode" Value="Display"/>
-			<Setter Property="toms:StyleBindings.Behaviors">
-				<Setter.Value>
-					<toms:BehaviorCollection>
-						<themes:WindowStyleManagerBehavior />
-					</toms:BehaviorCollection>
-				</Setter.Value>
-			</Setter>
-		</Style>
+    <Application.Resources>
```
tom-englert, line 9: It's hard to review if you completely reformat whole files. (Applies to many others, too)

LibScrollingOptimization/SmoothScrollingBehavior.cs
```diff
@@ -0,0 +1,468 @@
...
+		{
+			obj.SetValue(AlwaysHandleMouseWheelScrollingProperty, value);
+		}
+
+
+		/// <summary>
+		/// The DependencyProperty of <see cref="ScrollWithWheelDelta"/> property.
+		/// </summary>
+		public static readonly DependencyProperty ScrollWithWheelDeltaProperty =
+			DependencyProperty.RegisterAttached(nameof(ScrollWithWheelDelta), typeof(bool), typeof(SmoothScrollingBehavior),
```
tom-englert, line 430: We don't need attached properties with inheritance, since the properties now can be directly set on the behavior.

tom-englert (changes requested):

ILSpy.sln
```diff
@@ -36,7 +36,9 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ICSharpCode.Decompiler.Test
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ICSharpCode.ILSpyX", "ICSharpCode.ILSpyX\ICSharpCode.ILSpyX.csproj", "{F8EFCF9D-B9A3-4BA0-A1B2-B026A71DAC22}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ICSharpCode.BamlDecompiler", "ICSharpCode.BamlDecompiler\ICSharpCode.BamlDecompiler.csproj", "{81A30182-3378-4952-8880-F44822390040}"
```
siegfriedpammer, line 39: Why is BamlDecompiler receiving a new GUID and the new project has the old one?

ILSpy/App.xaml
siegfriedpammer: If you append `?w=1` to the GitHub URL, e.g. https://github.com/icsharpcode/ILSpy/pull/3234/files?w=1 then you can review everything without whitespace changes.

siegfriedpammer: But I agree that this should be reverted, especially because the only change in this file is the indentation.

ILSpy.sln
tom-englert: This is a VS issue, when editing a solution, old guids will be updated, but new projects start with an old guid.

However the plan is to move the behavior as a public available standard item to toms-toolbox, so this change will be reverted anyway.

PR #3235, opened 2024-07-16
Add C# 12 primary constructors

siegfriedpammer:
Open questions:

* Do we want a separate setting for this language feature? Currently there is only a setting for primary constructors in records


christophwille:
I'd vote for a setting so class primary ctors can be enabled (default off)

Issue #3172, opened 2024-03-06
ILSpy Developer Meeting 2024, July 19th-21st


christophwille:
Location: Vienna, Austria.
Confirmed: @siegfriedpammer @dgrunwald @christophwille

christophwille:
# Thursday

Afternoon only, as participants are trickling in.

First review of #3235 

# Friday

Together with @Rpinski how-to add analyzers in [ilspy-vscode](https://github.com/icsharpcode/ilspy-vscode)

Noted down SharpTreeView TODOs if we ever move SharpTreeNodes into ILSpyX
* Move SharpTreeView into ILSpy
* In SharpTreeNode: add alternative for IDataObject (copy), also drag/drop needs alternatives, RoutedEventArgs
* SharpTreeNode/TreeFlattener/TreeTraversal -> ILSpyX
* Everything else -> ILSpy
* Kill SharpTreeView project
* Move TreeNodes from ILSpy to ILSpyX
  
Images class TODO:
* Add abstraction so that ILSpyX can access the icons: IIconFactory

Coding together: #829/ref readonly parameters, added finding: https://github.com/icsharpcode/ILSpy/issues/3237
	
Start work on upgrading Roslyn 4.9 to 4.10.0.

# Saturday

Roslyn 4.10 continued lifted operators, compound assignments (detour ILInlining)

# Sunday

* Compound assignments string.Concat, thus upgrading Roslyn 4.10 completed
* Tests for lambda optional parameters to mark language feature as done in #829
* Tests for ref readonly parameters, verify overload resolution to mark language feature as done in #829

# Notes for further TODOs

@siegfriedpammer complete #3235 with notes from @dgrunwald (expand record primary ctor detection for this case), as well as #3232 / #3233 / #3236

PR #3240, opened 2024-07-22
SharpTreeView: Split into ILSpyX.TreeView and ILSpy.Controls.TreeView

christophwille:
https://github.com/icsharpcode/ILSpy/issues/3172#issuecomment-2242135835

This is the basics - splitting xplat and WPF-specific parts of SharpTreeView and distribute to ILSpyX and ILSpy projects respectively.


christophwille:
@Rpinski please add your comments from the viewpoint of the ilspy-vscode extension

@jmacato @maxkatz6 from the point of view of Avalonia / Avalonia XPF - would such a tree node move/change as outlined in the issue comment plus this PoC PR be in line with that architecture or do the IPlatform* abstractions completely run counter what Avalonia would need?

jmacato:
Hi @christophwille !

The changes looks fine for my perspective on XPF, as it just abstracts out the needed interfaces for the treenode controls.

ICSharpCode.ILSpyX/TreeView/PlatformAbstractions/IPlatformDragDrop.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace ICSharpCode.ILSpyX.TreeView.PlatformAbstractions
+{
+	public interface IPlatformDragDrop
+	{
+		XPlatDragDropEffects DoDragDrop(object dragSource, object data, XPlatDragDropEffects allowedEffects);
```
maxkatz6, line 5: Note: this and other operations should be async, as they are async in nature.

On desktop (and Browser with multithreading and hacks enabled), we can solve it by using DispatcherFrame to sync-wait on async operations. But since you are refactoring this part, consider making it async from the start as well.

maxkatz6: Note, DispatcherFrame is usable with both XPF and pure Avalonia. That's what I was using so far in new port too.

maxkatz6: Except mobile, Avalonia doesn't have any way to use DispatcherFrame on mobile yet. But I assume ILSpy won't benefit much from mobile support either way. Browser might be useful though.

christophwille: Ah, interesting - I was only reshuffling existing code, and there DoDragDrop was used in exactly that fashion.

ILSpy/MainWindow.xaml
```diff
@@ -144,7 +146,8 @@
 						<Style TargetType="{x:Type MenuItem}" BasedOn="{StaticResource {x:Type MenuItem}}">
 							<Setter Property="Command" Value="{x:Static local:ILSpyCommands.SetTheme}" />
 							<Setter Property="CommandParameter" Value="{Binding}" />
-							<Setter Property="IsCheckable" Value="True" /> <!-- Required by AvalonDock's MenuItem style to show the checkmark -->
+							<Setter Property="IsCheckable" Value="True" />
```
siegfriedpammer, line 149: Comment was moved to the wrong line...

siegfriedpammer (approved):

PR #3244, opened 2024-07-27
Add smooth scrolling behavior to settings panels and DecompilerTextView

tom-englert:
supersedes #3234 by @SlimeNull; code has been cleaned up and moved to [TomsToolbox](https://github.com/tom-englert/TomsToolbox/blob/master/src/TomsToolbox.Wpf/Interactivity/SmoothScrollingBehavior.cs)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -45,7 +45,7 @@
     <PackageVersion Include="System.Reflection.Metadata" Version="8.0.0" />
     <PackageVersion Include="System.Resources.Extensions" Version="8.0.0" />
     <PackageVersion Include="System.Runtime.CompilerServices.Unsafe" Version="6.0.0" />
-    <PackageVersion Include="TomsToolbox.Wpf.Styles" Version="2.15.0" />
+    <PackageVersion Include="TomsToolbox.Wpf.Styles" Version="2.16.0" />
     <PackageVersion Include="coverlet.collector" Version="6.0.2" />
   </ItemGroup>
 </Project>
--- a/ILSpy/Controls/ZoomScrollViewer.cs
+++ b/ILSpy/Controls/ZoomScrollViewer.cs
@@ -108,7 +108,7 @@ namespace ICSharpCode.ILSpy.Controls
 				z.ComputedZoomButtonCollapsed = (z.AlwaysShowZoomButtons == false) && (z.CurrentZoom == 1.0);
 		}
 
-		protected override void OnMouseWheel(MouseWheelEventArgs e)
+		protected override void OnPreviewMouseWheel(MouseWheelEventArgs e)
 		{
 			if (!e.Handled && Keyboard.Modifiers == ModifierKeys.Control && MouseWheelZoom)
 			{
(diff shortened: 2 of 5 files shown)


christophwille:
@SlimeNull @ltrzesniewski everyone happy with this?

ltrzesniewski:
I like it, it works nice! Thanks @tom-englert for adding it to your toolbox. 🙂 

I tried to add the new behavior to the `AssemblyTreeView` component (for consistency), and didn't spot any difference. I ran the debugger, and noticed the following line gets executed and changes the `duration` to a very small value (0.01 sec):

```csharp
duration = new(TimeSpan.FromTicks((long)(duration.TimeSpan.Ticks * absDiff / Mouse.MouseWheelDeltaForOneLine)));
```

This line isn't executed for AvalonEdit, as `if (absDiff < Mouse.MouseWheelDeltaForOneLine)` is false. `absDiff` is 120 for AvalonEdit, but only 6.9 for the tree view.

I'm not sure what to do about this but it's not a blocker I suppose.

tom-englert:
@ltrzesniewski it has no visual effect if the `VirtualizingPanel.ScrollUnit` is `Item`, which is the default for e.g. `ListBox` and probably also the `TreeView`. 
I tried to change the scroll unit to `Pixel`, but the visual experience is much worse even with the smooth animation, so I did not investigate further.
However maybe it makes sense to use it also for the tree view to improve the touch pad experience?

ltrzesniewski:
Oh ok I see, I've also had bad experience with `Pixel` in the past, so I suppose it's better to leave it as-is indeed.

I can't really comment on the touchpad though.

christophwille:
I am going to merge the PR at this point in time, should further improvements / changes come up, we'll handle that with another PR.

SlimeNull:
@tom-englert I have a question, why am I not in the list of contributors? It doesn't matter if it's ILSpy or TomsToolbox. 🤓👆

tom-englert:
I'm just a simple contributor, I don't manage this project.

christophwille:
"and moved to TomsToolbox" - that would mean a mention would be in order there.

fawdlstty:
> I'm just a simple contributor, I don't manage this project.

You should fork from @SlimeNull's repository and do work so that everyone who contributed to the project can get the credit they deserve

christophwille:
> > I'm just a simple contributor, I don't manage this project.
> 
> You should fork from @SlimeNull's repository and do work so that everyone who contributed to the project can get the credit they deserve

this is not a free-for-all discussion thread.

tom-englert:
@fawdlstty forking is not an option, since the final solution has not too much in common with the original one

@SlimeNull If you like to be listed, I'd suggest to add a comment with the link to your project here https://github.com/tom-englert/TomsToolbox/blob/master/src/TomsToolbox.Wpf/Interactivity/AdvancedScrollWheelBehavior.cs. If you do it as a PR, you'll be also automatically listed as a contributor. 
Also maybe add some links to your own sources, since the basic principle of using an animation has been described earlier in several articles, e.g. on Stack Overflow e.a.

PR #3248, opened 2024-08-01
Fix #3246: Choppy scrolling when scrolling fast after smooth scrolling introduction

### LINKED ISSUE #3246: Choppy scrolling when scrolling fast after smooth scrolling introduction
### Steps to reproduce
1. Open ILSpy compiled from master (https://github.com/icsharpcode/ILSpy/commit/d1c7a51a1d823983afe93bae1d0c70a6912cf3da)
2. Decopile a large class like `System.Array` from `System.Private.CoreLib`
3. Scroll fast on the scroll wheel
4. Observe weird choppiness and snapping when scrolling

See video below which demonstrates the behavior:
https://www.youtube.com/watch?v=py0OHKgWccw

tom-englert:
#3246

### Problem
Choppy scrolling when scrolling fast after smooth scrolling introduction

### Solution
Update TomsToolbox.Wpf package with fixed behavior.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -45,7 +45,7 @@
     <PackageVersion Include="System.Reflection.Metadata" Version="8.0.0" />
     <PackageVersion Include="System.Resources.Extensions" Version="8.0.0" />
     <PackageVersion Include="System.Runtime.CompilerServices.Unsafe" Version="6.0.0" />
-    <PackageVersion Include="TomsToolbox.Wpf.Styles" Version="2.16.0" />
+    <PackageVersion Include="TomsToolbox.Wpf.Styles" Version="2.17.1" />
     <PackageVersion Include="coverlet.collector" Version="6.0.2" />
   </ItemGroup>
 </Project>
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -45,7 +45,7 @@
     <PackageVersion Include="System.Reflection.Metadata" Version="8.0.0" />
     <PackageVersion Include="System.Resources.Extensions" Version="8.0.0" />
     <PackageVersion Include="System.Runtime.CompilerServices.Unsafe" Version="6.0.0" />
-    <PackageVersion Include="TomsToolbox.Wpf.Styles" Version="2.17.1" />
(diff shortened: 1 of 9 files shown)


christophwille:
@ElektroKill Can you please give the PR's action build artefacts a try?

ElektroKill:
I've compiled the build from the source and here are my observations:
- The smoothness of the scrolling is greatly improved
- The scrolling appears to be WAY faster than before (scrolling many more lines of text per scroll operation)
- The number of lines scrolled per scroll operation no longer matches what is configured in Windows scrolling speed settings

I have my Windows settings so that one scroll = 3 lines. Before the smooth scrolling changes it would scroll 3 lines, now it scrolls around 36 lines which is quite an increase. This increase in scrolling affects other components, such as the treeview.

tom-englert:
@ElektroKill can you try the latest version? Should now cover all different scenarios.

ElektroKill:
I took a look at the latest commit and the scrolling in the main text view seems to behave correctly when it comes to how much a single scroll operation is scrolling, however, the smoothness is back to what it used to be before the smooth scrolling changes.

When it comes to scrolling in the Treeview, I can observe some weird jumping and snapping occasionally when the Treeview will jump/scroll much further than expected. This can be observed in the video found below.

Throughout the video, I'm scrolling with the same speed on the scroll wheel. First, the behavior in the latest build from this PR is demonstrated, and later a build from before any of the scrolling changes is shown.
https://www.youtube.com/watch?v=8v7dLZ8LfQ0

tom-englert:
It's now also activated on the test view.

What you call "jumping" and "snapping" is a dynamic acceleration when you turn the mouse wheel very fast, so it's easier to scroll large lists.

ElektroKill:
I can confirm the smooth scrolling is now working in the text view.

Regarding the dynamic acceleration? Would it be possible to add an option in settings to enable/disable it? I'm really not a fan of mouse pointer acceleration and scrolling acceleration as in my opinion it makes things more unpredictable.

tom-englert:
@ElektroKill I just removed it

christophwille:
Thanks for sorting out smooth scrolling!

Issue #3261, opened 2024-08-15
Empty dotnet version folders lead to problems with resolving core assemblies in DotNetCorePathFinder


Herrmel:
### Steps to reproduce
1. Create empty Version directory in C:\Program Files\dotnet\shared\Microsoft.NETCore.App\ that is between the base-version of the target framework of the decompiled assembly (in my case 8.0.0) and the currently installed version (in my case 8.0.8). In my case a version directory for v8.0.2 and v8.0.5 were already there. It seemes they were left over by an update of dotnet.
2. decompile an assembly of mentioned framework, click on References and select a net assembly e.g. System.Runtime
3. Output shows `Error: Could not find reference: System.Runtime, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a`

### Error message shown

![image](https://github.com/user-attachments/assets/f08ee339-aae2-494a-9470-698fe1c3a497)

### Details
* Product in use: ILSpy and ICSharpCode.Decompiler nuget package
* Version in use: ILSpy and Nuget both v8.2.0.7535

Same behaviour happens if I'll try to find System.Runtime with the class DotNetCorePathFinder. If I remove the empty version folder the problem disappears. Of course it is easy to delete said empty folders, but it would be nice to be safe in case future updates of dotnet do the same thing again. I'm not sure why these folders exist for me since the only [issue I found](https://github.com/dotnet/installer/issues/7368) regarding this problem is older and closed but they also mentioned that most consumers were made more resilliant so maybe just no one notices? I think an easy and safe solution is simply to filter empty folders out [on this line](https://github.com/icsharpcode/ILSpy/blob/cb63f3c873cf5721b12f37292c09eb0eb949b198/ICSharpCode.Decompiler/Metadata/DotNetCorePathFinder.cs#L251). What do you think?

siegfriedpammer:
The proposed solution seems like a good idea. Would you be interested in providing the fix in a PR?

PR #3265, opened 2024-08-19
Ignore empty version directories of dotnet

Herrmel:
#3261 

### Problem
Dotnet core assemblies can not be resolved if empty folders of previously installed versions exist.

### Solution
Ignore empty Directories
I changed the result of ConvertToVersion only check for empty directories where the directory is considered a fitting version. I hope that is fine.
In my case the version directories where completly empty but maybe it would make sense to look for at least one file that does have .dll or .exe extension?
I'm not sure how to test things that are environment dependend.

### DIFF
--- a/ICSharpCode.Decompiler/Metadata/DotNetCorePathFinder.cs
+++ b/ICSharpCode.Decompiler/Metadata/DotNetCorePathFinder.cs
@@ -249,21 +249,22 @@ namespace ICSharpCode.Decompiler.Metadata
 		static string GetClosestVersionFolder(string basePath, Version version)
 		{
 			var foundVersions = new DirectoryInfo(basePath).GetDirectories()
-				.Select(d => ConvertToVersion(d.Name))
+				.Select(ConvertToVersion)
 				.Where(v => v.version != null);
 			foreach (var folder in foundVersions.OrderBy(v => v.version))
 			{
 				if (folder.version >= version)
-					return folder.directoryName;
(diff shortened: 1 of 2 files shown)


ICSharpCode.Decompiler/Metadata/DotNetCorePathFinder.cs
```diff
@@ -249,21 +249,22 @@ public string TryResolveDotNetCoreShared(IAssemblyReference name, out string run
...
 		{
 			var foundVersions = new DirectoryInfo(basePath).GetDirectories()
-				.Select(d => ConvertToVersion(d.Name))
+				.Select(ConvertToVersion)
 				.Where(v => v.version != null);
 			foreach (var folder in foundVersions.OrderBy(v => v.version))
 			{
 				if (folder.version >= version)
-					return folder.directoryName;
+					if(folder.directory.EnumerateFiles().Any())
```
siegfriedpammer, line 257: It seems that your changes broke the reference assembly resolver logic that is used by our tests. The structure of reference assembly folder is slightly different. A possible fix would be:

				if (folder.version >= version
					&& folder.directory.EnumerateFiles("*.dll", SearchOption.AllDirectories).Any())
				{
					return folder.directory.Name;
				}

This way we search for libraries in sub-folders as well.

Herrmel: I adjusted this part to your suggestion.

Issue #3221, opened 2024-06-12
ArgumentOutOfRangeException: Index was out of range when clearing the assembly list


provegard:
### Steps to reproduce
1. Have a very long list of assemblies in the Assemblies list.
2. Click `Clear assembly list` in the `File` menu

### Error message shown

```
---------------------------
Sorry, we crashed
---------------------------
System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
   at ICSharpCode.TreeView.LinesRenderer.OnRender(DrawingContext dc)
   at System.Windows.UIElement.Arrange(Rect finalRect)
   at System.Windows.Controls.Grid.ArrangeOverride(Size arrangeSize)
   at System.Windows.FrameworkElement.ArrangeCore(Rect finalRect)
   at System.Windows.UIElement.Arrange(Rect finalRect)
   at System.Windows.Controls.Control.ArrangeOverride(Size arrangeBounds)
   at System.Windows.FrameworkElement.ArrangeCore(Rect finalRect)
   at System.Windows.UIElement.Arrange(Rect finalRect)
   at System.Windows.Controls.Border.ArrangeOverride(Size finalSize)
   at System.Windows.FrameworkElement.ArrangeCore(Rect finalRect)
   at System.Windows.UIElement.Arrange(Rect finalRect)
   at System.Windows.Controls.Border.ArrangeOverride(Size finalSize)
   at System.Windows.FrameworkElement.ArrangeCore(Rect finalRect)
   at System.Windows.UIElement.Arrange(Rect finalRect)
   at System.Windows.Controls.Control.ArrangeOverride(Size arrangeBounds)
   at System.Windows.FrameworkElement.ArrangeCore(Rect finalRect)
   at System.Windows.UIElement.Arrange(Rect finalRect)
   at System.Windows.Controls.VirtualizingStackPanel.ArrangeOtherItemsInExtendedViewport(Boolean isHorizontal, UIElement child, Size childDesiredSize, Double arrangeLength, Int32 index, Rect& rcChild, Size& previousChildSize, Point& previousChildOffset, Int32& previousChildItemIndex)
   at System.Windows.Controls.VirtualizingStackPanel.ArrangeOverride(Size arrangeSize)
   at System.Windows.FrameworkElement.ArrangeCore(Rect finalRect)
   at System.Windows.UIElement.Arrange(Rect finalRect)
[...]

PR #3271, opened 2024-08-26
Fix/issue 3221 argument out of range exception

1samrand:
issue : [](https://github.com/icsharpcode/ILSpy/issues/3221)

### Solution
* Handle out of bound array index

### DIFF
--- a/ICSharpCode.ILSpyX/TreeView/SharpTreeNodeCollection.cs
+++ b/ICSharpCode.ILSpyX/TreeView/SharpTreeNodeCollection.cs
@@ -73,7 +73,10 @@ namespace ICSharpCode.ILSpyX.TreeView
 
 		public SharpTreeNode this[int index] {
 			get {
-				return list[index];
+				if (index >= 0)
+					return list[index];
+				else
+					return null;
 			}
 			set {
 				ThrowOnReentrancy();
--- a/ICSharpCode.ILSpyX/TreeView/SharpTreeNodeCollection.cs
+++ b/ICSharpCode.ILSpyX/TreeView/SharpTreeNodeCollection.cs
@@ -73,7 +73,7 @@ namespace ICSharpCode.ILSpyX.TreeView
 
 		public SharpTreeNode this[int index] {
 			get {
-				if (index >= 0)
+				if (index >= 0 && index < list.Count)
 					return list[index];
 				else
 					return null;
--- a/ILSpy/Controls/TreeView/SharpTreeView.cs
+++ b/ILSpy/Controls/TreeView/SharpTreeView.cs
@@ -561,8 +561,12 @@ namespace ICSharpCode.ILSpy.Controls.TreeView
 			{
 				if (node.IsExpanded && node.Children.Count > 0)
 				{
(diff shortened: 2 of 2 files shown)


christophwille:
This looks like fixing the symptom, but not the root cause - because some code must be picking an out-of-bounds index (maybe a race condition or some other problem). Preferably we should try to debug/find the actual culprit that hands over the out of bounds index.

1samrand:
Thank you for your feedback , I understand your concern about addressing the root cause rather than just the symptom. I'll take some time to investigate further to identify the underlying issue that might be causing the out-of-bounds index

1samrand:
Hi @christophwille,

After further investigation, it seems the issue is related to OnRender from WPF for a disconnected item. In this case, the Node is not null but has a count of 0.

However, as mentioned earlier, I'm still unsure why this is happening. If you have any insights, I would appreciate it if you could share with me. 

The orange area highlights the specific case where this occurs (which I added to check):

image

siegfriedpammer:
```diff
diff --git a/ILSpy/Controls/TreeView/SharpTreeNodeView.cs b/ILSpy/Controls/TreeView/SharpTreeNodeView.cs
index a7cca8838..d144ca5c5 100644
--- a/ILSpy/Controls/TreeView/SharpTreeNodeView.cs
+++ b/ILSpy/Controls/TreeView/SharpTreeNodeView.cs
@@ -67,7 +67,6 @@ static SharpTreeNodeView()
                public override void OnApplyTemplate()
                {
                        base.OnApplyTemplate();
-                       LinesRenderer = Template.FindName("linesRenderer", this) as LinesRenderer;
                        UpdateTemplate();
                }

@@ -148,6 +147,7 @@ void OnIsEditingChanged()

                void UpdateTemplate()
                {
+                       LinesRenderer = Template.FindName("linesRenderer", this) as LinesRenderer;
                        var spacer = Template.FindName("spacer", this) as FrameworkElement;
                        spacer.Width = CalculateIndent();
```

@1samrand I found that this change fixes the issue and is likely the root cause. Can you confirm?

1samrand:
@christophwille I have tested by this change, but the **ArgumentOutOfRangeException** still occurs when accessing list[index].

siegfriedpammer:
How did you test it? I would like to take a look as well. I am following the steps given in the issue and I can no longer reproduce the error.

1samrand:
I have tested the scenario described here: https://github.com/icsharpcode/ILSpy/issues/3221#issuecomment-2310906975. 
It seems that the issue persists, so perhaps there have been some changes in your branch that could be affecting the results.

If you're working on a specific branch, could you please share it with me so I can clone and test it as well?

siegfriedpammer:
Thanks for getting back to me. You are right, the patch I posted above does not fix the issue. Yeah, unfortunately there is no "nice" fix for the weird issues with LinesRenderer.

An idea for a simple non-invasive fix I had previously was:

image

We can either do your second suggestion (the additional sanity check in OnRender) or go with my fix... not sure which is better.

1samrand:
I have tested the changes you made, and I can confirm that the issue no longer occurs. I also prefer to use your solution moving forward

PR #3274, opened 2024-09-03
WPF Refactoring

tom-englert:
- Decouple AssemblyListPane from MainWindow
- Move menu/toolbar logic from MainWindow to separate service.


tom-englert:
@christophwille I think this was the biggest part (maybe bigger than I have expected), moving the business logic out of the views and reducing component dependencies.

Next steps are to get rid of the singletons and replace them with DI, and to migrate the DI framework to Microsoft.Extensions.DependencyInjection

Of course there will always be some obstacles left that will be fixed when stumbling upon them.

christophwille:
Ok, then let's first take on this PR as-is, and do the DI separately next. The PR is way big anyways.

tom-englert:
btw. what was the intention behind the ILSpyX project? Just to make sure I don't break things.

christophwille:
It is code that can be shared between frontends. Eg the VS code extension or the Avalonia based one.

ILSpy/AssemblyTree/AssemblyTreeModel.cs
```diff
@@ -0,0 +1,932 @@
...
+				if (!cmdArgs.NoActivate && window is { WindowState: WindowState.Minimized })
+				{
+					window.WindowState = WindowState.Normal;
+				}
+
+				HandleCommandLineArgumentsAfterShowList(cmdArgs);
+			});
+		}
+
+		public async void NavigateOnLaunch(string navigateTo, string[] activeTreeViewPath, ISettingsProvider spySettings, List<LoadedAssembly> relevantAssemblies)
```
siegfriedpammer, line 184: Please don't use `async void` unless it's absolutely necessary. And, if it's necessary call-sites need a comment hinting at the fact and we need to make sure it does not cause problems.

siegfriedpammer:
@tom-englert thank you very much for your work. However, can we please in the future split the work in smaller PRs? 144 changed files is a bit much... currently I cannot allocate the necessary time to thoroughly review PRs of this size. Thanks!

ILSpy/AssemblyTree/AssemblyTreeModel.cs
tom-englert: Fully agree, however this is not new code, it is just moved from `MainWindow.xaml.cs` an I did not want to introduce too much functional changes in this PR, but just decouple UI and model, so maybe we can go with this and clean it up later since this PR is already huge.

tom-englert:
@siegfriedpammer just take every single commit as an individual PR, maybe then it's easier to follow. (Stacking PRs that depend on each other would not have made it easier, and if I had to wait between related refactorings until parts are reviewed, I would have lost track of what I wanted to do.)

E.g. the first three commits are about moving the tree view logic out of `MainWindow.xaml.cs` into `AssemblyTreeModel.cs` and `MenuService.cs`, which looks like a huge change, but if you compare the old `MainWindow.xaml.cs` with the new  `AssemblyTreeModel.cs` and `MenuService.cs` you will see that there is not too much new code, it just has moved.
And most of the other file changes are only side effects because the path to reach the tree model has changed, so this should be easy to review.

The other commits should be mostly self explanatory by their comment, however if there are still questions just give me some feedback.

tom-englert:
> It is code that can be shared between frontends. Eg the VS code extension or the Avalonia based one. It says so on the tin https://github.com/icsharpcode/ILSpy/blob/master/ICSharpCode.ILSpyX/PackageReadme.md

So when this refactoring is finished, there another big pile of code that no longer depends on WPF and can be moved to ILSpyX

siegfriedpammer:
Minor comment: please add a file/license header to any new files. Thanks!

siegfriedpammer:
I quickly took a look at all your changes and they seem fine. Thank you for doing all this work and sorry again for taking so long to review.

tom-englert:
That's all fine - this was a really big chunk with all the side effects it created.

PR #3285, opened 2024-09-23
Fix #3284: Focus, Tabs and Title issues

### LINKED ISSUE #3284: Focus always moving to the assembly tree view when changing tabs
Using the latest released version the focus behavior was as shown in the following gif:

focus1

Using latest master the behavior is as follows:

focus2

I have yet to determine the change that causes this deviation. For me this is only a minor distraction, because switching between tabs and tree nodes makes the screen flicker a bit more than usual. However, as [we have seen previously](https://github.com/icsharpcode/ILSpy/issues/2577), there are people who depend on accessibility software to use ILSpy, and I can image that this change causes some frustration, because the focus move is unanticipated.

While creating the above screen recordings, I noticed that the label of the assembly tree view content (previously labelled `Assemblies`) is now labelled `ILSpy <version> - <assembly list name>`, which was previously in the window title, not sure if this has some negative impact on accessibility...

/cc @tom-englert

tom-englert:
Fix #3284: 
- Focus always moving to the assembly tree view when changing tabs
- don't decompile when just switching tabs
- AssemblyTree title is wrong

### DIFF
--- a/ILSpy/AssemblyTree/AssemblyListPane.xaml.cs
+++ b/ILSpy/AssemblyTree/AssemblyListPane.xaml.cs
@@ -1,7 +1,6 @@
 ﻿using System.ComponentModel.Composition;
 using System.Windows;
-
-using ICSharpCode.ILSpyX.TreeView;
+using System.Windows.Threading;
 
 using TomsToolbox.Wpf.Composition.Mef;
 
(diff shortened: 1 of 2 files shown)


siegfriedpammer:
Thank you very much!

PR #3292, opened 2024-09-27
Refresh should also refresh the content of the active tab 

tom-englert:
to reflect e.g. indent size change in options tab.


siegfriedpammer:
Thank you very much!

PR #3294, opened 2024-09-28
Fix possible ObservableCollection re-entrance crash when trying to na…

tom-englert:
…vigate to a collapsed item.


siegfriedpammer:
Thank you very much!

siegfriedpammer:
Looking at ILSpy/AssemblyTree/AssemblyTreeModel.cs as a whole I see that we sometimes do a null check (or ?) on `activeView` and sometimes don't. Is this intentional?

tom-englert:
I think that should be always, I'm just not used to work without NRTs enabled 😔

tom-englert:
I'll fix this

siegfriedpammer:
you can enable NRT on any file you touch, if you think it makes work easier... however, please not all files at once... that will be a pain to review and there already is some work done in #3295

PR #3298, opened 2024-10-06
Dev/wpf refactoring

### LINKED ISSUE #3293: Right panel shows old info in case all is deleted in the left-hand tree
### Steps to reproduce
1. Delete all assemblies
2. rihgt panel show old info
3. 

### Error message shown

eg:

image

```
Copy and paste any error messages / relevant text and/or screenshot the incorrect behaviour.
```
### Details
* ILSpy version 9.0.0.7785-preview2+c1a42ccf7205f41d9ba3988926162feb6e57ca98
.NET version 8.0.8+08338fcaa5c9b9a8190abb99222fed12aaba956c

tom-englert:
Fix #3293

### DIFF
--- a/ILSpy/AssemblyTree/AssemblyTreeModel.cs
+++ b/ILSpy/AssemblyTree/AssemblyTreeModel.cs
@@ -716,7 +716,12 @@ namespace ICSharpCode.ILSpy.AssemblyTree
 
 		private void TreeView_SelectionChanged()
 		{
-			if (SelectedItems.Length > 0)
+			if (SelectedItems.Length <= 0)
+			{
+				// To cancel any pending decompilation requests and show an empty tab
+				DecompileSelectedNodes();
+			}
+			else
 			{
 				var activeTabPage = DockWorkspace.Instance.ActiveTabPage;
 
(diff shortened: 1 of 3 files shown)


siegfriedpammer:
I think I fixed the other issue in 916a3fa83... Also I noticed that we are calling `SelectNode(Root)` at https://github.com/icsharpcode/ILSpy/blob/master/ILSpy/AssemblyTree/AssemblyTreeModel.cs#L416

@tom-englert is that something you added or did we do that already in the past? I noticed this, while testing my fix and I can't remember to have ever seen the output of `AssemblyListTreeNode.Decompile` (aka the root) in action :)... Would it make more sense to select the first assembly in the list and show its decompilation?

tom-englert:
I didn't change that, it was the same before:
https://github.com/icsharpcode/ILSpy/blob/cf4f68bf3a4c7d78209d00f4a354dc0adc4eedcc/ILSpy/MainWindow.xaml.cs#L1005-L1014

And yes, selecting the first assembly sounds more reasonable, I'll change that.

tom-englert:
The old code had so many side effects, maybe this one also was obscured by another side effect?

siegfriedpammer:
> The old code had so many side effects, maybe this one also was obscured by another side effect?

Yeah, that's highly likely.

PR #3302, opened 2024-10-10
Move active code from LanguageSettings to LanguageService

tom-englert:
make `LanguageSettings `independent of other services injected via DI

Settings is the first service we need at startup, and it should not depend (even deferred) on any other service for a fast startup

There are two central changes in this PR, `LanguageSettings` and `LanguageService`:
- renamed `Languages` to `LanguageService`
- moved active code from `LanguageSettings` to `LanguageService`

All other changes are just side effects of the renaming/moving.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -147,3 +147,5 @@ dotnet_naming_rule.private_fields_rule.symbols = private_fields_symbols
 
 # MEF006: No importing constructor
 dotnet_diagnostic.MEF006.severity = silent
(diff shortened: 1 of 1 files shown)


siegfriedpammer:
Is it possible to fix or suppress these before we merge the PR? Thanks!

image

siegfriedpammer (approved):

tom-englert:
Sure, done

siegfriedpammer:
Thank you very much!

PR #3314, opened 2024-10-27
Replace singletons with DI patterns...

tom-englert:
... and decouple the direct dependencies of the services.

Again this is only architectural changes in the UI layer, no functional changes.

I tried to break it down into comprehensible steps per commit, however due to the many side effects the first ones are still big, sorry for that.


clean.cmd
```diff
@@ -0,0 +1,22 @@
+@ECHO OFF
```
siegfriedpammer, line 1: what's the purpose of this script? Isn't `git clean` sufficient?

siegfriedpammer:
Thank you for the good work... just a few notes:

- some files are missing the file headers (we definitely need an analyzer for that...)
- can you please fix the warnings reported in the last commit?
- the "fix formatting" commit could probably be merged into other commits that modify the code (if clean.cmd would be removed)

ILSpy/AboutPage.cs
```diff
@@ -44,10 +44,7 @@ public sealed class AboutPageCommand(AssemblyTreeModel assemblyTreeModel) : Simp
 	{
 		public override void Execute(object parameter)
 		{
-			assemblyTreeModel.NavigateTo(
-				new RequestNavigateEventArgs(new Uri("resource://aboutpage"), null),
-				inNewTabPage: true
-			);
+			assemblyTreeModel.NavigateTo(new(new("resource://aboutpage"), null), inNewTabPage: true);
```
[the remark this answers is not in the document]
tom-englert: I'll try to change the .editorconfig defaults to not suggest removing types with the next commit

clean.cmd
tom-englert: Oops, that was not intended to be included, I'll remove it

tom-englert: But git clean does either too much or too less... this one removes exactly the bin and obj folders.

tom-englert: e.g. It would also remove my Resharper configuration which I'm not allowed to commit.

siegfriedpammer: that's why I use `git clean -X`, which removes only files that match .gitignore entries, but does NOT touch uncommitted files. I have lots of uncommitted extra files in my ILSpy folder (test cases, diff files, TODO lists, notes, etc.), which some might argue is a bad thing, but I don't care :)

see https://git-scm.com/docs/git-clean

image

Usually, I combine that with `d` to remove empty directories and `q` for "quiet mode" and `f` is mandatory anyway. So, I use `git clean -Xdfq`.

Just configure your .gitignore correctly and `git clean` behaves exactly as it should.

siegfriedpammer: Just remember to always use upper-case X and you will be fine.

siegfriedpammer: If you think our .gitignore file is missing something, feel free to extend it to match your needs.

siegfriedpammer:
Thank you very much!

PR #3325, opened 2024-11-10
Dev/wpf refactoring

tom-englert:
- Move menu and tool-bar from MainWindow to separate controls.
- Add more defaults to .editorconfig to better match coding style


Directory.Build.props
```diff
@@ -0,0 +1,5 @@
+<Project>
+  <PropertyGroup>
+    <WarningsAsErrors>IDE2000</WarningsAsErrors>
```
christophwille, line 3: Why?

tom-englert: Because double blank lines cause the check in to fail, which is very annoying, so better catch this early

christophwille: But thats because of what you introduced with dotnet_style_allow_multiple_blank_lines_experimental = false:warning? Or does dotnet format even today complain without this setting?

tom-englert: Yes, it failed already when dotnet format was running after check in, it seems to be a hard coded rule

tom-englert: The new rules just better reflect what dotnet format has built in.

christophwille:
Thanks!

PR #3334, opened 2024-11-17
Update NuGet Packages, some to 9.0 outside of ICSD

christophwille:
Not touched (first one is block list anyways)

image


christophwille:
Ok, funny - the branch build failed on release for dotnet format, and the PR build failed for debug. Both however with the same error:

image

christophwille:
Seems R2R plugin has problems with SRM 9

Issue #3337, opened 2024-11-17
Update R2R Addin ILCompiler.Reflection.ReadyToRun.Experimental


christophwille:
Currently using
![image](https://github.com/user-attachments/assets/42b66da1-918e-44cb-a339-1de5af7a2575)
@cshung @eduardo-vp any newer versions published that could be used in net80?

cshung:
Thanks for asking, @christophwille.

It looks like we have some pending work here. We do have some additions in the ReadyToRun area for .NET 9, but they're not implemented in the package yet, @eduardo-vp.

christophwille:
Closing now, as I forgot to close it together with PR #3340

PR #3340, opened 2024-11-25
Update ReadyToRun to use 9.0 package

cshung:
Link to issue(s) this covers #3337 

### Problem
This update improves the disassembly for https://github.com/dotnet/runtime/pull/98623, https://github.com/dotnet/runtime/pull/101761. These PRs added new helper calls to the runtime from ready to run native code, which could appear as opaque call to random addresses if not properly annotated.

### Solution
* The diassembly for methods having calls to these new helper functions will be properly annotated with the call target
* Updated the DummyOutput to match the updated signature so that the code compile with `STRESS` defined in `ReadyToRunLanguage.cs`, and
* Tested the diassembly functionality with the .NET 9 `System.Private.CoreLib.dll` and no error found.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -4,7 +4,7 @@
     <clear />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
     <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
-    <add key="dotnet8-transport" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8-transport/nuget/v3/index.json" />
+    <add key="dotnet9-transport" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet9-transport/nuget/v3/index.json" />
   </packageSources>
   <packageSourceMapping>
     <packageSource key="nuget.org">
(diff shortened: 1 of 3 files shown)


NuGet.config
```diff
@@ -4,7 +4,7 @@
     <clear />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
     <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
-    <add key="dotnet8-transport" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8-transport/nuget/v3/index.json" />
+    <add key="dotnet9-transport" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet9-transport/nuget/v3/index.json" />
```
christophwille, line 7: You also need to fix line 15 for package source mapping

ILSpy.ReadyToRun/ILSpy.ReadyToRun.csproj
```diff
@@ -35,12 +35,12 @@
 
   <ItemGroup>
     <PackageReference Include="Iced" />
-    <PackageReference Include="ILCompiler.Reflection.ReadyToRun.Experimental" />
     <!-- ILCompiler.Reflection.ReadyToRun has dependencies on System.Reflection.Metadata and
```
cshung, line 38: Is this comment still valid?

christophwille: Most likely... #3334 is also open in parallel

PR #3348, opened 2024-12-08
Fix high Dpi, close #3347

### LINKED ISSUE #3347: CustomDialog doesn't display correctly when scale isn't 100%
There's a lot of blank space when windows scale > 100%

100%
image

125%
image

150%
image

175%
image

200%
image

CreateAndInject:
100%
image

125%
image

150%
image

175%
image

200%
image

### DIFF
--- a/ILSpy/Controls/CustomDialog.cs
+++ b/ILSpy/Controls/CustomDialog.cs
@@ -53,7 +53,7 @@ namespace ICSharpCode.ILSpy.Controls
 			{
 				Rectangle screen = Screen.PrimaryScreen.WorkingArea;
 				SizeF size = g.MeasureString(message, label.Font, screen.Width - 20);
-				Size clientSize = size.ToSize();
+				Size clientSize = new Size((int)(size.Width * 96 / g.DpiX), (int)(size.Height * 96 / g.DpiY));
 				Button[] buttons = new Button[buttonLabels.Length];
 				int[] positions = new int[buttonLabels.Length];
 				int pos = 0;
@@ -66,7 +66,7 @@ namespace ICSharpCode.ILSpy.Controls
 					newButton.Text = buttonLabel;
 					newButton.Click += new EventHandler(ButtonClick);
 					SizeF buttonSize = g.MeasureString(buttonLabel, newButton.Font);
-					newButton.Width = Math.Max(newButton.Width, ((int)Math.Ceiling(buttonSize.Width / 8.0) + 1) * 8);
+					newButton.Width = Math.Max(newButton.Width, ((int)Math.Ceiling(buttonSize.Width * 96 / g.DpiX / 8.0) + 1) * 8);
 					positions[i] = pos;
 					buttons[i] = newButton;
 					pos += newButton.Width + 4;
--- a/ILSpy/Controls/CustomDialog.cs
+++ b/ILSpy/Controls/CustomDialog.cs
@@ -51,9 +51,8 @@ namespace ICSharpCode.ILSpy.Controls
 
 			using (Graphics g = this.CreateGraphics())
 			{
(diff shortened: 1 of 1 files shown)


siegfriedpammer:
To me it looks like there are some lines/parts missing if scaling is <200%... Is this something you could fix as well?

christophwille:
I know that might be out of scope for this PR - what-if we modernized the whole dialog to a Textbox with Scrollbars?

CreateAndInject:
> To me it looks like there are some lines/parts missing if scaling is <200%... Is this something you could fix as well?

@siegfriedpammer I move CustomDialog.cs to a new project, test without any changes, this issue doesn't happen on .NET Framework, only happen on .NET Core(even if I use 1920*1080 100%)

So there's no such issue when ILSpy make the code first time, it's exposed when ILSpy migrate to .NET Core.

My test code: [WinFormsApp48.zip](https://github.com/user-attachments/files/18067435/WinFormsApp48.zip)
```
        [STAThread]
        static void Main()
        {
            Application.SetCompatibleTextRenderingDefault(false);
            Application.EnableVisualStyles();
            string message = @"Test

   at ICSharpCode.ILSpy.OpenCommand.OnExecute(Object sender, ExecutedRoutedEventArgs e) in E:\Src\ILSpy\ILSpy\Commands\OpenCommand.cs:line 45
   at System.Windows.Input.CommandBinding.OnExecuted(Object sender, ExecutedRoutedEventArgs e)
   at System.Windows.Input.CommandManager.HandleCommandBinding(Object sender, RoutedEventArgs e, CommandBinding commandBinding, Boolean execute)
   at System.Windows.Input.CommandManager.FindCommandBinding(Object sender, RoutedEventArgs e, ICommand command, Boolean execute)
   at System.Windows.Input.CommandManager.OnExecuted(Object sender, ExecutedRoutedEventArgs e)
   at System.Windows.RoutedEventArgs.InvokeHandler(Delegate handler, Object target)
   at System.Windows.EventRoute.InvokeHandler...";
            string[] buttonTexts = { "Throw", "Debug", "Ignore", "Ignore All" };
            Application.Run(new ICSharpCode.ILSpy.Controls.CustomDialog("Assertion Failed", message, 0, 0, buttonTexts));
        }
```

ElektroKill:
I'd think the best way to address this problem would be to let WPF layout engine handle the placement of controls by using a `Grid` or some panel in which the controls are placed and then let WPF do the rest of the magic. Perhaps the `CustomDialog` could also be migrated to be defined in .xaml?

Moving it to .xaml would also allow the design to be previewed and easily updated

CreateAndInject:
> I'd think the best way to address this problem would be to let WPF layout engine handle the placement of controls by using a `Grid` or some panel in which the controls are placed and then let WPF do the rest of the magic. Perhaps the `CustomDialog` could also be migrated to be defined in .xaml?
> 
> Moving it to .xaml would also allow the design to be previewed and easily updated

I want to know, why this code has such issue on .NET Core when Label.Height = MeasureString's result? There's bug on MeasureString?

CreateAndInject:
I got part of the truth:
This is not .NET Core issue, the reason .NET Core and .NET Framework have different result is because they have different font by default, if we use the same font family and size, .NET Core and .NET Framework have the exact same effect.
Seems the issue is from `MeasureString`, all of you can test different font family and size, sometimes there's enough space, sometimes there miss a line, and sometimes miss half a line.

```c#
                label.Font = new Font(...);
                SizeF size = g.MeasureString(message, label.Font);
```

CreateAndInject:
@siegfriedpammer I got the whole truth: We should use TextRenderer.MeasureText instead of Graphics.MeasureString to get exact result.

100%
image

125%
image

150%
image

175%
image

200%
image

siegfriedpammer:
Thank you very much!

PR #3350, opened 2024-12-17
Fix Mnemonic

CreateAndInject:
If there's `&` in text(eg. ByRef parameter in stack trace), TextRenderer will treat it as Mnemonic by default, this will cause no enough space and the text to wrap automatically.


siegfriedpammer:
Thanks a lot!

PR #3359, opened 2025-01-03
Enable repeatable package restores

christophwille:
https://devblogs.microsoft.com/nuget/enable-repeatable-package-restores-using-a-lock-file/ via https://devblogs.microsoft.com/nuget/openssf-scorecard-for-net-nuget/#pinned-dependencies

Maybe a better idea is to lock certain projects only


christophwille:
See https://github.com/NuGet/Home/issues/8287#issuecomment-1059228563

christophwille:
Locking restricted to projects that create NuGet packages
* ICSharpCode.Decompiler
* ICSharpCode.ILSpyX
* ILSpyCmd
* ICSharpCode.BamlDecompiler

siegfriedpammer (approved):

PR #3370, opened 2025-01-15
Switch back to FluentAssertions 7.0

christophwille:
https://github.com/fluentassertions/fluentassertions/releases/tag/8.0.0

We will need to stay on 7 series or switch to a different library or approach like we did with mocking libraries.


siegfriedpammer:
Definitely needs a comment remind/documenting that we cannot upgrade in the future due to the incompatible license.

PR #3373, opened 2025-01-16
fix #3372 empty tree for assembly

### LINKED ISSUE #3372: Loading special library lead to empty tree for this assembly
### Steps to reproduce
1. load Telerik.Windows.Controls.dll of current years (I tested from 2023)
2. try to expand in `Assemblies` view
3. nothing expands

### Error message shown
none

but with attached Debugger shows:

`System.BadImageFormatException: "Invalid COR20 header signature."` with Stacktrace:
```
 	System.Reflection.Metadata.dll!System.Reflection.Metadata.MetadataReader.ReadMetadataHeader(ref System.Reflection.Metadata.BlobReader memReader, out string versionString)	Unknown
 	System.Reflection.Metadata.dll!System.Reflection.Metadata.MetadataReader.MetadataReader(byte* metadata, int length, System.Reflection.Metadata.MetadataReaderOptions options, System.Reflection.Metadata.MetadataStringDecoder utf8Decoder, object memoryOwner)	Unknown
 	System.Reflection.Metadata.dll!System.Reflection.Metadata.MetadataReaderProvider.GetMetadataReader(System.Reflection.Metadata.MetadataReaderOptions options, System.Reflection.Metadata.MetadataStringDecoder utf8Decoder)	Unknown
	ICSharpCode.Decompiler.dll!ICSharpCode.Decompiler.Metadata.MetadataFile.MetadataFile(ICSharpCode.Decompiler.Metadata.MetadataFile.MetadataFileKind kind, string fileName, System.Reflection.Metadata.MetadataReaderProvider metadata, System.Reflection.Metadata.MetadataReaderOptions metadataOptions, int metadataOffset, bool isEmbedded) Line 234	C#
 	ICSharpCode.ILSpyX.dll!ICSharpCode.ILSpyX.FileLoaders.MetadataFileLoader.Load(string fileName, System.IO.Stream stream, ICSharpCode.ILSpyX.FileLoaders.FileLoadContext settings) Line 39	C#
 	ICSharpCode.ILSpyX.dll!ICSharpCode.ILSpyX.LoadedAssembly.LoadAsync(System.Threading.Tasks.Task<System.IO.Stream> streamTask) Line 339	C#
 	ICSharpCode.ILSpyX.dll!ICSharpCode.ILSpyX.LoadedAssembly..ctor.AnonymousMethod__0() Line 90	C#
```

### Details
* in Version 7.2.0.6702-preview2 it works
* Version 8.2.0.7535 is last released version where it works
* ILSpy shows in front of assmebly name a bluish icon
* the text of the exception comes from `System.Reflection.Metadata.dll`, I replaced with newest, but still failing
* source of exception is probably https://github.com/dotnet/runtime/blob/dde00d376e4b40674488c5c688899d7a343f4eca/src/libraries/System.Reflection.Metadata/src/System/Reflection/Metadata/MetadataReader.cs#L186

trivalik:
fix https://github.com/icsharpcode/ILSpy/issues/3372

### Problem
Loading as Zip-file works for every library and should be the last resort. The regression was in commit 65e30c0bd68924560e45d001caabe41235917057.

### Solution
* in case zip loader was successful, we have to check before proceed with `PEFile`

### DIFF
--- a/ICSharpCode.ILSpyX/LoadedAssembly.cs
+++ b/ICSharpCode.ILSpyX/LoadedAssembly.cs
@@ -353,7 +353,7 @@ namespace ICSharpCode.ILSpyX
 				}
 			}
 
-			if (result?.IsSuccess != true)
+			if (result?.IsSuccess != true || result.Package?.Kind == LoadedPackage.PackageKind.Zip) // workaround to check this before zip
 			{
 				stream.Position = 0;
 				try
@@ -367,7 +367,8 @@ namespace ICSharpCode.ILSpyX
 				}
 				catch (Exception ex)
 				{
-					result = new LoadResult { FileLoadException = ex };
+					if (result == null)
+						result = new LoadResult { FileLoadException = ex };
 				}
 			}
 


ICSharpCode.ILSpyX/LoadedAssembly.cs
```diff
@@ -353,7 +353,7 @@ async Task<LoadResult> LoadAsync(Task<Stream?>? streamTask)
 				}
 			}
 
-			if (result?.IsSuccess != true)
+			if (result?.IsSuccess != true || result.Package?.Kind == LoadedPackage.PackageKind.Zip) // workaround to check this before zip
```
siegfriedpammer, line 356: we cannot do this, because LoadedAssembly should not need to know about any format other than PE.

siegfriedpammer:
Thank you for investigating the bug and providing a fix, anyway.

PR #3380, opened 2025-01-24
Replace `RuntimeHelpers.CreateSpan<T>(LdMemberToken)` with `new T[] { }`

### LINKED ISSUE #3379: Static Initialization for Non-Byte Spans
### Input code

```c#
internal class Class
{
	public static ReadOnlySpan<int> Data => new int[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 };
}
```

### Erroneous output

```c#
internal class Class
{
	public static ReadOnlySpan<int> Data => RuntimeHelpers.CreateSpan<int>((RuntimeFieldHandle)/*OpCode not supported: LdMemberToken*/);
}
```

### Details

* Product in use: ILSpy
* Version in use: `9.0` (the latest release)
* SDK: `9.0.200-preview.0.25057.12` (the latest preview of .NET 9)
* I'm open to contributing a fix, but I may need some guidance on how it should be implemented into the codebase. I understand the IL side of things, but I'm unfamiliar with the structure of ILSpy.

ds5678:
### Problem

Resolves #3379 

### Solution
  * I tried to make it very similar to the code for `Span<T>(void*, int)`.
  * Testing. I'm happy to write additional tests and modify my existing test.

### DIFF
--- a/ICSharpCode.Decompiler/IL/Transforms/TransformArrayInitializers.cs
+++ b/ICSharpCode.Decompiler/IL/Transforms/TransformArrayInitializers.cs
@@ -143,6 +143,40 @@ namespace ICSharpCode.Decompiler.IL.Transforms
 			return false;
 		}
 
+		internal static bool TransformRuntimeHelpersCreateSpanInitialization(Call inst, StatementTransformContext context, out ILInstruction replacement)
+		{
+			replacement = null;
+			if (!context.Settings.ArrayInitializers)
+				return false;
+			if (MatchRuntimeHelpersCreateSpan(inst, context, out var elementType, out var field))
+			{
+				if (field.HasFlag(System.Reflection.FieldAttributes.HasFieldRVA))
+				{
+					var valuesList = new List<ILInstruction>();
+					var initialValue = field.GetInitialValue(context.PEFile, context.TypeSystem);
+					var elementTypeSize = elementType.GetSize();
+					if (elementTypeSize <= 0 || initialValue.Length % elementTypeSize != 0)
+						return false;
+
+					var size = initialValue.Length / elementTypeSize;
+					if (context.Settings.Utf8StringLiterals &&
+						elementType.IsKnownType(KnownTypeCode.Byte) &&
+						DecodeUTF8String(initialValue, size, out string text))
+					{
+						replacement = new LdStrUtf8(text);
+						return true;
+					}
+					if (DecodeArrayInitializer(elementType, initialValue, new[] { size }, valuesList))
+					{
+						var tempStore = context.Function.RegisterVariable(VariableKind.InitializerTarget, new ArrayType(context.TypeSystem, elementType));
+						replacement = BlockFromInitializer(tempStore, elementType, new[] { size }, valuesList.ToArray());
+						return true;
+					}
+				}
+			}
+			return false;
+		}
+
 		private static unsafe bool DecodeUTF8String(BlobReader blob, int size, out string text)
 		{
 			if (size > blob.RemainingBytes)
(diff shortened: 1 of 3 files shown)


ICSharpCode.Decompiler.Tests/TestCases/Correctness/InitializerTests.cs
```diff
@@ -375,6 +375,18 @@ public static void StructInitializerWithinObjectInitializer()
...
+		{
+			return new byte[] { 1, 2, 3 };
+		}
+
+		public static ReadOnlySpan<int> ReadOnlySpanInitializer_Int32Array()
+		{
+			return new int[] { 1, 2, 3 };
+		}
+#endif
+
```
ds5678, line 389: Should I have put these in a new file and modified the options? It currently fails for Roslyn v1.

ICSharpCode.Decompiler/IL/Transforms/TransformArrayInitializers.cs
```diff
@@ -143,6 +143,40 @@ internal static bool TransformSpanTArrayInitialization(NewObj inst, StatementTra
 			return false;
 		}
 
+		internal static bool TransformRuntimeHelpersCreateSpanInitialization(Call inst, StatementTransformContext context, out ILInstruction replacement)
```
ds5678, line 146: Thoughts on the method name?

ICSharpCode.Decompiler/IL/Transforms/TransformArrayInitializers.cs
```diff
@@ -143,6 +143,40 @@ internal static bool TransformSpanTArrayInitialization(NewObj inst, StatementTra
...
+					{
+						replacement = new LdStrUtf8(text);
+						return true;
+					}
+					if (DecodeArrayInitializer(elementType, initialValue, new[] { size }, valuesList))
+					{
+						var tempStore = context.Function.RegisterVariable(VariableKind.InitializerTarget, new ArrayType(context.TypeSystem, elementType));
+						replacement = BlockFromInitializer(tempStore, elementType, new[] { size }, valuesList.ToArray());
+						return true;
+					}
```
ds5678, line 174: This is the same as for `Span<T>(void*, int)` handler. Should split this out into a shared function?

ICSharpCode.Decompiler/IL/Transforms/TransformArrayInitializers.cs
```diff
@@ -187,6 +221,27 @@ static bool MatchSpanTCtorWithPointerAndSize(NewObj newObj, StatementTransformCo
 			return true;
 		}
 
+		static bool MatchRuntimeHelpersCreateSpan(Call inst, StatementTransformContext context, out IType elementType, out FieldDefinition field)
+		{
+			field = default;
+			elementType = null;
+			IType type = inst.Method.DeclaringType;
+			if (type.Namespace != "System.Runtime.CompilerServices" || type.Name != "RuntimeHelpers" || type.TypeParameterCount != 0)
```
ds5678, line 229: Is this the right way to check? Should I have created a "known" type?

ICSharpCode.Decompiler/IL/Transforms/TransformArrayInitializers.cs
```diff
@@ -187,6 +221,27 @@ static bool MatchSpanTCtorWithPointerAndSize(NewObj newObj, StatementTransformCo
...
+				return false;
+			if (inst.Arguments.Count != 1)
+				return false;
+			IMethod method = inst.Method;
+			if (method.Name != "CreateSpan" || method.TypeArguments.Count != 1)
+				return false;
```
ds5678, line 235: Should I be checking the full method signature? I noticed that the other handler doesn't check argument types.

siegfriedpammer: checking more is usually better than checking less... note that the code in that file is very old and thus is not using some of the helpers and is not following the style of later transform implementations

ICSharpCode.Decompiler/IL/Transforms/TransformArrayInitializers.cs
siegfriedpammer: You could do that, yes...

ICSharpCode.Decompiler/IL/Transforms/TransformArrayInitializers.cs
siegfriedpammer: Not really, I am bad at naming things... 😅 ... luckily I had nine months to name my children...

ICSharpCode.Decompiler.Tests/TestCases/Correctness/InitializerTests.cs
siegfriedpammer: Is there a reason why these tests are "correctness" tests and not "pretty" tests?

It seems like this is a relatively new language feature... you can simply use the `CSx` preprocessor symbol that matches the earliest version of C# where this feature is supported.

ICSharpCode.Decompiler/IL/Transforms/TransformArrayInitializers.cs
siegfriedpammer: Addendum: for methods we usually check the name, the number of parameters and then validate all the arguments, so checking the parameter types is usually not necessary.

ds5678:
Sorry, I meant to come back to this. It just slipped my mind.

ds5678:
I appreciate you working on it. If you want, I can finish things up. I saw that the CI failed for the unit test. I'll resolve that.

ds5678:
If there's any other changes you want me to make, let me know.

siegfriedpammer:
I have fixed some more things locally, I will try to re-integrate your recent additions as well...

siegfriedpammer:
Thank you very much for your contribution! Sorry for the trouble!

ds5678:
Thank you as well!

Issue #3343, opened 2024-11-30
Lazy loaded highlight definition failing when clear and add assemblies


sgkoishi:
### Steps to reproduce
It just happened when I cleared the assembly list and added a new one. With default light theme I got this error, and it works if I switch to any other theme and back to light theme again.

### Error message shown
```
ICSharpCode.AvalonEdit.Highlighting.HighlightingDefinitionInvalidException: Error delay-loading highlighting definition
 ---> ICSharpCode.AvalonEdit.Highlighting.HighlightingDefinitionInvalidException: Error at position (line 47, column 26):
There is an unclosed literal string. Line 47, position 26.
 ---> System.Xml.XmlException: There is an unclosed literal string. Line 47, position 26.
   at System.Xml.XmlTextReaderImpl.Throw(Exception e)
   at System.Xml.XmlTextReaderImpl.Throw(String res, String arg)
   at System.Xml.XmlTextReaderImpl.ParseAttributeValueSlow(Int32 curPos, Char quoteChar, NodeData attr)
   at System.Xml.XmlTextReaderImpl.ParseAttributes()
   at System.Xml.XmlTextReaderImpl.ParseElement()
   at System.Xml.XmlTextReaderImpl.ParseElementContent()
   at System.Xml.XmlCharCheckingReader.Read()
   at System.Xml.XmlCharCheckingReader.Read()
   at System.Xml.XsdValidatingReader.Read()
   at ICSharpCode.AvalonEdit.Highlighting.Xshd.V2Loader.ParseElements(ICollection`1 c, XmlReader reader)
   at ICSharpCode.AvalonEdit.Highlighting.Xshd.V2Loader.ParseDefinition(XmlReader reader)
   at ICSharpCode.AvalonEdit.Highlighting.Xshd.V2Loader.LoadDefinition(XmlReader reader, Boolean skipValidation)
   at ICSharpCode.AvalonEdit.Highlighting.Xshd.HighlightingLoader.LoadXshd(XmlReader reader, Boolean skipValidation)
   --- End of inner exception stack trace ---
   at ICSharpCode.AvalonEdit.Highlighting.Xshd.HighlightingLoader.LoadXshd(XmlReader reader, Boolean skipValidation)
   at ICSharpCode.AvalonEdit.Highlighting.Xshd.HighlightingLoader.LoadXshd(XmlReader reader)
[...]

PR #3389, opened 2025-02-07
Load resource stream outside of delegate

snorrk:
Fixes [Lazy loaded highlight definition failing when clear and add assemblies](https://github.com/icsharpcode/ILSpy/issues/3343).

### Problem
Reading from the Manifest Resource Stream inside of the RegisterHighlighting delegate seems to result in malformed reads (possibly due to threading issues).

### Solution
Workaround is to read the stream before registering the delegate.


siegfriedpammer:

Issue #3390, opened 2025-02-09
Consider enabling `<RollForward>major</RollForward>` for `ILSpy.exe`


nil4:
**Is your feature request related to a problem? Please describe.**

The latest release notes state:

> ILSpy 9.x is based on .NET 8.0. All artifacts except the self-contained distribution are built framework-dependent, which means <mark>[.NET 8.0.x or later](https://dotnet.microsoft.com/en-us/download/dotnet/8.0) must be installed</mark> prior to starting ILSpy.

Trying to start `ILSpy.exe` with a later version (e.g. .NET 9) installed, but no .NET 8.x available on the machine, currently fails. Unfortunately, a "*You must install or update .NET to run this application*" dialog appears and the program is blocked from launching.

End-users can work around this by editing the `ILSpy.runtimeconfig.json` file to add:

```diff
{
  "runtimeOptions": {
    "tfm": "net8.0",
+   "rollForward": "major",
```

Doing so allows `ILSpy.exe` to start on .NET 8, 9 or any later version.

**Describe the solution you'd like**

Consider adding `<RollForward>major</RollForward>` to `ILSpy.csproj`  so that future releases do not need this workaround, and ILSpy starts on  .NET 8 **or any later major versions** out of the box.

ref. https://learn.microsoft.com/en-us/dotnet/core/versions/selection#control-roll-forward-behavior

**Additional context**

ref. https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#rollforward

PR #3391, opened 2025-02-09
Enable `<RollForward>major</RollForward>` for `ILSpy.exe`

nil4:
Allow `ILSpy.exe` to launch when .NET 8 (the version it currently targets) is not installed, but some later major version (e.g. .NET 9) is available instead.

ref. https://learn.microsoft.com/en-us/dotnet/core/versions/selection#control-roll-forward-behavior

### Problem

https://github.com/icsharpcode/ILSpy/issues/3390


siegfriedpammer:
Thank you very much!

PR #3394, opened 2025-02-11
Add option to turn off smooth scrolling

### LINKED ISSUE #3393: Turn off smooth scrolling
I _heavily_ dislike smooth scrolling and there's seemingly no way to turn it off.

Add an option to turn it off to turn it back into fast and snappy scrolling.

**Additional context**
Scrolling seems extremely slow, almost like a bug. I'm scrolling really fast here and it's extremely sluggish. It may be hard to see in the video but you can definitely feel it. It feels like it has to scroll to a specific position before allowing any more scrolling. 

https://github.com/user-attachments/assets/07ca42ac-6eff-4685-873c-5456dc84a72e

Hertzole:
Fixes #3393

### Problem
Smooth scrolling was recently added and not everyone enjoys that.

### Solution
Add an option to toggle smooth scrolling. I added it under Display -> Other options
Smooth scrolling option

The option applies to everywhere ScrollViewer is used and ZoomScrollView, which seems to just be the options dialog and the main text editor.

~~This PR is currently blocked due to needing a bug fix in TomsToolbox https://github.com/tom-englert/TomsToolbox/pull/20, hence the draft~~

### DIFF
--- a/ILSpy/App.xaml
+++ b/ILSpy/App.xaml
@@ -3,7 +3,8 @@
 			 xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
 			 xmlns:styles="urn:TomsToolbox.Wpf.Styles"
 			 xmlns:toms="urn:TomsToolbox"
-			 xmlns:themes="clr-namespace:ICSharpCode.ILSpy.Themes">
+			 xmlns:themes="clr-namespace:ICSharpCode.ILSpy.Themes"
+			 xmlns:ilSpy="clr-namespace:ICSharpCode.ILSpy">
 	<Application.Resources>
 		<Style x:Key="DialogWindow" TargetType="{x:Type Window}">
 			<Setter Property="ShowInTaskbar" Value="False" />
@@ -25,6 +26,12 @@
 
 		<Style TargetType="ScrollViewer">
 			<Setter Property="toms:AdvancedScrollWheelBehavior.Attach" Value="WithAnimation" />
+			<Style.Triggers>
+				<!-- Allow for disabling smooth scrolling -->
+				<DataTrigger Binding="{Binding Path=DisplaySettings.EnableSmoothScrolling, Source={x:Static ilSpy:App.SettingsService}}" Value="False">
+					<Setter Property="toms:AdvancedScrollWheelBehavior.Attach" Value="None"/>
+				</DataTrigger>
+			</Style.Triggers>
 		</Style>
 
 	</Application.Resources>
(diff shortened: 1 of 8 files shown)


ILSpy/App.xaml.cs
```diff
@@ -59,6 +59,10 @@ public partial class App : Application
 
 		public static IExportProvider ExportProvider { get; private set; }
 
+		private readonly SettingsService settingsService;
+		
+		public static SettingsService SettingsService => Current.settingsService;
```
siegfriedpammer, line 64: I am not a fan of new static state... @tom-englert what would be the current WPF best practice to solve this problem? Given that you have refactored pretty much everything UI and DI-related in ILSpy to follow best practices. Thank you very much!

ILSpy/App.xaml
```diff
@@ -25,6 +26,12 @@
 
 		<Style TargetType="ScrollViewer">
 			<Setter Property="toms:AdvancedScrollWheelBehavior.Attach" Value="WithAnimation" />
+			<Style.Triggers>
```
tom-englert, line 29: As written in the documentation, `Attach` is a shortcut and should only be used if you want to go with the default only.
If you want to control the behavior, you should not attach and detach it, but use the full behavior notation and simply configure the properties of interest. (now I understand the intend of https://github.com/tom-englert/TomsToolbox/pull/20, this was not necessary at all)

ILSpy/App.xaml.cs
tom-englert: Yes, settings must only be used via DI.
@Hertzole, @siegfriedpammer a fixed version is here: https://github.com/icsharpcode/ILSpy/pull/3405. Either adapt this PR to match #3405 and abandon #3405, or abandon this and use #3405, whatever you prefer.

tom-englert (changes requested):

ILSpy/App.xaml
Hertzole: I was not aware of the method you used in your implementation, hence this rudimentary implementation.

Hertzole:
I'd say to go with #3405 instead of this since it's already fixed and Tom seems to know more about this than I do. I'm just glad there will be an option to turn off smooth scrolling. :)

ILSpy/App.xaml
tom-englert: Agree this is a bit hidden, in xaml VS does not show the documentation directly when hovering...

ILSpy/App.xaml.cs
siegfriedpammer: @tom-englert thank you very much for your fast response and taking the time to look into this.

PR #3404, opened 2025-02-20
Fix #3382: Support compiler-generated throw-helper invocations in switch-expression implicit default-case.

### LINKED ISSUE #3382: Invalid decompilation of switch expression with no default case
### Input code
```cs
using System;
class IncompleteSwitchExpressionTest
{
    static int Test(StringComparison c)
    {
        return c switch {
            StringComparison.Ordinal => 0,
            StringComparison.OrdinalIgnoreCase => 1,
        };
    }
}
```

### Erroneous output
```cs
internal class IncompleteSwitchExpressionTest
{
	private static int Test(StringComparison c)
	{
		switch (c)
		{
		case StringComparison.Ordinal:
			return 0;
		case StringComparison.OrdinalIgnoreCase:
			return 1;
		default:
		{
			global::<PrivateImplementationDetails>.ThrowSwitchExpressionException(c);

			int result = default(int);

			return result;
		}
		}
	}
}
```

When a switch expression does not specify a default case the compiler will add one that throws a `SwitchExpressionException` or a `InvalidOperationException` on older runtimes.
The decompiled code cannot recompile as `<PrivateImplementationDetails>` is not a valid identifier.

### Details
* Product in use: ILSpy 9.0 VS extension

siegfriedpammer:
### Solution

### DIFF
--- a/ICSharpCode.Decompiler.Tests/PrettyTestRunner.cs
+++ b/ICSharpCode.Decompiler.Tests/PrettyTestRunner.cs
@@ -141,6 +141,20 @@ namespace ICSharpCode.Decompiler.Tests
 			CompilerOptions.Optimize | CompilerOptions.UseRoslynLatest,
 		});
 
+		static readonly CompilerOptions[] roslyn3OrNewerWithNet40Roslyn4Options = Tester.SupportedOnCurrentPlatform(new[]
+		{
+			CompilerOptions.UseRoslyn4_14_0 | CompilerOptions.TargetNet40,
+			CompilerOptions.Optimize | CompilerOptions.UseRoslyn4_14_0 | CompilerOptions.TargetNet40,
+			CompilerOptions.UseRoslynLatest | CompilerOptions.TargetNet40,
+			CompilerOptions.Optimize | CompilerOptions.UseRoslynLatest | CompilerOptions.TargetNet40,
+			CompilerOptions.UseRoslyn3_11_0,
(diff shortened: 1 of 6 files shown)


ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
```diff
@@ -106,6 +106,7 @@ public static List<IILTransform> GetILTransforms()
 				new SwitchDetection(),
 				new SwitchOnStringTransform(),
 				new SwitchOnNullableTransform(),
+				new SwitchExpressionDefaultCaseTransform(),
```
siegfriedpammer, line 109: I think that doing this as a post-processing transform of switch detection is a problem. At this point there is no separate block container for `switch` (`BlockContainerKind.Switch` is not yet set)... this transformation only happens later in `LoopDetection`. So this transform would have to perform all the validation that happens in the other `switch` detection transforms again.

ICSharpCode.Decompiler.Tests/PrettyTestRunner.cs
```diff
@@ -141,6 +141,25 @@ public void AllFilesHaveTests()
 			CompilerOptions.Optimize | CompilerOptions.UseRoslynLatest,
 		});
 
+		// Roslyn 3.x and Roslyn 4.x or newer, with TargetNet40 variants only for Roslyn 4.x
```
siegfriedpammer: this comment is in the wrong place, it should go next to the test

siegfriedpammer (commented):
needs changes

PR #3413, opened 2025-02-25
Use type hint in pointer arithmetic

### LINKED ISSUE #3411: Pointer Arithmetic Prettiness

I'm decompiling code that uses lots of pointer casts. Although the decompiled expressions using `byte*` are functionally correct, they're not very pretty.

Source:

```cs
public static class C
{
    public unsafe static void Assignment(void* ptr)
    {
        ((int*)ptr)[2] = 1;
    }

    public unsafe static int Access(void* ptr)
    {
        return ((int*)ptr)[2];
    }

    public unsafe static void Assignment(long* ptr)
    {
        ((int*)ptr)[2] = 1;
    }

    public unsafe static int Access(long* ptr)
    {
        return ((int*)ptr)[2];
    }

    public unsafe static void Assignment(int* ptr)
    {
        ptr[2] = 1;
    }

    public unsafe static int Access(int* ptr)
    {
        return ptr[2];
    }
}
```

Decompiled:

```cs
public static class C
{
    public unsafe static void Assignment(void* ptr)
    {
        *(int*)((byte*)ptr + (nint)2 * (nint)4) = 1;
    }

    public unsafe static int Access(void* ptr)
    {
        return *(int*)((byte*)ptr + (nint)2 * (nint)4);
    }

    public unsafe static void Assignment(long* ptr)
    {
        *(int*)((byte*)ptr + (nint)2 * (nint)4) = 1;
    }

    public unsafe static int Access(long* ptr)
    {
        return *(int*)((byte*)ptr + (nint)2 * (nint)4);
    }

    public unsafe static void Assignment(int* ptr)
    {
        ptr[2] = 1;
    }

    public unsafe static int Access(int* ptr)
    {
        return ptr[2];
    }
}
```

https://sharplab.io/#v2:CYLg1APgAgzABFAjANgQJjgYQLACgDeecxC8ArgHYDOAhgGYCmCKCALHAIJVUCWA5hQC2DCgBcAFFFYAqOAAdRAJwCUREoVwktcceJ5jpyhSoDaaALpwAvHEQBuNcQC+eR6TiVajZqn2jOAMYBDNySMvJKqprqblpQAOw6egZGkWbmDtHOrlnunvRMSKhSnNz8QiISADYA9hR8ssZRWhraJLp+hk3p1raZWi64brAe1AU+cH6BwaG19Y2Rbq1tCIkdKd0W/SSDw+Rj3kVspbwCwmLJogsqS7Ekxj029m67uSP5hyxTHEEhVJfXZoxXJxRIPLYvPBOIA=

Ideally, ILSpy could detect this pattern and decompile the code in a more pretty way.

**Additional context**

I'm open to working on this. I think it can probably be fixed somewhere around here:

https://github.com/icsharpcode/ILSpy/blob/e94eb917395e50c74ea363a79df3514538989daa/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs#L1252-L1253

ds5678:
Resolves #3411 

### Problem

This makes pointer arithmetic prettier with the help of available type hints.

### Solution
  * I tried to be as surgical as possible.
  * There was a slight behavior change involving `sbyte*` vs `byte*`. I want to know if that's okay or not. Previously, `byte*` got picked because it's the default when things go wrong. However, now the type hint (which is often signed) prevents some of those situations.
  * I think the type hints could be better if they had sign information in more cases, but I was hesitant to change the code that sets the type hint.

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
@@ -1220,7 +1220,7 @@ namespace ICSharpCode.Decompiler.CSharp
 		/// Returns null if 'inst' is not performing pointer arithmetic.
 		/// 'ptr - ptr' is not handled here, but in HandlePointerSubtraction()!
 		/// </summary>
-		TranslatedExpression? HandlePointerArithmetic(BinaryNumericInstruction inst, TranslatedExpression left, TranslatedExpression right)
+		TranslatedExpression? HandlePointerArithmetic(BinaryNumericInstruction inst, TranslatedExpression left, TranslatedExpression right, TranslationContext context)
 		{
 			if (!(inst.Operator == BinaryNumericOperator.Add || inst.Operator == BinaryNumericOperator.Sub))
 				return null;
@@ -1248,6 +1248,20 @@ namespace ICSharpCode.Decompiler.CSharp
 			else
 			{
 				return null;
+			}
+			if (context.TypeHint.Kind == TypeKind.Pointer)
+			{
+				// We use the type hint if one of the following is true:
+				// * The current element type is a non-primitive struct.
+				// * The current element type has a different size than the type hint element type.
+				// This prevents the type hint from overriding in undesirable situations (eg changing char* to short*).
+
+				var typeHint = (PointerType)context.TypeHint;
+				int elementTypeSize = pointerType.ElementType.GetSize();
+				if (elementTypeSize == 0 || typeHint.ElementType.GetSize() != elementTypeSize)
+				{
+					pointerType = typeHint;
+				}
 			}
 			TranslatedExpression offsetExpr = GetPointerArithmeticOffset(byteOffsetInst, byteOffsetExpr, pointerType.ElementType, inst.CheckForOverflow)
 				?? FallBackToBytePointer();
@@ -1534,7 +1548,7 @@ namespace ICSharpCode.Decompiler.CSharp
 			}
 			if (left.Type.Kind == TypeKind.Pointer || right.Type.Kind == TypeKind.Pointer)
 			{
-				var ptrResult = HandlePointerArithmetic(inst, left, right);
+				var ptrResult = HandlePointerArithmetic(inst, left, right, context);
 				if (ptrResult != null)
 					return ptrResult.Value;
 			}
--- a/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
@@ -1249,6 +1249,7 @@ namespace ICSharpCode.Decompiler.CSharp
 			{
 				return null;
 			}
+			TranslatedExpression? offsetExpressionFromTypeHint = null;
 			if (context.TypeHint.Kind == TypeKind.Pointer)
 			{
 				// We use the type hint if one of the following is true:
@@ -1258,13 +1259,17 @@ namespace ICSharpCode.Decompiler.CSharp
 
 				var typeHint = (PointerType)context.TypeHint;
 				int elementTypeSize = pointerType.ElementType.GetSize();
-				if ((elementTypeSize == 0 || typeHint.ElementType.GetSize() != elementTypeSize)
-					&& GetPointerArithmeticOffset(byteOffsetInst, byteOffsetExpr, typeHint.ElementType, inst.CheckForOverflow) != null)
+				if (elementTypeSize == 0 || typeHint.ElementType.GetSize() != elementTypeSize)
 				{
-					pointerType = typeHint;
+					offsetExpressionFromTypeHint = GetPointerArithmeticOffset(byteOffsetInst, byteOffsetExpr, typeHint.ElementType, inst.CheckForOverflow);
+					if (offsetExpressionFromTypeHint != null)
+					{
+						pointerType = typeHint;
+					}
 				}
 			}
-			TranslatedExpression offsetExpr = GetPointerArithmeticOffset(byteOffsetInst, byteOffsetExpr, pointerType.ElementType, inst.CheckForOverflow)
+			TranslatedExpression offsetExpr = offsetExpressionFromTypeHint
+				?? GetPointerArithmeticOffset(byteOffsetInst, byteOffsetExpr, pointerType.ElementType, inst.CheckForOverflow)
 				?? FallBackToBytePointer();
 
 			if (left.Type.Kind == TypeKind.Pointer)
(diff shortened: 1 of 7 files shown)


ElektroKill:
Could we have an option for this? I think it's a matter of preference which version of the decompilation is preferred (using only byte offsets or the offsets using the size of the type of the pointer), and it could be something that is configurable.

ds5678:
An option for what specifically?

* My change (using expression result type hints to influence code generation of pointer arithmetic)
* Assuming all pointers are `byte*`
* Something else?

ElektroKill:
> An option for what specifically?
> 
>     * My change (using expression result type hints to influence code generation of pointer arithmetic)
> 
>     * Assuming all pointers are `byte*`
> 
>     * Something else?

Your change makes the decompiler change the pointer arithmetic to be based on the size of the pointer type, while previously, it used to use byte offsets and then cast it to the necessary type pointer. I suggest an option to make this new behavior a toggleable option since, in my opinion, it's really a matter of preference.

TLDR: Make the change in PR toggleable with a deecompiler option.

ds5678:
I don't see why this:

```cs
*(int*)((byte*)ptr + (nint)2 * (nint)4) = 1;
```

would be preferable to:

```cs
((int*)ptr)[2] = 1;
```

ElektroKill:
> I don't see why this:
> 
> ```cs
> *(int*)((byte*)ptr + (nint)2 * (nint)4) = 1;
> ```
> 
> would be preferable to:
> 
> ```cs
> ((int*)ptr)[2] = 1;
> ```

Yes, this is clearer. However, this is not what initially sparked my suggestion for the option to be added.

What made me suggest the option was the changes that occurred in this test as a effect of the change:
https://github.com/icsharpcode/ILSpy/pull/3413/files#diff-efb55f39bb23d80b28a5b409a98149cef496df7c6cf653dbdf26f22dc6f46226L225-R226

The lines:
```cs
*(int*)(num - 4) = 2;
*(int*)(num - 8) = 3;
```
got changed to:
```cs
*((int*)num - 1) = 2;
*((int*)num - 2) = 3;
```
In my opinion, the code decompilation before the change was clearer as we are working on a `byte*` pointer in this context, and it would make more sense to do the pointer math on the byte pointer rather than first casting the pointer.

ds5678:
While there is room for tweaking the rules about when to use the type hint instead of the natural type, I disagree that it carries enough importance to be elevated to a configuration option.

I have a few suggestions for a rule change:

* When the natural type is `byte*`, ignore the type hint.
* When the natural type and type hint are both numeric primitives, ignore the type hint if the natural type is the same size or smaller.
* When the natural type and type hint are both numeric primitives, ignore the type hint.
* When the natural type and type hint are both numeric primitives, ignore the type hint unless one is integral and the other is floating point.

What are your thoughts?

ElektroKill:
> While there is room for tweaking the rules about when to use the type hint instead of the natural type, I disagree that it carries enough importance to be elevated to a configuration option.
> 
> I have a few suggestions for a rule change:
> 
>     * When the natural type is `byte*`, ignore the type hint.
> 
>     * When the natural type and type hint are both numeric primitives, ignore the type hint if the natural type is the same size or smaller.
> 
>     * When the natural type and type hint are both numeric primitives, ignore the type hint.
> 
>     * When the natural type and type hint are both numeric primitives, ignore the type hint unless one is integral and the other is floating point.
> 
> 
> What are your thoughts?

I think tweaking the output of pointer arithmetic is indeed the best option. In my opinion something like this would be good:
* If the pointer type is `byte*` and the type of the pointer operation is a primitive, we should keep the pointer arithmetic on `byte*` pointer.
* If the pointer is `void*`, then we should just use the type of the pointer operation since we can't do arithmetic on `void*`.
* If the pointer is of a primitive type and the pointer operation is of the same type, or of the same size but different sign, I think using the pointer type is best.
* if the pointer operation type is different than the pointer type, then we use the operation type for arithmetic.

Let me know what you think!

ds5678:
> When the natural type is `byte*`, ignore the type hint.

It sounds like you prefer this modification to the current rules. The rest of your bullet points are already part of the rule set.

ds5678:
Actually, I'm not sure that's the best change. It results in your desired outcome.

```cs
// Source and decompiled the same
*(int*)(ptr - 8) = 3;
```

However, it also makes this code decompile poorly.

```cs
// Source
*((int*)ptr - 2) = 3;

// Decompiled
*(int*)(ptr - (nint)2 * (nint)4) = 3;

**Comment by ElektroKill** [comment]
Oh, I'd expect the decompiler here to simplify the multiplication there, considering it already does simplify some simple arithmetic when converting the ILAst to C#. I think this is because of the `nint` cast coming from the `conv.i` instruction inserted by the C# compiler. I'm puzzled as to why the decompiler chooses to maintain these casts since they are not changing the behavior.
These two are equivalent.
```cs
*(int*)(ptr - (nint)2 * (nint)4) = 3;
```
```cs
*(int*)(ptr - 8) = 3;
```

There really is no way to decompile these kinds of expressions the same way they are written, and we just need to find the best compromise.

siegfriedpammer:
@ds5678 I don't want to comment on the changes in this PR, because it's @dgrunwald's area of expertise. Thank you for your understanding...

ds5678:
> I don't want to comment on the changes in this PR, because it's dgrunwald's area of expertise.

I notice that he doesn't seem to be active on GitHub. To set proper expectations, how long should we wait for his review before moving forward?

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -1249,6 +1249,21 @@ protected internal override TranslatedExpression VisitBinaryNumericInstruction(B
...
+			{
+				// We use the type hint if one of the following is true:
+				// * The current element type is a non-primitive struct.
+				// * The current element type has a different size than the type hint element type.
+				// This prevents the type hint from overriding in undesirable situations (eg changing char* to short*).
+
+				var typeHint = (PointerType)context.TypeHint;
+				int elementTypeSize = pointerType.ElementType.GetSize();
+				if ((elementTypeSize == 0 || typeHint.ElementType.GetSize() != elementTypeSize)
+					&& GetPointerArithmeticOffset(byteOffsetInst, byteOffsetExpr, typeHint.ElementType, inst.CheckForOverflow) != null)
```
dgrunwald, line 1262: The repeated call to `GetPointerArithmeticOffset` is a bit unfortunate, as it includes the recursive call to `Translate` for the offset instruction -> potential for exponential runtime (in the unlikely case that the offsetExpr contains further pointer arithmetic in a nested expression).

I think a call to `PointerArithmeticOffset.Detect` would be sufficient here, as `GetPointerArithmeticOffset` returns null iff the detection returned null.

Alternatively you could use the `TranslatedExpression` returned here without calling `GetPointerArithmeticOffset` again.

dgrunwald (commented):
I think this change looks good except for the small detail with the repeated call.

ds5678:
Thank you for the review! I implemented the change you requested.

PR #3418, opened 2025-03-04
Fix exception when writing resx files without adding any resources

jwfx:
### Problem

Version [9.0.0.7889](https://www.nuget.org/packages/ICSharpCode.Decompiler/9.0.0.7889) throws a NullReferenceException when trying to write a resx file without calling AddResource() due to the uninitialized writer.

### Solution
Initialize the writer during Generate, in case it was not initialized yet.


siegfriedpammer:
Thank you very much!

jwfx:
> Thank you very much!

Thanks for the quick reaction time on review and merge, much appreciated. 👍

PR #3427, opened 2025-03-16
log full instruction on onsupported OpCode

zed-0xff:
provides more context on analyzing obfuscated binaries for instance:
before:
```
((delegate*<Array, RuntimeFieldHandle, void>)Class195.methods[1])(array, (RuntimeFieldHandle)
/*OpCode not supported: LdMemberToken*/);
```

after:
```
((delegate*<Array, RuntimeFieldHandle, void>)Class195.methods[1])(array, (RuntimeFieldHandle)
/*OpCode not supported: ldmembertoken method_5678 at IL_000d*/);
```

PS sorry I don't have VS to run tests `ICSharpCode.Decompiler.Tests/TestCases/Ugly/NoArrayInitializers.Expected.cs` should be fixed too.

Link to issue(s) this covers

### Problem
Link to, or brief information about the issue

### Solution

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
@@ -4815,7 +4815,7 @@ namespace ICSharpCode.Decompiler.CSharp
 
 		protected override TranslatedExpression Default(ILInstruction inst, TranslationContext context)
 		{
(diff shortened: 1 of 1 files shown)


siegfriedpammer:
You don't need VS to run the tests, `dotnet test ILSpy.sln` should work.

PR #3435, opened 2025-03-20
PSCmdLet to output target framework for an assembly

snorrk:
### Problem
Sometimes it is helpful to validate that all the assemblies in a directory are targeting the expected framework(s).

### Solution
Adding a simple PS cmdlet to output the target framework for an assembly using PowerShell.


siegfriedpammer:
I am not exactly sure whether you want the extra behavior of `DetectTargetFrameworkId`. It does some extra inference/guessing, if no TargetFrameworkAttribute is present in the binary. The ultimate fallback would be .NET Framework.

Do you want all the extra logic or just "give me what's embedded in the binary"?

Issue #3438, opened 2025-03-28
Wrong ref expression : CS8374


CreateAndInject:
Original:
```C#
using System;
using System.Collections.Generic;

namespace Lsj.Util.Collections
{
    public static class ArrayHelper
    {
        public static (T value, int row, int col)[] ToThreeValueTuples<T>(this T[][] array)
        {
            var result = new List<ValueTuple<T, int, int>>();
            for (int i = 0; i < array.Length; i++)
            {
                for (int j = 0; j < array[i].Length; j++)
                {
                    if (!array[i][j].Equals(default(T)))
                        result.Add((array[i][j], i, j));
                }
            }
            return result.ToArray();
        }
    }
}

```

ILSpy:
```C#
	public static (T value, int row, int col)[] ToThreeValueTuples<T>(this T[][] array)
	{
		List<(T, int, int)> result = new List<(T, int, int)>();
		for (int i = 0; i < array.Length; i++)
		{
			for (int j = 0; j < array[i].Length; j++)
			{
				ref readonly T reference = ref array[i][j];
				T val = default(T);
				if (val == null)
				{
					val = reference;
					reference = ref val;
				}
				if (!reference.Equals(default(T)))
				{
					result.Add((array[i][j], i, j));
				}
			}
		}
		return result.ToArray();
	}
```

greenozon:
is it somehow similar to this ?  https://github.com/icsharpcode/ILSpy/issues/3419

CreateAndInject:
> is it somehow similar to this ? [#3419](https://github.com/icsharpcode/ILSpy/issues/3419)

Yes, it's the same issue, but now it can be reproduced from source file rather than binary file.

CreateAndInject:
2 more samples:
```C#
      public static void DirectInsertionSort<T>(this IList<T> list) where T : IComparable
      {
          var length = list.Count;
          for (int i = 1; i < length; i++)
          {
              var currentvalue = list[i];
              int j;
              for (j = i - 1; currentvalue.CompareTo(list[j]) < 0 && j > 0; j--)
              {
                  list[j] = list[j - 1];
              }
              if (j + 1 != i)
              {
                  list[j + 1] = currentvalue;
              }
          }
      }
```

```C#
	[ComVisible(true)]
	public class PriorityQueue<T> where T : IComparable<T>
	{
		private T[] array_0;

		private int int_0 = 7;

		private int int_1 = 8;

		public int Count { get; private set; }

[...]

Issue #3419, opened 2025-03-05
Error CS8374	Cannot ref-assign 'val' to 'reference' because 'val' has a narrower escape scope than 'reference'


greenozon:
### Input code
Can't compile decompiled code

### Erroneous output
```c#
		public static int CountDifferences<T>(T[] enumerable1, T[] enumerable2)
		{
			if (enumerable1 == null)
			{
				throw new ArgumentNullException("enumerable1");
			}
			if (enumerable2 == null)
			{
				throw new ArgumentNullException("enumerable2");
			}
			if (enumerable1.Length != enumerable2.Length)
			{
				throw new ArgumentException("Sizes of arrays are different");
			}
			int num = 0;
			for (int i = 0; i < enumerable1.Length; i++)
			{
				ref readonly T reference = ref enumerable1[i];
				T val = default(T);
				if (val == null)
				{
					val = reference;
					reference = ref val;
				}
				if (!reference.Equals(enumerable2[i]))
				{
					num++;
				}
			}
			return num;
		}
```

![Image](https://github.com/user-attachments/assets/45ab461f-e92b-4716-a08f-3237821f90e6)

### Details
* ILSpy version 9.0.0.7915+53522c45f887c4dc8f490815f40a724e4e9e1559
.NET version 8.0.11+9cb3b725e3ad2b57ddc9fb2dd48d2d170563a8f5

greenozon:
as a quick fix I made this (is it OK?)

```
if (!enumerable1[i].Equals(enumerable2[i]))
				{
					num++;
				}
```

PR #3440, opened 2025-03-29
Fix changes due to eval order fix

siegfriedpammer:
Reason: Recent changes in the runtime as well as Roslyn fixing an eval order bug of constrained calls.

Link to issue(s) this covers
#3438 
#3419

I will clean up the PR before merge.


siegfriedpammer:
@dgrunwald I was able to fix most of the problems. However, there are a few test cases still failing because of weird codegen with redundant assignments such as `val = val;` and I would like your opinion/guidance on how to approach those. Thanks!

PR #3444, opened 2025-04-01
Support params keyword on non-array collections

### LINKED ISSUE #3431: Params keyword on non-array collections
https://github.com/dotnet/csharplang/issues/7700

I don't feel confident about implementing full support for this feature, but I think I could ensure that method parameters get correctly marked as `params`.

ds5678:
Resolves #3431 

### Problem

This adds support for showing the `params` keyword on non-array collection parameters. It should not affect call sites.

### Solution
  * I'm not sure if this should be part of `TypeSystemOptions`, but that seemed consistent.
  * I'm not certain that I did language versioning correct.


ds5678:
Apparently, the CI didn't like that I bumped the language version to 13. I'm not sure how to fix that.

siegfriedpammer:
Don't worry about the build fail... We will take care of the infrastructure for the new language version

christophwille:
For reference - this happens regularly #2873

christophwille:
I only had a cursory look comparing #2873 to your changes - global.json and ICSharpCode.Decompiler.TestRunner.csproj actually should be set to net90. Also, Helpers/Tester.cs then needs different output paths. Maybe if you want to implement that transition for tests to net90 it would be best to compare side-by-side with that old PR.

christophwille:
I just saw the break for dotnet format; that didn't bomb back then with the move to net70. 

Maybe this applies then here https://github.com/icsharpcode/ILSpy/pull/3238/files

https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet9/nuget/v3/index.json exists with a version of 9.0.520307

ds5678:
Bumping to the .NET 9 SDK also caused some issues with tests now failing. It changed how some code is compiled (eg overload resolution priority). I'm not sure how to fix that, other than to implement support for the features it's using.

christophwille:
@siegfriedpammer oh, that is even more involved then. We need to have a look.

siegfriedpammer:
> I'm not sure how to fix that, other than to implement support for the features it's using.

Yeah, it's similar to when we upgrade to a newer version of Roslyn. We need to update the decompiler engine in lots of different places.

I think it would be a good idea to do the dotnet upgrade in a separate PR before anything else. 

Unfortunately, I am blocked by family obligations due to the holidays, so I won't have time to look at this soon.

christophwille:
Update @ds5678 - @siegfriedpammer and I had a longer deliberation on how to tackle the net90 and upgrade. We came to the conclusion it is better to right away go net10. We'll keep you posted once that has happened (shouldn't be too far out)

ds5678:
Thanks for the status update!

siegfriedpammer:
@ds5678 Just to let you know, the decompiler is missing some smaller transforms related to InlineArray and Span<T>, which are used extensively in net10, blocking the net10 upgrade currently.

This will take some more time to resolve. I will create a branch/PR in the next few days to make progress more visible. Also, if you want to help, we can discuss further steps then.

siegfriedpammer:
So, I have spent some time looking into the whole thing and here's what I found:

The .NET 10 upgrade fails due to missing support for

* Some inlining/variable splitting special cases
* InlineArrays
* Collection Expressions (*)

(*) some of the special cases mentioned in the first point can only be generally represented using Collection Expressions...

ds5678:
I think this was superseded by #3500.

PR #3449, opened 2025-04-02
Fix nested namespace directories

### LINKED ISSUE #3448: Regression: `UseNestedDirectoriesForNamespaces` doesn't work anymore
### Steps to reproduce
1. Open ILSpy.
2. Enable UseNestedDirectoriesForNamespaces in the options.
3. Save an assembly's code.

### Issue

`List<T>` will be in `System.Collections.Generic/`, rather than `System/Collections/Generic/`.

### Details
* Product in use: e.g. ILSpy
* Version in use: fd6070d2292edcf817763b191ea4d491368970ec
* I think this was caused by 292c21d68efd776ff658de4a941f36df14d2bc19, 31bbcf41bcb0dcd0b3e019bc56fa158d8ac37f70, or 53522c45f887c4dc8f490815f40a724e4e9e1559.

ds5678:
Resolves #3448 

### Problem

31bbcf41bcb0dcd0b3e019bc56fa158d8ac37f70 introduced a regression, preventing the `UseNestedDirectoriesForNamespaces` from functioning as intended.

### Solution
  * As part of my testing, I added two API methods to `WholeProjectDecompiler`.
  * ~~`BamlResourceNodeFactory::WriteResourceToFile` also uses this method, which I think is incorrect, but I'm not sure what to change it to.~~

### DIFF
--- a/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
+++ b/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
@@ -131,6 +131,7 @@
     <Compile Include="Helpers\TestsAssemblyOutput.cs" />
     <Compile Include="Output\InsertParenthesesVisitorTests.cs" />
     <Compile Include="ProjectDecompiler\TargetFrameworkTests.cs" />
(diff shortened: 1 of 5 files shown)


siegfriedpammer:

PR #3456, opened 2025-04-09
Update to _wix4 packages for installer

christophwille:
* Format of wxl has changed
* Build now has a PostBuild step
* Global tool for wix is used

image

Additional bonus: size is reduced


christophwille:
Not ready for merge!

PR #3463, opened 2025-04-27
Parenthesize interpolations containing `global::`

### LINKED ISSUE #3447: `global::` in string interpolation
### Input code

```cs
return $"<b>{(global::ParadoxNotion.ReflectionTools.FriendlyName(varType))}</b>";
```

### Erroneous output

```c#
return $"<b>{global::ParadoxNotion.ReflectionTools.FriendlyName(varType)}</b>";
```

The output fails to re-compile and produces the following compiler error message:

```
CS0103: The name 'global' does not exist in the current context
```

The parentheses are necessary because otherwise `:` specifies a numeric format.

### Details
* Product in use: ICSharpCode.Decompiler nuget package
* Version in use: fd6070d2292edcf817763b191ea4d491368970ec
* As always, I am willing to contribute a fix. However, I may need advice on the best way to solve this issue.

ds5678:
Resolves #3447 

### Solution
  * As recommended, I implemented the fix in the `InsertParenthesesVisitor` class.
  * I used recursion to check if any child has `global::`. Is there a better (more performant) way to check?

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/GloballyQualifiedTypeInStringInterpolation.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/GloballyQualifiedTypeInStringInterpolation.cs
@@ -4,7 +4,11 @@
 	{
 		public static string Root => $"Prefix {(global::System.DateTime.Now)} suffix";
 		public static string Cast => $"Prefix {((int)global::System.DateTime.Now.Ticks)} suffix";
+#if CS100 && !NET40
+		public static string Lambda1 => $"Prefix {(() => global::System.DateTime.Now)} suffix";
+#else
 		public static string Lambda1 => $"Prefix {(global::System.Func<global::System.DateTime>)(() => global::System.DateTime.Now)} suffix";
+#endif
 		public static string Lambda2 => $"Prefix {((global::System.Func<global::System.DateTime>)(() => global::System.DateTime.Now))()} suffix";
 		public static string Method1 => $"Prefix {M(global::System.DateTime.Now)} suffix";
 		public static string Method2 => $"Prefix {(global::System.DateTime.Now.Ticks)} suffix";
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/GloballyQualifiedTypeInStringInterpolation.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/GloballyQualifiedTypeInStringInterpolation.cs
@@ -4,7 +4,7 @@
 	{
 		public static string Root => $"Prefix {(global::System.DateTime.Now)} suffix";
 		public static string Cast => $"Prefix {((int)global::System.DateTime.Now.Ticks)} suffix";
-#if CS100 && !NET40
+#if CS100 && NET60
 		public static string Lambda1 => $"Prefix {(() => global::System.DateTime.Now)} suffix";
 #else
 		public static string Lambda1 => $"Prefix {(global::System.Func<global::System.DateTime>)(() => global::System.DateTime.Now)} suffix";
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/GloballyQualifiedTypeInStringInterpolation.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/GloballyQualifiedTypeInStringInterpolation.cs
@@ -1,0 +1,7 @@
+﻿namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
+{
+	public static class GloballyQualifiedTypeInStringInterpolation
+	{
+		public static string CurrentDateTime => $"Time: {(global::System.DateTime.Now)}";
+	}
+}
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/GloballyQualifiedTypeInStringInterpolation.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/GloballyQualifiedTypeInStringInterpolation.cs
@@ -2,6 +2,20 @@
 {
 	public static class GloballyQualifiedTypeInStringInterpolation
 	{
-		public static string CurrentDateTime => $"Time: {(global::System.DateTime.Now)}";
+		public static string Root => $"Prefix {(global::System.DateTime.Now)} suffix";
+		public static string Cast => $"Prefix {((int)global::System.DateTime.Now.Ticks)} suffix";
+		public static string Lambda1 => $"Prefix {(global::System.Func<global::System.DateTime>)(() => global::System.DateTime.Now)} suffix";
+		public static string Lambda2 => $"Prefix {((global::System.Func<global::System.DateTime>)(() => global::System.DateTime.Now))()} suffix";
+		public static string Method1 => $"Prefix {M(global::System.DateTime.Now)} suffix";
+		public static string Method2 => $"Prefix {(global::System.DateTime.Now.Ticks)} suffix";
+		public static string Method3 => $"Prefix {(global::System.DateTime.Equals(global::System.DateTime.Now, global::System.DateTime.Now))} suffix";
+		public static string ConditionalExpression1 => $"Prefix {(Boolean ? global::System.DateTime.Now : global::System.DateTime.UtcNow)} suffix";
+		public static string ConditionalExpression2 => $"Prefix {(Boolean ? global::System.DateTime.Now : global::System.DateTime.UtcNow).Ticks} suffix";
+
+		private static bool Boolean => false;
+		private static long M(global::System.DateTime time)
+		{
+			return time.Ticks;
+		}
 	}
 }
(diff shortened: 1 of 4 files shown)


siegfriedpammer:
Ad recursion:

You can avoid stack overflows by using `AstNode.Dsecendants`, see https://github.com/icsharpcode/ILSpy/blob/03b7444943e720b3134d296c0c8dd3876f8ea4ce/ICSharpCode.Decompiler/CSharp/Syntax/AstNode.cs#L291-L337

siegfriedpammer:
Note to self: do not review pull requests when not entirely awake :(

sorry, I gave wrong feedback. We cannot prevent stack overflows in this case, because the whole thing is using the Visitor pattern. So the problem remains and the performance might even be worse due to the use of LINQ.

So, is there actually anything we could do to limit the number of nodes to check? Can we, for example, stop at anonymous functions/lambdas? Do we still need to disambiguate inside invocation argument lists? Might we be able to just look at the top-level expression?

This is fine according to [SharpLab](https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQMwAJboMLoN7LpGYZRwAM6AsgBRmUCGAlPoce1AOzoAkARABUApgGcALiHy0A5gBsA9gCMGskCDIA6AKIA7AG4BLAE7ydAWyE6xG7AFcjRy2IAixoQGMx8owE90YdGYAXz4AbjYiIIiSTApqOgBWAB56ABp0RXl5WQA+dAAPFgIkdg5ufmFxSTxgIVkhaQYxITo4nyLMbjklFTVNXUMTcycbe0crV0dPbx8NABlLaTEAC3Q88lD0IJDwkuIopCCgA=)...

```
    public string M(string a) {
        return $"Test: {M(global::System.Environment.CurrentDirectory + a)}";
    }
    public string M() {
        return $"Test: {delegate(string y) { return global::System.Environment.CurrentDirectory.Length > 0; }}";
    }
```

So I guess, we can stop descending as soon as we find:
* Parentheses
* An argument
* An anonymous method/lambda expression

What do you think?

ds5678:
I implemented what you suggested. However, I encountered another issue that's preventing my tests from passing.

siegfriedpammer:
While fixing #3464 I noticed that `public static string Lambda1 => $"Prefix {() => global::System.DateTime.Now} suffix";` doesn't require a cast, but still requires parentheses.

ICSharpCode.Decompiler/CSharp/OutputVisitor/InsertParenthesesVisitor.cs
```diff
@@ -389,6 +390,36 @@ public override void VisitAsExpression(AsExpression asExpression)
...
+			base.VisitInterpolation(interpolation);
+
+			static bool InterpolationNeedsParenthesized(AstNode node)
+			{
+				if (node is MemberType { IsDoubleColon: true })
+					return true;
+
+				if (node is ConditionalExpression)
+					return false; // Conditional expressions are parenthesized in their own visit method.
+				if (node is AnonymousMethodExpression or LambdaExpression)
```
siegfriedpammer, line 407:
```suggestion
				if (node is AnonymousMethodExpression or LambdaExpression { Body: BlockStatement })
```

ICSharpCode.Decompiler/CSharp/OutputVisitor/InsertParenthesesVisitor.cs
```diff
@@ -389,6 +390,36 @@ public override void VisitAsExpression(AsExpression asExpression)
...
 		}
 
+		public override void VisitInterpolation(Interpolation interpolation)
+		{
+			// If an interpolation contains global::, we need to parenthesize the expression.
+			if (InterpolationNeedsParenthesized(interpolation))
+				Parenthesize(interpolation.Expression);
+			base.VisitInterpolation(interpolation);
+
+			static bool InterpolationNeedsParenthesized(AstNode node)
```
siegfriedpammer, line 400:
```suggestion
			static bool InterpolationNeedsParenthesis(AstNode node)
```

ICSharpCode.Decompiler/CSharp/OutputVisitor/InsertParenthesesVisitor.cs
```diff
@@ -389,6 +390,36 @@ public override void VisitAsExpression(AsExpression asExpression)
...
+					return true;
+
+				if (node is ConditionalExpression)
+					return false; // Conditional expressions are parenthesized in their own visit method.
+				if (node is AnonymousMethodExpression or LambdaExpression)
+					return false;
+				if (node is InvocationExpression invocation)
+					return InterpolationNeedsParenthesized(invocation.Target);
+				if (node is CastExpression cast)
+					return InterpolationNeedsParenthesized(cast.Expression);
```
siegfriedpammer, line 412: There is a whole lot of special cases for cast-expressions: see `InsertParenthesesVisitor.VisitCastExpression`.

ds5678: I reordered the visitation of children, so that those special cases should get hit before we reason about whether or not to parenthesize the interpolation expression.

ds5678:
> While fixing #3464 I noticed that `public static string Lambda1 => $"Prefix {() => global::System.DateTime.Now} suffix";` doesn't require a cast, but still requires parentheses.

This is a recent change caused by you fixing #3464. Previously, the decompiler would emit the cast always. I added a pragma directive in the unit test to choose the appropriate format based on framework and C# version.

ICSharpCode.Decompiler.Tests/TestCases/Pretty/GloballyQualifiedTypeInStringInterpolation.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
+{
+	public static class GloballyQualifiedTypeInStringInterpolation
+	{
+		public static string Root => $"Prefix {(global::System.DateTime.Now)} suffix";
+		public static string Cast => $"Prefix {((int)global::System.DateTime.Now.Ticks)} suffix";
+#if CS100 && !NET40
```
ds5678, line 7: Should this have been `#if CS100 && NET60` instead?

ds5678: I think so.

ds5678:
Also, I notice that you seem to normal merge all pull requests, even though mine almost always have a few "junk" commits. Is there a reason you don't squash merge?

ICSharpCode.Decompiler.Tests/TestCases/Pretty/GloballyQualifiedTypeInStringInterpolation.cs
siegfriedpammer: I think it makes more sense to use `!NET40` because classic framework is the "legacy edge case" and we want to minimize necessary changes when updating the framework version.

ds5678: This happens because `DefaultInterpolatedStringHandler` is only available on .NET 6 and above. If any of our tests ran on .NET 5, they would fail.

ds5678: Do you want me to change it back?

siegfriedpammer: Hmmm... yes, and because we emit conditional compilation symbols for all subsumed versions used, it works. OK, we can leave it like this...

siegfriedpammer:
> This is a recent change caused by you fixing #3464.

Yes, my comment was imprecise: While playing around with your test case (in order to decide whether I can pass `allowImplicitConversion: true` to `ConvertTo`), I noticed that Visual Studio highlights the cast to `Func<...>` as unnecessary, so I tried removing it and it seemed to work, but the parenthesis was no longer optional, if the body of the lambda was just an expression containing `global::`.

> Also, I notice that you seem to normal merge all pull requests, even though mine almost always have a few "junk" commits. Is there a reason you don't squash merge?

I usually organize my commits into smaller self-contained units/changes to make bisecting after bug reports easier. If a PR has somewhat organized commits I usually just do a normal merge, however, if there is junk in there, I do squash-merges too. It all the depends on the person creating the PR.

ds5678:
Is there something else I need to do for this pull request to be merged?

ICSharpCode.Decompiler/CSharp/OutputVisitor/InsertParenthesesVisitor.cs
```diff
@@ -17,6 +17,7 @@
 // DEALINGS IN THE SOFTWARE.
 
 using System;
+using System.Linq;
```
siegfriedpammer, line 20: Nit: no longer necessary? 

```suggestion
```

siegfriedpammer:
> Is there something else I need to do for this pull request to be merged?

No, I just wasn't sure whether you are done yet.

siegfriedpammer:
Thank you very much for your continued contributions!

ds5678:
No problem. I appreciate your reviews.

PR #3467, opened 2025-05-04
Part 1 of support for C# 12 InlineArray

siegfriedpammer:


siegfriedpammer:
@ds5678 this PR covers the basics of InlineArrays... before we can continue with params collections, we will have to add minimal support for collection expressions:

* Only use collection expression syntax, if the array type is an inline array type (target-typed to span) or a single-element span

Would you be interested in taking this on?

See some experiments here: https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQAwAIpwHQCUCuAdgC4CWAtgKbYDCA9uQA6kA2lATgMocBupAxpQDOAbmRQAzJhjoa6AN7J0y9AG0AkoRalClAILt2AQwCeACgAsASgC6SlZPRDi7fP2LoAsic4u3xe2VFJBVQ9EZ2Uh4jYkp0HQ8AfTEQlQBfJUDMKSwMAAUjY3IhTkYjQk1tXQNjcysFLNCoAHZMOFRsACk6HTMAIgAadEHhoxG+gCNx/j6rFNCMkKzHXPRS8voWNndSOgrCUjIjbQAvDjN64LCHVtyunsJ+oZHVPrHnqeeZmzmsxdDllJvL5XO50AAxOh0C5ZK7XdAAegRmFaqjgQxgQwkQwsNhE6AA7nR8CxgIQAOQeInsADWQwmlH4RnwQjiOiqcUKtSE6EKcUIdA8fX4dC2jLIewAtJQAB4RYRCXaEdAiwjOUFkCZsPqNFTAvxgrmmdAAXnQwEoADNmSxiPN4UaTKpUDZTeg4Pbro60a6zTBPWFvTBfegJAHQt6JCGLOGbrzDKYA4s0kA

```csharp
using System;
public class C {
    [InlineArray(4)]
    public struct MyStruct
    {
        private int _;
    }
 
    public string ParamsSpanInlineArray() {
        return string.Join(", ", "a", "b", "c");
    }
 
    public string SpanCollectionInitializer() {
        return string.Join(", ", ["a", "b", "c"]);
    }
    
    public MyStruct Foo()
    {
        // return [1, 2, 3, 4]; wouldn't work, because inline arrays are not "collection-expression constructible"
        MyStruct array = default;
        array[0] = 1;
        array[1] = 2;
        array[2] = 3;
        array[3] = 4;
        return array;
    }
}
```

ds5678:
> Would you be interested in taking this on?

Sure, I can work on this.

ICSharpCode.Decompiler/CSharp/CallBuilder.cs
```diff
@@ -435,6 +435,29 @@ public ExpressionWithResolveResult Build(OpCode callOpCode, IMethod method,
 						argumentList.GetArgumentResolveResults(), isExpandedForm: argumentList.IsExpandedForm, isDelegateInvocation: true));
 			}
 
+			if (settings.InlineArrays && method.DeclaringType.FullName == "<PrivateImplementationDetails>")
+			{
+				var unwrappedTarget = argumentList.Arguments[0].Expression;
```
dgrunwald, line 440: `argumentList.Arguments` might be empty here.

ICSharpCode.Decompiler/CSharp/CallBuilder.cs
```diff
@@ -435,6 +435,29 @@ public ExpressionWithResolveResult Build(OpCode callOpCode, IMethod method,
...
+				if (unwrappedTarget is DirectionExpression dirExpr)
+					unwrappedTarget = dirExpr.Expression;
+				switch (method.Name)
+				{
+					case "InlineArrayAsSpan":
+					case "InlineArrayAsReadOnlySpan":
+						var arrayResolveResult = unwrappedTarget.GetResolveResult();
+						unwrappedTarget.RemoveAnnotations<ResolveResult>();
+						return unwrappedTarget.Detach().WithRR(new InlineArrayResolveResult(arrayResolveResult, method.ReturnType));
```
dgrunwald, line 449: This doesn't make sense. A ResolveResult describes the semantics of a C# expression we have generated. If the C# expression doesn't change, it doesn't make sense to swap out the associated ResolveResult?

dgrunwald: Note that our conversion logic needs to consider all possible conversions (for overload resolution), not just the conversion actually picked in the IL program (that wouldn't allow us to detect if we're about to emit C# that would pick the wrong overload).

dgrunwald: Your case is a bit of a red herring: the return value of `GetArray()` cannot be converted to `ReadOnlySpan<int>` (neither with an explicit cast nor with an implicit conversion), because this conversion only allows lvalues.
We need to treat that particular call combination (conversion to span followed by indexing) as a special case, maybe even introducing a special ILAst instruction for this combination.
For thinking about the conversion modelling, the indexing special case is actively misleading. There's it's more useful to consider `Foo(some_array)`, where Foo has multiple overloads (some of which involve the conversion).

dgrunwald: In particular for conversion modeling you also you need to think about cases where the IL does not perform the conversion to span, but the C# code might unless we add explicit casts.
For example: [in this example](https://sharplab.io/#v2:C4LglgNgNAJiDUAfAAgJgIwFgBQyAMABMugHQBKArgHbBgC2ApiQMID2dADpAwE4DKvAG5gAxgwDOAbhzIAzEVQFmBAN44CGggG0AklQhgqDAII8eAQwCeACgAsASgC6BccB4URwAgFlLpi5aqBBw8YILmwAwEhl4MEAx0kgQAvjjqmsi2BABirKzWrABGAFYMngSs9qqp2JpEWbn5ZAzmMADy+pZ8HOZUADwxAHwVVSo16RqZBAAqEsDWvv5WBOZmoxN1OXnWBSVlwParPPbStZo1yUA===) the explicit cast to `(object)` is required because the overload would otherwise be ambiguous. If our conversion machinery does not consider the span conversion to be possible, OR would think that the cast is not required, and we generate invalid code. Thus our conversion machinery needs to support span conversions in general, not just for a special resolve result.

siegfriedpammer:
@dgrunwald please review again... I think this could be merged.

ICSharpCode.Decompiler/IL/Transforms/InlineArrayTransform.cs
```diff
@@ -0,0 +1,234 @@
...
+			{
+				isReadOnly = true;
+
+				if (inst.Arguments is not [AddressOf { Value: Call targetInst, Type: var typeInfo }, var indexInst])
+					return false;
+
+				if (!MatchInlineArrayHelper(targetInst.Method, "InlineArrayAsReadOnlySpan", out var inlineArrayType))
+					return false;
+
+				if (targetInst.Arguments is not [var addrInst, LdcI4])
```
dgrunwald: The value of the LdcI4 needs to be checked against the length of the inline array type.

ICSharpCode.Decompiler/IL/Transforms/InlineArrayTransform.cs
```diff
@@ -0,0 +1,234 @@
...
+
+			if (MatchInlineArrayElementRef(inst, out type, out addr, out index, out isReadOnly))
+			{
+				if (isReadOnly)
+				{
+					context.Step("call InlineArrayElementRefReadOnly(addr, index) -> readonly.ldelema.inlinearray(addr, index)", inst);
+				}
+				else
+				{
+					context.Step("call InlineArrayElementRef(addr, index) -> ldelema.inlinearray(addr, index)", inst);
```
dgrunwald, line 53: Does `InlineArrayElementRef` perform bounds-checking? If not, this transform is not semantics-preserving (we'd need to check that `index` is a constant and inbounds)

siegfriedpammer: https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQAwAIpwHQCUCuAdgC4CWAtgKbYDCA9uQA6kA2lATgMocBupAxpQDOAbmRQAzJhjoa6AN7J0y9AG0AkoRalClAILt2AQwCeACgAsASgC6SlZPRDi7fP2LoAsic4u3xe2VFJBVQ9EZ2Uh4jYkp0HQ8AfTEQlQBfJUDMKQT0AA0zb19Xd3QADwAaeJJ4q3QAXgA+ctVUGxTQx1yATUKfP1LK6o9SOqaWiXasrpqALT7i/3Kq3NGG5rLVUinU5Wmcmu6AUUM6dgWBjyHVsY3WgHoAKgA2cPY6YDdhdCNCdA53uwfh5+AxmGx0GQqOhHvcdukgA

My quick tests show that InlineArrayElementRef is only used for constants where compile-time checking occurs/is possible. Will adjust the transform accordingly.

siegfriedpammer: In the other case MemoryMarshal.CreateSpan (https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.memorymarshal.createspan?view=net-9.0) is used where the documentation says it does not perform bounds checking either.

I think LdElemaInlineArray can be specified as never performing a bounds check. Or am I missing something?

dgrunwald: If LdElemaInlineArray is specified as never perform a bounds checks, then the `MatchSpanIndexerWithInlineArrayAsSpan` transform would be invalid as it'd be removing the `get_Item` bounds check.
Also you wouldn't be able to turn `ldelema.inlinearray(addr, index)` into C# `array[index]` as the latter does perform a bounds check.

siegfriedpammer: Ah yeah the get_Item call performs the bounds check. Sorry, this slipped my attention. Small mobile phone screen...

siegfriedpammer:
@dgrunwald I just added one new commit, I had to rebase to get the WholeProjectDecompilerTest thread-safety fix.

PR #3470, opened 2025-05-05
Handle explicit optional parameter after default parameter

### LINKED ISSUE #3469: Optional parameters with implicit default values
### Input code

```il
.assembly extern System.Runtime
{
  .publickeytoken = (B0 3F 5F 7F 11 D5 0A 3A )                         // .?_....:
  .ver 4:0:0:0
}

.class public auto ansi abstract sealed beforefieldinit ICSharpCode.Decompiler.Tests.TestCases.ILPretty.Issue3469
    extends [System.Runtime]System.Object
{
    // Methods
    .method public hidebysig static 
        void M (
            [opt] int32 i,
            [opt] valuetype [System.Runtime]System.DateTime d
        ) cil managed 
    {
        .param [1] = int32(0)
        // Roslyn inserts, but Mono doesn't:
        // .param [2] = nullref
        // Method begins at RVA 0x2050
        // Code size 1 (0x1)
        .maxstack 8

        IL_0000: ret
    } // end of method Issue3469::M

} // end of class ICSharpCode.Decompiler.Tests.TestCases.ILPretty.Issue3469
```

### Erroneous output

```c#
using System;
namespace ICSharpCode.Decompiler.Tests.TestCases.ILPretty
{
	public static class Issue3469
	{
		public static void M(int i = 0, [Optional] DateTime d)
		{
		}
 	}
}
```

```
CS1737: Optional parameter cannot precede required parameters
```

### Expected output

```c#
using System;
namespace ICSharpCode.Decompiler.Tests.TestCases.ILPretty
{
	public static class Issue3469
	{
		public static void M(int i = 0, DateTime d = default(DateTime))
		{
		}
 	}
}
```

### Details
* Product in use: IL pretty unit test
* Version in use: latest commit (aff9649711fa3579f46e40336e7dc3fb8192302c)
  * I am always willing to contribute fixes for my issues.
  * This assembly was compiled with the Unity Mono compiler.
  * If this is a step too far, I can automatically preprocess the assembly with AsmResolver before using ILSpy. I already do this in other situations, like removing public keys from `InternalsVisibleToAttribute`.

ds5678:
Resolves #3469 

### Solution
  * I implemented this in an extensions class, which is a little bit removed from the code that uses it.
  * Do you think my extension method needs more documentation?
  * Should there be any additional tests involving decimal constants?

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/OptionalArguments.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/OptionalArguments.cs
@@ -308,11 +308,15 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 		{
 		}
 #endif
-		public static void Issue3469([Optional][DefaultParameterValue(0)] int i, [Optional] DateTime d)
+		public static void Issue3469a([Optional][DefaultParameterValue(0)] int i, [Optional] DateTime d)
 		{
 		}
-
 #if CS120
+		public static Action<int, DateTime> Issue3469b()
+		{
+			return ([Optional][DefaultParameterValue(0)] int i, [Optional] DateTime d) => {
+			};
+		}
 		public static D LambdaWithOptionalParameter()
 		{
 			return (int x = 10) => x;
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/OptionalArguments.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/OptionalArguments.cs
@@ -314,8 +314,10 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 #if CS120
 		public static Action<int, DateTime> Issue3469b()
 		{
+#pragma warning disable CS9099 // Parameter 1 has default value 'default(int)' in lambda but '<missing>' in the target delegate type
 			return ([Optional][DefaultParameterValue(0)] int i, [Optional] DateTime d) => {
 			};
+#pragma warning restore CS9099
 		}
 		public static D LambdaWithOptionalParameter()
 		{
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/OptionalArguments.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/OptionalArguments.cs
@@ -16,6 +16,7 @@
 // OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
+using System;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
@@ -307,6 +308,9 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 		{
 		}
 #endif
+		public static void Issue3469([Optional][DefaultParameterValue(0)] int i, [Optional] DateTime d)
+		{
+		}
 
 #if CS120
 		public static D LambdaWithOptionalParameter()
--- a/ICSharpCode.Decompiler/TypeSystem/IParameter.cs
+++ b/ICSharpCode.Decompiler/TypeSystem/IParameter.cs
@@ -105,44 +105,4 @@ namespace ICSharpCode.Decompiler.TypeSystem
 		/// </summary>
 		IParameterizedMember? Owner { get; }
 	}
-
-	internal static class IParameterExtensions
-	{
-		/// <summary>
-		/// Checks if the parameter is allowed to be assigned a default value.
-		/// </summary>
-		/// <remarks>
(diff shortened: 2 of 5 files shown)


ICSharpCode.Decompiler/TypeSystem/IParameter.cs
```diff
@@ -105,4 +105,34 @@ public interface IParameter : IVariable
 		/// </summary>
 		IParameterizedMember? Owner { get; }
 	}
+
+	internal static class IParameterExtensions
+	{
+		public static bool GetDefaultValueAssignmentAllowed(this IParameter parameter)
```
siegfriedpammer, line 111: IsDefaultValueAssignmentAllowed?

ICSharpCode.Decompiler/TypeSystem/IParameter.cs
```diff
@@ -105,4 +105,34 @@ public interface IParameter : IVariable
 		/// </summary>
 		IParameterizedMember? Owner { get; }
 	}
+
+	internal static class IParameterExtensions
```
siegfriedpammer, line 109: We already have `TypeSystemExtensions`, wouldn't it make sense to keep them all in the same unit?

ICSharpCode.Decompiler/TypeSystem/IParameter.cs
```diff
@@ -105,4 +105,34 @@ public interface IParameter : IVariable
...
+	{
+		public static bool GetDefaultValueAssignmentAllowed(this IParameter parameter)
+		{
+			if (!DefaultValueAssignmentAllowedIndividual(parameter))
+				return false;
+
+			if (parameter.Owner == null)
+				return true;
```
siegfriedpammer, line 117: if the parameter belongs to an anonymous method, we'd still have to figure out whether we can use assignment syntax, I think...

Also I am not sure, whether the `Owner` is actually null in practice, I'd expect the `IParameter` instance that is used for the lambda parameters just to be "owned" by the compiler-generated method from which the lambda was constructed. But I might be wrong.

ICSharpCode.Decompiler/TypeSystem/IParameter.cs
ds5678: I can move it over

ICSharpCode.Decompiler/TypeSystem/IParameter.cs
ds5678: I can write a unit test?

siegfriedpammer: Yes, you could do that... writing a pretty test that contains a lambda with parameter attributes should be possible in recent versions of C#

Looking at the code that actually creates the anonymous method expressions/lambdas, it seems I remember it correctly: https://github.com/icsharpcode/ILSpy/blob/df1f1051d4fd69595b00fbe829b52826505e4ea6/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs?plain=1#L2406C1-L2406C75

So maybe the documentation needs to be updated :)

ds5678: > Also I am not sure, whether the `Owner` is actually null in practice, I'd expect the `IParameter` instance that is used for the lambda parameters just to be "owned" by the compiler-generated method from which the lambda was constructed. But I might be wrong.

This intuition was correct.

siegfriedpammer: > So maybe the documentation needs to be updated :)

No, I think, we should keep it that way, otherwise it might break all code that uses `DefaultParameter`s...

ICSharpCode.Decompiler.Tests/TestCases/Pretty/OptionalArguments.cs
```diff
@@ -307,8 +308,15 @@ public static void Issue2920p(in int x = 3)
 		{
 		}
 #endif
-
+		public static void Issue3469a([Optional][DefaultParameterValue(0)] int i, [Optional] DateTime d)
+		{
+		}
 #if CS120
+		public static Action<int, DateTime> Issue3469b()
+		{
```
siegfriedpammer, line 316:
```suggestion
		{
		#pragma warning disable warning CS9099 // Parameter 1 has default value 'default(int)' in lambda but '<missing>' in the target delegate type
```

ICSharpCode.Decompiler.Tests/TestCases/Pretty/OptionalArguments.cs
```diff
@@ -307,8 +308,15 @@ public static void Issue2920p(in int x = 3)
...
+			return ([Optional][DefaultParameterValue(0)] int i, [Optional] DateTime d) => {
+			};
```
siegfriedpammer, line 318:
```suggestion
			};
			#pragma warning restore CS9099
```

ds5678:
Is there anything else that needs done for this pull request?

siegfriedpammer:
Sorry for the wait... I had second thoughts about the use of an extension method (and then other things came up) instead of implementing this directly in the type system, but it's fine, because this way it's independent from the actual implementation.

Issue #3468, opened 2025-05-05
Ref locals before C# 7


ds5678:
### Input code

```cs
public struct StructWithStringField
{
	public string Id;
	public static void ModifyArray(StructWithStringField[] array)
	{
		int num = 0;
		while (num < 10)
		{
			array[0].Id = (++num).ToString();
		}
	}
}
```

### Erroneous output

```c#
public struct StructWithStringField
{
	public string Id;
	public static void ModifyArray(StructWithStringField[] array)
	{
		int num = 0;
		while (num < 10)
		{
			ref StructWithStringField reference = ref array[0];
			int num2 = ++num;
			reference.Id = num2.ToString();
		}
	}
 }
```

This produces an error before C# 7:

```
CS1525: Unexpected symbol 'ref'
```

### Details
* Product in use: pretty unit test
* Version in use: latest commit (aff9649711fa3579f46e40336e7dc3fb8192302c)
* Any other relevant information to the issue, or your interest in contributing a fix.
  * I, of course, am willing to work on this.
  * I think I've seen other situations with similar `ref` local issues, but I don't have any examples at this time. I'm checking to see if I can find some.

siegfriedpammer:
This is an ILInlining problem.

ds5678:
I narrowed the problem down with tests that are more specific. However, I'm struggling to figure out how to resolve the problem.

```cs
public static string Test2(int value)
{
	return (++value).ToString();
}
public static string Test3(int value)
{
	M(++value);
	return value.ToString();
}
private static void M(in int value)
{
}
```

I did a lot of debugging, both in `ILInlining` and `TransformAssignment`. I could not find a clean way to solve the problem in `ILInlining`, and my modifications to `TransformAssignment` (#3472) feel like they're the wrong way to solve this. I could use some guidance.

dgrunwald:
One idea might be to bring back the aggressive forms of copy propagation that we removed when we added support for ref locals. (but only enable aggressive copy propagation if UseRefLocalsForAccurateOrderOfEvaluation is disabled).

Basically the idea there was to turn
`ref StructWithStringField reference = ref array[0];`
into
```
var x = array;
var y = 0;
```
and then turn every use of `reference` into `x[y]`.

This way we are guaranteed to get rid of ref locals no matter how good inlining is.
[...]

PR #3472, opened 2025-05-06
WIP preincrement improvements

ds5678:
### Problem

#3468

### Solution
  * I tried to copy and adapt the code used for post-increment, but I have doubts that this is the right approach.
  * Everything :)

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/CompoundAssignmentTest.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/CompoundAssignmentTest.cs
@@ -4946,5 +4946,21 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 			CustomStruct2 customStruct = GetStruct();
 			customStruct.IntProp += value;
 		}
+
+		public static string PreincrementWithMethodCall(int value)
+		{
+			return (++value).ToString();
+		}
+
+#if CS72
+		public static string PreincrementWithInParameter(int value)
+		{
+			PreincrementWithInParameter_Helper(++value);
+			return value.ToString();
+		}
+		public static void PreincrementWithInParameter_Helper(in int value)
+		{
+		}
+#endif
 	}
 }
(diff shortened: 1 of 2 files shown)


ds5678:
Hi @siegfriedpammer, I'm running into a small problem with this pull request. My tests pass, except for Roslyn 1.3, which compiles the pre-increment like this:

```
{Block IL_0000 (incoming: 1) {
	stloc V_0(binary.add.i4(ldloc value, ldc.i4 1))
	stloc value(ldloc V_0)
	leave IL_0000 (call ToString(ldloca V_0))
} at IL_0000}
```

I want to write a separate transform to make the IL into this:

```
{Block IL_0000 (incoming: 1) {
	stloc value(binary.add.i4(ldloc value, ldc.i4 1))
	stloc V_0(ldloc value)
	leave IL_0000 (call ToString(ldloca V_0))
} at IL_0000}
```

In other words, I want the second transform to switch the position of `value` and `V_0`, so that other transforms can reduce it further:

```
{Block IL_0000 (incoming: 1) {
	stloc V_0(stloc value(binary.add.i4(ldloc value, ldc.i4 1)))
	leave IL_0000 (call ToString(ldloca V_0))
} at IL_0000}
```

My current pattern matching would work on this final transformed IL.

However, I'm concerned about unintended consequences and unsure of how this should be implemented. In particular:

* What rules should I use when deciding whether or not to do this transformation?
* When should the second transform be run?
* Does the second transform violate semantics in some way that might affect the use of ILSpy in other tools like Visual Studio?

In regards to rules, my idea was:

* There are two instructions, `stloc x1(expression)` followed by `stloc x2(ldloc x1)`.
* `expression` references `x2` but not `x1`.
  * Or maybe it references only `x2` and no other variables?
* `x1` and `x2` have the same type.

What are your thoughts?

dgrunwald:
I think your order-swapping only makes sense if you can guarantee that it'll be combined into a preincrement.
If this doesn't happen, it's confusing to show the code in a different order than it's actually executed. If there's multiple statements in the "wrong" order, that might also be problematic for the generated sequence points. But if the preincrement transform can combine it into an inline assignment, then there's only one statement and the sequence points won't be an issue.

So I think this doesn't make sense as a separate statement-swapping transform, but should be a transform that directly creates the inline assignment.

dgrunwald:
But if it only impacts Roslyn 1.3, then maybe we don't need to support preincrements for that particular version?
You could just use `#if` in the testcase to avoid testing with that Roslyn version.

ds5678:
> But if it only impacts Roslyn 1.3, then maybe we don't need to support preincrements for that particular version? You could just use `#if` in the testcase to avoid testing with that Roslyn version.

I ended up doing this. I'll come back and implement a second pull request if I discover that 1.3 support is necessary for my use case.

christophwille:
@ds5678 so it is no longer WIP (PR title)?

ds5678:
Fixed

PR #3481, opened 2025-05-19
Correctly handle use of `cgt` and others for object refs (fixes #3465)

### LINKED ISSUE #3465: error CS0019: Operator '>' cannot be applied to operands of type 'Image' and 'Image'
### Erroneous output
```c#

			foreach (PdfName key in pdfDictionary.Keys)
			{
				PdfObject pdfObject = pdfDictionary.Get(key);
				PdfObject pdfObject2 = pdfReader.GetPdfObject(((PRIndirectReference)pdfObject).Number);
				if (smethod_2(pdfObject2) > Class13.image_0)
				{
					list_1.Add(smethod_2(pdfObject2));
				}
			}


IL:


			// PdfObject pdfObject2 = pdfReader.GetPdfObject(((PRIndirectReference)pdfObject).Number);
			IL_0074: ldloc.0
			IL_0075: ldloc.s 4
			IL_0077: castclass [itextsharp]iTextSharp.text.pdf.PRIndirectReference /* 010000F9 */
			IL_007c: callvirt instance int32 [itextsharp]iTextSharp.text.pdf.PdfIndirectReference::get_Number() /* 0A00020B */
			IL_0081: callvirt instance class [itextsharp]iTextSharp.text.pdf.PdfObject [itextsharp]iTextSharp.text.pdf.PdfReader::GetPdfObject(int32) /* 0A00020C */
			IL_0086: stloc.s 5
			// if (smethod_2(pdfObject2) > Class13.image_0)
			IL_0088: ldloc.s 5
			IL_008a: call class [System.Drawing]System.Drawing.Image ns0.DashboardForm::smethod_2(class [itextsharp]iTextSharp.text.pdf.PdfObject) /* 060000B5 */
			IL_008f: ldsfld class [System.Drawing]System.Drawing.Image ns0.Class13::image_0 /* 0400020F */
			IL_0094: cgt.un
			IL_0096: brfalse.s IL_00aa

```

error CS0019: Operator '>' cannot be applied to operands of type 'Image' and 'Image'

dnSpy gens OK code:

```
		foreach (PdfName pdfName in pdfDictionary.Keys)
		{
			PdfObject pdfObject = pdfDictionary.Get(pdfName);
			PdfObject pdfObject2 = pdfReader.GetPdfObject(((PRIndirectReference)pdfObject).Number);
			if (smethod_2(pdfObject2) != Class13.image_0)
			{
				this.list_1.Add(DashboardForm.smethod_2(pdfObject2));
			}
		}
```

### Details
* ILSpy version 9.1.0.7988+03b7444943e720b3134d296c0c8dd3876f8ea4ce
* .NET version 8.0.15+50c4cb9fc31c47f03eac865d7bc518af173b74b7

ElektroKill:
Link to issue(s) this covers
https://github.com/icsharpcode/ILSpy/issues/3465

### Problem
ILSpy did not handle use of `cgt` for object reference comparison where neither side was `ldnull`.

This is allowed as stated in ECMA:
"cgt.un is allowed and verifiable on ObjectRefs (O). This is commonly used when comparing an ObjectRef with null (there is no “compare-not-equal” instruction, which would otherwise be a more obvious solution)" - ECMA III.1.5 - Operand type table, Table III.4: Binary Comparison or Branch Operations, Note 2

Code checks for other comparison types just like it did before for `ldnull` but now for all object refs.

### Solution
* No test currently as I'm yet to see a file using this.

@greenozon Could you verify if this changes fixes your issue on your file?

### DIFF
--- a/ICSharpCode.Decompiler/IL/Transforms/EarlyExpressionTransforms.cs
+++ b/ICSharpCode.Decompiler/IL/Transforms/EarlyExpressionTransforms.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) 2017 Daniel Grunwald
+// Copyright (c) 2017 Daniel Grunwald
 // 
 // Permission is hereby granted, free of charge, to any person obtaining a copy of this
 // software and associated documentation files (the "Software"), to deal in the Software
@@ -52,29 +52,16 @@ namespace ICSharpCode.Decompiler.IL.Transforms
 			{
 				return;
 			}
-			if (inst.Right.MatchLdNull())
+			if (inst.Left.ResultType == StackType.O && inst.Right.ResultType == StackType.O)
 			{
-				if (inst.Kind == ComparisonKind.GreaterThan)
+				if (inst.Kind == ComparisonKind.GreaterThan || inst.Kind == ComparisonKind.LessThan)
 				{
-					context.Step("comp(left > ldnull)  => comp(left != ldnull)", inst);
+					context.Step($"comp(left {inst.Kind.GetToken()} right) => comp(left != right)", inst);
 					inst.Kind = ComparisonKind.Inequality;
 				}
-				else if (inst.Kind == ComparisonKind.LessThanOrEqual)
+				else if (inst.Kind == ComparisonKind.GreaterThanOrEqual || inst.Kind == ComparisonKind.LessThanOrEqual)
 				{
-					context.Step("comp(left <= ldnull) => comp(left == ldnull)", inst);
-					inst.Kind = ComparisonKind.Equality;
-				}
-			}
-			else if (inst.Left.MatchLdNull())
-			{
-				if (inst.Kind == ComparisonKind.LessThan)
-				{
-					context.Step("comp(ldnull < right)  => comp(ldnull != right)", inst);
-					inst.Kind = ComparisonKind.Inequality;
-				}
-				else if (inst.Kind == ComparisonKind.GreaterThanOrEqual)
-				{
-					context.Step("comp(ldnull >= right) => comp(ldnull == right)", inst);
+					context.Step($"comp(left {inst.Kind.GetToken()} right) => comp(left == right)", inst);
 					inst.Kind = ComparisonKind.Equality;
 				}
 			}


greenozon:
Thanks! 
will give it a try tomorrow and ping back!

dgrunwald:
But is this correct? Is the JIT silently switching `>` to `!=`? Or is actually comparing the numeric values of the object pointers? (which would only have reliable runtime behavior if at least one if the input pointers is null)

ElektroKill:
> But is this correct? Is the JIT silently switching `>` to `!=`? Or is actually comparing the numeric values of the object pointers? (which would only have reliable runtime behavior if at least one if the input pointers is null)

Looks like I rushed this a bit too much and didn't really read into it! ECMA says its only valid for null as implemented in ILSpy already and modern C# JIT appears to treat them as pointers (A quick glance at dotnet/runtime reveals no special handling of `cgt.un` for object refs) so this change is not valid on IL level. This would probably explain why I couldn't get the C# compiler to emit the pattern seen in the snippet in the issue.

The current way of rewriting these comparisons on the IL level is incorrect. The only other alternative I see to decompile this properly would be to handle this case on the C# AST level or when translating ILAst to C# AST. Would this be appropriate, and if so, which way would be preferred?

greenozon:
test is positive, but well, comments are not so..

image

dgrunwald:
It's probably an obfuscator hiding nulls in static fields?

Am I guessing correctly that `Class13.image_0` is always `null`?

I'm not sure what we should do about this, if anything. `Unsafe.IsAddressGreaterThan` only works for references, not object pointers. So I'm not sure if there is any semantics-preserving C# expression for this IL.
On the other hand, the IL only has semantics worth preserving if one of the inputs is `null` (the output of the comparison is nondeterministic otherwise).
Still, someone might use ILSpy to debug a compiler they're writing, in which case we shouldn't silently fix bugs in the generated IL code.
Maybe we could output `!=` but with a warning comment?

ElektroKill:
> It's probably an obfuscator hiding nulls in static fields?
> 
> Am I guessing correctly that `Class13.image_0` is always `null`?

This is actually something I have not considered! It may very well be the case here, as it is something decently common in obfuscation. @greenozon, could you confirm? If this is, in fact, the case, we can close the PR and issue, I'd say, since it's not something the C# compiler will/has produced.

> Maybe we could output `!=` but with a warning comment?

If this doesn't end up as an obfuscation issue, I think this is the way that makes the most sense.

ds5678:
> `Unsafe.IsAddressGreaterThan` only works for references, not object pointers. So I'm not sure if there is any semantics-preserving C# expression for this IL.

This is one way to do it. However, it depends on the memory layout of `TypedReference`.

```cs
public static unsafe bool GreaterThan(object left, object right)
{
    TypedReference leftTypedReference = __makeref(left);
    void* leftPointer = **(void***)&leftTypedReference;
    TypedReference rightTypedReference = __makeref(right);
    void* rightPointer = **(void***)&rightTypedReference;
    return leftPointer > rightPointer;
}
```

https://sharplab.io/#v2:D4AQTAjAsAUAxAOwK4BsUEMBGKCmACAEwEsBnLXWEAZj3DwGE8BvWPN2mkCANjyQTIAzfJgD2olHgDiAJxzoALjhkAVABboEAClGYAVjgDGCvLkEKANHl0HjeGUQDmahQEpW7FjHY+8KgJ4ADjgEAEo4wnIIhvhmCgHBYRHKONH4ALx4APpZALboANYpglpxrgDcHr60ACwAVKYRCgAKokQISjJ4mXV1WiD1va4AZHEJIeGRqTGV3tV+QRPJUTH2Ti7jSVNp3dl5hcVaDs5us/O1Dccure2du739g3UjV/GLWylpZ/MgAOyN5huHWUeAAfGsTkDOt82ABfWCwoA=

ds5678:
Or maybe just this?

```cs
public static unsafe bool GreaterThan(object left, object right)
{
    return *(void**)&left > *(void**)&right;
}
```

https://sharplab.io/#v2:D4AQTAjAsAUAxAOwK4BsUEMBGKCmACAEwEsBnLXWEAZj3DwGE8BvWPN2mkCANjyQTIAzfJgD2olHgDiAJxzoALjhkAVABboEAClGYAVjgDGCvLkEKANHl0HjeGUQDmahQEpW7FjHY/aAdjwAKi0QABZAwNcAMjMTAD4gkPDIqIdnBQBuDzYAX1gcoA==

dgrunwald:
Yes, `*(void**)&left > *(void**)&right` should work.
`&left` could maybe be replaced with `Unsafe.AsPointer(ref left)` for older C# versions?

Both of these ideas have the problem of requiring an lvalue object pointer, so we'd need to avoid inlining in these cases; or un-inline later (via `ILInstruction.Extract`, though this isn't guaranteed to succeed for arbitrarily complex constructs).

dgrunwald:
Actually there's no reason to use an unmanaged pointer for the second level of indirection.
`Unsafe.As<object, UIntPtr>(ref left) > Unsafe.As<object, UIntPtr>(ref right)` seems like it'd be more reliable, as now we'd support the `left` expression pointing to the heap (e.g. parameter in function with yield-return) without getting in trouble with the GC.

dgrunwald:
Fixed with the `Unsafe.As<object, UIntPtr>(ref left) > Unsafe.As<object, UIntPtr>(ref right)` approach in 317b4e8add865decb7626af1d833b6056cdce8ca.

Issue #3486, opened 2025-05-31
UI Performance Issue


ds5678:
### Steps to reproduce
1. Open ILSpy
2. View a class
3. Click on another class in the tree view

On ILSpy 9.1, there is multiple-second lag in step 3.

### Details
* Product in use: ILSpy
* Version in use: release 9.0 and 9.1
* Any other relevant information to the issue, or your interest in contributing a fix.
  * I thought it might have been caused by a592169bb358b0d5afaf849ca74b8aec719f30ba, but I reverted that locally, and the lag was still present.
  * I initially posted about this in #3485 before I did any investigation.
  * 3b8302903a191c543c6945df1414f1379ea533f5 is currently the oldest commit with an unexpired actions run. The lag is present on that commit.

siegfriedpammer:
I cannot reproduce this.

![Image](https://github.com/user-attachments/assets/bf32b439-ea68-4250-b93a-a50983125338)

ds5678:
Might need a bigger assembly to be noticeable. Try this one.

https://www.nuget.org/packages/AssetRipper.SourceGenerated

siegfriedpammer:
Is there a specific type I should look at?

ds5678:
For me, there does not seem to be a difference. Any type is sufficient.

siegfriedpammer:
There is no perceivable delay on my machine.

ds5678:
https://github.com/user-attachments/assets/6a4110e2-540a-4b1b-9cff-02dad5638b29

I'm going to bisect the changes.

ds5678:
I finished my bisection.

https://github.com/icsharpcode/ILSpy/actions/workflows/build-ilspy.yml?query=branch%3Amaster

51522c448784d31b4ce0058261436f0de573ef14 is the last actions run without the problem.

96caa4ecb793c2e3215f44b7c05f1bd4d9ccf08e is the first actions run with the problem.

siegfriedpammer:
So the problem seems to be 128f83d74cdef3cf6065977b2df7414ba950e22d?

ds5678:
Yes, I think so.

siegfriedpammer:
Can you try to profile it? I don't see anything particularly expensive in that change...

ds5678:
The vast majority of time is spent in `ICSharpCode.ILSpy.Options.DecompilerSettings::Clone()`.

ds5678:
[...]

PR #3488, opened 2025-06-01
Convert DecompilerSettings to a record class

ds5678:
### Problem

#3486

### Solution
  * `[Browseable(false)]` properties

### DIFF
--- a/ILSpy/Search/SearchPane.xaml.cs
+++ b/ILSpy/Search/SearchPane.xaml.cs
@@ -474,11 +474,11 @@ namespace ICSharpCode.ILSpy.Search
 				request.RegEx = regex;
 				request.SearchResultFactory = new SearchResultFactory(language);
 				request.TreeNodeFactory = this.treeNodeFactory;
-				var decompilerSettings = settingsService.DecompilerSettings.Clone();
+				var decompilerSettings = settingsService.DecompilerSettings.DeepClone();
 				if (!Enum.TryParse(this.languageVersion?.Version, out Decompiler.CSharp.LanguageVersion languageVersion))
 					languageVersion = Decompiler.CSharp.LanguageVersion.Latest;
 				decompilerSettings.SetLanguageVersion(languageVersion);
-				request.DecompilerSettings = settingsService.DecompilerSettings.Clone();
+				request.DecompilerSettings = settingsService.DecompilerSettings.DeepClone();
 
 				return request;
 			}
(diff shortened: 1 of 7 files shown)


ILSpy/Search/SearchPane.xaml.cs
```diff
@@ -474,11 +474,11 @@ Regex CreateRegex(string s)
 				request.RegEx = regex;
 				request.SearchResultFactory = new SearchResultFactory(language);
 				request.TreeNodeFactory = this.treeNodeFactory;
-				var decompilerSettings = settingsService.DecompilerSettings.Clone();
+				var decompilerSettings = settingsService.DecompilerSettings.DeepClone();
 				if (!Enum.TryParse(this.languageVersion?.Version, out Decompiler.CSharp.LanguageVersion languageVersion))
 					languageVersion = Decompiler.CSharp.LanguageVersion.Latest;
 				decompilerSettings.SetLanguageVersion(languageVersion);
-				request.DecompilerSettings = settingsService.DecompilerSettings.Clone();
+				request.DecompilerSettings = settingsService.DecompilerSettings.DeepClone();
```
ds5678, line 481:
```suggestion
				request.DecompilerSettings = decompilerSettings;
```

I think the original line is bugged.

siegfriedpammer:
Ah, now I see it. There is XML serialization/deserialization happening on ever clone?

`MemberwiseClone` should be fast enough as it should be a simple `memcpy` (I'd hope).

If you change your PR to just get rid of the `SaveToXml` call, does that solve the problem already? I don't think changing `DecompilerSettings` to a record is necessary.

Also, why rename `Clone` to `DeepClone`?

ILSpy/Search/SearchPane.xaml.cs
siegfriedpammer: yes, my bad

ds5678:
> Ah, now I see it. There is XML serialization/deserialization happening on ever clone?

Yes

> `MemberwiseClone` should be fast enough as it should be a simple `memcpy` (I'd hope).
> 
> If you change your PR to just get rid of the `SaveToXml` call, does that solve the problem already? I don't think changing `DecompilerSettings` to a record is necessary.

I tried your commit. I had to add a cast in `DecompilationOptions`, but otherwise it works the same as this.

> Also, why rename `Clone` to `DeepClone`?

Not allowed in records

siegfriedpammer:
> > `MemberwiseClone` should be fast enough as it should be a simple `memcpy` (I'd hope).
> > If you change your PR to just get rid of the `SaveToXml` call, does that solve the problem already? I don't think changing `DecompilerSettings` to a record is necessary.
> 
> I tried your commit. I had to add a cast in `DecompilationOptions`, but otherwise it works the same as this.

Yes, my bad. :( now fixed in the latest commit

> > Also, why rename `Clone` to `DeepClone`?
> 
> Not allowed in records

Ah, yeah, one more reason to *not* use a record, then ;)

Issue #3495, opened 2025-06-17
Add an option to not transform `Activator.CreateInstance<T>()` to `new T()`


DoctorKrolic:
**Is your feature request related to a problem? Please describe.**
`new()` generic constraint was introduced in the same C# version as generics themselves (C# 2). Its implementation under the hood requres to call `Activator.CreateInstance<T>()` for what user writes as `new T()` in the source code. ILSpy _always_ decompiles `Activator.CreateInstance<T>()` into `new T()` and there is no setting to disable that. Low-level C# viewers like [sharplab](https://sharplab.io/), [.NET Lab](https://lab.razor.fyi) or [Compiler developer SDK](https://marketplace.visualstudio.com/items?itemName=333fred.net-compiler-developer-sdk) would benifit is there was an option to disable such transformation to be able to present as low-level C# decompilation view as possible

**Describe the solution you'd like**
New API like:
```diff
namespace ICSharpCode.Decompiler;

class DecompilerSettings
{
+    public bool GenericTypeInstantiation { get; set; }
}
```
The name is obviously not final here, maybe you can suggest something better. By default the setting is `true` for C# 2 and above, which represents current behavior, but can be set to `false`, in which case `Activator.CreateInstance<T>()` are not transformed to `new T()` even for a generic `T` with `new()` constraint

**Additional context**
The use case here seems very specific, but I believe implementation and maintaining this option is very straight-forward, this is probably a matter of 1 `if` somewhere in the codebase

PR #3497, opened 2025-06-19
Add an option to not transform `Activator.CreateInstance<T>()` to `new T()`

DoctorKrolic:
Closes https://github.com/icsharpcode/ILSpy/issues/3495

This is my first contribution to this project, so to found places in code I needed I used full text search on the solution. Besides the place I changed there is also this piece:

https://github.com/icsharpcode/ILSpy/blob/0bfe222cf45b468f618a91c038693b234ee7d838/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs#L3420-L3426

I was never able to hit this code path when I debugged added test case. Does this logic also need to be changed and if so, what test should I add to cover it?


dgrunwald:
I think that extra logic you found is used when dealing with `new T { 1, 2, 3 }` (`where T : IList<int>, new()`).
In that case you can't avoid the transformation to `new T` syntax; you could only disable collection initializer syntax.

DoctorKrolic:
> In that case you can't avoid the transformation to `new T` syntax

Why not? If both collection initializer syntax and `new T` are disabled, we can get this:
Before:
```cs
static T Test<T>()
    where T : IList<int>, new()
{
    return new T()
    {
        1, 2, 3
    };
}
```
After:
```cs
static T Test<T>()
    where T : IList<int>, new()
{
    T t = Activator.CreateInstance<T>();
    t.Add(1);
    t.Add(2);
    t.Add(3);
    return t;
}
```

siegfriedpammer:
I am not a fan of the setting name `GenericTypeInstantiation`...

Copilot says:
> "GenericTypeInstantiation" is the act of creating a constructed type from a generic type definition by supplying actual type arguments (e.g., turning `List<T>` into `List<string>`).

System.Reflection.Metadata uses the term to refer to construction of generic type instances, see `ISignatureTypeProvider.GetGenericInstantiation`. It's also used in that way throughout ILSpy's source code.

We should probably use something like `UseObjectCreationOfGenericTypeParameter` but that is not really succinct.

> Does this logic also need to be changed and if so, what test should I add to cover it?

It's sufficient to add the check here:

https://github.com/icsharpcode/ILSpy/blob/51a26180bba108bfe1266d0d0282ce059c672466/ICSharpCode.Decompiler/IL/Transforms/TransformCollectionAndObjectInitializers.cs#L78-L81

just like we already do for `with` initializers:

https://github.com/icsharpcode/ILSpy/blob/51a26180bba108bfe1266d0d0282ce059c672466/ICSharpCode.Decompiler/IL/Transforms/TransformCollectionAndObjectInitializers.cs#L82-L86

Then, if the setting is disabled, the initializer transform won't be applied in this specific case only.

siegfriedpammer:
Upon reading the code referenced above again, I think this should be implemented by adding `if (!context.Settings.YourOptionName) { return; }` in the respective `case` section, to make it easier to read.

siegfriedpammer (approved):

siegfriedpammer:

PR #3500, opened 2025-06-23
Upgrade dotnet target framework to net10

siegfriedpammer:


ICSharpCode.Decompiler/CSharp/CallBuilder.cs
```diff
@@ -1057,30 +1046,46 @@ private bool TransformParamsArgument(ExpectedTargetDetails expectedTargetDetails
...
+						if (args is [] && method is { IsStatic: true, FullName: "System.Array.Empty", TypeArguments: [var type] })
+						{
+							elementType = type;
+							arguments = new();
+							parameters = new();
+							return true;
+						}
+						if (paramsArgument.Expression is ObjectCreateExpression oce
+							&& method is { IsConstructor: true, DeclaringType: { TypeArguments: [var type2] } declaringType }
+							&& declaringType.IsKnownType(KnownTypeCode.ReadOnlySpanOfT))
```
dgrunwald: This matches every overload of `new ReadOnlySpan<T>`; but I think it's meant to only apply to the `(in T)` overload?

PR #3501, opened 2025-06-23
Make TargetFramework and TargetServices public

nikitalita:
### Problem

Custom project writers cannot currently make use of TargetFramework and TargetServices because they're not public.

My specific use case is: I am using ILSpy in my Godot decompiler to add support for decompiling games with C# scripts. One of the quirks about building a C# project for Godot is that the SDK in the `csproj` is labelled as `Godot.NET.Sdk`, and there are custom assemblies that are automatically included without references, so I had to make a custom project writer for this and pass it into `WholeProjectDecompiler`.

I currently have to copy a bunch of files, including these, because they're not public; since we explicitly provide the ability to pass in a custom project writer to the decompiler, I think that they should be made public so that others can make use of them.

### Solution
* What it says on the tin


siegfriedpammer:

Issue #2875, opened 2022-12-23
Decompile multiple projects under single solution and use ProjectReference's


gerneio:
### Steps to reproduce
1. Use either **a)** the ILSpy WPF desktop app, or **b)** ILSpyCmd w/ 8.0 prev 1 (CLI requires #2364 to support multiple assemblies)
2. Make sure you have at least two DLL's where one of them has a dependency on the other (i.e. `First.dll` depends on `Second.dll`)
3. Using CLI in terminal, execute command: `ilspycmd First.dll Second.dll -p -o TEMP`
4. Open the newly generated solution in VS from the `TEMP` directory
5. Notice in project `First` how the reference to `Second` is a DLL reference, NOT a project reference.

### Expectation
* `First` should see `Second` as a project reference

### Details
The main culprit of this issue is in `ICSharpCode.Decompiler.Solution.SolutionCreator` with the use of `ProjectFileNamespace` which is coded for how this would work with the old project file structure. 

https://github.com/icsharpcode/ILSpy/blob/94982b5368c46f58f0af2bdbd3ba300574e0f8e4/ICSharpCode.Decompiler/Solution/SolutionCreator.cs#L32

https://github.com/icsharpcode/ILSpy/blob/94982b5368c46f58f0af2bdbd3ba300574e0f8e4/ICSharpCode.Decompiler/Solution/SolutionCreator.cs#L163-L165

https://github.com/icsharpcode/ILSpy/blob/94982b5368c46f58f0af2bdbd3ba300574e0f8e4/ICSharpCode.Decompiler/Solution/SolutionCreator.cs#L176-L193

Since the new SDK style is used by default (from CLI or UI), our only option is to manually correct this after decompilation, which is a ton of work. From some quick testing, simply using `string.Empty` in replace of the XML namespace works just fine for the expected behavior to work. However, the correct solution would probably involve finding a way to pull in `ICSharpCode.Decompiler.DecompilerSettings` into `SolutionCreator` somehow so that we can reference the `UseSdkStyleProjectFormat` boolean. 

https://github.com/icsharpcode/ILSpy/blob/94982b5368c46f58f0af2bdbd3ba300574e0f8e4/ICSharpCode.Decompiler/DecompilerSettings.cs#L1868

The only public method of that class is one of the `WriteSolutionFile` overloads...

[...]

PR #3502, opened 2025-07-05
Fix SDK-style (modern) inter-project references

petercrabtree:
[Link to issue(s) this covers](https://github.com/icsharpcode/ILSpy/issues/2875)

### Problem
The XML namespace isn't correctly configured when using new-style solution files.

### Solution

There are three changes in this PR; I bundled them here as they are near in code and I didn't think anyone would object to the cleanup. (Please let me know if you'd rather I split up the changes.)

* Cleanup use of repeated enumeration of an IEnumerable<T> in a public API
* Change project guids to be written in lower-case (this fixes compatibility with some old Microsoft tools, I believe no other common tools care either way, and the uuid RFC specifies lower-case)
* Detects whether the solution is an SDK-style by examining its xml and chooses the XML namespace appropriately (this is the fix for issue 2875)

### DIFF
--- a/ICSharpCode.Decompiler/Solution/SolutionCreator.cs
+++ b/ICSharpCode.Decompiler/Solution/SolutionCreator.cs
@@ -193,7 +193,12 @@ namespace ICSharpCode.Decompiler.Solution
 		static void FixProjectReferences(string projectFilePath, XElement itemGroup,
 			Dictionary<string, ProjectItem> projects, bool sdkStyle)
 		{
-			var referenceTagName = sdkStyle ? "Reference" : ProjectFileNamespace + "Reference";
+			XName GetElementName(string localName) => sdkStyle ? localName : ProjectFileNamespace + localName;
+
+			var referenceTagName = GetElementName("Reference");
+			var projectReferenceTagName = GetElementName("ProjectReference");
+			var projectTagName = GetElementName("Project");
+			var nameTagName = GetElementName("Name");
 
 			foreach (var item in itemGroup.Elements(referenceTagName).ToList())
 			{
@@ -202,9 +207,9 @@ namespace ICSharpCode.Decompiler.Solution
 				{
 					item.Remove();
 
-					var projectReference = new XElement(ProjectFileNamespace + "ProjectReference",
-						new XElement(ProjectFileNamespace + "Project", referencedProject.Guid.ToString("B").ToLowerInvariant()),
-						new XElement(ProjectFileNamespace + "Name", referencedProject.ProjectName));
+					var projectReference = new XElement(projectReferenceTagName,
+						new XElement(projectTagName, referencedProject.Guid.ToString("B").ToLowerInvariant()),
+						new XElement(nameTagName, referencedProject.ProjectName));
 					projectReference.SetAttributeValue("Include", GetRelativePath(projectFilePath, referencedProject.FilePath));
 
 					itemGroup.Add(projectReference);
--- a/ICSharpCode.Decompiler/Solution/SolutionCreator.cs
+++ b/ICSharpCode.Decompiler/Solution/SolutionCreator.cs
@@ -29,7 +29,7 @@ namespace ICSharpCode.Decompiler.Solution
 	/// </summary>
 	public static class SolutionCreator
 	{
-		static readonly XNamespace ProjectFileNamespace = XNamespace.Get("http://schemas.microsoft.com/developer/msbuild/2003");
+		static readonly XNamespace NonSDKProjectFileNamespace = XNamespace.Get("http://schemas.microsoft.com/developer/msbuild/2003");
 
 		/// <summary>
 		/// Writes a solution file to the specified <paramref name="targetFile"/>.
@@ -63,7 +63,7 @@ namespace ICSharpCode.Decompiler.Solution
 				WriteSolutionFile(writer, projects, targetFile);
 			}
 
-			FixProjectReferences(projects);
+			FixAllProjectReferences(projects);
 		}
 
 		static void WriteSolutionFile(TextWriter writer, List<ProjectItem> projects, string solutionFilePath)
@@ -153,7 +153,7 @@ namespace ICSharpCode.Decompiler.Solution
 			writer.WriteLine("\tEndGlobalSection");
 		}
 
-		static void FixProjectReferences(List<ProjectItem> projects)
+		static void FixAllProjectReferences(List<ProjectItem> projects)
 		{
 			var projectsMap = projects.ToDictionary(
 				p => p.ProjectName,
@@ -170,9 +170,11 @@ namespace ICSharpCode.Decompiler.Solution
 						$"no <Project> at the root; could not fix project references.");
 				}
 
+				// sdk style projects don't use a namespace for the elements,
+				// but we still need to use the namespace for non-sdk style projects.
 				var sdkStyle = projectDoc.Root.Attribute("Sdk") != null;
-				var itemGroupTagName = sdkStyle ? "ItemGroup" : ProjectFileNamespace + "ItemGroup";
-				var referenceTagName = sdkStyle ? "Reference" : ProjectFileNamespace + "Reference";
+				var itemGroupTagName = sdkStyle ? "ItemGroup" : NonSDKProjectFileNamespace + "ItemGroup";
+				var referenceTagName = sdkStyle ? "Reference" : NonSDKProjectFileNamespace + "Reference";
 
 				var referencesItemGroups = projectDoc.Root
 					.Elements(itemGroupTagName)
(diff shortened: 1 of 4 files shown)


ICSharpCode.Decompiler/Solution/SolutionCreator.cs
```diff
@@ -52,20 +53,22 @@ public static void WriteSolutionFile(string targetFile, IEnumerable<ProjectItem>
 				throw new ArgumentNullException(nameof(projects));
 			}
 
-			if (!projects.Any())
+			var projectList = projects.ToList();
```
siegfriedpammer: I am not exactly sure why the parameter is using `IEnumerable` here, I think it would make sense to just change the parameter type and/or add an overload, while we're at it. This change is going into the next major release, so breaking changes are allowed.

ICSharpCode.Decompiler/Solution/SolutionCreator.cs
```diff
@@ -164,30 +165,45 @@ static void FixProjectReferences(List<ProjectItem> projects)
...
 				projectDoc.Save(project.FilePath);
 			}
 		}
 
-		private static void FixProjectReferences(string projectFilePath, XElement itemGroup, IDictionary<string, ProjectItem> projects)
+		static void FixProjectReferences(string projectFilePath, XElement itemGroup,
+			Dictionary<string, ProjectItem> projects, bool sdkStyle)
 		{
-			foreach (var item in itemGroup.Elements(ProjectFileNamespace + "Reference").ToList())
+			var referenceTagName = sdkStyle ? "Reference" : ProjectFileNamespace + "Reference";
```
siegfriedpammer: This condition could be extracted into a local function like this?

`string GetElementName(string name) => sdkStyle ? name : ProjectFileNamespace + name;`

siegfriedpammer: ah, never mind, the other occurrences are in different methods.

ICSharpCode.Decompiler/Solution/SolutionCreator.cs
petercrabtree: I'm happy to just change it to take a List<> if breaking changes are fine. An API purist might suggest IReadOnlyCollection<> but here I slightly prefer giving the compiler all the help it can get devirtualizing (in software like this where perf is very nice).

(Incidentally, the "IEnumerable<> API which is .ToList() immediately" pattern is less bad than it might seem at first glance -- .ToList() won't reallocate if the IEnumerable<> reference represents a List<> already (at least, that's how it worked last time I checked).)

ICSharpCode.Decompiler/Solution/SolutionCreator.cs
```diff
@@ -152,38 +155,55 @@ private static void WriteProjectConfigurations(
...
 			{
 				var assemblyName = item.Attribute("Include")?.Value;
 				if (assemblyName != null && projects.TryGetValue(assemblyName, out var referencedProject))
 				{
 					item.Remove();
 
 					var projectReference = new XElement(ProjectFileNamespace + "ProjectReference",
-						new XElement(ProjectFileNamespace + "Project", referencedProject.Guid.ToString("B").ToUpperInvariant()),
+						new XElement(ProjectFileNamespace + "Project", referencedProject.Guid.ToString("B").ToLowerInvariant()),
 						new XElement(ProjectFileNamespace + "Name", referencedProject.ProjectName));
```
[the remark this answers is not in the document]
petercrabtree: I think this is a good point. Might as well get the project file syntax right. I'll fix this too.

petercrabtree:
I haven't tested these two changes yet, but thought I'd throw them up here so you can see.

ICSharpCode.Decompiler/Solution/SolutionCreator.cs
petercrabtree: turns out it was useful after all :)

ILSpy/Util/SettingsService.cs
petercrabtree, line 1: don't know why dotnet format whitespace ilspy.sln thought this needed doing, but here it is anyway

petercrabtree:
This version isn't right; sdk-style projects shouldn't be writing `<Project>` or `<Name>` tags under `<ProjectReference>`. Marking draft until I fix.

ICSharpCode.Decompiler/Solution/SolutionCreator.cs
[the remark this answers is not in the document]
petercrabtree: Or, rather, was adjacent to a good point, copilot actually pointed out an issue that didn't exist. However, the `<Project>` and `<Name>` elements shouldn't be written at all for SDK-style projects.

ICSharpCode.Decompiler/Solution/SolutionCreator.cs
```diff
@@ -152,47 +153,80 @@ private static void WriteProjectConfigurations(
...
+						new XAttribute("Include", GetRelativePath(projectFilePath, referencedProject.FilePath)));
+
+					// SDK-style projects do not use the <Project> and <Name> elements for project references.
+					// (Instead, those get read from the .csproj file in "Include".)
+					if (!sdkStyle)
+					{
+						projectReference.Add(
+							// no ToUpper() for uuids, most Microsoft tools seem to emit them in lowercase
+							// (no .ToLower() as .ToString("B") already outputs lowercase)
+							new XElement(NonSDKProjectFileNamespace + "Project", referencedProject.Guid.ToString("B")),
```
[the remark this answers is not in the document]
petercrabtree: wrong

petercrabtree:
The pull request was misnamed -- it does affect ilspycmd (when decompiling to a solution, that is, using --project/-p with more than 1 assembly), but it changes [ICSharpCode.Decompiler/Solution/SolutionCreator.cs](https://github.com/icsharpcode/ILSpy/pull/3502/files/9788e44cfd256607951b8fe6f28c280a561efbdf#diff-3d92b3e87a47692c30571d1ab326351173358a67f9fe309b755d7d0fdf44a714), which is also called by the ILSpy GUI.

# PR Summary

Fixes the `<ProjectReference>` XML written to csproj files for SDK-style projects. Addresses #2875.

## Breaking Change
- **API Cleanup** – `WriteSolutionFile` (a *public* method downstream frontends presumably use) now accepts a `List<T>` instead of `IEnumerable<T>`; the signature change cascades to all dependent methods.

## Enhancement
- **Lowercase ProjectReference Project Guids** – Non-SDK `<ProjectReference><Project>{GUID}</Project>` values are now emitted *lower-case*, matching how Visual Studio 2022 (17.14.5) writes them (and the UUID RFC, 4122).  
  _Reference files:_ [csproj_and_sln_files_from_ms_tools.zip](https://github.com/user-attachments/files/21110140/csproj_and_sln_files_from_ms_tools.zip)

## Fixes
- **Correct XML namespace for SDK projects** – The writer now detects SDK style via XML inspection and chooses the proper namespace.
- **Clean SDK-style references** – `<Project>` and `<Name>` elements are no longer emitted for SDK `<ProjectReference>` entries, aligning with MSBuild conventions. (Nothing seems to object to those elements being there in my testing, so this is more of a cleanup.)

petercrabtree:
... aaand it looks like my lf vs crlf fixes aren't quite right (I'm using an SCM that doesn't respect git's autocrlf setting). Apologies, fixing.

petercrabtree:
The pull request was misnamed -- it does affect ilspycmd (when decompiling to a solution, that is, using --project/-p with more than 1 assembly), but it changes [ICSharpCode.Decompiler/Solution/SolutionCreator.cs](https://github.com/icsharpcode/ILSpy/pull/3502/files/9788e44cfd256607951b8fe6f28c280a561efbdf#diff-3d92b3e87a47692c30571d1ab326351173358a67f9fe309b755d7d0fdf44a714), which is also called by the ILSpy GUI.

# PR Summary

Fixes the `<ProjectReference>` XML written to csproj files for SDK-style projects. Addresses #2875.

## Breaking Change
- **API Cleanup** – `WriteSolutionFile` (a *public* method downstream frontends presumably use) now accepts a `List<T>` instead of `IEnumerable<T>`; the signature change cascades to all dependent methods.

## Enhancement
- **Lowercase ProjectReference Project Guids** – Non-SDK `<ProjectReference><Project>{GUID}</Project>` values are now emitted *lower-case*, matching how Visual Studio 2022 (17.14.5) writes them (and the UUID RFC, 4122).  
  _Reference files:_ [csproj_and_sln_files_from_ms_tools.zip](https://github.com/user-attachments/files/21110140/csproj_and_sln_files_from_ms_tools.zip)

## Fixes
- **Correct XML namespace for SDK projects** – The writer now detects SDK style via XML inspection and chooses the proper namespace.
- **Clean SDK-style references** – `<Project>` and `<Name>` elements are no longer emitted for SDK `<ProjectReference>` entries, aligning with MSBuild conventions. (Nothing seems to object to those elements being there in my testing, so this is more of a cleanup.)

petercrabtree:
They say 7th force push is the charm, right?

Ok, finally correctly removed the spurious change to SettingsService.cs, the diffs in github.com look good, and the build actions in my fork run cleanly. Everything should be good for review.

siegfriedpammer:

Issue #3494, opened 2025-06-15
decompiler: stable attributes ordering


StalkR:
Using: ilspycmd 9.1.0.8007 from latest HEAD 57cd68fc1ffff48d8307d304e321e4de07d3394d as of today.

### Problem

I'm decompiling a (private) program across updates and reviewing the differences. The program uses attributes, and for some surely valid technical reasons either on the compiler or decompiler side, the order changes in the decompiler output.

Example 1:
```
[A]
[B]
void Method()
```

Example 2:
```
[B]
[A]
void Method()
```

Diff:
```
-[A]
 [B]
+[A]
void Method()
```

This produces unnecessary noise in the difference, because nothing significantly changed. I'd like to reduce this noise to make code reviews across updates easier.

According to the [C# attribute specification](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/attributes#223-attribute-specification), the order is not significant:

> The order in which attributes are specified in such a list, and the order in which sections attached to the same program entity are arranged, is not significant. For instance, the attribute specifications [A][B], [B][A], [A, B], and [B, A] are equivalent.

### Proposed solution

Implement a stable attribute ordering, for instance a regular alphabetical sort.

If you agree in principle, I'd be happy to work on this and send a pull request. However, I'm not familiar with the codebase, so I welcome any pointers where the code changes would be best made.

Thank you!

siegfriedpammer:
Duplicate of #2716

![Image](https://github.com/user-attachments/assets/8539738d-53e0-4f3f-8656-29cf1ac67005)

siegfriedpammer:
@christophwille do we have a good strategy to get this into ilspycmd as well? I don't think it's sustainable to always keep patching this on demand.

StalkR:
Thanks Siegfried for the quick reply. My bad, but good news, didn't realize this was already done! So it's just a matter of plumbing it from ilspycmd (like a cmd line flag) to the decompiler object options? I'll wait for Christoph's response.

Any reason why we shouldn't make this the default? (at least for ilspycmd, but maybe ILSpy too) Unlike some of the other options, it doesn't seem too controversial given the specification's position on ordering being non-significant. I doubt anyone would rely on non-ordering, but, who knows.

christophwille:
[...]

PR #3503, opened 2025-07-05
Add sort custom attributes option to ILSpyCmd

MortalAndTry:
Link to issue(s) this covers

### Problem
ILSpyCmd: Add parameter to support attribute sorting

### Solution
Add sort custom attributes option to ILSpyCmd

### DIFF
--- a/ICSharpCode.ILSpyCmd/IlspyCmdProgram.cs
+++ b/ICSharpCode.ILSpyCmd/IlspyCmdProgram.cs
@@ -113,6 +113,9 @@ Examples:
 		[Option("--no-dead-stores", "Remove dead stores.", CommandOptionType.NoValue)]
 		public bool RemoveDeadStores { get; }
 
+		[Option("--sort-attributes", "Sort custom attributes.", CommandOptionType.NoValue)]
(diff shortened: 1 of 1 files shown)


siegfriedpammer:
See #3494 @christophwille I think we want an option that enables reading all options from an ILSpy.xml file, right?

christophwille:
Yes, the idea was to not add options piecemeal but in one go via the standard options file format.

MortalAndTry:
I feel it would be better to have both configuration methods available. If complex configurations can only be done through files, it feels a bit bloated and cumbersome, especially in automation scenarios. This means users need to maintain one or multiple configuration files.  Anyway,  respect your thoughts

christophwille:
> I feel it would be better to have both configuration methods available. If complex configurations can only be done through files, it feels a bit bloated and cumbersome, especially in automation scenarios. This means users need to maintain one or multiple configuration files. Anyway, respect your thoughts

The only way to properly do "one cmd line switch per decompiler option" would be to add them dynamically (if that is possible at all). That way it would always be up to date (think "--dco-REFLECTEDNAMEOFOPTION")

PR #3504, opened 2025-07-05
Fix incorrect method match for static methods with instance signature

MSchmoecker:
### Problem
ILSpy could match an instance method call with a static method reference. This can happen when resolving a method whose *current* definition is `static` but the IL actually calls an instance method. I.e. the method was originally an instance method but was changed to `static`.
This results in the decompiled code to silently drop the instance argument on the evaluation stack. While this produces valid/compilible C# code, it does not accurately decompile the given code.

Example:

the reference code was changed from
```cs
class A {
    static A instance;
    void Ref() {}
}
```
to
```cs
class A {
    static A instance;
    static void Ref() {}
}
```

the original code is
```cs
A.instance.Ref();
```
but gets wrongly decompiled to
```cs
A.Ref(); // should still be A.instance.Ref();
```

### Solution
* The given code searches for a matching reference method. Since static and non-static are not matching, we are excluding those as potential matches. In our case this means we don't match the `Ref` method, so ILSpy decompiles the code without a reference and produces `A.instance.Ref()`
* I am not sure how to write a test because the compile and decompile step require different reference assemblies

### DIFF
--- a/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
+++ b/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
@@ -96,6 +96,7 @@
     <None Include="TestCases\ILPretty\Issue2260SwitchString.il" />
     <None Include="TestCases\ILPretty\Issue3442.il" />
     <None Include="TestCases\ILPretty\Issue3466.il" />
+    <None Include="testcases\ilpretty\Issue3504.il" />
     <None Include="TestCases\ILPretty\MonoFixed.il" />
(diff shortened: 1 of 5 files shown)


siegfriedpammer:
~~You can write a correctness test based on IL files.~~

Yes, it needs two separate assemblies.

dgrunwald:
I think you could maybe test this with a `.il` based test that calls a method in the .NET stdlib, e.g. a hand-written `call` instruction to `Console.WriteLine` that expects WriteLine to be an instance method.

This would have to be a pretty-test, as it would not be possible to recompile the decompiled code.

siegfriedpammer:

MSchmoecker:
Thanks so much for providing the test case and merging!

PR #3505, opened 2025-07-06
Add dynamic ILSpy settings options to ILSpyCmd

### LINKED ISSUE #3494: decompiler: stable attributes ordering

siegfriedpammer:
Fixes #3494

Still not really user-friendly, but it's a start...

### DIFF
--- a/ILSpy/Util/SettingsService.cs
+++ b/ILSpy/Util/SettingsService.cs
@@ -16,71 +16,18 @@
 // OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
-using System;
-using System.Collections.Concurrent;
 using System.ComponentModel;
-using System.Xml.Linq;
 
 using ICSharpCode.ILSpy.Options;
 using ICSharpCode.ILSpyX;
 using ICSharpCode.ILSpyX.Settings;
 
-using DecompilerSettings = ICSharpCode.ILSpy.Options.DecompilerSettings;
+using DecompilerSettings = ICSharpCode.ILSpyX.Settings.DecompilerSettings;
 
 #nullable enable
 
 namespace ICSharpCode.ILSpy.Util
 {
-	public interface IChildSettings
-	{
-		ISettingsSection Parent { get; }
-	}
-
-	public interface ISettingsSection : INotifyPropertyChanged
-	{
-		XName SectionName { get; }
-
-		void LoadFromXml(XElement section);
-
-		XElement SaveToXml();
-	}
-
-	public abstract class SettingsServiceBase(ISettingsProvider spySettings)
-	{
-		protected readonly ConcurrentDictionary<Type, ISettingsSection> sections = new();
-
-		protected ISettingsProvider SpySettings { get; set; } = spySettings;
-
-		public T GetSettings<T>() where T : ISettingsSection, new()
-		{
-			return (T)sections.GetOrAdd(typeof(T), _ => {
-				T section = new T();
-
-				var sectionElement = SpySettings[section.SectionName];
-
-				section.LoadFromXml(sectionElement);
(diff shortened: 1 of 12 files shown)


siegfriedpammer:
@StalkR @MortalAndTry you might want to try this before it gets merged. Any feedback is appreciated!

ILSpy/Util/SettingsService.cs
```diff
@@ -173,7 +120,8 @@ protected override void Section_PropertyChanged(object? sender, PropertyChangedE
 					SpySettings.Update(root => {
 						SaveSection(section, root);
 					});
-				};
+				}
+				;
```
StalkR: is the empty line with a single `;` intended?

StalkR:
Thanks @siegfriedpammer!

I haven't tried `--ilspy-settingsfile` but `-ds|--decompiler-setting name=value` is most convenient for command-line users, thank you.
I tried by taking the `name=value` from the names in https://github.com/icsharpcode/ILSpy/blob/master/ICSharpCode.Decompiler/DecompilerSettings.cs#L2220 and it seems to be accepted, while other invalid values are rejected as expected:
```
$ dotnet build --configuration Release ICSharpCode.ILSpyCmd/ICSharpCode.ILSpyCmd.csproj

$ ICSharpCode.ILSpyCmd/bin/Release/net10.0/ilspycmd --disable-updatecheck -ds SortCustomAttributes=true ICSharpCode.ILSpyCmd/bin/Release/net10.0/ilspycmd.dll

$ ICSharpCode.ILSpyCmd/bin/Release/net10.0/ilspycmd --disable-updatecheck -ds invalid=true ICSharpCode.ILSpyCmd/bin/Release/net10.0/ilspycmd.dll
Decompiler setting 'invalid' is unknown.
[...]
```

I'll do more testing and report if anything, but otherwise LGTM so far! Thanks again.

ILSpy/Util/SettingsService.cs
siegfriedpammer: This is a dotnet format quirk... Also occurring here: https://github.com/icsharpcode/ILSpy/pull/3502#pullrequestreview-2990850667

StalkR: ok, I noticed format just changed `};` to `}\n;` but I thought the `;` remained unnecessary here either way as it's after an `if`. I wanted to try if it still compiles without the `;` but .NET 10 preview is being difficult (https://github.com/dotnet/sdk/issues/48949).

siegfriedpammer: ah, it's an `if` 😅 ... then `dotnet format` is correct, let's remove the empty statement. 👍

siegfriedpammer: this even confuses GitHub Copilot 😆 

image

PR #3387, opened 2025-02-03
New NUnit runner for Microsoft.Testing.Platform

christophwille:
https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-nunit-runner-intro

Microsoft Testing Platform 
https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-platform-intro?tabs=dotnetcli

image

PR #3507, opened 2025-07-07
.NET 10 MTP dotnet test

christophwille:
PoC see https://github.com/christophwille/poc-oh/tree/main/src/MTPEntrypoint

Docs

* https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-nunit-runner-intro
* https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-test?tabs=dotnet-test-with-vstest

Using trx extension because https://github.com/icsharpcode/ILSpy/commit/c6b6c09cfcc8a10940d913c2d6f11342b7fddb54 switched from junit format to trx format.

Supersedes #3387

Optional in VS Preview

image


christophwille:
Local output:

```
  In process file artifacts produced:
    - ILSpy.Tests\bin\Debug\net10.0-windows\TestResults\test-results.trx
    - ILSpy.BamlDecompiler.Tests\bin\Debug\net10.0-windows\win-x64\TestResults\test-results.trx
    - ICSharpCode.Decompiler.Tests\bin\Debug\net10.0-windows\win-x64\TestResults\test-results.trx

Test run summary: Passed!
  ILSpy.Tests\bin\Debug\net10.0-windows\ILSpy.Tests.exe (net10.0|x64) passed (1s 590ms)
  ICSharpCode.Decompiler.Tests\bin\Debug\net10.0-windows\win-x64\ICSharpCode.Decompiler.Tests.exe (net10.0|x64) passed (9m 45s 002ms)
  ILSpy.BamlDecompiler.Tests\bin\Debug\net10.0-windows\win-x64\ILSpy.BamlDecompiler.Tests.exe (net10.0|x64) passed (2s 708ms)

  total: 2685
  failed: 0
  succeeded: 2672
  skipped: 13
  duration: 9m 46s 681ms
```
Previously, a single trx file was created. @Youssef1313 is it possible like before to create a single trx for the entire `dotnet test` run?

Youssef1313:
So far we don't have post-processing of artifacts, so it's a TRX per test assembly. But you could search for tools out there that merges TRX files, you may be able to find some third-party tools that do it.

christophwille:
> So far we don't have post-processing of artifacts, so it's a TRX per test assembly. But you could search for tools out there that merges TRX files, you may be able to find some third-party tools that do it.

I switched to --results-directory - but the files are named quite ugly

```
  In process file artifacts produced:
    - test-results\christophw_ZORT_2025-07-07_09_14_52.226.trx
    - test-results\christophw_ZORT_2025-07-07_09_14_52.910.trx
    - test-results\christophw_ZORT_2025-07-07_09_23_12.688.trx
```

Is there a way to include the project name?

Youssef1313:
@christophwille We don't yet have placeholders for `--report-trx-filename`, BUT, you can still definitely achieve it. Something in Directory.Build.props like this:

```xml
<PropertyGroup Condition="'$(UseReportTrx)' == 'true'">
    <TestingPlatformCommandLineArguments>$(TestingPlatformCommandLineArguments) --report-trx --report-trx-filename $(MSBuildProjectName)</TestingPlatformCommandLineArguments>
</PropertyGroup>
```

Then test with `dotnet test -p:UseReportTrx=true`

Note I'm writing directly on GitHub so I could have errors.

christophwille:
@siegfriedpammer is it a problem if the test-results files have funky names? I mean we know the biggest one is from ics.d.

ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
```diff
@@ -13,7 +13,12 @@
     <RuntimeIdentifier Condition="$(IsWindowsARM64) == true">win-arm64</RuntimeIdentifier>
 
     <IsPackable>false</IsPackable>
-    <StartupObject>AutoGeneratedProgram</StartupObject>
+    
+    <GenerateTestingPlatformEntryPoint>true</GenerateTestingPlatformEntryPoint>
+    <StartupObject>ICSharpCode.Decompiler.Tests.MicrosoftTestingPlatformEntryPoint</StartupObject>
+    <EnableNUnitRunner>true</EnableNUnitRunner>
+    <TestingPlatformDotnetTestSupport>true</TestingPlatformDotnetTestSupport>
```
Youssef1313: This property is no longer required if you are using the new `dotnet test` in .NET 10 SDK.

ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
```diff
@@ -13,7 +13,12 @@
     <RuntimeIdentifier Condition="$(IsWindowsARM64) == true">win-arm64</RuntimeIdentifier>
 
     <IsPackable>false</IsPackable>
-    <StartupObject>AutoGeneratedProgram</StartupObject>
+    
+    <GenerateTestingPlatformEntryPoint>true</GenerateTestingPlatformEntryPoint>
+    <StartupObject>ICSharpCode.Decompiler.Tests.MicrosoftTestingPlatformEntryPoint</StartupObject>
+    <EnableNUnitRunner>true</EnableNUnitRunner>
```
Youssef1313: This should be moved to Directory.Build.props

ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
christophwille: I took 19-20 from https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-nunit-runner-intro. So you are saying 20 is no longer needed?

ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
christophwille: And what would be a good condition for it to only apply for test projects?

ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
Youssef1313: Yup, `TestingPlatformDotnetTestSupport` is only needed if you are using the old dotnet test (either because you are using .NET 9 SDK or older, or you didn't opt-in via dotnet.config). But as you are using the new dotnet test (using .NET 10 SDK and opting-in in dotnet.config), this property is irrelevant.

ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
Youssef1313: You don't need to apply it to test projects. For non-test projects, the property will never be read so it's okay to be there.

siegfriedpammer (approved):

ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
```diff
@@ -13,7 +13,10 @@
     <RuntimeIdentifier Condition="$(IsWindowsARM64) == true">win-arm64</RuntimeIdentifier>
 
     <IsPackable>false</IsPackable>
-    <StartupObject>AutoGeneratedProgram</StartupObject>
+    
+    <GenerateTestingPlatformEntryPoint>true</GenerateTestingPlatformEntryPoint>
```
Youssef1313, line 17: This is true by default. It's strange if you don't get the generation automatically and there may be a bug somewhere.

christophwille: Yes, I know - this is just to make it explicit that we want the override for the generated one.

PR #3517, opened 2025-07-17
Test Roslyn 5 from dotnet-tools transport feed

christophwille:
PoC to check if anything major is coming our way.

The VSIX release build break is expected bc only official nuget feed has release versions that are usually referenced. We are interested in test breaks for this one


christophwille:
image

Looking at that ... no more SCI and SRM deps?

dgrunwald:
System.Collections.Immutable and System.Reflection.Metadata are part of the .NET runtime nowadays, so you only see a dependency if a newer version is needed than .NET provides.

PR #3523, opened 2025-07-25
Slot-based C# syntax tree via a Roslyn source generator

siegfriedpammer:
## Overview

This reworks the C# AST in `ICSharpCode.Decompiler/CSharp/Syntax/` from the NRefactory
**role + doubly-linked-list** child model to a **slot-based model** (in the spirit of the ILAst),
with the per-node boilerplate emitted by a Roslyn **source generator**, and sheds the NRefactory
machinery the decompiler does not need. The end state also turns on nullable reference types across
the AST and its consumers.

Each child of a node is declared once as a `[Slot]` partial property; the generator owns everything
mechanical derived from those declarations.

## What changes

- **Source generator** for the AST: from `[DecompilerAstNode]`/`[Slot]` declarations it emits the
  visitor interface, `AcceptVisitor`, `DoMatch`, the slot accessors and flattened child-index API
  (`GetChildCount`/`GetChild`/`SetChild`/`GetChildSlotInfo`/`GetCollectionByKind`/`CloneChildrenInto`),
  a single-pass child enumerator, the per-node constructors (parameters in source order), and
  `node.Slot`, the per-node `CSharpSlotInfo<T>`, and a generated `Slots` holder of the canonical kinds. A `[Slot]` on an `AstNode`-typed property
  is a child slot; a `[Slot]` on a `string` property is a *name* -- a convenience string over a
  backing `Identifier` token slot the generator owns (the property type disambiguates, so no separate
  attribute is needed).
- **Storage flip**: children move from a per-node linked list to real slot backing fields, with
  `AstNodeCollection<T>` backed by a `List<T>`.
- **`CheckInvariant`** (DEBUG only, the analog of `ILInstruction.CheckInvariant`): runs before the
  transform pipeline and after every transform. It recursively verifies the slot structure (each
  child's parent back-pointer, its flattened index, and its slot type) and that every required
  (non-optional) slot is filled, and node types override it to assert their own scalar constraints --
  the invariants formerly enforced by throwing setters (e.g. `ComposedType.PointerRank >= 0`,
  `ArraySpecifier.Dimensions >= 1`) now read as auto-properties checked there.
- **Drop token nodes**: `CSharpTokenNode`/`CSharpModifierToken` and `InsertSpecialsDecorator` are
  gone. Punctuation/keyword/operator text lives on the nodes; modifiers, specifiers, the await flag,
  etc. become scalar fields. Source locations are stored at print time instead of recomputed from
  child token nodes.
- **Comments and directives become trivia** on a leading/trailing side-channel, off the child-index
  space (like annotations), rather than positional children.
- **Null-objects to nullable reference types**: optional single-child slots become `T?` (with the C#
  grammar as the oracle for optionality), the null-object pattern is deleted, and `#nullable enable`
  is turned on across the transforms, the AST consumer layer (output visitor, the IL-to-C# builders,
  `TypeSystemAstBuilder`, the translation-result wrappers, sequence-point/namespace collectors, the
  annotation helpers), and the syntax node files themselves. Optional *names* follow suit: a name that
  may be absent is typed `string?` and reads as `null` when absent (rather than a non-null empty-string
  sentinel); required names stay non-null `string`.
- **Delete `Role`/`Role<T>` and `TokenRole`**: child positions are keyed on typed slot objects --
  `node.GetChild(Slots.Left)` infers the child type and `node.Slot.Kind == Slots.Initializer` is matched
  by object identity (there is no kind enum); token text is carried as plain constants.
- **Pattern matcher** ported to list-indexed slot children (slim `INode`, list-by-index collection
  matching, with a nullable `other` so a pattern can match against an absent child); generated
  `DoMatch` matches every structural member, which fixed a few latent under-matching bugs.
- **Misc**: AST node classes are annotated with their C# grammar productions; trivial scalar
  properties become auto-properties; leaf node classes are sealed; dead parsing-only state and the
  `NodeType` enum are removed. The now-unused per-node `flags` word was dropped (its only user,
  `Identifier.IsVerbatim`, is a plain `bool`), which the CLR's field-padding makes free on the node
  carrying it while shrinking every other node by 8 bytes -- a measured win that a planned bit-packing
  scheme did not deliver (the CLR already packs scalars into padding).

## Behavior and testing

This is a representation change, not an output change: the **Pretty test suite stays byte-identical**
throughout, and the full decompiler suite is green on top of current `master`. Each commit builds the
decompiler libraries, `ilspycmd`, and the ILSpy UI, and the suite is green per logical step. Generated
`DoMatch` made several matchers stricter where the hand-written ones under-matched; those are captured
by the pattern-matching tests.

## Notes

- The history is organised into logical, self-contained commits; the early commits are the original
  generator bring-up, the later ones the migration on top.
- Nullable reference types are enabled across the AST and its `CSharp/` consumers; the semantic
  resolver layer (`CSharp/Resolver/`) is intentionally out of scope here.
- [`CSharp/Syntax/README.md`](ICSharpCode.Decompiler/CSharp/Syntax/README.md) documents the generator,
  its attributes, the slot/kind model, `CheckInvariant`, and how to add a node.


mkhall300-code (approved):

ICSharpCode.Decompiler.Generators/DecompilerSyntaxTreeGenerator.cs
```diff
@@ -0,0 +1,843 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Text;
+
+namespace ICSharpCode.Decompiler.Generators;
+
+[Generator]
+internal class DecompilerSyntaxTreeGenerator : IIncrementalGenerator
+{
+	record AstNodeAdditions(string NodeName, bool NeedsVisitor, bool IsAbstract, bool BaseHasDefaultConstructor, bool NeedsPatternPlaceholder, string VisitMethodName, string VisitMethodParamType, EquatableArray<(string Member, string TypeName, bool RecursiveMatch, bool MatchAny, bool Nullable)>? MembersToMatch, EquatableArray<(bool IsCollection, string PropertyName, string PropertyType, string ElementType, bool IsOverride, bool IsNullable, string KindName, bool IsPartial)>? Slots, EquatableArray<(string StringName, string TokenName, bool IsOptional)>? NameAccessors, EquatableArray<(string PropertyName, string ParamType, string ElementType, bool IsCollection, bool IsOptional)>? CtorParams);
```
siegfriedpammer: Extract the nested tuples into record structs for readability

ICSharpCode.Decompiler.Generators/DecompilerSyntaxTreeGenerator.cs
```diff
@@ -0,0 +1,843 @@
...
+
+	// Derives the shared SlotKind name from a [Slot] string. Those strings are already bare
+	// kind names (e.g. "Body", "Getter"), so this is mostly identity; the last-dotted-segment and
+	// trailing-"Role" stripping below is defensive against any legacy dotted/Role-suffixed form. Names
+	// shared across node types (aliases, or the same logical slot on different nodes) intentionally
+	// collapse to one kind, so node.Slot.Kind is shared and consumers compare against SlotKind.X.
+	static string SlotKindName(string roleExpr)
```
siegfriedpammer: I think this is no longer necessary, since we already stripped any dots and "Role" from the slot names? If there are still slots with dots or "Role" in them change the use-sites and then remove this method.

ICSharpCode.Decompiler.Generators/RoslynHelpers.cs
```diff
@@ -0,0 +1,21 @@
+using Microsoft.CodeAnalysis;
```
siegfriedpammer, line 19: missing file header

ICSharpCode.Decompiler.Generators/IsExternalInit.cs
```diff
@@ -0,0 +1,3 @@
+namespace System.Runtime.CompilerServices;
```
siegfriedpammer, line 19: missing file header

ICSharpCode.Decompiler.Tests/Output/LocationsInAstSamples.cs
```diff
@@ -0,0 +1,72 @@
+using System;
```
siegfriedpammer, line 19: move this file next to "Output/LocationsInAstTests.cs` .

siegfriedpammer, line 19: license header

ROLES_FREE_SLOT_AST_DESIGN.md
```diff
@@ -0,0 +1,925 @@
+# Roles-free slot-based C# AST -- complete design + implementation plan
```
siegfriedpammer: remove this

PHASE_3C_TODO.md
```diff
@@ -0,0 +1,103 @@
+# Phase 3c -- storage flip (linked list -> slot fields). Fresh-session execution checklist.
```
siegfriedpammer: remove this

Spikes/phaseS-workflow.js
```diff
@@ -0,0 +1,166 @@
+export const meta = {
```
siegfriedpammer: remove this

ICSharpCode.Decompiler.Generators/DecompilerSyntaxTreeGenerator.cs
```diff
@@ -0,0 +1,849 @@
+using System.Collections;
```
siegfriedpammer, line 19: file header missing

ICSharpCode.Decompiler/CSharp/Syntax/ComposedType.cs
```diff
@@ -198,19 +120,22 @@ public override AstType MakeRefType()
 			this.HasRefSpecifier = true;
 			return this;
 		}
+
+		internal override void CheckInvariant()
+		{
+			base.CheckInvariant();
+			// PointerRank is the number of '*' specifiers; its setter rejects negatives, so a transform
+			// that corrupts it (e.g. an unbalanced decrement) is caught here at the offending transform.
```
siegfriedpammer: Stale comment: `PointerRank` is now a plain auto-property (line 74) with no rejecting setter — line 73 already states the actual mechanism ("verified by CheckInvariant rather than an eager setter guard"). This comment describes a guard that no longer exists.

siegfriedpammer (commented):

---

Reviewed the full diff with focused passes over the source generator, the AST runtime, the pattern matcher, output/locations, the null-object→nullable migration, and the node declarations.

Built clean in Release and Debug (0 warnings / 0 errors, no `packages.lock.json` churn). Ran the in-memory AST tests locally: PatternMatching and LocationsInAst are green (35/36; the one failure is a missing legacy reference-assembly in my environment, not the change). The matcher's index-alignment, checkpoint save/restore, and null handling also pass an out-of-process probe harness.

No correctness blockers. The inline notes are minor: one stale comment, two `DoMatch` tightenings that have no live consumer, an enumerator-semantics narrowing, a narrowed backreference contract, and latent (currently-unread) metadata in the generator.

Non-blocking coverage gaps (all pass when probed, just not pinned): the matcher's absent-single-optional-slot, null-capture, `Repeat`+`OptionalNode`, and backreference-across-`Repeat`-backtrack paths in `PatternMatchingTests`; and the bare `catch {` (type-less) sequence-point span in `LocationsInAstSamples`.

ICSharpCode.Decompiler/CSharp/Syntax/AstNodeCollection.cs
```diff
@@ -174,18 +267,17 @@ bool ICollection<T>.IsReadOnly {
 			get { return false; }
 		}
 
+		// Tolerates removing or replacing the current element during enumeration (the common
+		// transform pattern), matching the previous linked-list collection's behavior.
```
siegfriedpammer: This enumerator tolerates remove/replace of the *current* node, but inserting before the cursor or removing a *different* element shifts indices and can re-yield/skip. The old linked-list enumerator tracked node identity and was immune to both, so "matching the previous behavior" holds only for the remove/replace-current case.

ICSharpCode.Decompiler/CSharp/Syntax/Expressions/InterpolatedStringExpression.cs
```diff
@@ -1,184 +1,75 @@
...
-		}
-
-		public Expression Expression {
-			get { return GetChildByRole(Roles.Expression); }
-			set { SetChildByRole(Roles.Expression, value); }
-		}
+		[Slot("Expression")]
+		public partial Expression Expression { get; set; }
 
 		public int Alignment { get; }
```
siegfriedpammer, line 47: The generated `DoMatch` now compares `Alignment`/`Suffix`; the hand-written `DoMatch` on master compared only `Expression`. Verified that `Interpolation` is only ever constructed, never a pattern/`IsMatch` target, so no behavioral effect — flagging in case the tightening should be made explicit (e.g. `[ExcludeFromMatch]`).

ICSharpCode.Decompiler/CSharp/Syntax/GeneralScope/NamespaceDeclaration.cs
```diff
@@ -23,36 +23,31 @@
...
-		public static readonly Role<AstNode> MemberRole = SyntaxTree.MemberRole;
-		public static readonly Role<AstType> NamespaceNameRole = new Role<AstType>("NamespaceName", AstType.Null);
-
-		public override NodeType NodeType {
-			get {
-				return NodeType.Unknown;
-			}
-		}
-
 		public bool IsFileScoped { get; set; }
```
siegfriedpammer, line 44: Same as the `Interpolation` note: `IsFileScoped` now participates in the generated `DoMatch` where master compared only Name + Members. No consumer matches `NamespaceDeclaration` structurally, so no behavioral effect.

ICSharpCode.Decompiler/CSharp/Syntax/PatternMatching/IdentifierExpressionBackreference.cs
```diff
@@ -39,15 +41,14 @@ public IdentifierExpressionBackreference(string referencedGroupName)
...
-		public override bool DoMatch(PatternMatching.INode other, PatternMatching.Match match)
+		public override bool DoMatch(INode? other, Match match)
 		{
-			IdentifierExpression ident = other as IdentifierExpression;
+			var ident = other as IdentifierExpression;
 			if (ident == null || ident.TypeArguments.Any())
 				return false;
-			AstNode referenced = (AstNode)match.Get(referencedGroupName).Last();
-			if (referenced == null)
+			if (match.Get(referencedGroupName).Last() is not IdentifierExpression referenced)
```
siegfriedpammer: This requires the referent to be an `IdentifierExpression`; master read `GetChildByRole(Roles.Identifier).Name` on any referent. The doc-comment (line 27) still says "variable/type definition/method definition". All in-tree callers capture via `IdentifierExpression`, so no regression — but the contract and its doc now diverge.

ICSharpCode.Decompiler.Generators/DecompilerSyntaxTreeGenerator.cs
```diff
@@ -0,0 +1,883 @@
...
+		foreach (var node in source)
+		{
+			if (node.Slots is { } slots)
+			{
+				foreach (var s in slots)
+				{
+					if (!kindTypes.TryGetValue(s.KindName, out var set))
+						kindTypes[s.KindName] = set = new SortedSet<string>(StringComparer.Ordinal);
+					set.Add(s.ElementType);
+					kindIsCollection[s.KindName] = s.IsCollection;
```
siegfriedpammer: `kindIsCollection[kind]` is last-write-wins across `source` order. For kinds used as a collection on one node and a single child on another (`Expression`, `Initializer`, `Variable`, ...), the canonical `Slots.X` constant ends up with an order-dependent `IsCollection` (and the hard-coded `isOptional: false` at line 818). Verified unread — consumers read these flags only off the per-node `XSlot` — so latent, not live.

ICSharpCode.Decompiler/CSharp/Syntax/ComposedType.cs
siegfriedpammer: Fixed in c8d320e62 — dropped the stale "setter rejects negatives" wording; the comment now just notes CheckInvariant catches a corrupted PointerRank.

_(Reply written by an AI agent (Claude) on Siegfried's behalf.)_

ICSharpCode.Decompiler/CSharp/Syntax/AstNodeCollection.cs
siegfriedpammer: Adjusted in 9a8df114f — the comment no longer claims full equivalence with the old linked-list enumerator. It now states the tolerance covers the current element only, and that inserting before the cursor or removing a different element can re-yield or skip.

_(Reply written by an AI agent (Claude) on Siegfried's behalf.)_

ICSharpCode.Decompiler/CSharp/Syntax/Expressions/InterpolatedStringExpression.cs
siegfriedpammer: Intended — the generated DoMatch deliberately matches every structural member. Interpolation is only ever constructed, never a pattern/IsMatch target, so this has no behavioral effect. Leaving as-is.

_(Reply written by an AI agent (Claude) on Siegfried's behalf.)_

ICSharpCode.Decompiler/CSharp/Syntax/GeneralScope/NamespaceDeclaration.cs
siegfriedpammer: Intended — same as the Interpolation note. No consumer matches NamespaceDeclaration structurally, so the tightening has no behavioral effect. Leaving as-is.

_(Reply written by an AI agent (Claude) on Siegfried's behalf.)_

ICSharpCode.Decompiler/CSharp/Syntax/PatternMatching/IdentifierExpressionBackreference.cs
siegfriedpammer: Fixed in 95c987737 — restored the broader contract. It now reads the referent's Identifier child generically via GetChild(Slots.Identifier) (the slot analog of master's GetChildByRole(Roles.Identifier)), so any identifier-bearing referent matches again. Verified no in-tree caller regressed and the matcher tests pass.

_(Reply written by an AI agent (Claude) on Siegfried's behalf.)_

ICSharpCode.Decompiler.Generators/DecompilerSyntaxTreeGenerator.cs
siegfriedpammer: Documented in 4219507ca — added a note that the shared kind's IsCollection (last-write-wins) and hard-coded isOptional: false are not authoritative; the precise per-position flags live on the per-node slots, which is where consumers read them.

_(Reply written by an AI agent (Claude) on Siegfried's behalf.)_

ICSharpCode.Decompiler/CSharp/Syntax/AstNodeCollection.cs
siegfriedpammer: Follow-up: I dug into this and the earlier "can re-yield or skip" framing was overstated. Both fast paths in the resume logic are guarded by ReferenceEquals(next), so they only fire when the captured successor genuinely sits at that index; otherwise it falls back to an O(n) IndexOf(next). The cursor therefore always lands on the captured successor by identity -- the same hand-over-hand walk the old linked-list enumerator did. Inserting before the cursor or removing another element only shifts backing-list indices, which the identity resume absorbs; the one case it cannot follow is removing the captured successor itself, which stops enumeration (the linked-list enumerator lost track of it too).

Added in f6c2dbc96: a DEBUG runtime guard in MoveNext asserting the chosen index equals the authoritative identity position (so any future change that lets the cursor mislocate fails loudly under the test suite), the corrected doc comment, and an AstNodeCollectionEnumeratorTests fixture pinning the mutation-during-enumeration cases (remove/replace current, remove an earlier element, insert before the cursor, remove the successor).

_(Reply written by an AI agent (Claude) on Siegfried's behalf.)_

Issue #3430, opened 2025-03-17
Ref Struct Interfaces/allows ref struct constraint


ds5678:
https://github.com/dotnet/csharplang/issues/7608

PR #3527, opened 2025-07-27
Support `allows ref struct` constraint

jjonescz:
### Problem
Part of https://github.com/icsharpcode/ILSpy/issues/3430. Decompiling to IL doesn't show `byreflike` constraint. Similarly, decompiling to C# doesn't show `allows ref struct` constraint.

### Solution


christophwille:
Thanks!

PR #3533, opened 2025-08-05
Fix incorrect argument removal if OptionalArguments is false

MSchmoecker:
### Problem
Setting `OptionalArguments = false` (i.e. keeping optional arguments) did not work correctly for some cases:
1. If the method has both mandatory and optional arguments:
```cs
void MixedArguments(string msg, int a = 0, int b = 0) {}

MixedArguments("123", 0, 0);
// Gets decompiled to:
MixedArguments("123"); // We want to preserve all 3 arguments
```

2. If the method has params arguments
```cs
ParamsArgument(params string[] args) {}

ParamsArgument(Array.Empty<string>());
// Gets decompiled to:
ParamsArgument(); // We want to preserve the empty array creation
```

### Solution
* First case is a simple bug. If the first parameter is mandatory, firstOptionalArgumentIndex was changed from -1 to -2 and thus allowed subsequent optional parameters to be found
* The second case is rather a design decision. Since we don't want to discard optional arguments, it's better to retain explicit params array creation (to distinguish it from IL that doesn't create an array)


siegfriedpammer:
I would argue that the second case is not related to optional arguments at all and thus should not be controlled by the same setting.

MSchmoecker:
That's a fair point. I can drop the second commit and open a new PR that adds a decompile setting to toggle params argument expansion (of course enabled by default).

siegfriedpammer:
Yes, please! I think that would make a lot of sense.

siegfriedpammer:

PR #3538, opened 2025-08-14
Update TomsToolbox to fix #3462

### LINKED ISSUE #3462: Scrolling with scroll wheel breaks after using file search
### Steps to reproduce
1. Open any decompiled file.
2. Use Ctrl-F to search within the file for code that is below the fold.
3. Hover over the field and mousewheel down to scroll.
4. The window will snap all the way back up to the top of the file before it starts scrolling.

### Error message shown

No errors are happening, but this is happening with 100% consistency for me.

### Details

This was happening in 9.0.0 and after updating to 9.1.0 today it exhibits the same issue.

ILSpy standalone in Windows 11.

tom-englert:
fix #3462: Scrolling with scroll wheel breaks after using file search #3462


siegfriedpammer:
Thank you very much for the quick fix.

tom-englert:
@siegfriedpammer Thanks for pointing me directly to the root cause, could have been searching for hours to spot this 😃

PR #3544, opened 2025-08-24
Minor Dev Environment Fixes (.editorconfig cleanup, typo fix, .gitignore tweak)

petercrabtree:
Minor fixes that seemed like they're small enough to bundle together for review (there should be no run-time change in behavior):

* Removed duplicate settings in .editorconfig.
* Fix invalid .editorconfig value (valid values for csharp_space_around_declaration_statements seem to be "false" and "ignore"; "do_not_ignore" currently is equivalent to false based on https://github.com/dotnet/roslyn/issues/28022).
* Change .gitignore to ignore all .vscode folders, not just a top-level one.
* Typo fix (ReportExludedTypes -> ReportExcludedTypes)


siegfriedpammer:

PR #3546, opened 2025-08-27
Normalize UTF-8 BOM Marks and ps1 Indention

petercrabtree:
## Problem

The repo mixes UTF-8 **with and without** a byte order mark (BOM). Some tools in my workflow write using their own defaults instead of preserving a file’s existing encoding, so **routine edits flip the BOM and create noisy diffs**. This PR **establishes one policy, aligns scripts to it, then applies a mechanical cleanup**.

**Why you may not encounter this:** if your tools preserve a file’s current encoding (as *most* of mine do...), you won’t hit churn. The noise appears only with non-preserving writers.

## Changes

No functional/runtime changes.

### Commit 1 — policy + tooling

* **`.editorconfig`**: default to UTF-8 **without** BOM; **`.resx`** stays UTF-8 **with** BOM. The ResX toolchain (`ResXResourceWriter`, Visual Studio designer, `resgen`) emits a BOM and ignores `.editorconfig`, so we configure `.resx` accordingly to keep editors and build tools aligned.
* **`BuildTools/update-assemblyinfo.ps1`**: write UTF-8 **without** BOM to match policy. On PowerShell pre-v6, `Out-File` writes a BOM by default; this prevents that.
* New: `BuildTools/bom-strip.ps1`: remove all BOMs (except resx)
* New: `BuildTools/bom-classify-encoding.ps1`: inspect BOM presence; also used to derive the extension list consumed by `bom-strip`.

`dotnet format` honours `.editorconfig` and checks the charset (including BOM), so this should prevent BOM churn going forward by hooking into the existing formatting enforcement.

### Commit 2 — mechanical application

* Normalize UTF-8 BOMs across the repository by removing them from all text files other than resx

## Review notes

Commit 2 is the mechanical output of running `buildtools/bom-strip.ps1` after Commit 1. Its diff should contain only BOM removals.

It was done this way to make verifying the giant (>1000 files) commit easier--I suggest parsing the diff/patch or running `buildtools/bom-strip.ps1` on commit 1 if you want to make sure for yourself that there are no other changes in there.

### DIFF
--- a/BuildTools/update-assemblyinfo.ps1
+++ b/BuildTools/update-assemblyinfo.ps1
@@ -1,6 +1,6 @@
 if (-not ($PSVersionTable.PSCompatibleVersions -contains "5.0")) {
-    Write-Error "This script requires at least powershell version 5.0!";
-    return 255;
+	Write-Error "This script requires at least powershell version 5.0!";
+	return 255;
 }
 
 $ErrorActionPreference = "Stop"
@@ -14,11 +14,11 @@ $masterBranches = '^(master|release/.+)$';
 $decompilerVersionInfoTemplateFile = "ICSharpCode.Decompiler/Properties/DecompilerVersionInfo.template.cs";
 
 function Test-File([string]$filename) {
-    return [System.IO.File]::Exists((Join-Path (Get-Location) $filename));
+	return [System.IO.File]::Exists((Join-Path (Get-Location) $filename));
 }
 
 function Test-Dir([string]$name) {
-    return [System.IO.Directory]::Exists((Join-Path (Get-Location) $name));
+	return [System.IO.Directory]::Exists((Join-Path (Get-Location) $name));
 }
 
 function Find-Git() {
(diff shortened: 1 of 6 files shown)


BuildTools/update-assemblyinfo.ps1
```diff
@@ -175,7 +175,8 @@ try {
         $out = $out.Replace('$INSERTBUILDCONFIG$', $buildConfig);
 
         if ((-not (Test-File $file.Output)) -or (((Get-Content $file.Output) -Join [System.Environment]::NewLine) -ne $out)) {
-            $out | Out-File -Encoding utf8 $file.Output;
+			$utf8NoBom = New-Object System.Text.UTF8Encoding($false);
```
siegfriedpammer, line 178: indentation in ps1 should be spaces, I guess?

siegfriedpammer: could be added to .editorconfig as well :)

petercrabtree: Given that the existing ps1 files are a total mishmash of tabs vs spaces (even on the same line!), I think we should just choose.

And given that the project bucks the trend and uses tabs for .cs files (and I like tabs...), I've gone ahead and just standardized the ps1 files on tabs.

siegfriedpammer:
Thank you very much!

Issue #3521, opened 2025-07-20
Highlight reference line


greenozon:
**Is your feature request related to a problem? Please describe.**
I"d like to propose a new mini-feature that will definitely make users happy
So when you do some extensive searching, eg you have a list of some functions in the Analyze window
now if you double click on each & every line in it the expectations would be you go to that context as close as possible
eg - ILSpy would somehow highlight the line in which the reference 

**Describe the solution you'd like**
Somehow hightlight the used function in the code window whenever user double clicks the reference line in Analyze window

Thanks!

dgrunwald:
Clicking on a local variable will highlight all uses of that variable.
I think we should use that infrastructure to highlight all uses of the `ConvertFrom` method when navigating to the `get_WaitingReadCount` entry in the "Used By" list.

@siegfriedpammer Not sure why you marked this as related to a bookmarks pad.

siegfriedpammer:
When I first thought about solving this issue, I got the idea that we'd need similar infrastructure to implement this feature.

PR #3550, opened 2025-09-04
#3521: Add API to set an initially highlighted entity after navigation

siegfriedpammer:
@dgrunwald please review! I am not sure, if this is the best way to solve the state handling.


greenozon:
@siegfriedpammer   thanks a lot for sky-rocket feat!

PR #3553, opened 2025-09-08
Move from dotnet.config to global.json

Youssef1313:
If https://github.com/dotnet/sdk/pull/50673 gets merged, updating to RC2 with that change will require the change in this PR.

FYI @christophwille


christophwille:
Thanks for the heads-up! I surely would have missed that subtle change.

nikitalita:
it did get merged, so this is required

christophwille:
Thanks!

PR #3560, opened 2025-09-11
Fix regression in decompiling local functions with default parameters

### LINKED ISSUE #3541: Regression with optional arguments in local function closure
### Input code
```cs
class OptionalArgumentTest
{
    int data;

    void Test(string format)
    {
        TestLocal();

        void TestLocal(int a = 0)
        {
            a.ToString(format);
        }
    }
}
```

### Erroneous output
```cs

internal class OptionalArgumentTest
{
	private int data;

	private void Test(string format)
	{
		TestLocal();

		void TestLocal([Optional][DefaultParameterValue(0)] int a)
		{
			a.ToString(format);
		}
	}
}
```

### Details
Tested at commit 587a359e8c37b1745afdca9442c0eaa23834a9de
Regressed in #3470

The loop in `TypeSystemExtensions.IsDefaultValueAssignmentAllowed` does not account for the extra parameter generated for the closure.

### LINKED ISSUE #3558: Regression: Nested function parameters incorrectly have DefaultParameterValue attribute assigned vs. default value
Introduced in #3470

Before: 

Image

After: 
Image

It's because the reference capture parameter comes at the end of the parameters list: 

Image

### Details
* Version: [ILSpy 10.0 Preview 1](https://github.com/icsharpcode/ILSpy/releases/tag/v10.0-preview1)

ds5678:
### Problem

Resolves #3541 

Resolves #3558

### Solution

### DIFF
--- a/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
+++ b/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
@@ -154,6 +154,7 @@
     <Compile Include="TestCases\ILPretty\Issue3552.cs" />
     <Compile Include="TestCases\Pretty\ExpandParamsArgumentsDisabled.cs" />
     <Compile Include="TestCases\Pretty\ExtensionProperties.cs" />
+    <Compile Include="TestCases\Pretty\Issue3541.cs" />
     <None Include="TestCases\ILPretty\Issue3504.cs" />
     <Compile Include="TestCases\ILPretty\MonoFixed.cs" />
     <Compile Include="TestCases\Pretty\Comparisons.cs" />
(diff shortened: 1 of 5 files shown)


ds5678:
Weird. I wonder why the CI is failing on this pull request, but not my one from a couple days ago.

siegfriedpammer:
My guess would be that they updated the runner images and now .NET 10 RC1 is used by default, which has language version 14 as default, which probably turned the `field` keyword into a hard error instead of warning. But I am not entirely sure.

ds5678:
Hopefully, that fixes it.

ds5678:
I just ran all the unit tests locally. None failed.

siegfriedpammer:
Looks like this now depends on #3556 because GitHub decided to update the CI images... sorry for that.

PR #3572, opened 2025-09-29
Fix lower case type name conflicting with variable name

### LINKED ISSUE #3571: Variable naming doesn't find all accessible types with lower case names
### Input code

```cs
using System;

namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
{
	internal static class Issue4000
	{
		[StructLayout(LayoutKind.Sequential, Size = 1)]
		public readonly struct fsResult
		{
			public static fsResult Success => default(fsResult);
			public static fsResult Failure => default(fsResult);
			public bool Succeeded => true;
			public bool Failed => false;
			public static fsResult operator +(fsResult a, fsResult b)
			{
				return default(fsResult);
			}
		}

		public static fsResult M()
		{
			fsResult success = fsResult.Success;
			fsResult fsResult2 = success + fsResult.Success;
			if (fsResult2.Succeeded)
			{
				return success;
			}
			Console.WriteLine("Failed");
			return fsResult2 + fsResult.Failure;
		}
	}
}
```

### Erroneous output
```c#
using System;

namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
{
	internal static class Issue4000
	{
		[StructLayout(LayoutKind.Sequential, Size = 1)]
		public readonly struct fsResult
		{
			public static fsResult Success => default(fsResult);
			public static fsResult Failure => default(fsResult);
			public bool Succeeded => true;
			public bool Failed => false;
			public static fsResult operator +(fsResult a, fsResult b)
			{
				return default(fsResult);
			}
		}

		public static fsResult M()
		{
			fsResult success = fsResult.Success;
			fsResult fsResult = success + fsResult.Success;
			if (fsResult.Succeeded)
			{
				return success;
			}
			Console.WriteLine("Failed");
			return fsResult + fsResult.Failure;
		}
	}
}
```

```
CS0841: Cannot use local variable 'fsResult' before it is declared
```

### Details
* Product in use: ILSpy
* Version in use: 61f82d0c2dd9b77a4b5d76767637e24eaa4ee73f
  * I've implemented a fix in `AssignVariableNames` and am polishing it up for a pull request.

ds5678:
### Problem
Resolves #3571

### Solution
  * In `AssignVariableNames`, I made changes to `CollectReservedVariableNames`, `GenerateForeachVariableName`, and `GenerateVariableName` for consistent use between `CollectAllLowerCaseTypeNames(ITypeDefinition)` and the new `CollectAllLowerCaseTypeNames(UsingScope)`.
  * Is there any reason `CollectAllLowerCaseTypeNames(ITypeDefinition)` shouldn't have used Linq? I noticed that `CollectAllLowerCaseMemberNames(ITypeDefinition)` avoided Linq, but it made my code more concise and readable.


ds5678:
The last failing test is the `Discards` pretty test. I'm not sure how to fix it without violating the spirit of the test:

```cs
// Source
public void DiscardedOutVsLambdaParameter()
{
	GetOut(out var _);
	MakeValue((@_ _) => 5);
}

// Decompiled
public void DiscardedOutVsLambdaParameter()
{
	GetOut(out var _);
	MakeValue((@_ _2) => 5);
}
```

ICSharpCode.Decompiler.Tests/TestCases/Pretty/QualifierTests.cs
```diff
@@ -308,9 +308,9 @@ private void ParameterConflictsWithTypeName(string[] Array)
 
 		private void LocalConflictsWithTypeName()
 		{
-			for (int i = 0; i < 10; i++)
+			for (int j = 0; j < 10; j++)
 			{
-				QualifierTests.i.Test();
+				i.Test();
```
siegfriedpammer: hmm... doesn't this change defeat the purpose of the test?

ds5678: Are you suggesting this portion of the test be removed?

siegfriedpammer:
The issue you are fixing is actually not related to lower-case type names at all, but rather that nested types were not considered previously. This is true for both the nested type `Discards.@_` and `QualifierTests.i`.

I don't see a way to actually make the tests still work with this change. Maybe we should just remove them?

ICSharpCode.Decompiler.Tests/TestCases/Pretty/QualifierTests.cs
siegfriedpammer: I would like to keep the test, but it seems we cannot actually test the case where a local variable conflicts with a type name, because `AssignVariableNames` will no longer use nested type names as variable names :)

ds5678:
Good now?

siegfriedpammer:
Computer says yes! 😆

siegfriedpammer:

siegfriedpammer:
> * Is there any reason `CollectAllLowerCaseTypeNames(UserScope)` shouldn't have used Linq? I noticed that `CollectAllLowerCaseMemberNames(ITypeDefinition)` avoided Linq, but it made my code more concise and readable.

I forgot to answer the question. Actually, I no longer remember why I wrote it without using LINQ. Not sure, if there ever was a performance problem in that part of the code.

PR #3587, opened 2025-10-09
Improve inlining of boxed values

### LINKED ISSUE #3584: Ref locals from generic array access
### Input code

```cs
using System.Collections;

namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
{
	internal abstract class Issue4000<T> : IEnumerable<T>, IEnumerable
	{
		public uint Length;

		protected T[] results;

		public T this[int i] {
			get {
				if (i >= Length || i < 0)
				{
					return default(T);
				}
				if (results[i] != null && results[i].Equals(default(T)))
				{
					results[i] = CreateIthElement(i);
				}
				return results[i];
			}
		}

		protected abstract T CreateIthElement(int i);

		public IEnumerator<T> GetEnumerator()
		{
			for (int i = 0; i < Length; i++)
			{
				if (results[i] != null && results[i].Equals(default(T)))
				{
					results[i] = CreateIthElement(i);
				}
				yield return results[i];
			}
		}

		IEnumerator IEnumerable.GetEnumerator()
		{
			return GetEnumerator();
		}
	}
}
```

### Erroneous output

The decompiled output varies based on decisions like the compiler version, but almost all of the `defaultOptions` result in output that contains:

```cs
ref readonly T reference = ref results[i];
```

which causes issues on C# versions that don't support ref locals.

### Details
* Product in use: ILSpy
* Version in use: ed36ba1d8a11c1efda5c68b8a871c90ce502d5cb
  * As always, I'm willing to work on this. However, I would appreciate some guidance about where this might be fixed.

ds5678:
### Problem

Resolves #3584 

### Solution
  * I'm not sure my changes to `Box` are semantically correct.

### DIFF
--- a/ICSharpCode.Decompiler/IL/Instructions.cs
+++ b/ICSharpCode.Decompiler/IL/Instructions.cs
@@ -318,7 +318,6 @@ namespace ICSharpCode.Decompiler.IL
 			set {
 				ValidateChild(value);
 				SetChildInstruction(ref this.argument, value, 0);
-				InvalidateFlags();
 			}
 		}
 		protected sealed override int GetChildCount()
@@ -4356,24 +4355,6 @@ namespace ICSharpCode.Decompiler.IL
 			set { type = value; InvalidateFlags(); }
 		}
 		public override StackType ResultType { get { return StackType.O; } }
-		protected override InstructionFlags ComputeFlags()
-		{
-			var baseFlags = base.ComputeFlags();
-			if (baseFlags == InstructionFlags.None && Type.Equals(Argument.InferType(null)))
-			{
-				return InstructionFlags.None;
-			}
-			return baseFlags | InstructionFlags.SideEffect | InstructionFlags.MayThrow;
-		}
-		public override InstructionFlags DirectFlags {
-			get {
-				if (Flags == InstructionFlags.None)
-				{
-					return InstructionFlags.None;
-				}
-				return base.DirectFlags | InstructionFlags.SideEffect | InstructionFlags.MayThrow;
-			}
-		}
 		public override void WriteTo(ITextOutput output, ILAstWritingOptions options)
 		{
 			WriteILRange(output, options);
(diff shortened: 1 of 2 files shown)


dgrunwald:
* Instructions.cs is generated from Instructions.tt; please don't modify the former manually. (instead, use `CustomComputeFlags` in the .tt and a partial class for the actual implementation)
* SetChildInstruction already invalidates the flags; so there's no need for your added `InvalidateFlags()` call. 
* `box` not having any side-effects despite doing a heap-allocation, seems a bit questionable. But I don't see how it could cause any real issues (as the heap-allocation cannot be accessed without depending on the result of the `box` instruction, so incorrect reordering of heap-accesses still seems impossible).
* I don't really like the `InferType` call in the low-level flag computation -- that seems like something that might risk infinite recursion. Maybe just restrict this to `box T(default.value T)`?

dgrunwald:
Or maybe we don't need the type comparison at all -- if it's OK to ignore the heap-allocation and heap-write in some cases, it might also be OK to do so in all cases?
The argument that no problematic reordering is possible because the heap allocation is not otherwise accessible, doesn't depend on `Argument.Flags == None` or the type comparison...

ds5678:
I rebased, and CI is now failing, but I don't think that's my fault. It appears that CI has failed on every commit since #3594, though that might be a coincidence since I also saw https://github.com/github/codeql-action/issues/3208 and https://github.com/github/codeql-action/issues/3207 which claim that the release of .NET 10 RC2 caused the issue.

siegfriedpammer:
As long as the "Build ILSpy" action runs fine in debug and release there is no need to worry.

CodeQL has not once reported anything. @christophwille why do we have the action set up in the first place? I think, maybe it would make sense to just remove it completely?

christophwille:
The intent for it is to report nothing - it would be bad IF at some point it actually found some vulnerability.

ds5678:
@dgrunwald Can this be merged?

Issue #3539, opened 2025-08-16
Roslyn 5 changes (from 4.14)


christophwille:
This issue is mostly a PSA for those that use latest code from our repository that extension pattern detection is broken for latest lowering in Roslyn 5. 

#3530 we already knew from the spec that ExtensionMarkerAttribute is coming, and it is there now https://github.com/icsharpcode/ILSpy/actions/runs/16934497769

<img width="778" height="396" alt="Image" src="https://github.com/user-attachments/assets/21959ecf-1b85-42df-8f38-40731e52b37e" />

Additional change: ROS casts in expression trees.

To be fixed once VSnext and Roslyn 5 are available.

christophwille:
Depends on https://github.com/dotnet/roslyn/issues/80213

PR #3588, opened 2025-10-10
Roslyn 5.0.0

christophwille:
https://github.com/dotnet/roslyn/issues/80213

#3539

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/CallBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/CallBuilder.cs
@@ -1438,7 +1438,7 @@ namespace ICSharpCode.Decompiler.CSharp
 			var conversions = CSharpConversions.Get(expressionBuilder.compilation);
 			IType targetType = method.ReturnType;
 			var conv = conversions.ImplicitConversion(argument.Type, targetType);
-			if (!(conv.IsUserDefined && conv.IsValid && conv.Method.Equals(method)))
+			if (!(conv.IsUserDefined && conv.IsValid && conv.Method.Equals(method, NormalizeTypeVisitor.TypeErasure)))
 			{
 				// implicit conversion to targetType isn't directly possible, so first insert a cast to the argument type
 				argument = argument.ConvertTo(method.Parameters[0].Type, expressionBuilder);
--- a/ICSharpCode.Decompiler/CSharp/Resolver/CSharpConversions.cs
+++ b/ICSharpCode.Decompiler/CSharp/Resolver/CSharpConversions.cs
@@ -247,6 +247,10 @@ namespace ICSharpCode.Decompiler.CSharp.Resolver
 				if (IdentityConversion(elementType, spanElementType))
 					return Conversion.InlineArrayConversion;
 			}
+			if (IsImplicitSpanConversion(fromType, toType))
+			{
+				return Conversion.ImplicitSpanConversion;
+			}
 			return Conversion.None;
 		}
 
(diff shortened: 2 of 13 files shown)


christophwille:
Supersedes #3517

ICSharpCode.Decompiler/CSharp/CallBuilder.cs
```diff
@@ -1438,7 +1438,7 @@ private ExpressionWithResolveResult HandleImplicitConversion(IMethod method, Tra
 			var conversions = CSharpConversions.Get(expressionBuilder.compilation);
 			IType targetType = method.ReturnType;
 			var conv = conversions.ImplicitConversion(argument.Type, targetType);
-			if (!(conv.IsUserDefined && conv.IsValid && conv.Method.Equals(method)))
+			if (!(conv.IsUserDefined && conv.IsValid && conv.Method.Equals(method, NormalizeTypeVisitor.IgnoreNullabilityAndTuples)))
```
dgrunwald: Why only ignore nullability and tuples?
I think normally when comparing if a method we found in the type system matches a method present in IL, we use `NormalizeTypeVisitor.TypeErasure` which also handles dynamic/object and IntPtr/nint.

ICSharpCode.Decompiler/CSharp/Resolver/CSharpConversions.cs
```diff
@@ -1229,6 +1229,41 @@ List<OperatorInfo> GetApplicableConversionOperators(ResolveResult fromResult, IT
...
+			// see https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-14.0/first-class-span-types#span-conversions
+
+			switch (fromType)
+			{
+				case ArrayType { Dimensions: 1, ElementType: var elementType }:
+					if (toType.IsKnownType(KnownTypeCode.SpanOfT) ||
+						toType.IsKnownType(KnownTypeCode.ReadOnlySpanOfT))
+					{
+						return ImplicitReferenceConversion(elementType, toType.TypeArguments[0], 0)
+							|| IdentityConversion(elementType, toType.TypeArguments[0]);
```
dgrunwald: IdentityConversion is the cheaper check, so we usually try that one first.
Also you don't need the `, 0` (which just made me look up "what's that extra parameter")

dgrunwald: You need to distinguish between `Span<T>` and `ReadOnlySpan<T>` -- an array `T[]` only implicitly converts to `Span<U>` if there's an identity conversion from T to U; it's only for `ReadOnlySpan<U>` that a reference converstion T->U is allowed.

ICSharpCode.Decompiler/CSharp/Resolver/CSharpResolver.cs
```diff
@@ -2171,7 +2171,8 @@ static bool IsEligibleExtensionMethod(ICompilation compilation, CSharpConversion
 				thisParameterType = thisParameterType.AcceptVisitor(substitution);
 			}
 			Conversion c = conversions.ImplicitConversion(targetType, thisParameterType);
-			return c.IsValid && (c.IsIdentityConversion || c.IsReferenceConversion || c.IsBoxingConversion);
+			return (c.IsValid && (c.IsIdentityConversion || c.IsReferenceConversion || c.IsBoxingConversion))
+				|| conversions.IsImplicitSpanConversion(targetType, thisParameterType);
```
dgrunwald: Why re-check the conversion? I would expect `|| c.IsSpanConversion` here.

ICSharpCode.Decompiler/CSharp/TranslatedExpression.cs
```diff
@@ -650,6 +650,10 @@ bool CastCanBeMadeImplicit(Resolver.CSharpConversions conversions, Conversion co
 				return newTargetType.IsKnownType(KnownTypeCode.FormattableString)
 					|| newTargetType.IsKnownType(KnownTypeCode.IFormattable);
 			}
+			if (conversion.IsUserDefined)
+			{
+				return conversions.IsImplicitSpanConversion(inputType, newTargetType);
```
dgrunwald: This change doesn't make any sense: you're not checking `oldTargetType` so this is essentially saying that any user-defined conversion, no matter which operator method is called, can be replaced with an implicit span conversion (if that's valid).
e.g. you could end up replacing a `int[]`->`MyClass` user-defined conversion with a `int[]`->`Span<int>` conversion.

ICSharpCode.Decompiler/CSharp/Resolver/CSharpConversions.cs
siegfriedpammer: > Also you don't need the `, 0` (which just made me look up "what's that extra parameter")

But it's not optional and `IsImplicitReferenceConversion` would end up calling the same overload, so I thought I take the shortcut :)

PR #3589, opened 2025-10-10
List name first in ILSpy title for multiple instances

### LINKED ISSUE #3562: Put assembly list name first in the window title
I have multiple instances of ILSpy allowed, because I need to use see different set of assemblies at the same time. However, since the word "ILSpy" is shown first in the title, it is not easy to find out which window is the correct one, as the list name gets often cut of (especially if the title also contains the build number). 

I would suggest the assembly name list is put first and the application name second in the window title, like in all other applications that open documents. If that was too disruptive to users, the title can change this way only when multiple instances are enabled.

**Additional context**

Image

Image

Happy to implement if there was an agreement.

miloush:
Fixes #3562

### Problem
I have multiple instances of ILSpy allowed, because I need to use see different set of assemblies at the same time. However, since the word "ILSpy" is shown first in the title, it is not easy to find out which window is the correct one, as the list name gets often cut of (especially if the title also contains the build number).

### Solution
When multiple instances are allowed, swap the main window title from "ILSpy - List name" to "List name - ILSpy"

### DIFF
--- a/ILSpy/AssemblyTree/AssemblyTreeModel.cs
+++ b/ILSpy/AssemblyTree/AssemblyTreeModel.cs
@@ -469,9 +469,9 @@ namespace ICSharpCode.ILSpy.AssemblyTree
 #endif
 			else
 #if DEBUG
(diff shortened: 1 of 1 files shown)


christophwille:
Thanks!

Issue #3582, opened 2025-10-06
Can we have menus for back/previous like in web browsers?


aybe:
**Is your feature request related to a problem? Please describe.**

No.

**Describe the solution you'd like**

Can we have menus for back/previous like in web browsers?

**Additional context**

For this:

<img width="325" height="161" alt="Image" src="https://github.com/user-attachments/assets/e47ebb88-e3f2-45fa-aea3-6c3c892ebba8" />

But like this:

<img width="854" height="251" alt="Image" src="https://github.com/user-attachments/assets/3197032c-bf7a-4154-83fe-3f7bbd6d7347" />

PR #3591, opened 2025-10-11
Navigate visible history

miloush:
Implements #3582 to show history items:

image

@aybe

### DIFF
--- a/ILSpy/Commands/BrowseBackCommand.cs
+++ b/ILSpy/Commands/BrowseBackCommand.cs
@@ -57,7 +57,7 @@ namespace ICSharpCode.ILSpy
 
 		public IEnumerable ParameterList => assemblyTreeModel.GetNavigateHistory(false).Reverse();
 
-		public object GetParamaterText(object parameter)
+		public object GetParameterText(object parameter)
 		{
 			return (parameter as NavigationState)?.NavigationText;
 		}
--- a/ILSpy/Commands/BrowseForwardCommand.cs
+++ b/ILSpy/Commands/BrowseForwardCommand.cs
@@ -57,7 +57,7 @@ namespace ICSharpCode.ILSpy
 
 		public IEnumerable ParameterList => assemblyTreeModel.GetNavigateHistory(true).Reverse();
 
-		public object GetParamaterText(object parameter)
+		public object GetParameterText(object parameter)
 		{
 			return (parameter as NavigationState)?.NavigationText;
 		}
--- a/ILSpy/Commands/SimpleCommand.cs
+++ b/ILSpy/Commands/SimpleCommand.cs
@@ -46,6 +46,6 @@ namespace ICSharpCode.ILSpy
 	public interface IProvideParameterList
 	{
 		IEnumerable ParameterList { get; }
-		object GetParamaterText(object parameter);
+		object GetParameterText(object parameter);
 	}
 }
--- a/ILSpy/Controls/MainToolBar.xaml.cs
+++ b/ILSpy/Controls/MainToolBar.xaml.cs
@@ -148,7 +148,7 @@ namespace ICSharpCode.ILSpy.Controls
 				BindingOperations.SetBinding(dropDownToggle, IsEnabledProperty,
 					new Binding(nameof(IsEnabled)) { Source = toolbarItem });
 
-				// When the toggle button is checked, clicking it to uncheck will dimiss the menu first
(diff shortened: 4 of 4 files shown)


siegfriedpammer:
Why do we need a `NavigationText` property?

miloush:
Because it differs from what is in the tree, it shows full types, members including type, and assembly name for some of the folders (screenshot first item)

miloush:
I can also imagine it could be a method where you put current state and the text would differ on where you are (like same namespace or same assembly)

siegfriedpammer:
I think I am fine with that. Does need to be in the `SharpTreeNode` base class? Also definitely needs a separator other than space...

siegfriedpammer:
It's also important to use the settings-aware implementation for the node text that is used everywhere else.

miloush:
It has a default implementation returning `Text` so no extra work for inheritors if the entry is to be the same as in the tree. `NavigationState` is holding `SharpTreeNode` nodes so that's why I put it there.

You can see in the changed files that it follows what `Text` is doing. The only "new syntax" is the space for the folders. What would you prefer? ` > `?

With that said, I don't mind using `Text` directly if that was preferred. I just didn't find it too helpful for navigation.

siegfriedpammer:
Maybe put the folder in parentheses?

I am not sure. I can have a closer look when I get home. Reviewing code on the phone is not ideal. 😅

miloush:
I put the folder first and where it belongs in parentheses, looks better, see the updated screenshot.

miloush:
I also tried to include module name in the metadata nodes but feel free to skip that commit. Done working on this PR until there is feedback to address.

siegfriedpammer:
One thought I had: is it possible to get an actual dropdown toolbar button in WPF? Is this only included in some community library?

miloush:
There is no built-in dropdown button in WPF. I considered creating a control for it (or retemplating combobox) but since the theme and styles are external it didn't seem worth it. We could also just have it as context menu on the normal buttons and that's it.

ICSharpCode.Decompiler/Metadata/MetadataFile.cs
```diff
@@ -73,9 +73,12 @@ public string Name {
 				if (value == null)
 				{
 					var metadata = Metadata;
-					value = metadata.IsAssembly
-						? metadata.GetString(metadata.GetAssemblyDefinition().Name)
-						: metadata.GetString(metadata.GetModuleDefinition().Name);
+					if (metadata.IsAssembly)
+						value = metadata.GetString(metadata.GetAssemblyDefinition().Name);
+					else if (metadata.DebugMetadataHeader == null) // standalone debug metadata does not contain module table
```
siegfriedpammer, line 78: probably should check `Kind` instead?

siegfriedpammer:
I am playing around with your PR and something I noticed:

grafik

this doesn't seem to be very useful... I think each item should include the top-level ancestor, maybe?

ICSharpCode.Decompiler/Metadata/MetadataFile.cs
miloush: The GetModuleDefinition throws on non-null DebugMetadataHeader, so that felt like an appropriate check before invoking it.

I don't mind changing it to Kind check, but I would need some hint about which Kinds guarantee DebugMetadataHeader, as I don't know how to produce e.g. WebCIL to check.

miloush:
Yeah I am aware of that. I thought it is a rare if not adverse scenario where you would jump through the same nodes in different assemblies. I thought about adding the top-level info, but then all the items become cluttered with full assembly name for minimal benefit in rare cases. Next thing you come with will be same assembly names using different file locations, so I think the line needs to be drawn somewhere.
(to be fair I only considered adding assembly name rather than top node text, and some of the nodes, especially in the metadata bits, do not have assembly/module associated, so it didn't really help)

This is not intended to replace user's memory. You still know what you have been doing recently so you will have an idea where the items come from.

With that said, with the `NavigationText` we can fine tune this easily on per node basis. For example, for namespaces we could include assembly name.

I could also do some analysis when building the history list, and if the previous item "looks the same" but has different top level, append the top level.

siegfriedpammer:
You are right. It's probably fine the way it's now.

miloush:
I addressed what I intend to.

siegfriedpammer:
I ran another quick test and I think we can merge it like this. Thanky you very much for your contribution!

PR #3596, opened 2025-10-14
Fixed some inherited record types being incorrectly identified as cla…

sonyps5201314:
There's a problem with the `ComputeIsRecord` function. It appears to only check whether the current class has functions like `ToString`, `PrintMembers`, and `GetHashCode`.
```cs
public abstract Parameter <Clone>$();
public sealed override string ToString();
protected virtual bool PrintMembers(StringBuilder builder);
```
In reality, as shown above, functions like `ToString` and `PrintMembers` are inherited, so as long as the class isn't a base class, these member functions don't need to exist in the current class.

https://github.com/restsharp/RestSharp/blob/dev/src/RestSharp/Parameters/BodyParameter.cs
before:
image
after:
image


siegfriedpammer:
Unfortunately, your solution breaks the laziness of `MetdataTypeDefinition` by having to materialize all properties and methods when evaluating the `IsRecord` property. This is not acceptable.

Issue #3611, opened 2025-11-13
Further investigate regression in primary constructor decompilation


siegfriedpammer:
> After the recent fixes the issue with `PCFieldInitTest` becomes the same as the C3 test in #3452. The primary constructor parameter wrongly gets referenced outside of field initialization and generates an extra backing field when recompiled. For `PCFieldInitTest` it now looks like this:
> ```cs
> internal abstract class PCFieldInitTest(StringComparison value)
> {
> 	private StringComparison _value = value;
> 
> 	public bool Func()
> 	{
> 		// Should be `_value` not `value`
> 		return value == StringComparison.Ordinal;
> 	}
> }
> ```
> Sorry if the naming difference was too subtle, I could have made it clearer. 

 _Originally posted by @mmusu3 in [#3610](https://github.com/icsharpcode/ILSpy/issues/3610#issuecomment-3526190809)_

mmusu3:
I don't like generating extra work for people but I figure it's better to get this all done now while it's still fresh in mind. I spent the day going through my diffs from the changes in commits 9c8d1e48d93bb21a30f566d2cae4ab245a8d59e1 through 7a5a2a00a9fcd5f002d3b28801d68b8e2e1b5ccb and there are now many improvements but I found some extra test cases that still need work. Some are still recent regressions but others are older cases that existed before.

**Input**
```cs
class C4(string value)
{
    public object Obj { get; } = new object();
    public string Value { get; } = value;
}

class C5(C5.ValueArray array)
{
    public struct ValueArray
    {
        bool b;
        public bool[] ToArray() => null;
    }

    public bool[] Values = array.ToArray();
}

class BaseClass
{
    protected BaseClass(int value) { }
}

class C6(C6.Data2 data) : BaseClass(data.Value)
{
    public struct Data2 { public int Value { get; set; } }

    public Data2 Data => data;
}

struct S3<T>(T v)
{
    public T Value => v;
}

interface I1
{
    int Number { get; }
}

// May be the same issue as S3
struct S4<T>(int number)
    : IComparable<T> where T : I1
{
    public int CompareTo(T other) => number.CompareTo(other.Number);
}
```

**Output**
```cs
private class C4
{
    public object Obj { get; } = new object();

    public string Value { get; }

    public C4(string value)
    {
        Value = value;
        base..ctor();
    }
}

private class C5
{
	public struct ValueArray
	{
		private bool b;

		public bool[] ToArray()
		{
			return null;
		}
	}

[...]

PR #3614, opened 2025-11-19
Fixes for primary constructors

mmusu3:
Adds the test cases from #3611 and fixes them.

The changes to `IntroduceUsingDeclarations` fix missing type name qualifiers in primary constructor parameters when the parameter type is or contains a nested type. However, I'm not sure if my changes are the best way to handle it, what do you think?

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/Transforms/IntroduceUsingDeclarations.cs
+++ b/ICSharpCode.Decompiler/CSharp/Transforms/IntroduceUsingDeclarations.cs
@@ -271,8 +271,15 @@ namespace ICSharpCode.Decompiler.CSharp.Transforms
 				if (typeDeclaration.HasPrimaryConstructor)
 				{
 					inPrimaryConstructor = true;
-					typeDeclaration.PrimaryConstructorParameters.AcceptVisitor(this);
-					inPrimaryConstructor = false;
+
+					try
+					{
+						typeDeclaration.PrimaryConstructorParameters.AcceptVisitor(this);
+					}
+					finally
+					{
+						inPrimaryConstructor = false;
+					}
 				}
 
 				var previousResolver = resolver;
@@ -294,6 +301,11 @@ namespace ICSharpCode.Decompiler.CSharp.Transforms
 
 			public override void VisitParameterDeclaration(ParameterDeclaration parameterDeclaration)
 			{
+				// Parameters of primary constructors are visited separately from the rest of the
+				// type declaration since their types are at the same scope as the type declaration
+				// and so need to use the outer resolver. This check ensures that the visitor only
+				// runs once per parameter since their AstNodes will get revisited by the call to
+				// `base.VisitTypeDeclaration(typeDeclaration)` in `VisitTypeDeclaration` above.
 				if (inPrimaryConstructor || parameterDeclaration.Parent is not TypeDeclaration)
 					base.VisitParameterDeclaration(parameterDeclaration);
 			}


siegfriedpammer:
oh, no, I just spent the last week rewriting that part of the decompiler.

mmusu3:
I'm sorry I was late with this. Most of it was done days ago but I was having a troublesome week.

siegfriedpammer:
Oh, no worries. For me it was a stressful week too, I only finished my refactoring today.

mmusu3:
Oh I forgot to mention in the main post but I still had one unsolved issue that isn't part of #3611. It is a variable naming conflict. I had trouble creating a simple case that doesn't involve pattern matching so this will have to do. It only occurs once in my full decompiled code project so it's a low priority.

Input
```cs
private class C8(object obj)
{
    public int Test()
    {
        if (obj is int i)
            return i;
        return 0;
    }
}
```
Output
```cs
private class C8(object obj)
{
    public int Test()
    {
        object obj = obj; // Name conflict
        if (obj is int)
            return (int)obj;
        return 0;
    }
}
```

siegfriedpammer:
Can you extract the changes to `IntroduceUsingDeclarations` into a separate commit? Once that is done, I can start reviewing. Thanks!

mmusu3:
Done

siegfriedpammer:
I am working on this right now... please wait

ICSharpCode.Decompiler/CSharp/Transforms/IntroduceUsingDeclarations.cs
```diff
@@ -280,6 +292,12 @@ public override void VisitTypeDeclaration(TypeDeclaration typeDeclaration)
 				}
 			}
 
+			public override void VisitParameterDeclaration(ParameterDeclaration parameterDeclaration)
+			{
+				if (inPrimaryConstructor || parameterDeclaration.Parent is not TypeDeclaration)
```
siegfriedpammer, line 309: this looks suspicious... we shouldn't skip those... instead we should just make sure that the correct resolver is used when visiting...

ICSharpCode.Decompiler/CSharp/Transforms/IntroduceUsingDeclarations.cs
```diff
@@ -265,9 +267,19 @@ public override void VisitTypeDeclaration(TypeDeclaration typeDeclaration)
 					base.VisitTypeDeclaration(typeDeclaration);
 					return;
 				}
+
+				if (typeDeclaration.HasPrimaryConstructor)
+				{
+					inPrimaryConstructor = true;
```
siegfriedpammer, line 273: I think it is not necessary to track the primary constructor separately.

siegfriedpammer:
```
❯ git diff
diff --git a/ICSharpCode.Decompiler/CSharp/Transforms/IntroduceUsingDeclarations.cs b/ICSharpCode.Decompiler/CSharp/Transforms/IntroduceUsingDeclarations.cs
index a2321bf10..1acee8776 100644
--- a/ICSharpCode.Decompiler/CSharp/Transforms/IntroduceUsingDeclarations.cs
+++ b/ICSharpCode.Decompiler/CSharp/Transforms/IntroduceUsingDeclarations.cs
@@ -189,8 +189,6 @@ sealed class FullyQualifyAmbiguousTypeNamesVisitor : DepthFirstAstVisitor
                        CSharpResolver resolver;
                        TypeSystemAstBuilder astBuilder;

-                       bool inPrimaryConstructor;
-
                        public FullyQualifyAmbiguousTypeNamesVisitor(TransformContext context, UsingScope usingScope)
                        {
                                this.ignoreUsingScope = !context.Settings.UsingDeclarations;
@@ -268,11 +266,11 @@ public override void VisitTypeDeclaration(TypeDeclaration typeDeclaration)
                                        return;
                                }

+                               // special case: primary constructor parameters are resolved in the outer context
+                               // (they cannot directly refer to the type being declared)
                                if (typeDeclaration.HasPrimaryConstructor)
                                {
-                                       inPrimaryConstructor = true;
                                        typeDeclaration.PrimaryConstructorParameters.AcceptVisitor(this);
-                                       inPrimaryConstructor = false;
                                }

                                var previousResolver = resolver;
@@ -292,12 +290,6 @@ public override void VisitTypeDeclaration(TypeDeclaration typeDeclaration)
                                }
                        }

-                       public override void VisitParameterDeclaration(ParameterDeclaration parameterDeclaration)
-                       {
-                               if (inPrimaryConstructor || parameterDeclaration.Parent is not TypeDeclaration)
-                                       base.VisitParameterDeclaration(parameterDeclaration);
-                       }
-
                        public override void VisitMethodDeclaration(MethodDeclaration methodDeclaration)
                        {
                                Visit(methodDeclaration, base.VisitMethodDeclaration);
```

if I change your code like this, it still works... so no need to track the primary constructor separately, just visit it in a different context.

ICSharpCode.Decompiler/CSharp/Transforms/IntroduceUsingDeclarations.cs
siegfriedpammer: if you want to prevent a mostly harmless "double visit" you can keep it like this, but I guess the `inPrimaryConstructor` field is unnecessary, you can just skip if the parent is a TypeDeclaration...

siegfriedpammer: if we keep it, it needs a comment

mmusu3:
I'm trying out your different suggestions and, while removing the `VisitParameterDeclaration` overload does seem to work it doesn't really feel like it should. The adjusted type name gets preserved through the second visit but it seems like the kind of thing that would break if some assumptions changed in the future. I'm no ILSpy expert though.
Your other suggestion to skip visiting parameters that are children of typeDecls also doesn't work unless I misunderstand your comment?
My original implementation still seems like the neatest option even though it uses the extra field but I will go with removing the overload if that's what you prefer.

mmusu3:
After some more thorough testing just removing the `VisitParameterDeclaration` overload does not work. I am getting regressions on my decompilation project that involve parameters with generic type arguments.

Edit.
Here's an example:
**Input**
```cs
private class C10<T>(C10<T>.DataType data)
{
    public struct DataType { public int Value { get; set; } }

    public DataType Data => data;
}
```
**Output**
```cs
// Missing type argument V
private class C10<T>(C10<>.DataType data)
{
    public struct DataType
    {
        public int Value { get; set; }
    }

    public DataType Data => data;
}
```

siegfriedpammer:
that's a different bug; now fixed see [45efc73](https://github.com/icsharpcode/ILSpy/pull/3614/commits/45efc732a726e4efe82930510492bd73b23e2e28)

mmusu3:
There are also more problems from your new commit. `StructDefaultConstructorsAndFieldInitializers = false` doesn't seem to work anymore and XML comments are quite broken now.

Edit. Moved case to new issue #3617

siegfriedpammer:
Please create issues for new problems, otherwise the problems might get lost in a PR comment, if they are not fixed immediately. Let's keep the discussion on this PR focused on the PR topic. Thanks a lot for your understanding.

mmusu3:
Ok, I'll make another issue. Was there anything more you wanted from me in this one?

siegfriedpammer:
> I'm trying out your different suggestions and, while removing the `VisitParameterDeclaration` overload does seem to work it doesn't really feel like it should. The adjusted type name gets preserved through the second visit but it seems like the kind of thing that would break if some assumptions changed in the future. I'm no ILSpy expert though.

I fixed the underlying issue that the annotations created by the `TypeSystemAstBuilder` where not using the correct types. This now makes the the `FullyQualifyAmbiguousTypeNamesVisitor` idempotent. This can be verified by duplicating the `AcceptVisitor` call at the end of the `Run` method.

> Your other suggestion to skip visiting parameters that are children of typeDecls also doesn't work unless I misunderstand your comment?

Yeah, I got it wrong. Sorry!

> My original implementation still seems like the neatest option even though it uses the extra field but I will go with removing the overload if that's what you prefer.

Yes, you are right, seems like we cannot get rid of the extra state. But could you please use try-finally for the assignments and, add a comment/explanation in the VisitParameterDeclaration override?

Thank you very much!

ICSharpCode.Decompiler/CSharp/Transforms/IntroduceUsingDeclarations.cs
siegfriedpammer: Wrong.

mmusu3:
Done. Is that change and comment what you had in mind?

siegfriedpammer:

PR #3619, opened 2025-11-20
Support batch PDB generation.

sonyps5201314:
Added support for multi-select batch PDB generation. When multiple assemblies are selected, the user is prompted to choose a target folder, and files are generated as `{ShortName}.pdb` based on the assembly's short name.

### DIFF
--- a/ILSpy/Util/ShellHelper.cs
+++ b/ILSpy/Util/ShellHelper.cs
@@ -17,11 +17,12 @@
 // DEALINGS IN THE SOFTWARE.
 
 using System;
+using System.Collections.Generic;
+using System.ComponentModel;
 using System.Diagnostics;
 using System.IO;
-using System.Runtime.InteropServices;
 using System.Linq;
-using System.Collections.Generic;
+using System.Runtime.InteropServices;
 
 #pragma warning disable CA1060 // Move pinvokes to native methods class
 
@@ -43,6 +44,7 @@ namespace ICSharpCode.ILSpy.Util
 
 		public static void OpenFolder(string folderPath)
 		{
+			nint folderPidl = IntPtr.Zero;
 			try
 			{
 				if (string.IsNullOrEmpty(folderPath))
@@ -50,122 +52,122 @@ namespace ICSharpCode.ILSpy.Util
 				if (!Directory.Exists(folderPath))
 					return;
 
-				IntPtr folderPidl = IntPtr.Zero;
-				uint attrs;
-				int hr = SHParseDisplayName(folderPath, IntPtr.Zero, out folderPidl, 0, out attrs);
-				if (hr == 0 && folderPidl != IntPtr.Zero)
-				{
-					SHOpenFolderAndSelectItems(folderPidl, 0, null, 0);
-					CoTaskMemFree(folderPidl);
-				}
-				else
-				{
-					// fallback
-					Process.Start(new ProcessStartInfo { FileName = folderPath, UseShellExecute = true });
-				}
+				int hr = SHParseDisplayName(folderPath, IntPtr.Zero, out folderPidl, 0, out var attrs);
+				Marshal.ThrowExceptionForHR(hr);
+
+				hr = SHOpenFolderAndSelectItems(folderPidl, 0, null, 0);
+				Marshal.ThrowExceptionForHR(hr);
 			}
-			catch
+			catch (Exception ex) when (ex is COMException or Win32Exception)
 			{
-				// ignore
+				// fall back to Process.Start
+				OpenFolderFallback(folderPath);
+			}
+			finally
+			{
+				if (folderPidl != IntPtr.Zero)
+					CoTaskMemFree(folderPidl);
+			}
+		}
+
+		static void OpenFolderFallback(string path)
+		{
+			try
+			{
+				Process.Start(new ProcessStartInfo { FileName = path, UseShellExecute = true });
+			}
+			catch (Exception)
+			{
+				// Process.Start can throw several errors (not all of them documented),
+				// just ignore all of them.
 			}
 		}
 
 		public static void OpenFolderAndSelectItem(string path)
 		{
 			// Reuse the multi-item implementation for single item selection to avoid duplication.
-			try
+			if (string.IsNullOrEmpty(path))
+				return;
+			if (Directory.Exists(path))
 			{
-				if (string.IsNullOrEmpty(path))
-					return;
-				if (Directory.Exists(path))
-				{
-					OpenFolder(path);
-					return;
-				}
-
-				if (!File.Exists(path))
-					return;
-
-				OpenFolderAndSelectItems(new[] { path });
+				OpenFolder(path);
+				return;
 			}
-			catch
-			{
-				// ignore
-			}
+
+			if (!File.Exists(path))
+				return;
+
+			OpenFolderAndSelectItems(path);
 		}
 
-		public static void OpenFolderAndSelectItems(IEnumerable<string> paths)
+		public static void OpenFolderAndSelectItems(params IEnumerable<string> paths)
 		{
-			try
+			if (paths == null)
+				return;
+			// Group by containing folder
+			var files = paths.Distinct(StringComparer.OrdinalIgnoreCase).Where(p => !string.IsNullOrEmpty(p) && File.Exists(p)).ToList();
+			if (files.Count == 0)
+				return;
+
+			var itemPidlAllocs = new List<IntPtr>();
+			var relativePidls = new List<IntPtr>();
+
+			foreach (var group in files.GroupBy(Path.GetDirectoryName))
 			{
-				if (paths == null)
-					return;
-				// Group by containing folder
-				var files = paths.Distinct(StringComparer.OrdinalIgnoreCase).Where(p => !string.IsNullOrEmpty(p) && File.Exists(p)).ToList();
-				if (files.Count == 0)
-					return;
+				string folder = group.Key;
+				if (string.IsNullOrEmpty(folder) || !Directory.Exists(folder))
+					continue;
 
-				var groups = files.GroupBy(p => Path.GetDirectoryName(p));
-				foreach (var group in groups)
+				IntPtr folderPidl = IntPtr.Zero;
+
+				try
 				{
-					string folder = group.Key;
-					if (string.IsNullOrEmpty(folder) || !Directory.Exists(folder))
-						continue;
+					int hrFolder = SHParseDisplayName(folder, IntPtr.Zero, out folderPidl, 0, out uint attrs);
+					Marshal.ThrowExceptionForHR(hrFolder);
 
-					IntPtr folderPidl = IntPtr.Zero;
-					uint attrs;
-					int hrFolder = SHParseDisplayName(folder, IntPtr.Zero, out folderPidl, 0, out attrs);
-					if (hrFolder != 0 || folderPidl == IntPtr.Zero)
+					foreach (var file in group)
 					{
-						// fallback: open folder normally
-						OpenFolder(folder);
-						continue;
-					}
-
-					var itemPidlAllocs = new List<IntPtr>();
-					var relativePidls = new List<IntPtr>();
-					try
-					{
-						foreach (var file in group)
+						int hrItem = SHParseDisplayName(file, IntPtr.Zero, out var itemPidl, 0, out attrs);
+						if (hrItem == 0 && itemPidl != IntPtr.Zero)
 						{
-							IntPtr itemPidl = IntPtr.Zero;
-							int hrItem = SHParseDisplayName(file, IntPtr.Zero, out itemPidl, 0, out attrs);
-							if (hrItem == 0 && itemPidl != IntPtr.Zero)
+							IntPtr relative = ILFindLastID(itemPidl);
+							if (relative != IntPtr.Zero)
 							{
-								IntPtr relative = ILFindLastID(itemPidl);
-								if (relative != IntPtr.Zero)
-								{
-									relativePidls.Add(relative);
-									itemPidlAllocs.Add(itemPidl);
-									continue;
-								}
+								relativePidls.Add(relative);
+								itemPidlAllocs.Add(itemPidl);
+								continue;
 							}
-							if (itemPidl != IntPtr.Zero)
-								CoTaskMemFree(itemPidl);
 						}
-
-						if (relativePidls.Count > 0)
-						{
-							SHOpenFolderAndSelectItems(folderPidl, (uint)relativePidls.Count, relativePidls.ToArray(), 0);
-						}
-						else
-						{
-							// nothing to select - open folder
-							OpenFolder(folder);
-						}
+						if (itemPidl != IntPtr.Zero)
+							CoTaskMemFree(itemPidl);
 					}
-					finally
+
+					if (relativePidls.Count > 0)
 					{
-						foreach (var p in itemPidlAllocs)
-							CoTaskMemFree(p);
-						if (folderPidl != IntPtr.Zero)
-							CoTaskMemFree(folderPidl);
+						int hr = SHOpenFolderAndSelectItems(folderPidl, (uint)relativePidls.Count, relativePidls.ToArray(), 0);
+						Marshal.ThrowExceptionForHR(hr);
+					}
+					else
+					{
+						// nothing to select - open folder
+						OpenFolder(folder);
 					}
 				}
-			}
-			catch
-			{
-				// ignore
+				catch (Exception ex) when (ex is COMException or Win32Exception)
+				{
+					// fall back to Process.Start
+					OpenFolderFallback(folder);
+				}
+				finally
+				{
+					foreach (var p in itemPidlAllocs)
+						CoTaskMemFree(p);
+					if (folderPidl != IntPtr.Zero)
+						CoTaskMemFree(folderPidl);
+
+					itemPidlAllocs.Clear();
+					relativePidls.Clear();
+				}
 			}
 		}
 	}
--- a/ILSpy/Commands/ExtractPackageEntryContextMenuEntry.cs
+++ b/ILSpy/Commands/ExtractPackageEntryContextMenuEntry.cs
@@ -178,7 +178,7 @@ namespace ICSharpCode.ILSpy
 				else
 				{
 					if (isFile && File.Exists(path))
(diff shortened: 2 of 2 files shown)


ILSpy/Commands/GeneratePdbContextMenuEntry.cs
```diff
@@ -105,6 +116,109 @@ internal static void GeneratePdbForAssembly(LoadedAssembly assembly, LanguageSer
...
+
+						string fileName = Path.Combine(targetFolder, WholeProjectDecompiler.CleanUpFileName(assembly.ShortName, ".pdb"));
+
+						try
+						{
+							using (FileStream stream = new FileStream(fileName, FileMode.OpenOrCreate, FileAccess.Write))
+							{
+								var decompiler = new CSharpDecompiler(file, assembly.GetAssemblyResolver(options.DecompilerSettings.AutoLoadAssemblyReferences), options.DecompilerSettings);
+								decompiler.CancellationToken = ct;
+								PortablePdbWriter.WritePdb(file, decompiler, options.DecompilerSettings, stream, progress: options.Progress);
```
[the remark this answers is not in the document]
siegfriedpammer: @sonyps5201314 this is a legitimate bug that is in the existing code as well. May I ask you to fix it by switching `FileMode.Create`?

ILSpy/Commands/GeneratePdbContextMenuEntry.cs
```diff
@@ -105,6 +116,109 @@ internal static void GeneratePdbForAssembly(LoadedAssembly assembly, LanguageSer
...
+						}
+						catch (Exception ex)
+						{
+							output.WriteLine(string.Format(Resources.GenerationFailedForAssembly, assembly.FileName, ex.Message));
+						}
+						processed++;
+						if (options.Progress != null)
+						{
+							options.Progress.Report(new DecompilationProgress {
+								Title = "Generating portable PDB...",
```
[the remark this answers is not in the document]
siegfriedpammer: @sonyps5201314 if you want, you can add this. There are other locations where this same string is used, that might need updating as well.

ILSpy/Commands/GeneratePdbContextMenuEntry.cs
```diff
@@ -105,6 +116,109 @@ internal static void GeneratePdbForAssembly(LoadedAssembly assembly, LanguageSer
...
+								UnitsCompleted = processed
+							});
+						}
+					}
+
+					totalWatch.Stop();
+					output.WriteLine();
+					output.WriteLine(Resources.GenerationCompleteInSeconds, totalWatch.Elapsed.TotalSeconds.ToString("F1"));
+					output.WriteLine();
+					output.AddButton(null, Resources.OpenExplorer, delegate { Process.Start("explorer", "/select,\"" + targetFolder + "\""); });
```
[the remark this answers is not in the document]
siegfriedpammer: @sonyps5201314 Yes, this a legitimate bug: `explorer C:\path\to\folder` is the syntax for folders.

ILSpy/Commands/GeneratePdbContextMenuEntry.cs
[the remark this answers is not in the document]
sonyps5201314: Okay, I just noticed this issue too. I'll try to verify it first, and then submit it.

ILSpy/Commands/GeneratePdbContextMenuEntry.cs
```diff
@@ -105,6 +116,109 @@ internal static void GeneratePdbForAssembly(LoadedAssembly assembly, LanguageSer
...
+		internal static void GeneratePdbForAssemblies(System.Collections.Generic.IEnumerable<LoadedAssembly> assemblies, LanguageService languageService, DockWorkspace dockWorkspace)
+		{
+			var assemblyArray = assemblies?.Where(a => a != null).ToArray();
+			if (assemblyArray == null || assemblyArray.Length == 0)
+				return;
+
+			// Ensure at least one assembly supports PDB generation
+			if (!assemblyArray.Any(a => PortablePdbWriter.HasCodeViewDebugDirectoryEntry(a.GetMetadataFileOrNull() as PEFile)))
+			{
+				MessageBox.Show(Resources.CannotCreatePDBFile);
```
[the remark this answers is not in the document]
siegfriedpammer: Maybe you can create a combined message for all files?

like `string.Join(Environment.NewLine, assemblyArray.Where(a => !HasCodeViewDebugDirectory(....)).Select(a => string.Format(Resources...., Path.GetFileName(assembly.FileName)))` (untested, please verify!)

ILSpy/Commands/GeneratePdbContextMenuEntry.cs
```diff
@@ -105,6 +116,109 @@ internal static void GeneratePdbForAssembly(LoadedAssembly assembly, LanguageSer
...
+				return;
+			}
+
+			// Ask for target folder
+			using (var dlg = new System.Windows.Forms.FolderBrowserDialog())
```
siegfriedpammer: Can we switch to a more modern kind of dialog, I think this is the really old one from Windows 95?

ILSpy/Commands/GeneratePdbContextMenuEntry.cs
```diff
@@ -105,6 +116,109 @@ internal static void GeneratePdbForAssembly(LoadedAssembly assembly, LanguageSer
 				return output;
 			}, ct)).Then(dockWorkspace.ShowText).HandleExceptions();
 		}
+
+		internal static void GeneratePdbForAssemblies(System.Collections.Generic.IEnumerable<LoadedAssembly> assemblies, LanguageService languageService, DockWorkspace dockWorkspace)
```
siegfriedpammer: just a nit: I think it would make sense to add a using for `System.Collections.Generic` and simplify the line.

ILSpy/Commands/GeneratePdbContextMenuEntry.cs
```diff
@@ -105,6 +116,109 @@ internal static void GeneratePdbForAssembly(LoadedAssembly assembly, LanguageSer
...
+		{
+			var assemblyArray = assemblies?.Where(a => a != null).ToArray();
```
siegfriedpammer:
```suggestion
			var assemblyArray = assemblies?.Where(a => a != null).ToArray() ?? [];
```

then you can omit the null checks later.

also I would assume that nobody passes null in here... at some point this needs nullable annotations

siegfriedpammer:
Looks good to me! Thank you very much  for contributing this feature!

ILSpy/Commands/GeneratePdbContextMenuEntry.cs
sonyps5201314: It's already a modern dialog since we're on .NET 10.
image

image

ILSpy/Commands/GeneratePdbContextMenuEntry.cs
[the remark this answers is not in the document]
sonyps5201314: https://github.com/icsharpcode/ILSpy/pull/3619/commits/ef52899b0bd7122f7bc3ec922d2bc46534e1521a resolved

ILSpy/Commands/GeneratePdbContextMenuEntry.cs
siegfriedpammer: Got it, thanks!

riQQ: Alternatively, as of .NET 8, there's also an OpenFolderDialog in WPF:
https://learn.microsoft.com/en-us/dotnet/api/microsoft.win32.openfolderdialog?view=windowsdesktop-8.0

siegfriedpammer: I think they both use the same OS code in the end?

@sonyps5201314 if it's not too much work, would you mind trying to use the non-WinForms one?

ILSpy/Commands/GeneratePdbContextMenuEntry.cs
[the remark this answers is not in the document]
sonyps5201314: https://github.com/icsharpcode/ILSpy/pull/3619/commits/71cfce3a486152db9e3139b98ab6c4806459e053 resolved.

ILSpy/Commands/GeneratePdbContextMenuEntry.cs
sonyps5201314: https://github.com/icsharpcode/ILSpy/pull/3619/commits/c5c3505a9851370726add16ce4f48a260ec0d51f resolved.

ILSpy/Commands/GeneratePdbContextMenuEntry.cs
sonyps5201314: https://github.com/icsharpcode/ILSpy/pull/3619/commits/c5c3505a9851370726add16ce4f48a260ec0d51f resolved.

ILSpy/Commands/GeneratePdbContextMenuEntry.cs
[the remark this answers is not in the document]
sonyps5201314: https://github.com/icsharpcode/ILSpy/pull/3619/commits/1f13b80a5f90abebd055c4710b89d344802d39a6 resolved.
This also incidentally resolves another small 'unmodern' issue in ILSpy: every time 'Open or Select File/Folder' was executed, a new `explorer.exe` process was launched, and these processes would not automatically exit even after closing the opened windows.
image

ILSpy/Util/ShellHelper.cs
```diff
@@ -0,0 +1,152 @@
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Linq;
+using System.Collections.Generic;
+
+namespace ICSharpCode.ILSpy.Util
+{
+	static class ShellHelper
```
siegfriedpammer, line 31: @sonyps5201314 why not just keep Process.Start("explorer.exe", ....)? We'd prefer to not introduce more P/Invoke dependencies.

sonyps5201314: Please refer to my explanation at https://github.com/icsharpcode/ILSpy/pull/3619#discussion_r2549765086. It causes significant memory and resource waste, which doesn't align with the pursuit of ultimate performance or code cleanliness (perfectionism). I think that if the intention is indeed to make `ILSpy GUI` cross-platform, the functions inside `ShellHelper.cs` can be implemented using conditional compilation symbols in the future, so this shouldn't be a major issue.

sonyps5201314: Moreover, directly invoking `explorer.exe` does not support opening a folder with multiple files selected. If the output directory already exists and contains other files, the user won't be able to distinguish which ones are the newly generated files.

Additionally, https://github.com/icsharpcode/ILSpy/pull/3619/commits/cae002273cd79ccde6da9fe546ba18c0672d95fc aims to resolve the issue where selecting multiple DLLs located in the same directory results in opening that folder multiple times when directly calling `explorer.exe`.

ILSpy/Commands/GeneratePdbContextMenuEntry.cs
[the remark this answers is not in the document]
sonyps5201314: https://github.com/icsharpcode/ILSpy/pull/3619/commits/5625db2a96d8f05ba6c916dec26c5cfbe69781dc resolved.

ILSpy/Commands/GeneratePdbContextMenuEntry.cs
sonyps5201314: https://github.com/icsharpcode/ILSpy/pull/3619/commits/a973b2e71d1d606980d3cade0df8ced9a91d4901 resolved.

sonyps5201314:
@siegfriedpammer I believe I have addressed all the points you mentioned. The functionality of `GeneratePdbForAssembly` is now fully covered by `GeneratePdbForAssemblies`, so `GeneratePdbForAssembly` can be removed to simplify maintenance.

Additionally, I intentionally did not set an initial directory for `OpenFolderDialog` because I prefer to use the user's last selected directory. If we default to the target DLL's directory, there is a risk of overwriting an existing official PDB, which could break the software's integrity verification.

ILSpy/Util/ShellHelper.cs
sonyps5201314: I noticed today that dotPeek does not have this issue, and I observed the same with TortoiseGit yesterday. The problem where `explorer.exe` fails to exit does not affect every machine; it occurs on my physical development machine (Win11 23H2), but not on my VMs (Win11 24H2 and Win10 22H2). Regardless, switching to the Shell API definitely provides a better user experience, so I believe the change is worthwhile.

ILSpy/Util/ShellHelper.cs
```diff
@@ -0,0 +1,152 @@
+using System;
```
siegfriedpammer, line 19: Needs license header

ILSpy/Util/ShellHelper.cs
```diff
@@ -0,0 +1,152 @@
...
+								IntPtr relative = ILFindLastID(itemPidl);
+								if (relative != IntPtr.Zero)
+								{
+									relativePidls.Add(relative);
+									itemPidlAllocs.Add(itemPidl);
+									continue;
+								}
+							}
+							if (itemPidl != IntPtr.Zero)
+								CoTaskMemFree(itemPidl);
```
siegfriedpammer: I think, this needs try-finally too?

ILSpy/Util/ShellHelper.cs
```diff
@@ -0,0 +1,152 @@
...
+				}
+
+				if (!File.Exists(path))
+					return;
+
+				OpenFolderAndSelectItems(new[] { path });
+			}
+			catch
+			{
+				// ignore
```
siegfriedpammer: what kind of exceptions are we ignoring here? "catch all" is considered bad pratice.

ILSpy/Util/ShellHelper.cs
```diff
@@ -0,0 +1,152 @@
...
+				}
+				else
+				{
+					// fallback
+					Process.Start(new ProcessStartInfo { FileName = folderPath, UseShellExecute = true });
+				}
+			}
+			catch
+			{
+				// ignore
```
siegfriedpammer: what kind of exceptions are we ignoring here? "catch all" is considered bad practice.

ILSpy/Util/ShellHelper.cs
```diff
@@ -0,0 +1,152 @@
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Linq;
+using System.Collections.Generic;
+
```
siegfriedpammer, line 26:
```suggestion

#pragma warning disable CA1060 // Move pinvokes to native methods class

```

ILSpy/TreeNodes/AssemblyTreeNode.cs
```diff
@@ -690,15 +690,18 @@ public void Execute(TextViewContext context)
 		{
 			if (context.SelectedTreeNodes == null)
 				return;
+			var paths = new List<string>();
```
siegfriedpammer, line 693: maybe it would make sense to use a HashSet<string> here? Note that this would need a `StringComparer.OrdinalIgnoreCase` on Windows :)

ILSpy/Util/ShellHelper.cs
sonyps5201314: https://github.com/icsharpcode/ILSpy/pull/3619/commits/e027a1a38a9a367ab6d8fdc2bb4fe536d637d6fb resolved.

ILSpy/TreeNodes/AssemblyTreeNode.cs
sonyps5201314: I don't think a change is strictly necessary here. If multiple instances of the same file appear in the Assembly List, it implies that the insertion logic (`AssemblyList.OpenAssembly`) isn't handling things correctly. The restriction should be enforced at that layer, rather than becoming a burden for internal calls.

 Given that `AssemblyList` defines:
 `readonly Dictionary<string, LoadedAssembly> byFilename = new Dictionary<string, LoadedAssembly>(StringComparer.OrdinalIgnoreCase);`
 we can infer that ILSpy doesn't suffer from this issue. However, I did notice a few days ago that `dnSpyEx` indeed has a problem with duplicate file entries.

 That being said, your point is valid. Since relying on external restrictions increases the cognitive load on the caller (requiring the user to pay special attention), I went ahead and handled it inside `OpenFolderAndSelectItems`.
https://github.com/icsharpcode/ILSpy/pull/3619/commits/752e6ed9c049e95923581a3d16d997d5e36e6be5 resolved.

ILSpy/Util/ShellHelper.cs
sonyps5201314: I didn't actually get this warning when I compiled, but I still added it as you suggested.
https://github.com/icsharpcode/ILSpy/pull/3619/commits/b631b55054dc90e2faf2f415c4e4c451202793a4 resolved.

ILSpy/Util/ShellHelper.cs
sonyps5201314: Unlike C, which typically uses return values to signal errors, C# is a language that relies heavily on exceptions. Since almost any method call has the potential to throw, adding a try-catch block is certainly an effective measure to prevent the application from crashing.

These potential exceptions include `Process.Start` failures, `DllImport("shell32.dll")` errors (such as failing to load the DLL or finding the specific entry point), and even managed exceptions thrown from within P/Invoke during extreme scenarios (e.g., when hooked by tools like EasyHook).

Given that the consequence of a failure in `ShellHelper`'s public API is merely that a folder fails to open—which is a minor impact—I believe adding a general exception handler is acceptable here. (Actually, I didn't add these catch blocks myself; I simply retained them because I agree they are necessary.)

ILSpy/Util/ShellHelper.cs
sonyps5201314: https://github.com/icsharpcode/ILSpy/pull/3619#discussion_r2558955460 answered.

sonyps5201314:
@siegfriedpammer Thank you for taking the time to review the code again. I have addressed all your questions and made the necessary changes. Is there anything else that needs modification?

PR #3620, opened 2025-11-22
Improve decompilation of unmanaged function pointers

ds5678:
### Problem

Previously, decompilation for `ldftn` treated all methods as managed with no custom calling conventions.

```cs
return (delegate* unmanaged[Cdecl]<void>)(delegate*<void>)&LocalFunction;
```

### Solution

  * ExpressionBuilder::VisitLdFtn now properly constructs the calling conventions.
  * FunctionPointerType::FromSignature now checks whether a modopt type affects the calling convention.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/FunctionPointers.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/FunctionPointers.cs
@@ -129,6 +129,32 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 			}
 		}
 
+		public unsafe delegate* unmanaged[Cdecl, Fastcall]<void> AddressOfLocalFunction_CDeclAndFastcall()
+		{
+			return &LocalFunction;
+
+			[UnmanagedCallersOnly(CallConvs = new Type[] {
+				typeof(CallConvCdecl),
+				typeof(CallConvFastcall)
+			})]
+			static void LocalFunction()
+			{
+			}
+		}
+
+		public unsafe delegate* unmanaged[Fastcall, Cdecl]<void> AddressOfLocalFunction_FastcallAndCDecl()
+		{
+			return &LocalFunction;
+
+			[UnmanagedCallersOnly(CallConvs = new Type[] {
+				typeof(CallConvFastcall),
+				typeof(CallConvCdecl)
+			})]
+			static void LocalFunction()
+			{
+			}
+		}
+
 #if NET60
 		public unsafe delegate* unmanaged[Cdecl, SuppressGCTransition]<void> AddressOfLocalFunction_CDeclAndSuppressGCTransition()
 		{
--- a/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
@@ -4611,7 +4611,7 @@ namespace ICSharpCode.Decompiler.CSharp
 					var builder = ImmutableArray.CreateBuilder<IType>(array.Length);
 					foreach (var type in array.Select(a => a.Value).OfType<IType>())
 					{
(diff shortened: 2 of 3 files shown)


ds5678:
I'm not sure my implementation is entirely correct.

```cs
// Source
public unsafe delegate* unmanaged[Cdecl, Fastcall]<void> AddressOfLocalFunction_CDeclAndFastcall()
{
	return &LocalFunction;

	[UnmanagedCallersOnly(CallConvs = new Type[] {
		typeof(CallConvCdecl),
		typeof(CallConvFastcall)
	})]
	static void LocalFunction()
	{
	}
}

// Decompiled
public unsafe delegate* unmanaged[Fastcall]<void> AddressOfLocalFunction_CDeclAndFastcall()
{
	return &LocalFunction;

	[UnmanagedCallersOnly(CallConvs = new Type[] {
		typeof(CallConvCdecl),
		typeof(CallConvFastcall)
	})]
	static void LocalFunction()
	{
	}
}
```

ds5678:
There's still an annoyance.

```cs
// Source
public unsafe delegate* unmanaged[Fastcall, Cdecl]<void> AddressOfLocalFunction_FastcallAndCDecl2()
{
	return &LocalFunction;

	[UnmanagedCallersOnly(CallConvs = new Type[] {
		typeof(CallConvCdecl),
		typeof(CallConvFastcall)
	})]
	static void LocalFunction()
	{
	}
}
public unsafe delegate* unmanaged[Fastcall, Cdecl]<void> AddressOfLocalFunction_FastcallAndCDecl3()
{
	return AddressOfLocalFunction_CDeclAndFastcall();
}

// Decompiled
public unsafe delegate* unmanaged[Fastcall, Cdecl]<void> AddressOfLocalFunction_FastcallAndCDecl2()
{
	return (delegate* unmanaged[Fastcall, Cdecl]<void>)(delegate* unmanaged[Cdecl, Fastcall]<void>)(&LocalFunction);

	[UnmanagedCallersOnly(CallConvs = new Type[] {
		typeof(CallConvCdecl),
		typeof(CallConvFastcall)
	})]
	static void LocalFunction()
	{
	}
}
public unsafe delegate* unmanaged[Fastcall, Cdecl]<void> AddressOfLocalFunction_FastcallAndCDecl3()
{
	return (delegate* unmanaged[Fastcall, Cdecl]<void>)AddressOfLocalFunction_CDeclAndFastcall();
}
```

It originates from the fact that the compiler treats the calling convention attributes as an unordered set, but we use ordered list equality.

However, I'm not sure this matters much, since such code is likely to be [extremely rare](https://github.com/search?q=%22unmanaged%5BCdecl%2C+Stdcall%5D%22&type=code), and the decompiled code is semantically correct, albeit ugly.

ds5678:
@siegfriedpammer Is this a pull request you can review, or does it need someone else's eyes?

siegfriedpammer:
> However, I'm not sure this matters much, since such code is likely to be [extremely rare](https://github.com/search?q=%22unmanaged%5BCdecl%2C+Stdcall%5D%22&type=code), and the decompiled code is semantically correct, albeit ugly.

I am not exactly sure why you think it is ugly? If the compiler treats the calling conventions as an unordered set, then there is no way to know the source-defined order, right? Also, there are many examples of information getting lost in the compilation process, different syntaxes producing the same IL code.

I don't think "normalizing" the order of calling conventions is a big deal. I'd even say it's good to be deterministic/consistent, as some people use text comparison to compare different versions of the same assembly and have requested an option that custom attributes should always be sorted and not reproduced in metadata order (which is default behavior of the decompiler).

siegfriedpammer:
> @siegfriedpammer Is this a pull request you can review, or does it need someone else's eyes?

I will try to look into it this evening as the changes are relatively small.

siegfriedpammer:

ds5678:
Thank you for reviewing it!

PR #3598, opened 2025-10-15
Fixed the issue where IsPrimaryConstructor's judgment was too conservative, which resulted in the inability to generate correct code in many scenarios.

sonyps5201314:
some bad test cases:
```cs
    class WebPair(string name)
    {
        public string Name { get; } = name;
    }
```

```cs
    class WebPair(string name, string? value, ref readonly object encode)
    {
        public string Name { get; } = name;
    }
```
```cs
    class WebPair(string name, string? value, bool encode = false)
    {
        public string Name { get; } = name;
        public string? Value { get; } = value;
        string? WebValue { get; } = encode ? "111" : value;
        //string? WebValue { get; } = "111" + value;
    }
```

```cs
    class WebPair(string name, string? value, ref readonly object encode)
    {
        public string Name { get; } = name;
        public string? Value { get; } = value;
        string? WebValue { get; } = encode is null ? "111" : value;
        string? WebValue2 { get; } = encode.ToString();
    }
```

In short:
You cannot assume that the number of constructor parameters and properties is equal, nor can you assume the relationship between the number of instructions and the number of parameters. You cannot assume that `xxx` in an assignment statement: `var property = xxx;` is necessarily a value or object. It could be a complex expression like a conditional expression or a concatenation call. You only need to ensure that the value is ultimately assigned to the member field.

PR #3629, opened 2025-12-01
Fix the issue where PDBs generated by ILSpy do not match certain asse…

sonyps5201314:
image

As shown in the image above, the file `C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\IDE\Extensions\Microsoft\Copilot\Conversations.Service\Microsoft.VisualStudio.Copilot.Service.dll` contains two sets of CodeView entries in its Debug Directory:

| Type | Age | GUID \| Timestamp | PDB Name |
| :--- | :--- | :--- | :--- |
| RSDS | 00000001 | {FD8F8A2F-2C21-C8C0-51B2-72A6D9F1736B} | Microsoft.VisualStudio.Copilot.Service.ni.pdb |
| RSDS | 00000001 | {9DD21ECB-9171-4F44-B337-CF21CA9180B7} | D:\a\_work\1\s\obj\src\Copilot.Service\x64\Release\net8.0\Microsoft.VisualStudio.Copilot.Service.pdb |

The first entry points to a non-Portable PDB filename (specifically a Native Image `.ni.pdb`). We must skip this entry; otherwise, Visual Studio will fail to load the generated PDB for debugging.

Since ILSpy has lagged significantly behind the rapid evolution of C# standards in recent years, many assemblies cannot be perfectly decompiled into source projects that can be successfully recompiled. In this situation, PDB debugging serves as a critical mitigation for these issues. Therefore, I hope that the PDB feature is at least perfectly usable.

### DIFF
--- a/ICSharpCode.Decompiler/DebugInfo/PortablePdbWriter.cs
+++ b/ICSharpCode.Decompiler/DebugInfo/PortablePdbWriter.cs
@@ -235,8 +235,9 @@ namespace ICSharpCode.Decompiler.DebugInfo
 
 			if (pdbId == null)
 			{
-				var debugDir = file.Reader.ReadDebugDirectory().FirstOrDefault(dir => dir.Type == DebugDirectoryEntryType.CodeView);
+				var debugDir = file.Reader.ReadDebugDirectory().LastOrDefault(dir => dir.Type == DebugDirectoryEntryType.CodeView);
 				var portable = file.Reader.ReadCodeViewDebugDirectoryData(debugDir);
+				Debug.Assert(!portable.Path.EndsWith(".ni.pdb"));
 				pdbId = new BlobContentId(portable.Guid, debugDir.Stamp);
 			}
 


sonyps5201314:
The code has been updated to fix the issue where `C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\IDE\PrivateAssemblies\vsdebugeng.manimpl.dll` was no longer supported.

sonyps5201314:
I have updated the code to fix the lack of support for ReadyToRun assemblies, such as `"C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\ServiceHub\SharedAssemblies\Microsoft.ServiceHub.HostLib.dll"`.

I must admit that `dotPeek` has very high code quality; the issues I've encountered recently are not present there, and this fix actually draws some inspiration from `dotPeek`'s approach.

That said, `ILSpy`'s code quality is also excellent. Through our discussions and my review of the codebase, I can see that the author is very meticulous. The current minor issues seem to stem mostly from the test cases being too limited in scope.

christophwille:
Test cases that could be added as unit tests would be very welcome to not regress in the future.

ICSharpCode.Decompiler/DebugInfo/PortablePdbWriter.cs
```diff
@@ -235,7 +235,7 @@ string BuildFileNameFromTypeName(TypeDefinitionHandle handle)
 
 			if (pdbId == null)
 			{
-				var debugDir = file.Reader.ReadDebugDirectory().FirstOrDefault(dir => dir.Type == DebugDirectoryEntryType.CodeView);
+				var debugDir = file.Reader.ReadDebugDirectory().LastOrDefault(dir => dir.Type == DebugDirectoryEntryType.CodeView);
```
siegfriedpammer, line 238: is it always guaranteed to be the last entry? Wouldn't it make more sense to check that the extension is not `.ni.pdb`?

I did some testing by scanning all the .dlls on my machine and reading the DebugDirectory and it seems that at least the Microsoft compiler puts the non-native image entry last.

sonyps5201314: Yes, I actually considered adding a line like `Debug.Assert(!portable.Path.EndsWith("ni.pdb"));` for more robust verification, or to alert us to unexpected scenarios in the future.

However, after checking `dotPeek`'s code, I noticed that they always iterate from the end. Trusting their implementation, I decided to simply omit that assertion.

siegfriedpammer: Sounds good, could you add a comment to document the assumption? Thank you!

siegfriedpammer:
@sonyps5201314 do you think there is a way to improve test coverage for PDBs?... would be nice to have some tests for this change.

ICSharpCode.Decompiler/DebugInfo/PortablePdbWriter.cs
sonyps5201314: Of course, feel free to modify this further, or we can wait until an unexpected scenario arises before making adjustments.

Arguably, the most reliable method would be to reverse engineer the relevant code in the VS debugger to confirm the exact matching logic. However, I suspect that `dotPeek` has already done this, which is why I decided to adopt an approach similar to theirs.

sonyps5201314:
> @sonyps5201314 do you think there is a way to improve test coverage for PDBs?... would be nice to have some tests for this change.

As I mentioned previously in https://github.com/icsharpcode/ILSpy/pull/3592#issuecomment-3396002140, using all managed DLLs from the latest Visual Studio as test targets prior to release would undoubtedly help create a near-perfect C# decompiler and PDB generator.

We could implement a new `Roundtrip` test case that specifically targets VS assemblies. However, this would certainly create a significant amount of workload in the short term and inevitably reveal many existing issues. That said, there is no need to rush; we can simply add these items to the roadmap and tackle them one by one.

One major challenge here is that manually verifying semantic correctness is a massive undertaking. Introducing the [IL comparison I implemented previously](https://github.com/icsharpcode/ILSpy/pull/3598/commits/f478f366bb3f3699ff5ac248ff46644e74222e85) would alleviate this burden.

However, ILSpy currently removes `NOP` instructions before generating the AST. This potentially creates issues for the subsequent [IL comparison](https://github.com/icsharpcode/ILSpy/pull/3598/commits/2a887cc5fd3442ef08aeef48d8fc1928ce80fce6) and [PDB sequence point generation](https://github.com/icsharpcode/ILSpy/pull/3623#issuecomment-3577124100). Therefore, should we consider **not** removing them?

sonyps5201314:
> Test cases that could be added as unit tests would be very welcome to not regress in the future.

I apologize, but I am unable to provide samples. I have currently only observed the issue of multiple debug entries in DLLs generated by Microsoft's proprietary toolchain. Since these files are protected by copyright and I do not have the right to redistribute them, I cannot share copies.

siegfriedpammer:
> However, ILSpy currently removes `NOP` instructions before generating the AST. This potentially creates issues for the subsequent [IL comparison](https://github.com/icsharpcode/ILSpy/pull/3598/commits/2a887cc5fd3442ef08aeef48d8fc1928ce80fce6) and [PDB sequence point generation](https://github.com/icsharpcode/ILSpy/pull/3623#issuecomment-3577124100). Therefore, should we consider **not** removing them?

`nop` instructions would not fit into the ILAst tree model, because they can exist at literally any point in the IL instruction stream and would cause a lot of problems for the pattern detection pipeline.

For the specific use case of PDB generation, I believe that we could detect nops from holes in the  `ILRange` attached to the syntax nodes.

siegfriedpammer:
@jacdavis of the VS debugger team implemented some "hacks" to improve sequence points for methods compiled in Release mode (where `nop`s don't exist at all); but I believe the PDB generator is already able to infer `nop`s and use them for opening/closing braces?

ICSharpCode.Decompiler/DebugInfo/PortablePdbWriter.cs
sonyps5201314: > Sounds good, could you add a comment to document the assumption? Thank you!

I didn't know how to describe this assumption, so I simply added a call to `Debug.Assert` in the latest commit to execute this assumption.

PR #3630, opened 2025-12-01
Replace GeneratePdbForAssembly with GeneratePdbForAssemblies to improve user experience and remove redundant code.

sonyps5201314:
The `GeneratePdbForAssembly` function internally sets the `InitialDirectory` of the `SaveFileDialog`, whereas `GeneratePdbForAssemblies` does not. This inconsistency causes confusion when users switch between batch and single-file PDB generation, as the behavior regarding the output directory (e.g., whether it remembers the last used folder) varies.

Since the functionality of `GeneratePdbForAssemblies` now fully covers that of `GeneratePdbForAssembly`, there is no longer any need to retain the latter.


siegfriedpammer:
Thank you very much, sorry for taking so long to look at this!

PR #3644, opened 2025-12-24
Do not use ObservableObject when ObservableObjectBase is sufficient.

tom-englert:
Architectural cleanup

### Problem
Classes derive from ObservableObject, but ObservableObjectBase is sufficient.

### Solution
Always use ObservableObjectBase to have minimal dependency


christophwille:
Minor nit. UIThread at least for me would imply a thread, not the dispatcher.

siegfriedpammer:

PR #3647, opened 2025-12-28
Add traditional chinese translation

new1271:
Add translation for traditional chinese(zh-Hant).

### DIFF
--- a/ILSpy/Properties/Resources.zh-Hant.resx
+++ b/ILSpy/Properties/Resources.zh-Hant.resx
@@ -694,7 +694,7 @@
     <value>ILSpyAboutPage_zh_Hant.txt</value>
   </data>
   <data name="ILSpyVersion" xml:space="preserve">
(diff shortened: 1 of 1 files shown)


christophwille:
Thank you for shining a bright light on an area of ILSpy we haven't had on our radar for years, and in fact dreaded to address. For one, there is the technical issue that adding another language is a hassle, as modifying csprojs should never be needed for making changes to supported languages.

Secondly, there are additional "costs" of having translations that bothered us for a long time:
* translation drift because of renames (menu items, options, error messages... you-name-it)
* terminology in translations / bug reports about untranslated strings
* no way for us (core team) to validate the translations for content
* no way for us (core team) to maintain translations, not even via automation (see #3551)
* given how highly technical the audience is, a translation shouldn't be necessary in the first place

Therefore, sorry, we cannot merge this PR and actually in addition we'll be removing existing translations.

PR #3653, opened 2026-01-18
Render operating system name as well as architecture if present

MatthewSteeples:
I was diagnosing an issue with our ReadyToRun assemblies and when I loaded them in ILSpy I noticed that instead of indicating which platform they were built for it was just a number.

Digging into it, I found [this comment](https://github.com/dotnet/runtime/issues/36364#issuecomment-628232472) that indicates that if the binaries are machine specific then the value is xor'd with a constant (like flags).

This Pull Request displays the OS (if present).

I'm happy for the format of the text output to be changed, or any other changes for that matter!

### DIFF
--- a/ILSpy/Languages/Language.cs
+++ b/ILSpy/Languages/Language.cs
@@ -387,7 +387,7 @@ namespace ICSharpCode.ILSpy
 
 			if (!Enum.IsDefined(architecture))
 			{
-				foreach ((var osEnum, var osText) in osMachineLookup)
+				foreach (var (osEnum, osText) in osMachineLookup)
 				{
 					var candidate = architecture ^ osEnum;
 					if (Enum.IsDefined(candidate))


ILSpy/Languages/Language.cs
```diff
@@ -367,28 +367,54 @@ public virtual CodeMappingInfo GetCodeMappingInfo(MetadataFile module, EntityHan
...
 			var headers = module.Reader.PEHeaders;
 			var architecture = headers.CoffHeader.Machine;
 			var characteristics = headers.CoffHeader.Characteristics;
 			var corflags = headers.CorHeader.Flags;
+
+			var modifier = string.Empty;
+
+			if (!Enum.IsDefined(architecture))
+			{
+				foreach (var kvp in osMachineLookup)
```
siegfriedpammer: I would've used tuple deconstruction syntax here for clarity

MatthewSteeples: Nice suggestion. Done

ILSpy/Languages/Language.cs
```diff
@@ -367,28 +367,54 @@ public virtual CodeMappingInfo GetCodeMappingInfo(MetadataFile module, EntityHan
...
 			var headers = module.Reader.PEHeaders;
 			var architecture = headers.CoffHeader.Machine;
 			var characteristics = headers.CoffHeader.Characteristics;
 			var corflags = headers.CorHeader.Flags;
+
+			var modifier = string.Empty;
+
+			if (!Enum.IsDefined(architecture))
+			{
+				foreach ((var osEnum, var osText) in osMachineLookup)
```
siegfriedpammer:
```suggestion
				foreach (var (osEnum, osText) in osMachineLookup)
```

PR #3666, opened 2026-02-17
Add ReferenceTrimmer and remove unused references

dfederm:
Add ReferenceTrimmer to enable detection of unnecessary project, package, and assembly references during build. Remove all unused references identified by ReferenceTrimmer.

Note: Because both ProjectReferences and PackageReferences are transitive, sometimes removing an unused dependency will cause build errors due to missing a dependency that was previously pull in transitively. In these cases, the reference is explicitly added. So despite this change having many added references, it's actually reducing references as those added references were previously underdefined but existing dependencies (due to transitivity).

Disclosure: I own ReferenceTrimmer, but I do think it will be helpful to this repo, as evidenced by the unused references it was able to identify. Adding it to the repo will help keep the references clean.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,6 +1,7 @@
 <Project>
   <PropertyGroup>
-    <NoWarn>$(NoWarn);NU1510</NoWarn>
+    <NoWarn>$(NoWarn);NU1510;CS1591</NoWarn>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
   <PropertyGroup>
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -56,6 +56,7 @@
     <PackageVersion Include="System.Text.RegularExpressions" Version="4.3.1" />
   </ItemGroup>
   <ItemGroup>
+    <GlobalPackageReference Include="ReferenceTrimmer" Version="3.4.5" />
     <GlobalPackageReference Include="TomsToolbox.Composition.Analyzer" Version="2.22.2" />
   </ItemGroup>
 </Project>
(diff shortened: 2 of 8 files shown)


Directory.Build.props
```diff
@@ -1,6 +1,7 @@
 <Project>
   <PropertyGroup>
-    <NoWarn>$(NoWarn);NU1510</NoWarn>
+    <NoWarn>$(NoWarn);NU1510;CS1591</NoWarn>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
siegfriedpammer, line 4: why did you add this?

dfederm: It improves the accuracy of the Roslyn APIs that ReferenceTrimmer uses. The compiler doesn't report accurate results without it.

More info: https://github.com/dfederm/ReferenceTrimmer?tab=readme-ov-file#c

siegfriedpammer: Ah, yeah that's weird Roslyn behavior. I don't think that the way comments are parsed should influence that, but OK.

Can we make this whole feature opt-in? I don't think we want to have ReferenceTrimmer listed as a dependency in the SBOM or the lock files...

Also the usefulness of this feature is not clear... The only projects where it found something are our unit test projects where we don't really have to be careful about dependencies.

dfederm: Sure, I can make it either opt in or I can just remove those bits and just remove the existing unused refs and call it a day. Ill defer to you.

The benefit of leaving it enabled though to tonstay clean wrt references. But ultimately its up tonyou. 

Note: it won't be listed as a dependency in nuget packages if thats what you mean. Its a dev dependency only. If you mean a dependency semantically, or of the repo as a whole, then sure.

christophwille:
I have been thinking about this for quite some time, conclusion: once we need reference trimming on stuff we ship, we'll reconsider adding it. At the moment it doesn't add value to customers.

Issue #2691, opened 2022-05-16
CultureNotFoundException when decompiling a satellite assembly


pvoosten:
Since version 7.2.0.6791-preview3, trying to extract the resources from a satellite assembly often results in the following stack trace:

```
System.Globalization.CultureNotFoundException: Only the invariant culture is supported in globalization-invariant mode. See https://aka.ms/GlobalizationInvariantMode for more information. (Parameter 'name')
it is an invalid culture identifier.
   at System.Globalization.CultureInfo..ctor(String name, Boolean useUserOverride)
   at System.Reflection.AssemblyName.set_CultureName(String value)
   at System.Reflection.Metadata.MetadataReader.GetAssemblyName(StringHandle nameHandle, Version version, StringHandle cultureHandle, BlobHandle publicKeyOrTokenHandle, AssemblyHashAlgorithm assemblyHashAlgorithm, AssemblyFlags flags)
   at System.Reflection.Metadata.AssemblyDefinition.GetAssemblyName()
   at ICSharpCode.Decompiler.Metadata.DotNetCorePathFinderExtensions.DetectTargetFrameworkId(PEReader assembly, String assemblyPath) in D:\GitWorkspace\ILSpy_72\ICSharpCode.Decompiler\Metadata\DotNetCorePathFinderExtensions.cs:line 56
   at ICSharpCode.Decompiler.Console.ILSpyCmdProgram.DecompileAsProject(String assemblyFileName, String outputDirectory) in D:\GitWorkspace\ILSpy_72\ICSharpCode.Decompiler.Console\IlspyCmdProgram.cs:line 217
   at ICSharpCode.Decompiler.Console.ILSpyCmdProgram.OnExecute(CommandLineApplication app) in D:\GitWorkspace\ILSpy_72\ICSharpCode.Decompiler.Console\IlspyCmdProgram.cs:line 101
```

In the previous version, 7.2.0.6702-preview2, the issue was not present yet.

To reproduce the issue, I used the following commands:
```PowerShell
dotnet tool uninstall ilspycmd -g
dotnet tool install ilspycmd --version <version> -g
rm -Recurse <output-directory>\*
ilspycmd <assembly-name>.resources.dll -p -o <output-directory>
```

A workaround is to use an older version of ilspycmd.

siegfriedpammer:
The error message hints at the following change: https://github.com/icsharpcode/ILSpy/commit/db6b08548494195b75240066151a0dd2be2e45e0

[...]

PR #3667, opened 2026-02-22
Replace FileVersionInfo.GetVersionInfo with DecompilerVersionInfo.Version constant

MattParkerDev:
### Problem
FileVersionInfo.GetVersionInfo requires a path to an assembly on disk. This fails in cases where the assembly has been loaded as a stream, rather than from a file.

### Solution
This PR replaces it with Assembly.GetName().Version, which I have verified returns the same string, in my case - "9.1.0.7988".


siegfriedpammer:
This is a good catch. I wonder why we are using `FileVersionInfo` in the first place, given that there is `DecompilerVersionInfo`, which is generated/updated on every build. So there should be no need for `FileVersionInfo` at runtime nor reflection.

Suggestion:
- You could add `public const string FileVersion = Major + "." + Minor + "." + Build + "." + Revision;` to `DecompilerVersionInfo.template.cs`
- Use it in `AssemblyInfo.cs` and `PortablePdbWriter.cs`

This would eliminate the need for a file path AND make it AOT-/single-file bundle compatible, since we don't rely on reflection.

siegfriedpammer:
Addendum: I think, I found another reason for not using reflection: #2691

Seems constructing `System.Reflection.AssemblyName` involves a lot more than one might think and may cause weird effects.

MattParkerDev:
Good idea, done! :)

PR #3678, opened 2026-03-15
Update TransformCollectionAndObjectInitializers to check for init-only properties

### LINKED ISSUE #3677: ILSpy does not use object initializers for class/structs with init-only properties
If an object with init-only properties is initialized in a function outside of a constructor, **and** it happens to be allocated locally and isn't in a compiler generated method/class, then ILSpy won't use an object initializer for it and will instead attempt to set the properties after using the default constructor, which results in the compile failing.

Please note: the following is dependent on whether or not the C# compiler decides to allocate the object assigned to `field1` on the stack or the heap, and I can't figure out a consistent method of forcing it to allocate on the heap.

### Input code
```
    public class Bar{
        public int? Prop1 { get; init; }
    }

    public class BarInitializer{
        public volatile Bar? field1 = null;
        public BarInitializer()
        {
            initialize();
        }

        private void initialize()
        {
            field1 = new Bar
            {
                Prop1 = 1
            };
        }
    }
```

### Erroneous output
```
    public class Bar{
        public int? Prop1 { get; init; }
    }

    public class BarInitializer{
        public volatile Bar? field1 = null;
        public BarInitializer()
        {
            initialize();
        }

        private void initialize()
        {
            field1 = new Bar();
            field1.Prop1 = 1; // Prop1 is init-only, so this is an compile error
        }
    }
```
If the output has the wrong behavior, explain how it differs from the expected behavior.

### Details
* Product in use: ILSpyCmd
* Version in use: current master

nikitalita:
### Problem
Fixes #3677

### Solution
The main culprit was here:

```c#
					if (newObjInst.ILStackWasEmpty && v.Kind == VariableKind.Local
						&& !currentMethod.IsConstructor
						&& !currentMethod.IsCompilerGeneratedOrIsInCompilerGeneratedClass())
```
Adding a simple check to see if the declaring type has init-only properties fixes the issue.

I would add a test that covers this, but I do not know how to force the C# compiler to allocate an object locally instead of on the stack consistently; if you have any tips, let me know.

### DIFF
--- a/ICSharpCode.Decompiler/IL/Transforms/TransformCollectionAndObjectInitializers.cs
+++ b/ICSharpCode.Decompiler/IL/Transforms/TransformCollectionAndObjectInitializers.cs
@@ -58,7 +58,7 @@ namespace ICSharpCode.Decompiler.IL.Transforms
 			{
 				case NewObj newObjInst:
 					if (newObjInst.ILStackWasEmpty && v.Kind == VariableKind.Local
-						&& !TypeContainsInitOnlyProperties(newObjInst.Method.DeclaringType.GetDefinition())
+						&& !TypeContainsInitOnlyProperties(newObjInst.Method.DeclaringTypeDefinition)
 						&& !currentMethod.IsConstructor
 						&& !currentMethod.IsCompilerGeneratedOrIsInCompilerGeneratedClass())
 					{


siegfriedpammer:
> I would add a test that covers this, but I do not know how to force the C# compiler to allocate an object locally instead of on the stack consistently; if you have any tips, let me know.

I couldn't find a definitive rule for when the Roslyn compiler uses local slots over stack slots. The main observation is that Roslyn prefers stack over locals especially in Release builds.

Some of the things just cannot be tested in C# pretty tests so you may have to resort to IL pretty tests.

ICSharpCode.Decompiler/NRExtensions.cs
```diff
@@ -73,6 +73,18 @@ public static bool ContainsAnonymousType(this IType type)
 			return visitor.ContainsAnonType;
 		}
 
+		public static bool ContainsInitOnlyProperties(this IType type)
```
siegfriedpammer: I don't think this makes sense as an extension method. Please move it to the `TransformCollectionAndObjectInitializers` class instead.

Also the receiver type should be `ITypeDefinition` since you are calling `GetDefinition()` on it immediately anyway... let the caller deal with getting the definition. Returning false for all types that are not `ITypeDefinition`s seems to be a bad idea.

nikitalita: > I don't think this makes sense as an extension method. Please move it to the TransformCollectionAndObjectInitializers class instead.
>
> Also the receiver type should be ITypeDefinition since you are calling GetDefinition() on it immediately anyway... let the caller deal with getting the definition. 

Done.

> Returning false for all types that are not ITypeDefinitions seems to be a bad idea.

What do you suggest instead? If there is no type definition, should we just return true? I don't believe there's a way to determine if it has InitOnly properties otherwise...

siegfriedpammer: My idea was that we change the parameter type from `IType` to `ITypeDefinition` and let the **caller** deal with resolving the definition and also let the caller decide what to do when it's not a definition at all.

ICSharpCode.Decompiler/IL/Transforms/TransformCollectionAndObjectInitializers.cs
```diff
@@ -58,6 +58,7 @@ void IStatementTransform.Run(Block block, int pos, StatementTransformContext con
 			{
 				case NewObj newObjInst:
 					if (newObjInst.ILStackWasEmpty && v.Kind == VariableKind.Local
+						&& !TypeContainsInitOnlyProperties(newObjInst.Method.DeclaringType.GetDefinition())
```
siegfriedpammer:
```suggestion
						&& !TypeContainsInitOnlyProperties(newObjInst.Method.DeclaringTypeDefinition)
```
makes it a bit shorter ;)

ICSharpCode.Decompiler/IL/Transforms/TransformCollectionAndObjectInitializers.cs
```diff
@@ -187,6 +188,18 @@ void IStatementTransform.Run(Block block, int pos, StatementTransformContext con
...
+		private static bool TypeContainsInitOnlyProperties(ITypeDefinition? typeDefinition)
+		{
+			foreach (var property in typeDefinition?.Properties ?? [])
+			{
+				if (property.Setter?.IsInitOnly ?? false)
+				{
+					return true;
+				}
+			}
+			return false;
```
siegfriedpammer, line 200: I think this is fine now, since it's a private static helper in this transform, we can just return false here.

PR #3680, opened 2026-03-18
Dedicated UI and Decompiler APIs for C# 14 extensions

siegfriedpammer:
grafik


christophwille:
@dgrunwald if you got a minute to spare, maybe you can have a look too. Thanks.

Issue #3578, opened 2025-10-05
`in` is erroneously applied to parameter when parameter is `this`


nikitalita:
If a templated function has a parameter is declared with `in`, and it is called with `this` as a parameter, ILSpy will erroneously apply `in` to it.

This only seems to happen on templated functions; if the function being called is a non-template, ILSpy will create a temporary variable set to `this` and use that as a parameter.

## Input
```C#
public class TestClass
{
    
    public void DoThing<T>(in T x)
    {
        Console.WriteLine(x.ToString());
    }

    public void DoThingToSelf()
    {
        DoThing(this);
    }
    
}
```

## Erroneous output
```C#
using System;

public class TestClass
{
	public void DoThing<T>(in T x)
	{
		Console.WriteLine(x.ToString());
	}

	public void DoThingToSelf()
	{
		DoThing(in this);
	}
}
```

### Details
* Product in use: IlSpyCmd
* Version in use: current master

PR #3682, opened 2026-03-20
Fix `in` erroneously being applied to `this`

nikitalita:
### Problem
Fixes https://github.com/icsharpcode/ILSpy/issues/3578

### Solution
* Fairly simple: in `ChangeDirectionExpressionTo()`, we just check if the parameter is a `this` expression and if the direction is `in` or `ref readonly`; if so, we don't change it.


siegfriedpammer:

PR #3694, opened 2026-03-26
Fix syntax error in CreateBoldRun method for CROSS_PLATFORM

lextm:
### Problem

A small break in the conditional statement

### Solution

Remove the extra `)`.


christophwille:
Sorry for that!

Issue #3689, opened 2026-03-23
The property 'ColumnDefinitions' is read-only and cannot be set from markup


CreateAndInject:
### Input code
```
   <Grid>
		<Grid.ColumnDefinitions></Grid.ColumnDefinitions>
    </Grid>
```

### Erroneous output
```
<Grid ColumnDefinitions="" />
```

PR #3701, opened 2026-03-31
Only move properties with setters to XAML attributes

ElektroKill:
Link to issue(s) this covers:
https://github.com/icsharpcode/ILSpy/issues/3689

### Problem
The `AttributeRewritePass` would create assignments via attributes despite the property not having a setter. 

### Solution
The problem was remedied by checking if the property has a setter.

I also cleaned up the code a little.

I did not add a test as the input code provided in the issue will not compile under .NET 1,0 which is the TFM of the BamlDecompiler test project. The provided input code does compile on .NET 8.


siegfriedpammer:

PR #3712, opened 2026-04-06
Fix SequencePointBuilder skipping moved member initializers

nikitalita:
### Problem
SequencePointBuilder was not visiting any member initializers that were lifted out of the constructor, so sequence points weren't being built for instructions corresponding to those.

### Solution
* This changes SequencePointBuilder such that, when it visits Member, Field, and Event declarations, and if their initializers are set, they visit those as sequence points.
* I would write a test for this, but it seems there aren't any SequencePointBuilder tests currently, so I'm not sure how to set that up with your current testing apparatus.

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/SequencePointBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/SequencePointBuilder.cs
@@ -163,11 +163,27 @@ namespace ICSharpCode.Decompiler.CSharp
 			}
 		}
 
+		public override void VisitFieldDeclaration(FieldDeclaration fieldDeclaration)
+		{
+			foreach (var variable in fieldDeclaration.Variables)
+			{
+				if (!variable.Initializer.IsNull)
+				{
+					VisitAsSequencePoint(variable.Initializer);
+				}
+			}
+			base.VisitFieldDeclaration(fieldDeclaration);
+		}
+
 		public override void VisitPropertyDeclaration(PropertyDeclaration propertyDeclaration)
 		{
 			if (!propertyDeclaration.ExpressionBody.IsNull)
 			{
 				VisitAsSequencePoint(propertyDeclaration.ExpressionBody);
+			}
+			else if (!propertyDeclaration.Initializer.IsNull)
+			{
+				VisitAsSequencePoint(propertyDeclaration.Initializer);
 			}
 			else
 			{
@@ -200,6 +216,18 @@ namespace ICSharpCode.Decompiler.CSharp
 				VisitAsSequencePoint(inc);
 			}
 			VisitAsSequencePoint(forStatement.EmbeddedStatement);
+		}
+
+		public override void VisitEventDeclaration(EventDeclaration eventDeclaration)
+		{
+			foreach (var variable in eventDeclaration.Variables)
+			{
+				if (!variable.Initializer.IsNull)
+				{
+					VisitAsSequencePoint(variable.Initializer);
+				}
+			}
+			base.VisitEventDeclaration(eventDeclaration);
 		}
 
 		public override void VisitSwitchStatement(SwitchStatement switchStatement)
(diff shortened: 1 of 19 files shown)


ICSharpCode.Decompiler/CSharp/SequencePointBuilder.cs
```diff
@@ -163,12 +163,28 @@ public override void VisitBlockStatement(BlockStatement blockStatement)
 			}
 		}
 
+		public override void VisitFieldDeclaration(FieldDeclaration fieldDeclaration)
+		{
+			foreach (var variable in fieldDeclaration.Variables)
+			{
+				if (!variable.AssignToken.IsNull || !variable.Initializer.IsNull)
```
siegfriedpammer: I don't think that there can be an initializer without an assign token?


```suggestion
				if (!variable.Initializer.IsNull)
```

ICSharpCode.Decompiler/CSharp/SequencePointBuilder.cs
```diff
@@ -202,6 +218,18 @@ public override void VisitForStatement(ForStatement forStatement)
 			VisitAsSequencePoint(forStatement.EmbeddedStatement);
 		}
 
+		public override void VisitEventDeclaration(EventDeclaration eventDeclaration)
+		{
+			foreach (var variable in eventDeclaration.Variables)
+			{
+				if (!variable.AssignToken.IsNull || !variable.Initializer.IsNull)
```
siegfriedpammer:
```suggestion
				if (!variable.Initializer.IsNull)
```

ICSharpCode.Decompiler/CSharp/SequencePointBuilder.cs
```diff
@@ -163,12 +163,28 @@ public override void VisitBlockStatement(BlockStatement blockStatement)
 			}
 		}
 
+		public override void VisitFieldDeclaration(FieldDeclaration fieldDeclaration)
+		{
+			foreach (var variable in fieldDeclaration.Variables)
+			{
+				if (!variable.AssignToken.IsNull || !variable.Initializer.IsNull)
+				{
+					VisitAsSequencePoint(variable);
```
siegfriedpammer:
```suggestion
					VisitAsSequencePoint(variable.Initializer);
```

is it on purpose that you pass the `VariableInitializer` instead of just the `Expression`?

ICSharpCode.Decompiler/CSharp/SequencePointBuilder.cs
```diff
@@ -202,6 +218,18 @@ public override void VisitForStatement(ForStatement forStatement)
 			VisitAsSequencePoint(forStatement.EmbeddedStatement);
 		}
 
+		public override void VisitEventDeclaration(EventDeclaration eventDeclaration)
+		{
+			foreach (var variable in eventDeclaration.Variables)
+			{
+				if (!variable.AssignToken.IsNull || !variable.Initializer.IsNull)
+				{
+					VisitAsSequencePoint(variable);
```
siegfriedpammer:
```suggestion
					VisitAsSequencePoint(variable.Initializer);
```

dgrunwald:
How does this work when there are multiple constructors?

siegfriedpammer:
I think it should work, because we are mapping multiple IL offsets to the same source locations. The other way around would be problematic.

siegfriedpammer:
we definitely need to find a good way to test PDB generation. The current setup is flawed.

dgrunwald:
> because we are mapping multiple IL offsets to the same source locations

Are we? I'm not sure what the annotations on the field initializer AST nodes are, but I would be surprised if this just works without mixing up IL offsets from different constructors.

dee9226 (approved):

siegfriedpammer:

PR #3726, opened 2026-04-26
Fix #3705: Code window is empty when select a .baml and refresh

### LINKED ISSUE #3705: Code window is empty when select a .baml and refresh
### Steps to reproduce
1. Open any WPF assembly
2. Navigate to a .baml in resources
3. Reload all assemblies

christophwille:
image
image
image
image
image

### DIFF
--- a/ILSpy/AssemblyTree/AssemblyTreeModel.cs
+++ b/ILSpy/AssemblyTree/AssemblyTreeModel.cs
@@ -947,11 +947,30 @@ namespace ICSharpCode.ILSpy.AssemblyTree
 
 		private void RefreshInternal()
 		{
+			_ = RefreshInternalAsync();
+		}
+
+		private async Task RefreshInternalAsync()
+		{
 			using (Keyboard.FocusedElement.PreserveFocus())
 			{
 				var path = GetPathForNode(SelectedItem);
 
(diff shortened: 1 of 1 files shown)


ILSpy/AssemblyTree/AssemblyTreeModel.cs
```diff
@@ -946,12 +946,31 @@ public void Refresh()
...
+
+		private async Task RefreshInternalAsync()
 		{
 			using (Keyboard.FocusedElement.PreserveFocus())
 			{
 				var path = GetPathForNode(SelectedItem);
 
 				ShowAssemblyList(settingsService.AssemblyListManager.LoadList(AssemblyList.ListName));
+
+				// Ensure assembly loaded before FindNodeByPath to allow lazy-loaded resource nodes to be found
```
siegfriedpammer: For consistency with the same idiom at line 270 in this file, consider matching the existing wording:

```suggestion
				// FindNodeByPath() blocks the UI if the assembly is not yet loaded,
				// so use an async wait instead.
```

The block at lines 261-275 uses this same loop+await pattern with that comment; aligning the phrasing makes the two sites read as obvious siblings.

PR #3731, opened 2026-05-07
Decompile C# 15 runtime-async (`/features:runtime-async=on`)

siegfriedpammer:
## Summary

Reverses Roslyn's C# 15 runtime-async lowering so methods compiled with `/features:runtime-async=on` round-trip back to their source form. The runtime-async feature replaces the compiler-generated state-machine class with a method-impl flag (`MethodImplOptions.Async` = 0x2000) and a small set of runtime helper calls; that means our existing async-decompilation pipeline (built around recognising the state machine) sees no state machine at all and bails. This branch adds the missing reversers as a single new IL transform `RuntimeAsyncExceptionRewriteTransform` invoked from `AsyncAwaitDecompiler` when the method has the runtime-async impl bit, plus minor type-system / helper changes.

What's covered:

- `Task` / `Task<T>` / `ValueTask` / `ValueTask<T>` / `[AsyncMethodBuilder]`-attributed custom task return types (via a new `TaskType.UnpackAnyTask`).
- `try { await … } catch (T ex) { …; [throw;] }` and `try { await … } finally { … }` — including the `ExceptionDispatchInfo.Capture/Throw` rethrow idiom and the captured-object machinery the runtime emits.
- `catch (T ex) when (filter)` and the typeless `catch when (filter)` form (Roslyn lowers both to `catch object when ({ isinst T; obj=ex; <filter> })`).
- Multi-handler try-catch (each handler sets a unique flag value, post-catch switch dispatches into the right body).
- `await` on Task awaitables (`AsyncHelpers.Await(task)` -> IL `Await`) and on non-Task awaitables (the manual `GetAwaiter` / `IsCompleted` / `UnsafeAwaitAwaiter` / `GetResult` pattern).
- `await foreach` (falls out of the existing `UsingTransform` once the manual await pattern is collapsed and we feed it IL `Await`).
- Async struct methods — Roslyn copies `*this` into a local at entry; we remap reads of that local back through the function's `this` parameter so `i++` doesn't decompile as `<copy>.i++`.
- Early `return value;` from inside a `try { … }` enclosed in a try-finally with await — runtime-async encodes it as a flag plus a post-finally `if (flag == K) return capture;`; we collapse it back to a direct return inside the try so the user's finally still runs.

The new transform shares the catch-handler match and the dominator-based block move with the existing `AwaitInFinallyTransform` (extracted as `internal static` helpers) so we're not duplicating the state-machine-async logic.

`ICSharpCode.Decompiler.Tests/Helpers/Tester.cs` now passes `/debug:portable` for runtime-async compiles — without the state machine, the user's local variable names only survive in the PDB.

The pretty-test sources (`Async.cs`, `AsyncForeach.cs`, `AsyncMain.cs`, `AsyncStreams.cs`, `AsyncUsing.cs`, `CustomTaskType.cs`) are reused by adding `RuntimeAsync*` test methods that compile the same source with `/features:runtime-async=on` and expect the same decompiled output.

## Test plan

- [ ] `dotnet test --filter \"FullyQualifiedName~PrettyTestRunner.RuntimeAsync\"` — 12/12 pass (the six new test methods × Optimize / non-Optimize).
- [ ] `dotnet test --filter \"Name=Async|Name=AsyncForeach|Name=AsyncMain|Name=AsyncStreams|Name=AsyncUsing|Name=CustomTaskType\"` — 98/98 classical-async pass (no regressions in the existing state-machine async path).
- [ ] Full `PrettyTestRunner` sweep — no regressions outside the async area.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/Helpers/Tester.cs
+++ b/ICSharpCode.Decompiler.Tests/Helpers/Tester.cs
@@ -501,6 +501,10 @@ namespace System.Runtime.CompilerServices
 				preprocessorSymbols.Add("LEGACY_CSC");
 				preprocessorSymbols.Add("LEGACY_VBC");
 			}
+			if (flags.HasFlag(CompilerOptions.EnableRuntimeAsync))
+			{
+				preprocessorSymbols.Add("RUNTIMEASYNC");
+			}
 			return preprocessorSymbols;
 		}
 
--- a/ICSharpCode.Decompiler.Tests/Helpers/Tester.cs
+++ b/ICSharpCode.Decompiler.Tests/Helpers/Tester.cs
@@ -848,6 +848,8 @@ namespace System.Runtime.CompilerServices
 				suffix += ".mcs2";
 			if ((cscOptions & CompilerOptions.UseMcs5_23) != 0)
 				suffix += ".mcs5";
+			if ((cscOptions & CompilerOptions.EnableRuntimeAsync) != 0)
+				suffix += ".runtimeasync";
 			return suffix;
 		}
 
--- a/ICSharpCode.Decompiler.Tests/Helpers/Tester.cs
+++ b/ICSharpCode.Decompiler.Tests/Helpers/Tester.cs
@@ -73,6 +73,7 @@ namespace ICSharpCode.Decompiler.Tests.Helpers
 		CheckForOverflowUnderflow = 0x20000,
 		ProcessXmlDoc = 0x40000,
 		UseRoslyn4_14_0 = 0x80000,
+		EnableRuntimeAsync = 0x100000,
 		UseMcsMask = UseMcs2_6_4 | UseMcs5_23,
 		UseRoslynMask = UseRoslyn1_3_2 | UseRoslyn2_10_0 | UseRoslyn3_11_0 | UseRoslyn4_14_0 | UseRoslynLatest
 	}
@@ -605,12 +606,17 @@ namespace System.Runtime.CompilerServices
 				if (roslynVersion != "legacy")
 				{
 					otherOptions += "/shared ";
-					if (!targetNet40 && Version.Parse(RoslynToolset.SanitizeVersion(roslynVersion)).Major > 2)
+					var version = Version.Parse(RoslynToolset.SanitizeVersion(roslynVersion));
+					if (!targetNet40 && version.Major > 2)
 					{
 						if (flags.HasFlag(CompilerOptions.NullableEnable))
 							otherOptions += "/nullable+ ";
 						else
 							otherOptions += "/nullable- ";
+					}
+					if (!targetNet40 && roslynVersion == roslynLatestVersion && flags.HasFlag(CompilerOptions.EnableRuntimeAsync))
+					{
+						otherOptions += "/features:runtime-async=on ";
 					}
 				}
 
--- a/ICSharpCode.Decompiler/IL/ControlFlow/AsyncAwaitDecompiler.cs
+++ b/ICSharpCode.Decompiler/IL/ControlFlow/AsyncAwaitDecompiler.cs
@@ -121,7 +121,11 @@ namespace ICSharpCode.Decompiler.IL.ControlFlow
 			awaitDebugInfos.Clear();
 			moveNextLeaves.Clear();
 			if (!MatchTaskCreationPattern(function) && !MatchAsyncEnumeratorCreationPattern(function))
+			{
+				if (function.IsAsync && context.Settings.RuntimeAsync)
+					RuntimeAsyncExceptionRewriteTransform.Run(function, context);
 				return;
+			}
 			try
 			{
 				AnalyzeMoveNext();
--- a/ICSharpCode.Decompiler/IL/ControlFlow/AsyncAwaitDecompiler.cs
+++ b/ICSharpCode.Decompiler/IL/ControlFlow/AsyncAwaitDecompiler.cs
@@ -124,6 +124,7 @@ namespace ICSharpCode.Decompiler.IL.ControlFlow
 			{
 				if (function.IsAsync && context.Settings.RuntimeAsync)
 				{
+					TranslateThisCopyAccess(function);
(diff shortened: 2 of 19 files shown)


ICSharpCode.Decompiler.Tests/Helpers/Tester.cs
```diff
@@ -605,13 +606,18 @@ public static async Task<CompilerResults> CompileCSharp(string sourceFileName, C
...
+					if (!targetNet40 && version.Major > 2)
 					{
 						if (flags.HasFlag(CompilerOptions.NullableEnable))
 							otherOptions += "/nullable+ ";
 						else
 							otherOptions += "/nullable- ";
 					}
+					if (!targetNet40 && roslynVersion == roslynLatestVersion && flags.HasFlag(CompilerOptions.EnableRuntimeAsync))
+					{
+						otherOptions += "/features:runtime-async=on ";
```
siegfriedpammer, line 623: @christophwille I think you found out, that this is no longer needed?

ICSharpCode.Decompiler/IL/ControlFlow/AsyncAwaitDecompiler.cs
```diff
@@ -121,7 +121,11 @@ public void Run(ILFunction function, ILTransformContext context)
 			awaitDebugInfos.Clear();
 			moveNextLeaves.Clear();
 			if (!MatchTaskCreationPattern(function) && !MatchAsyncEnumeratorCreationPattern(function))
+			{
+				if (function.Method?.IsAsync == true)
```
siegfriedpammer: I think this needs a decompiler settings check, should only run if Runtime Async is enabled.

ICSharpCode.Decompiler/IL/ControlFlow/RuntimeAsyncExceptionRewriteTransform.cs
```diff
@@ -0,0 +1,1400 @@
...
+				return;
+
+			bool changed = false;
+
+			// Convert `call AsyncHelpers.Await(value)` to the IL Await instruction. The state-machine
+			// async pipeline (AsyncAwaitDecompiler) produces IL Await for downstream transforms
+			// (UsingTransform's MatchDisposeBlock and friends pattern-match on it via UnwrapAwait);
+			// for runtime-async we have to do the same conversion ourselves so the rest of the
+			// pipeline behaves uniformly.
+			foreach (var call in function.Descendants.OfType<CallInstruction>().ToArray())
```
siegfriedpammer: this part needs to be extracted and run always, not just if `AwaitInCatchFinally` is active. Note, that there is a similar transform in the `ReplaceMethodCallsWithOperators` step. Can we deduplicate this? `ExpressionTransforms` might be too late? `EarlyExpressionTransforms` too early? Can you check that?

ICSharpCode.Decompiler.Tests/Helpers/Tester.cs
```diff
@@ -627,6 +633,12 @@ public static async Task<CompilerResults> CompileCSharp(string sourceFileName, C
 				{
 					otherOptions += "-debug:full ";
 				}
+				else if (flags.HasFlag(CompilerOptions.EnableRuntimeAsync))
```
siegfriedpammer: Revert this! If test cases need special handling because of renaming, you can use `#if OPT` to use different names in optimized tests.

ICSharpCode.Decompiler/IL/ControlFlow/RuntimeAsyncExceptionRewriteTransform.cs
```diff
@@ -0,0 +1,1400 @@
...
+						changed = true;
+						continue;
+					}
+				}
+			}
+
+			// Final pass: collapse the manual GetAwaiter/IsCompleted/UnsafeAwaitAwaiter/GetResult
+			// pattern that runtime-async emits for non-Task awaitables (Task.Yield(), MoveNextAsync,
+			// etc.) into a single Await IL instruction.
+			foreach (var block in function.Descendants.OfType<Block>().ToArray())
```
siegfriedpammer: Shouldn't this be a separate transform too? Can this pattern only occur in try-finally/try-catch awaits?

ICSharpCode.Decompiler/IL/ControlFlow/RuntimeAsyncExceptionRewriteTransform.cs
```diff
@@ -0,0 +1,1400 @@
...
+			// pauseBlock: call AsyncHelpers.UnsafeAwaitAwaiter(ldloc awaiter); br completedBlock
+			if (pauseBlock.Instructions.Count != 2)
+				return false;
+			if (pauseBlock.Instructions[0] is not Call pauseCall)
+				return false;
+			if (pauseCall.Method.Name != "UnsafeAwaitAwaiter" || pauseCall.Arguments.Count != 1)
+				return false;
+			if (!pauseCall.Arguments[0].MatchLdLoc(awaiterVar))
+				return false;
+			if (!pauseCall.Method.DeclaringType.FullName.EndsWith("AsyncHelpers"))
```
siegfriedpammer: Always check the full name of the declaring type! You could extract this into a helper method, because there are multiple members to verify.

ICSharpCode.Decompiler/TypeSystem/IMethod.cs
```diff
@@ -111,5 +111,11 @@ public interface IMethod : IParameterizedMember
 		/// If this method is already specialized, the new substitution is composed with the existing substition.
 		/// </summary>
 		new IMethod Specialize(TypeParameterSubstitution substitution);
+
+		/// <summary>
+		/// Returns <see langword="true"/>, if the method is <see cref="MethodImplAttributes.Async"/>.
+		/// <see langword="false"/> otherwise.
+		/// </summary>
+		bool IsAsync { get; }
```
siegfriedpammer: Remove this... why not use `MetadataMethod` directly and access MethodImplOptions from metadata?

PR #3744, opened 2026-05-17
feat(ui): add option to disable menu popup animations (#3737)

### LINKED ISSUE #3737: Disable menu animations
When opening a menu (eg, "view") by clicking on it, or mousing over a menu item that opens a submenu (eg "view/theme"), the newly appearing menu plays an animation where it slowly fades into existence over the course of a few seconds. This is obnoxious and not something I ever want any program to do. Ideally it would detect the system animation settings, which already work to disable similar window animations in most programs, but adding a dedicated option for it would also work. (If you do go the route of detecting system settings, I suggest you also use them to disable smooth scrolling (#3393) by default, for similar reasons.)

wyf027:
## Summary

Fixes #3737.

Adds **Enable menu animations** to Display options (under Other). When unchecked, menu and submenu popups use `PopupAnimation.None` instead of the default fade-in.

- New installs default to the current OS menu animation preference.
- The setting is persisted in `ILSpy.xml` and applied at startup and when changed in options.

## Test plan

- [ ] Open **View → Options → Display** and toggle **Enable menu animations**; open top-level menus and submenus (e.g. View → Theme) and confirm instant vs animated open.
- [ ] Restart ILSpy and confirm the preference is restored.
- [ ] CI build (no local .NET SDK in contributor environment; relying on GitHub Actions)

### DIFF
--- a/ILSpy/App.xaml.cs
+++ b/ILSpy/App.xaml.cs
@@ -29,6 +29,7 @@ using System.Windows.Threading;
 
 using ICSharpCode.ILSpy.AppEnv;
 using ICSharpCode.ILSpy.AssemblyTree;
+using ICSharpCode.ILSpy.Util;
 using ICSharpCode.ILSpyX.Analyzers;
 
 using Medo.Application;
@@ -109,6 +110,7 @@ namespace ICSharpCode.ILSpy
 			Resources.MergedDictionaries.Add(DataTemplateManager.CreateDynamicDataTemplates(ExportProvider));
 
 			var sessionSettings = settingsService.SessionSettings;
+			MenuPopupAnimationHelper.Apply(settingsService.DisplaySettings.EnableMenuAnimations);
 			ThemeManager.Current.Theme = sessionSettings.Theme;
 			if (!string.IsNullOrEmpty(sessionSettings.CurrentCulture))
 			{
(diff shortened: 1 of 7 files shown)


siegfriedpammer:
> [!NOTE]
> This comment was generated by Claude (claude-fable-5 via Claude Code) and is posted on its behalf after reading it.

Thanks for taking a look at #3737. We are closing this PR for two reasons:

**1. It does not compile — the central API call is hallucinated.** `MenuPopupAnimationHelper.cs` calls `SystemParameters.OverrideMetadata(SystemParameters.MenuPopupAnimationKey, new FrameworkPropertyMetadata(...))`, which does not exist in WPF: `OverrideMetadata` is a `DependencyProperty` method, and `MenuPopupAnimationKey` is a `ResourceKey`, not a dependency property. Building `ILSpy/ILSpy.csproj` at the PR head fails with CS0117. Please build before submitting; the repo compiles with `TreatWarningsAsErrors`, so anything that does not build locally cannot pass CI either.

**2. It no longer applies — ILSpy's UI is now Avalonia, not WPF.** This branch is based on a master from before the Avalonia 12 port landed; it patches files that no longer exist (`App.xaml.cs`, `DisplaySettingsPanel.xaml`, `Util/SettingsService.cs`) and relies on WPF-only concepts (`SystemParameters`, `PopupAnimation`) that have no Avalonia equivalent.

PR #3747, opened 2026-05-25
CustomDebugInformation: decode EmbeddedSource, split columns

fowl2:
### Problem
Embedded source custom debug info displays as hex:

image

### Solution
Decode it:
image

Also:
* split "Kind" (combined Guid heap offset, Guid, description) into Kind, KindGUID and KindString, respecively
* add "Info" with embedded source format and length
* add ValueLength with blob length

---

Future: It would probably be nice if there was some way to select/override the decoding method (eg. force a specific character encoding / base64 / hex), open in a tab, add syntax highlighting, save the raw content, etc.


siegfriedpammer:
I ported your proposed changes to Avalonia, fixed the warnings and added a few tests. Thank you for your contribution!

Issue #2716, opened 2022-06-16
Attributes sorting order


greenozon:
**Usually if you do export to project serveral times, ILSpy saves types with accompanying attributes in different order**
e.g.:

![image](https://user-images.githubusercontent.com/765343/174022762-806a20a0-6dd8-4335-8bd5-8090093961da.png)

**Describe the solution you'd like**
It'd be very nice to happen if the order of attributes would be sorted in some order and will be constant from export to export

Such feature was implemented in dnSpy and could be turned ON/OFF in Options:
![image](https://user-images.githubusercontent.com/765343/174024387-c2da528d-ceef-4e23-b586-03b72da19e58.png)

Overall benefit of this feature would be to have shorter time to find the diff between multiple exports thus making end-users even more happy :)

MeFisto94:
Was also going to report this issue, but not on the UI level but the nuget pkg/decompiler settings class.
I don't know if IL is preserving attribute order and this is a bug or if the order of attributes is just not defined.
In the latter case, I'd suggest adding a DecompilerSetting "sort attributes in alphabetical order" (or similar) and just sorting the set when rendering the code.

That would greatly help my tooling by reducing faulty diffs (that basically happen on every file that has more than two attributes on an element)

dgrunwald:
ILSpy should alway maintain the order of the attributes from the CustomAttributes metadata table.
I think it's the C# compiler that emits attributes in non-deterministic order. The C# compiler needs to sort the attribute table by the parent metadata token, so my guess is that they're using a non-stable sort for doing that.

dgrunwald:
Actually I'm not sure who's doing the sorting -- System.Reflection.Metadata also has logic to unstably sort the custom attribute table if they're not already sorted (CustomAttributeTableReader ctor).

xiaoxiao921:
Temp solution for anyone still intersted in this:
```diff
diff --git a/ICSharpCode.Decompiler/CSharp/OutputVisitor/CSharpOutputVisitor.cs b/ICSharpCode.Decompiler/CSharp/OutputVisitor/CSharpOutputVisitor.cs
index 96d591dab..057a5b385 100644
--- a/ICSharpCode.Decompiler/CSharp/OutputVisitor/CSharpOutputVisitor.cs
+++ b/ICSharpCode.Decompiler/CSharp/OutputVisitor/CSharpOutputVisitor.cs
[...]

Issue #2071, opened 2020-07-17
Add a option to sort switch cases


wwh1004:
If the order of basic blocks is changed, the decompilation result of the switch statement will change (e.g. de4dot cflow)
``` cs
after de4dot cflow:
switch (osType)
{
default:
	throw new InvalidOperationException("Shouldn't be here");
case OSType.Unix:
	Unix.Dispose(intPtr, origDataLength);
	break;
case OSType.Windows:
	Windows.Dispose(intPtr);
	break;
}

before de4dot cflow (original):
switch (osType)
{
case OSType.Windows:
	Windows.Dispose(intPtr);
	break;
case OSType.Unix:
	Unix.Dispose(intPtr, origDataLength);
	break;
default:
	throw new InvalidOperationException("Shouldn't be here");
}
```
So sorting cases by some rules may help, such as sorting by number value for byte short int long and sorting by string value for string

dgrunwald:
We intentionally sort by IL offset.

With non-obfuscated code, the C# compiler emits IL in the same order as the original source code; so we try to preserve that order.

PR #3750, opened 2026-05-29
Add opt-in SortSwitchSections setting (#3749)

### LINKED ISSUE #3749: Add a option to sort switch cases
Requesting an opt-in decompiler setting that sorts switch sections by their label value instead of by IL offset.

Today, `SortSwitchSections` in `ICSharpCode.Decompiler/IL/ControlFlow/SwitchDetection.cs` orders sections primarily by branch target IL offset, with the label value as a tiebreaker. For non-obfuscated assemblies that matches source order and is the correct default, which is exactly the position taken when #2071 was closed. I'm not asking to change that default.

The friction shows up with obfuscated assemblies. A lot of obfuscators reshuffle basic-block layout on every build, so the IL offsets behind a `switch` come out in arbitrary order even when the case-to-value mapping is identical. Decompiling the same logical method across two consecutive builds then produces a textual diff that is purely a reordering of the case arms, with no real behavior change. Diffing decompiled output across versions of a closed-source app is a real use case (it's what I'm doing here, and #3494 was filed for the same kind of workflow).

There's already a clean precedent for this. #2716 added the `SortCustomAttributes` setting because attribute order had the same "non-significant per the spec, but unstable in IL across rebuilds" problem, and #3494 / #3505 plumbed those settings through `ilspycmd` via `-ds Name=Value` and `--ilspy-settingsfile`. The end-to-end plumbing for a new opt-in setting already exists.

Concretely:

- New `SortSwitchSections` (or similar) bool on `DecompilerSettings`, defaulting to `false` so current output is unchanged.
- When `true`, `SortSwitchSections(sw)` orders by `s.Labels.Values.FirstOrDefault()` instead of by IL offset.
- No effect on non-obfuscated assemblies unless the user opts in.

danbopes:
Closes #3749.

### Problem

`SortSwitchSections` in `SwitchDetection.cs` orders sections primarily by the target block's IL offset, with the label value as a tiebreaker. That matches source order for non-obfuscated assemblies and is the right default (which is the position the maintainers took when #2071 was closed).

For obfuscated assemblies the situation is different: many obfuscators reshuffle basic-block layout on every build, so the IL offsets behind a `switch` come out in arbitrary order even when the case-to-value mapping is identical. Decompiling the same logical method across two consecutive builds then produces a textual diff that is purely a reordering of the case arms, with no real behavior change. Diffing decompiled output across versions of a closed-source app is a real workflow (it's what motivated the issue, and #3494 was filed for the same kind of use case).

### Solution

Add a new opt-in `SortSwitchSections` bool on `DecompilerSettings`, defaulting to `false` so current output is unchanged. When `true`, `SortSwitchSections(sw, context)` orders sections by `s.Labels.Values.FirstOrDefault()` instead of by IL offset. This mirrors the precedent set by `SortCustomAttributes` (#2716), which addressed the same "non-significant per the spec, but unstable in IL across rebuilds" problem for attribute order; the end-to-end plumbing through `ilspycmd` via `-ds Name=Value` (#3494 / #3505) already exists.

A few notes on the approach:

* `SortSwitchSections` now takes `ILTransformContext` so it can read `context.Settings`. The two call sites (`ProcessBlock`, `SimplifySwitchInstruction`) already had `context` in scope.
* Section reordering is purely cosmetic for the C# output. `StatementBuilder.TranslateSwitch` resolves the default section via `inst.GetDefaultSection()` (picked by label count, not position) and emits each `case X:` label explicitly, so position in the list is not load-bearing. Inter-section transitions (`goto case`, `goto default`) are explicit gotos at the IL level. The Resources.resx entry is added so the WPF settings UI shows a proper label rather than the raw key.

The most attention-worthy bit is the test fixture: the `.il` file uses a hand-written switch table whose targets are at non-monotonic IL offsets (simulating obfuscator block-reordering), and the `.cs` expected file is the sorted output with the setting enabled. The assembly was built locally with the runtime.linux-x64 ilasm package and the expected output verified end-to-end via `ilspycmd`; CI on this branch has confirmed it passes the test runner's formatter as well.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/ILPrettyTestRunner.cs
+++ b/ICSharpCode.Decompiler.Tests/ILPrettyTestRunner.cs
@@ -339,6 +339,12 @@ namespace ICSharpCode.Decompiler.Tests
 			await Run();
 		}
 
+		[Test]
+		public async Task SortSwitchSections()
+		{
+			await Run(settings: new DecompilerSettings { SortSwitchSections = true, FileScopedNamespaces = false });
+		}
+
 		async Task Run([CallerMemberName] string testName = null, DecompilerSettings settings = null,
 			AssemblerOptions assemblerOptions = AssemblerOptions.Library)
 		{
--- a/ICSharpCode.Decompiler.Tests/TestCases/ILPretty/SortSwitchSections.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/ILPretty/SortSwitchSections.cs
@@ -1,0 +1,35 @@
+// Copyright (c) AlphaSierraPapa for the SharpDevelop Team
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this
+// software and associated documentation files (the "Software"), to deal in the Software
+// without restriction, including without limitation the rights to use, copy, modify, merge,
+// publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons
+// to whom the Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all copies or
+// substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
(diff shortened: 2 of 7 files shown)


siegfriedpammer:

PR #3752, opened 2026-06-04
Fix anonymous-type lambda early-return emitting unresolvable cast

### LINKED ISSUE #3751: Decompilation of anonymous-type returning lambda with early return emits `<>f__AnonymousTypeN`
### Input code

A `return null;` statement inside a lambda whose return type is inferred to an anonymous type. Minimal repro:
```c#
using System;

class Repro
{
    static bool Cond;

    static T Infer<T>(Func<T> factory) => factory();   // T is inferred to the anonymous type

    public object Trigger()
    {
        return Infer(() =>
        {
            if (Cond)
            {
                Console.WriteLine();
                return null;
            }
            return new { Value = 1 };
        });
    }
}
```

### Erroneous output

```c#
public object Trigger()
{
    return Infer(delegate
    {
        if (Cond)
        {
            Console.WriteLine(); // prevent ternary
            return (<>f__AnonymousType0<int>)null;
        }
        return new { Value = 1 };
    });
}
```

`return (<>f__AnonymousType0<int>)null;` does not recompile — `<>f__AnonymousType0` is not a referenceable identifier:

```
error CS1525: Invalid expression term '<'
```

### Details

* Product in use: ICSharpCode.Decompiler (via ilspycmd)
* Version in use: 10.1.0.8388, built from master (`ddbb24f24`)

jakobhellermann:
Fixes #3751

### Problem

When a lambda's inferred return type contains an anonymous type and one branch returns null, the decompiler emitted an explicit cast such as `return (IEnumerable<<>f__AnonymousType0<int>>)null;`, which is invalid C#.

### Solution

Skip the cast in IsPossibleLossOfTypeInformation for null literals whenever the expected type contains an anonymous type:
null is implicitly convertible to any reference type, so no cast is needed, and the anonymous type has no nameable form to cast to anyway.

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/StatementBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/StatementBuilder.cs
@@ -412,7 +412,7 @@ namespace ICSharpCode.Decompiler.CSharp
 			if (expectedType == SpecialType.Dynamic)
 				return true;
 			if (givenType == SpecialType.NullType)
-				return true;
+				return !expectedType.ContainsAnonymousType();
 			return false;
 		}
 
--- a/ICSharpCode.Decompiler/CSharp/StatementBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/StatementBuilder.cs
@@ -382,6 +382,21 @@ namespace ICSharpCode.Decompiler.CSharp
 							.WithRR(new ConversionResolveResult(currentResultType, expr.ResolveResult, Conversion.IdentityConversion)).WithoutILInstruction();
 					}
 					return new ReturnStatement(expr).WithILInstruction(inst);
+
+					static bool IsPossibleLossOfTypeInformation(IType givenType, IType expectedType)
+					{
+						if (expectedType.ContainsAnonymousType())
+							return false;
+						if (NormalizeTypeVisitor.IgnoreNullability.EquivalentTypes(givenType, expectedType))
+							return false;
+						if (expectedType is TupleType { ElementNames.IsEmpty: false })
+							return true;
+						if (expectedType == SpecialType.Dynamic)
+							return true;
+						if (givenType == SpecialType.NullType)
+							return true;
+						return false;
+					}
 				}
 				else
 					return new ReturnStatement().WithILInstruction(inst);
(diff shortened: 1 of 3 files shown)


ICSharpCode.Decompiler/CSharp/StatementBuilder.cs
```diff
@@ -412,7 +412,7 @@ private bool IsPossibleLossOfTypeInformation(IType givenType, IType expectedType
 			if (expectedType == SpecialType.Dynamic)
 				return true;
 			if (givenType == SpecialType.NullType)
-				return true;
+				return !expectedType.ContainsAnonymousType();
```
siegfriedpammer: The net effect of ` IsPossibleLossOfTypeInformation`  returning ` true`  is adding a cast. An anonymous type can never be used in a cast expression. The ` !expectedType.ContainsAnonymousType()`  is actually a pre-condition for the whole transform, so you could extract that check to the top of the method. Also would make sense to inline ` IsPossibleLossOfTypeInformation`  as a static local function.

jakobhellermann: Done. Unless you mean 
```diff
-if (isLambdaOrExprTree && IsPossibleLossOfTypeInformation(expr.Type, currentResultType))
+if (isLambdaOrExprTree && !currentResultType.ContainsAnonymousType() && IsPossibleLossOfTypeInformation(expr.Type, currentResultType))
```
and keeping `IsPossibleLossOfTypeInformation` unchanged.

jakobhellermann:
CI seems to have failed during `Upload Test Logs`. Any idea why? or can you retry the action?

jakobhellermann:
CI seems to be green now, although `continuous-integration/appveyor/branchExpected` seems to be stuck at `Waiting for status to be reported`

christophwille:
> CI seems to be green now, although `continuous-integration/appveyor/branchExpected` seems to be stuck at `Waiting for status to be reported`

Sorry, that was a blast from the past - I enabled branch protection for master and missed that there was an old status check gating merges to master (we have moved on from AppVeyor for ages)

siegfriedpammer:

PR #3761, opened 2026-06-10
Stop the assembly list from disposing removed assemblies (release/10.1 backport)

siegfriedpammer:
Minimal backport of c5f8834e577898c445a2f6306de6dcd54085e603 from master.

`AssemblyList.Unload` / `Clear` / `ReloadAssembly` / `HotReplaceAssembly` all called `LoadedAssembly.Dispose()` on the assembly they removed, which disposes its `MetadataFile` and unmaps the underlying file. But open tabs and tree nodes can still hold that `MetadataFile`, and the list has no safe point at which to know those references are gone — so disposing risked unmapping a file out from under a live reader (use-after-dispose). Drop the removed assembly and let the GC reclaim it once nothing references it instead.

Compared to the master commit, this backport only takes the `ICSharpCode.ILSpyX/AssemblyList.cs` change: the `DockWorkspace.cs` hunk is a comment rewrite of Avalonia-only cancel-on-remove logic that does not exist in the WPF UI, and the new test relies on the `LoadedAssembly.isDisposed` field that only exists on master.

### DIFF
--- a/ICSharpCode.ILSpyX/AssemblyList.cs
+++ b/ICSharpCode.ILSpyX/AssemblyList.cs
@@ -340,7 +340,6 @@ namespace ICSharpCode.ILSpyX
 		{
 			VerifyAccess();
 			file = Path.GetFullPath(file);
-			LoadedAssembly evicted;
 			LoadedAssembly newAsm;
 			lock (lockObj)
 			{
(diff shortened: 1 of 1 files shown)


siegfriedpammer:
@christophwille 10.1.1 patch should be ready once this is merged.

PR #3780, opened 2026-06-12
Compute public-key tokens with a managed SHA-1

siegfriedpammer:
On FIPS-mode systems the platform crypto provider refuses to create SHA-1 instances (OpenSSL: `error:03000098 invalid digest`), so merely displaying a strong-named assembly's identity failed. The public-key token is a non-secret identity hash whose algorithm is fixed by ECMA-335 (last 8 bytes of the SHA-1 of the public key, reversed), so it should not depend on the host crypto policy.

## Changes

- Vendor dotnet/runtime's managed `Sha1ForNonSecretPurposes` (commit `6e51f762`) into `ICSharpCode.Decompiler/Util/`, license header intact. The only deviations from upstream are a provenance comment and `unchecked` blocks in `Finish`/`Drain`, required because this project compiles with `CheckForOverflowUnderflow` while SHA-1 relies on wrapping 32-bit arithmetic.
- Mark the vendored file `generated_code = true` in `.editorconfig` so `dotnet format` keeps it byte-identical to upstream and future syncs diff cleanly.
- Route both token sites (`MetadataExtensions.CalculatePublicKeyToken`, `AssemblyReferences.GetPublicKeyToken`) through it; `ICSharpCode.Decompiler` no longer references `System.Security.Cryptography` for metadata reading.
- Add `RotateLeft` to the existing `BitOperations` polyfill (netstandard2.0 has no `System.Numerics.BitOperations`).

`IncrementalHash` was considered and rejected: like `SHA1.Create()`, it resolves the digest through the platform provider, so it fails identically under restrictive crypto policies. Roslyn's token code (`CryptographicHashProvider.ComputeSha1`) also uses the platform SHA-1, so it offers no precedent for FIPS safety.

## Tests

New `Sha1ForNonSecretPurposesTests`: FIPS-180 vectors through both the one-shot and incremental paths, a cross-check against platform SHA-1 for all input lengths 0-130 (covers every block/padding boundary), `Start()` reuse, and the ECMA standard public key producing the well-known token `b77a5c561934e089`.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -275,3 +275,7 @@ dotnet_diagnostic.CA2242.severity = warning
 
 # MEF006: No importing constructor
 dotnet_diagnostic.MEF006.severity = silent
+
+# Files vendored from dotnet/runtime keep their upstream formatting so future syncs diff cleanly.
+[ICSharpCode.Decompiler/Util/Sha1ForNonSecretPurposes.cs]
+generated_code = true
--- a/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
+++ b/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
@@ -395,6 +395,7 @@
     <Compile Include="Util\BitSetTests.cs" />
     <Compile Include="Util\IntervalTests.cs" />
     <Compile Include="Util\LongSetTests.cs" />
+    <Compile Include="Util\Sha1ForNonSecretPurposesTests.cs" />
     <Compile Include="TestCases\Pretty\AssemblyCustomAttributes.cs" />
     <Compile Include="TestCases\Pretty\CustomAttributeSamples.cs" />
     <Compile Include="TestCases\Pretty\CustomAttributes.cs" />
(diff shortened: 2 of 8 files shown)


cscgg:
@christophwille when will this be included in a new release? I'm finding it very hard to run ILSpy on FIPS-enabled systems due to this. Thank you.

christophwille:
> @christophwille when will this be included in a new release? I'm finding it very hard to run ILSpy on FIPS-enabled systems due to this. Thank you.

You can download at any time from the master builds

cscgg:
@christophwille Thanks for the quick reply! Totally understand the master builds are available, but for my use case I'm consuming ILSpy as a library in a dotnet microservice, so pulling from CI artifacts isn't really practical... I need a proper NuGet package I can pin a version to and update deliberately. Would it be possible to get this into an official NuGet release sometime soon? Happy to help test if that's useful. Thanks.

christophwille:
So https://github.com/icsharpcode/ILSpy/pkgs/nuget/ICSharpCode.Decompiler doesn't work for you?

cscgg:
@christophwille thanks for sending out [v11.0-preview1](https://github.com/icsharpcode/ILSpy/releases/tag/v11.0-preview1), seems to be working a treat. cheers

PR #3787, opened 2026-06-16
dev: fix editorconfig parsing on some editors

mochaaP:
Some editors following the newer EditorConfig spec will not tolerate inline comments.

Link to issue(s) this covers
Fixes: 1d082e712863fa89e5dbd6afcaeabe3cd52e8b2d

### Problem
https://spec.editorconfig.org/#no-inline-comments

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,7 +2,8 @@
 root = true
 
 [*]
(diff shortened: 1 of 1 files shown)


christophwille:
Thanks!

PR #3795, opened 2026-06-20
Fix IEnumerator Decompilation When Block contains Conditional Wrappers

mrdhanz:
## Description

This PR fixes a decompilation failure where compiler-generated `IEnumerator` state machines are not folded back into clean C# `yield return` methods. This failure occurs when the helper method contains any additional control flow blocks (such as hotpatching wrappers/checks like `if (WrappersManagerImpl.IsPatched(id))`) enclosing or preceding the state machine instantiation block.

### The Bug

During the IL transformation phase, the decompiler identifies candidates for `yield return` reconstruction in `YieldReturnDecompiler.MatchEnumeratorCreationPattern`. 

Previously, this method expected the parent method's body to consist of a **single basic block**:

```csharp
Block body = SingleBlock(function.Body);
if (body == null || body.Instructions.Count == 0)
{
    return false;
}
```

If the method body contains multiple blocks (e.g., an `if` conditional wrapper or extra branches), `SingleBlock(function.Body)` returns `null`, and the yield-return decompilation pipeline bails out. As a result, the parent method is decompiled as a regular method returning a new state machine class instance, and the state machine class is exposed in the output instead of being hidden and simplified.

#### Before (Failing to decompile)
```csharp
[IteratorStateMachine(typeof(<Finish>d__82))]
private IEnumerator Finish()
{
	if (WrappersManagerImpl.IsPatched(373))
	{
		return WrappersManagerImpl.GetPatch(373).__Gen_Wrap_44(this);
	}
	return new <Finish>d__82(0)
	{
		<>4__this = this
	};
}

[CompilerGenerated]
private sealed class <Finish>d__82 : IEnumerator<object>, IDisposable, IEnumerator
{
	// Raw compiler-generated state machine logic here (MoveNext, Current, etc.)
}
```

#### After (Decompiled back to Clean `yield return`)
```csharp
[XID(373)]
private IEnumerator Finish()
{
	// Clean yield return body reconstructed from <Finish>d__82.MoveNext()
	if (_server != null)
	{
		_version = _server.ToString();
	}
	_server = null;
	_client = null;
	_buildin = null;
	if (_need_play_cg && _video != null)
	{
		_video.Play();
		yield return null;
	}
	while (_video.isPlaying)
	{
		yield return null;
	}
}
```

### The Solution

We split `MatchEnumeratorCreationPattern` into two overloads:
1. `MatchEnumeratorCreationPattern(ILFunction function)`: Checks if the body is a single block. If it has multiple blocks, it iterates through all blocks in the `BlockContainer` to find the one matching the state machine creation pattern.
2. `MatchEnumeratorCreationPattern(Block body)`: Contains the actual pattern matching logic.

Before matching each block, we reset the state machine matching fields (`isCompiledWithMono`, `isCompiledWithVisualBasic`, `isCompiledWithLegacyVisualBasic`, `stateField`, `fieldToParameterMap`, `enumeratorCtor`, `enumeratorType`) to ensure that partial failures do not pollute subsequent matching attempts.

This allows the decompiler to successfully recognize and fold the iterator state machine, discarding the compile-time/obfuscation wrapper noise and producing clean C# `yield return` code.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
+++ b/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
@@ -101,6 +101,7 @@
     <None Include="TestCases\IsolatedDecompilation\IsolatedStaticCtor.il" />
     <None Include="TestCases\ILPretty\ExtensionEncodingV2.il" />
     <None Include="testcases\ilpretty\ExtensionEncodingV1.il" />
+    <None Include="TestCases\ILPretty\IssueIEnumeratorPatchWrapper.il" />
     <None Include="TestCases\ILPretty\GuessAccessors.cs" />
     <None Include="TestCases\ILPretty\GuessAccessors.il" />
     <None Include="TestCases\ILPretty\Issue2260SwitchString.il" />
@@ -157,6 +158,7 @@
     <Compile Include="TestAssemblyResolver.cs" />
     <Compile Include="TestCases\ILPretty\ExtensionEncodingV2.cs" />
     <Compile Include="TestCases\ILPretty\ExtensionEncodingV1.cs" />
+    <None Include="TestCases\ILPretty\IssueIEnumeratorPatchWrapper.cs" />
     <Compile Include="TestCases\ILPretty\Issue3344CkFinite.cs" />
     <Compile Include="TestCases\ILPretty\Issue3421.cs" />
     <Compile Include="TestCases\ILPretty\Issue3465.cs" />
--- a/ICSharpCode.Decompiler.Tests/ILPrettyTestRunner.cs
+++ b/ICSharpCode.Decompiler.Tests/ILPrettyTestRunner.cs
@@ -76,6 +76,12 @@ namespace ICSharpCode.Decompiler.Tests
 		}
 
 		[Test]
+		public async Task IssueIEnumeratorPatchWrapper()
+		{
+			await Run();
+		}
+
+		[Test]
 		public async Task Issue982()
 		{
 			await Run();
--- a/ICSharpCode.Decompiler.Tests/TestCases/ILPretty/IssueIEnumeratorPatchWrapper.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/ILPretty/IssueIEnumeratorPatchWrapper.cs
@@ -1,0 +1,100 @@
+using System;
+using System.Collections;
+using IFix;
+using IFix.Core;
+using XUtliPoolLib;
+
+namespace IFix
+{
+	public class ILFixDynamicMethodWrapper
+	{
+		public IEnumerator __Gen_Wrap_44(object inst)
+		{
+			return null;
+		}
+	}
+
+	public class WrappersManagerImpl
+	{
+		public static bool IsPatched(int id)
+		{
+			return false;
+		}
+
+		public static ILFixDynamicMethodWrapper GetPatch(int id)
+		{
+			return null;
+		}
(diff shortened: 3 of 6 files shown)


siegfriedpammer:
Please provide test cases

mrdhanz:
> Please provide test cases

Hi @siegfriedpammer,

I have added the test cases to this branch:

1. [IssueIEnumeratorPatchWrapper.il](https://github.com/icsharpcode/ILSpy/pull/3795/changes/4d5e9d393a24180e81e4973540d0a795d9e3bb7d#diff-8cece2b4c5a2948ae3202c27dd1f8feeec2f1de2d9bab66520bdfb9dadb32ee8): Contains the compiler-generated IEnumerator state machine class (e.g. [\<Finish\>d__82](https://github.com/icsharpcode/ILSpy/pull/3795/changes/4d5e9d393a24180e81e4973540d0a795d9e3bb7d#diff-8cece2b4c5a2948ae3202c27dd1f8feeec2f1de2d9bab66520bdfb9dadb32ee8R122)) and a parent method Finish that includes a conditional wrapper check block (e.g. ```if (WrappersManagerImpl.IsPatched(373))```) before instantiating the state machine.
2. [IssueIEnumeratorPatchWrapper.cs](https://github.com/icsharpcode/ILSpy/pull/3795/changes/4d5e9d393a24180e81e4973540d0a795d9e3bb7d#diff-816fdcd66ca97700932c0b5e6cc755de82e7fa1aa00025a0ebf67554ce83de4f): The expected decompiled C# output. With the fix, the wrapper check is correctly preserved at the top of the method body, and the state machine is folded back into clean generator logic ending with yield break;.
3. [ICSharpCode.Decompiler.Tests.csproj](https://github.com/icsharpcode/ILSpy/pull/3795/changes/4d5e9d393a24180e81e4973540d0a795d9e3bb7d#diff-5c3028846493d69ec826d26c0f32a6029a8f89ff9151a75c8e7def5e5f830eadR161): Registered the expected C# output as a <None> item to prevent compiler errors during test assembly builds (since returning a value from an iterator block is not valid C# under standard specifications, but ILSpy decompiles it correctly).
4. [ILPrettyTestRunner.cs](https://github.com/icsharpcode/ILSpy/pull/3795/changes/4d5e9d393a24180e81e4973540d0a795d9e3bb7d#diff-8f85251997d0d066d021925eefef40ccd1507f6773547642fe1f8077f8e346a3R78): Registered the new test method ```IssueIEnumeratorPatchWrapper```.
The tests have been verified locally and pass successfully.

siegfriedpammer:
Thanks for the detailed write-up and repro.

We're going to decline this one. The trigger condition — an iterator factory whose body is more than a single basic block — isn't something any C# compiler emits; a real factory is always `return new <X>d__N(){ ... };`. The multi-block shape only appears when a post-compile tool (here IFix's hot-patch wrapper) rewrites the method. Handling injected/obfuscated/instrumented code shapes has historically and intentionally been out of scope for ILSpy's decompiler: we target compiler output, and deobfuscation belongs in dedicated tools (de4dot/dnSpy). Teaching the core iterator transform — and `StatementBuilder` — to admit this widens our strictest matcher to accept input no compiler produces, on the path that every method goes through, and the result isn't even legal C# (`return expr;` mixed with `yield`).

If there's appetite for reading IFix-patched assemblies, the right shape would be a separate, opt-in normalization pass that strips recognized wrappers before the standard pipeline runs, leaving the iterator matcher strict and the output legal — but that's a different change from this PR.

*(This comment was written by an agent on behalf of the maintainer.)*

PR #3796, opened 2026-06-21
Fix decompiler tests project inside VS

DoctorKrolic:
`System.Runtime.InteropServices.RuntimeInformation.RuntimeIdentifier` is available in .NET 5+ so .NET Framework hosts like VS fail to load the project. `NETCoreSdkRuntimeIdentifier` does the same job, but is an MSBuild property independent of host's .NET version


christophwille:
Thanks, goes to show how rarely we use VS these days...

PR #3806, opened 2026-06-23
Fix Export NullReferenceException for images/resourcexsd.baml

christophwille:
Trying to Export ilspy.dll 10.0.0.8345

image

```
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=ILSpy
  StackTrace:
   at ICSharpCode.ILSpy.Languages.CSharpLanguage.ResourceHandlerProjectDecompiler.WriteResourceToFile(String fileName, String resourceName, Stream entryStream) in Z:\GitWorkspace\ILSpy\ILSpy\Languages\CSharpLanguage.cs:line 565
```

Fix via VS Debugger Agent:

image

### DIFF
--- a/ILSpy/Languages/CSharpLanguage.cs
+++ b/ILSpy/Languages/CSharpLanguage.cs
@@ -561,8 +561,7 @@ namespace ICSharpCode.ILSpy.Languages
 					entryStream.Position = 0;
 					fileName = handler.WriteResourceToFile(assembly, fileName, entryStream, context);
 					var item = new ProjectItemInfo(handler.EntryType, fileName) { PartialTypes = context.PartialTypes };
-					foreach (var (k, v) in context.AdditionalProperties)
-						item.AdditionalProperties.Add(k, v);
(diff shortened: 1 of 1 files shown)


siegfriedpammer (approved):
Approved. Correct, minimal fix: `ProjectItemInfo.AdditionalProperties` defaults to null, so the old loop dereferenced null on the first `.Add()`; routing through `With(...)` lazily allocates the dictionary. No regression risk and matches the reported stack trace.

(This review was written by an AI agent on behalf of Siegfried Pammer.)

siegfriedpammer (approved):
Approved. Correct, minimal fix: `ProjectItemInfo` is a record struct whose `AdditionalProperties` defaults to null, so the old loop dereferenced null on the first `.Add()`. Routing through `With(...)` lazily allocates the dictionary via `??=`. Matches the reported stack trace; no regression risk (consumers already null-guard `AdditionalProperties`).

(This review was written by an AI agent on behalf of Siegfried Pammer.)

PR #3816, opened 2026-06-24
Allow overriding an assembly's target framework for reference resolution

siegfriedpammer:
## What

Lets the user override the target framework ILSpy resolves an assembly's **references** against, instead of the one detected from its `TargetFrameworkAttribute`. This helps when the attribute is missing, wrong, or when the user wants to force resolution against a specific framework, so references do not resolve against the wrong runtime pack or framework directory.

Resurrects a 2020 prototype (branch `tfmoverride`), re-implemented against the current `ICSharpCode.ILSpyX` / Avalonia code whose surrounding structures no longer matched.

## How

- **Core (`ICSharpCode.ILSpyX`):** `LoadedAssembly.TargetFrameworkIdOverride` short-circuits `GetTargetFrameworkIdAsync()` ahead of the cached detected value, so `LoadedAssembly`-based resolution paths read the effective TFM. The detected value remains available separately for UI/reporting. The override is persisted as a backward-compatible `TargetFramework` attribute on the `<Assembly>` element and carried across `ReloadAssembly`.
- **UI (`ILSpy`):** a "Set Target Framework..." context-menu entry on assembly nodes opens a dialog to edit the override, then reloads so references re-resolve and restores the tree selection. The dialog pairs a free-form text box with a list of common monikers; input is validated and converted from the short TFM users know (`net48`) to the long `FrameworkName` form the resolver consumes (`.NETFramework,Version=v4.8`) via NuGet. It returns `null` for cancel, an empty string to reset to auto-detection, and a non-empty string to set an override.
- **Visual distinction & reset:** an assembly carrying a manual override bolds its effective TFM in the tree label, so overridden assemblies stand out at a glance. A separate "Reset Target Framework" entry appears only while an override is in effect, clearing it back to auto-detection without reopening the dialog.
- **References listing:** the References folder prints both `Detected TargetFramework-Id` and `Effective TargetFramework-Id`, so overrides are visible without relabeling detected metadata.
- `NuGet.Frameworks` is promoted to an explicit reference (already resolved transitively at 7.6.0 via `NuGet.Protocol`); lock files regenerated.

## Screenshots

Set the override from the assembly context menu (no override yet, so only "Set Target Framework..." is offered):

Assembly context menu with the Set Target Framework entry, no override active

Enter the framework as the short moniker you know (`net48`); the presets list covers the common ones:

Set Target Framework dialog with net48 entered and highlighted in the presets list

After the reload, the effective TFM is bolded in the tree label and the References folder reports Detected vs Effective side by side:

Overridden assembly with the effective TFM bolded in the tree label and the References folder showing Detected versus Effective

While an override is active, a "Reset Target Framework" entry clears it back to auto-detection:

Assembly context menu with the Reset Target Framework entry while an override is active

## Scope

Only reference resolution is overridden. The direct `DetectTargetFrameworkId` callers in the decompiler core (project export, language version) keep reading the real attribute.

## Tests / verification

- Coverage includes override precedence, XML round-trip/backward compatibility, reload carry-over, short-to-long TFM conversion, invalid moniker rejection, and References listing headers.
- Focused local test run after the latest updates:
  - `OPENSSL_ENABLE_SHA1_SIGNATURES=1 dotnet test --project ILSpy.Tests/ILSpy.Tests.csproj --filter "FullyQualifiedName~DecompilerViewTests" --report-trx --no-restore -v:minimal`
  - Result: `Test run summary: Passed! failed: 0`
- Local build after the latest updates:
  - `OPENSSL_ENABLE_SHA1_SIGNATURES=1 pwsh ./build.ps1 -Configuration Debug --no-restore /m:1`
  - Result: `Build succeeded. 0 Warning(s) 0 Error(s)`
- Verified in the running app: the menu entry appears, the dialog opens prefilled with the detected TFM in short form, `net48` -> `.NETFramework,Version=v4.8` drives `GetTargetFrameworkIdAsync` after reload, invalid input shows an inline error and keeps the dialog open, and blank clears the override.

### DIFF
--- a/ILSpy/TreeNodes/ReferenceFolderTreeNode.cs
+++ b/ILSpy/TreeNodes/ReferenceFolderTreeNode.cs
@@ -17,6 +17,7 @@
 // DEALINGS IN THE SOFTWARE.
 
 using System.Linq;
+using System.Threading.Tasks;
 
 using Avalonia.Threading;
 
@@ -25,6 +26,7 @@ using ICSharpCode.Decompiler.Metadata;
 using ICSharpCode.ILSpy.Properties;
 
 using ICSharpCode.ILSpy.Languages;
+using ICSharpCode.ILSpyX;
 
 namespace ICSharpCode.ILSpy.TreeNodes
 {
@@ -61,9 +63,10 @@ namespace ICSharpCode.ILSpy.TreeNodes
 
 		public override void Decompile(Language language, ITextOutput output, DecompilationOptions options)
 		{
-			var targetFramework = parentAssembly.LoadedAssembly.GetTargetFrameworkIdAsync().GetAwaiter().GetResult();
-			var runtimePack = parentAssembly.LoadedAssembly.GetRuntimePackAsync().GetAwaiter().GetResult();
-			output.WriteLine($"Detected TargetFramework-Id: {targetFramework}");
+			var loadedAssembly = parentAssembly.LoadedAssembly;
+			var (detectedTargetFramework, effectiveTargetFramework, runtimePack) = GetFrameworkInfoAsync(loadedAssembly).GetAwaiter().GetResult();
+			output.WriteLine($"Detected TargetFramework-Id: {detectedTargetFramework}");
+			output.WriteLine($"Effective TargetFramework-Id: {effectiveTargetFramework}");
 			output.WriteLine($"Detected RuntimePack: {runtimePack}");
 
 			// Children realise lazily on the UI thread; we may run from a background decompile.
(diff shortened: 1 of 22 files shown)


ILSpy/TreeNodes/ReferenceFolderTreeNode.cs
```diff
@@ -61,9 +61,12 @@ protected override void LoadChildren()
 
 		public override void Decompile(Language language, ITextOutput output, DecompilationOptions options)
 		{
-			var targetFramework = parentAssembly.LoadedAssembly.GetTargetFrameworkIdAsync().GetAwaiter().GetResult();
-			var runtimePack = parentAssembly.LoadedAssembly.GetRuntimePackAsync().GetAwaiter().GetResult();
-			output.WriteLine($"Detected TargetFramework-Id: {targetFramework}");
+			var loadedAssembly = parentAssembly.LoadedAssembly;
+			var detectedTargetFramework = loadedAssembly.GetDetectedTargetFrameworkIdAsync().GetAwaiter().GetResult();
```
christophwille: Suggestion: put all awaited method call in a single internal async method and .GetAwaiter().GetResult() that single one. Should be more efficient

christophwille:
A visual distinction between auto-detected framework and manually locked-in ones would be nice.

Also, a context menu item to "Reset to auto" (only visible when manually locked-in) instead of having to go through the dialog when manually locked-in would be nice.

siegfriedpammer:
All review feedback is addressed as of 81704225e: blank/whitespace overrides normalize to null in the setter (detection is never suppressed), the TFM doc-comments use the comma form, the Debug-menu `Order` collision is gone (Set/Reset = 420/421), overridden assemblies bold their TFM fragment in the tree, and the References listing folds the awaits into one async call. Thanks!

_Written by Claude (claude-opus-4-8) via Claude Code, on Siegfried's behalf._

PR #3823, opened 2026-06-26
Fix #3821: keep ref-struct conditional as if/return, not ?. / ?? (CS8978)

### LINKED ISSUE #3821: Decompiler emits ?. / ?? over a ref-struct-returning member, producing CS8978 (cannot be made nullable)
# [Bug] Decompiler emits `?.` / `??` over a ref-struct-returning member, producing CS8978 ("cannot be made nullable")

**Component:** ICSharpCode.Decompiler
**Version:** 10.1.0.8386 (NuGet) — also reproduces in current `master`.

A conditional expression whose non-null branch yields a ref struct (e.g. `Span<T>` / `ReadOnlySpan<T>`) with a `default` fallback is decompiled using null-conditional access (`?.`) and the null-coalescing operator (`??`). A ref struct cannot be made nullable, so the decompiled code does not compile.

### Steps to reproduce

Compile this to any TFM where `Span<T>` is available (e.g. net48 + the `System.Memory` package, or net8.0):

```csharp
using System;

public class Holder
{
    public byte[] Data;
    public Span<byte> GetSpan() => Data;
}

public class C
{
    public static Span<byte> M(Holder h) => h != null ? h.GetSpan() : default;
}
```

Decompile the resulting assembly.

### Actual output

```csharp
public static Span<byte> M(Holder h)
{
    return h?.GetSpan() ?? default(Span<byte>);
}
```

Recompiling this output fails with:

```
error CS8978: 'Span<byte>' cannot be made nullable.
```

### Expected output

Something a ref struct can express, e.g.:

```csharp
public static Span<byte> M(Holder h)
{
    return (h != null) ? h.GetSpan() : default(Span<byte>);
}
```

### Root cause

`NullPropagationTransform.TryNullPropagation` (ICSharpCode.Decompiler/IL/Transforms/NullPropagationTransform.cs) takes the `NullableType.IsNonNullableValueType(returnType)` branch for `Span<byte>` — a ref struct has `IsReferenceType == false` and is not `Nullable<T>`, so `IsNonNullableValueType` returns `true` — and emits the `?. ?? fallback` shape. There is no guard for by-ref-like (ref struct) types.

### Suggested fix

Guard the value-type branch against ref structs:

```csharp
else if (!removedRewrapOrNullableCtor
         && NullableType.IsNonNullableValueType(returnType)
         && !returnType.IsByRefLike)   // ref structs cannot be made nullable
{
    ...
}
```

sailro:
Fixes #3821.

`NullPropagationTransform` rewrote `testedVar != null ? testedVar.AccessChain : default`
to `testedVar?.AccessChain ?? default` whenever the access-chain result was a
non-nullable value type. For a by-ref-like type (a ref struct such as `Span<T>`)
this does not compile: a ref struct cannot be wrapped in `Nullable<T>` (CS8978).

This excludes by-ref-like return types from the null-coalescing rewrite, leaving
the conditional as an `if`/return, which compiles.

### Test

A `Correctness/NullPropagation` case returns a ref struct from a
`holder != null ? holder.Get() : default` conditional (guarded `#if CS72`).
Without the fix the decompiled `holder?.Get() ?? default` fails to recompile
(CS8978); with it the decompiled output recompiles and behaves identically.

Verified against a build with this change by decompiling a minimal assembly:
`h != null ? h.GetSpan() : default` now decompiles to an `if`/return instead of
`h?.GetSpan() ?? default(Span<byte>)`.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Correctness/NullPropagation.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Correctness/NullPropagation.cs
@@ -7,6 +7,9 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Correctness
 		static void Main()
 		{
 			new NullPropagation().TestNotCoalescing();
+#if CS72
+			new NullPropagation().TestRefStructNotCoalescing();
+#endif
 		}
 
 		class MyClass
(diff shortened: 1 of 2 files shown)


siegfriedpammer:

PR #3828, opened 2026-06-26
Fix #3826: wrap an overflowing constant subexpression in unchecked()

### LINKED ISSUE #3826: Inlined hash accumulator becomes a constant overflow (CS0220) in decompiled GetHashCode
### Input code
The following compiles fine (default project settings, i.e. unchecked), but the decompiled output fails to recompile. The failing member is `HashRepro.GetHashCode`.

```c#

public class HashRepro
{
    public int Type { get; set; }
    public int? Position { get; set; }
    public string Text { get; set; }

    public override int GetHashCode()
    {
        var hashCode = 1688038063;
        hashCode = hashCode * -1521134295 + Type.GetHashCode();
        hashCode = hashCode * -1521134295 + EqualityComparer<int?>.Default.GetHashCode(Position);
        hashCode = hashCode * -1521134295 + EqualityComparer<string>.Default.GetHashCode(Text);
        return hashCode;
    }
}
```
This is the canonical IDE-generated `GetHashCode`. Because `hashCode` is a local, every `* -1521134295` is a runtime operation and compiles fine.

### Erroneous output
ILSpy inlines the accumulator into a single nested expression:
```c#
public override int GetHashCode()
{
    return ((1688038063 * -1521134295 + Type.GetHashCode()) * -1521134295
        + EqualityComparer<int?>.Default.GetHashCode(Position)) * -1521134295
        + EqualityComparer<string>.Default.GetHashCode(Text);
}
```
This does not compile:
```
error CS0220: The operation overflows at compile time in checked mode
```
Now `1688038063 * -1521134295` is a compile-time **constant** subexpression. In C#, constant expressions are always evaluated in a checked context regardless of the project's overflow setting, so the overflow is a compile error. Wrapping the arithmetic in `unchecked(...)` (the context the original had implicitly via the runtime accumulator) would fix it.

### Details
* Real-world occurrence: DiffPlex `DiffPiece.GetHashCode` (net45).
* Root cause: in `ExpressionBuilder`, an unchecked binary arithmetic op is annotated with `AddCheckedBlocks.UncheckedAnnotation` (not `ExplicitUncheckedAnnotation`). `AddCheckedBlocks` only emits an explicit `unchecked(...)` wrapper for the explicit annotation. When inlining turns the operands into compile-time constants whose product overflows, no wrapper is emitted and the result is CS0220. The explicit annotation is currently used only for native-int casts (`TranslatedExpression`); an analogous case is needed for a binary op whose constant result overflows.

sailro:
Fixes #3826.

A common hand-written `GetHashCode` uses a local accumulator:
```c#
var hashCode = 1688038063;
hashCode = hashCode * -1521134295 + Type.GetHashCode();
...
```
Because `hashCode` is a local, every `* -1521134295` is a runtime operation and
compiles fine. The decompiler inlines the accumulator into one nested expression,
so the leading `1688038063 * -1521134295` becomes a compile-time **constant**
subexpression. In C#, constant subexpressions are always evaluated in a checked
context regardless of the project's overflow setting, so the decompiled output
failed to compile with **CS0220** ("operation overflows at compile time in checked
mode").

### Fix
`HandleBinaryNumeric` now detects an unchecked binary operation whose result is a
compile-time constant that overflows in a checked context, and annotates it with
`ExplicitUncheckedAnnotation` so `AddCheckedBlocks` emits an explicit
`unchecked(...)` wrapper. This mirrors the existing handling of an overflowing
constant cast to `nint`/`nuint` (`TranslatedExpression.cs`).

Before:
```c#
return (1688038063 * -1521134295 + a) * -1521134295 + b;   // CS0220
```
After:
```c#
return (unchecked(1688038063 * -1521134295) + a) * -1521134295 + b;
```
Only the overflowing constant subexpression is wrapped; non-constant
multiplications are left as-is.

### Test
`Correctness/UncheckedConstants` exercises the inlined-accumulator hash. Without
the fix the decompiled output fails to recompile (CS0220); with it the output
recompiles and produces identical results.

Real-world occurrence: DiffPlex `DiffPiece.GetHashCode` (net45).

### DIFF
--- a/ICSharpCode.Decompiler.Tests/CorrectnessTestRunner.cs
+++ b/ICSharpCode.Decompiler.Tests/CorrectnessTestRunner.cs
@@ -359,6 +359,12 @@ namespace ICSharpCode.Decompiler.Tests
 		}
 
 		[Test]
+		public async Task UncheckedConstants([ValueSource(nameof(defaultOptions))] CompilerOptions options)
+		{
+			await RunCS(options: options);
+		}
+
+		[Test]
 		public async Task ConditionalAttr([ValueSource(nameof(defaultOptions))] CompilerOptions options)
 		{
 			await RunCS(options: options);
--- a/ICSharpCode.Decompiler.Tests/TestCases/Correctness/UncheckedConstants.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Correctness/UncheckedConstants.cs
@@ -1,0 +1,45 @@
+// Copyright (c) AlphaSierraPapa for the SharpDevelop Team
+// 
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this
+// software and associated documentation files (the "Software"), to deal in the Software
+// without restriction, including without limitation the rights to use, copy, modify, merge,
+// publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons
+// to whom the Software is furnished to do so, subject to the following conditions:
+// 
+// The above copyright notice and this permission notice shall be included in all copies or
+// substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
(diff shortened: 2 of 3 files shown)


dgrunwald (approved):

siegfriedpammer:

PR #3829, opened 2026-06-26
Fix #3824: fold a hoisted argument null-guard into the chained ctor call

### LINKED ISSUE #3824: Chained constructor call emitted as illegal in-body this..ctor(...) when a null-check is hoisted before it
### Input code
The following compiles fine (Release), but the chained constructor call is not reconstructed when decompiled. The failing member is the `CtorChild(string)` constructor.

```c#
using System;

public class CtorBase
{
    public CtorBase(int a, int b, int c, string s) { }
}

public sealed class CtorChild : CtorBase
{
    public string Value;

    public CtorChild(string value)
        : base(0, value?.Length ?? throw new ArgumentNullException(nameof(value)), value.Length, nameof(value))
    {
        Value = value;
    }
}
```

### Erroneous output
```c#
public CtorChild(string value)
{
    if (value == null)
    {
        throw new ArgumentNullException("value");
    }
    base..ctor(0, value.Length, value.Length, "value");
    Value = value;
}
```
`base..ctor(...)` is not valid C# (a parse error: `base.` followed by `.ctor`), so the output does not compile. The same happens with a `: this(...)` chain (emitted as `this..ctor(...)`).

Expected output:
```c#
public CtorChild(string value)
    : base(0, value?.Length ?? throw new ArgumentNullException("value"), value.Length, "value")
{
    Value = value;
}
```

### Details
* Real-world occurrence: IbanNet `IbanNet.Registry.Patterns.PatternToken(string)` (net462).
* Root cause: `TransformFieldAndConstructorInitializers.MoveConstructorInitializer` inspects only `Body.Statements.FirstOrDefault()`. When the compiler hoists an argument null-check (here from `value?.Length ?? throw ...` with `value` reused in a later argument) as `if (value == null) throw ...;`, that guard becomes the first statement, the `ThisCallClassPattern`/`ThisCallStructPattern` match fails, and the chained call stays in the body. Folding the guard back into the initializer argument (`value ?? throw ...`) would restore a valid initializer.
* Related but not duplicates: #3610, #3621, #950 (same symptom, different triggers).

sailro:
Fixes #3824.

When a constructor initializer argument contains argument-validation that throws
(e.g. `arg ?? throw new ArgumentNullException(...)`) and the argument is used by
more than one parameter, the compiler hoists the null-check in front of the
chained constructor call. `MoveConstructorInitializer` only inspected the first
body statement, so the hoisted guard defeated it and the chained call was left as
an illegal in-body `this..ctor(...)` / `base..ctor(...)` statement (a parse error).

### Fix
When locating the chained call, skip leading argument null-guards
(`if (param == null) throw ...;`) and fold each one back into the first argument
that uses the parameter, as `param ?? throw ...`. If a guard cannot be folded (its
parameter is not used by any argument), the constructor is left unchanged.

Before:
```c#
public PatternToken(string value)
{
    if (value == null)
    {
        throw new ArgumentNullException("value");
    }
    this..ctor(AsciiCategory.None, value.Length, value.Length, "value"); // parse error
    Value = value;
}
```
After:
```c#
public PatternToken(string value)
    : this(AsciiCategory.None, (value ?? throw new ArgumentNullException("value")).Length, value.Length, "value")
{
    Value = value;
}
```

### Test
`Pretty/ConstructorInitializers` gains a `#if CS70` case where the null-check is
hoisted because the parameter is used by more than one argument; it round-trips to
the folded initializer form.

Real-world occurrence: IbanNet `PatternToken(string)` (net462).

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/ConstructorInitializers.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/ConstructorInitializers.cs
@@ -232,6 +232,29 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 			}
 		}
 
+#if CS70
+		public class NullCheckedArgumentBase
+		{
+			public NullCheckedArgumentBase(int a, int b, int c, string s)
+			{
+			}
+		}
+
+		public class NullCheckedArgumentChain : NullCheckedArgumentBase
+		{
+			public string Value;
+
+			// The null-check for 'value' is hoisted in front of the chained constructor call because
+			// 'value' is used by more than one argument; it must be folded back into the first
+			// argument as 'value ?? throw ...' rather than emitted as an illegal in-body call.
+			public NullCheckedArgumentChain(string value)
+				: base(0, (value ?? throw new ArgumentNullException("value")).Length, value.Length, "value")
+			{
+				Value = value;
+			}
+		}
+#endif
+
 #if CS100
 		public class PrimaryCtorClassThisChain(Guid id)
 		{
--- a/ICSharpCode.Decompiler/CSharp/Transforms/TransformFieldAndConstructorInitializers.cs
+++ b/ICSharpCode.Decompiler/CSharp/Transforms/TransformFieldAndConstructorInitializers.cs
@@ -462,6 +462,25 @@ namespace ICSharpCode.Decompiler.CSharp.Transforms
 					return true;
 				}
 
+				// The chained constructor call is normally the first statement. However, when an
+				// initializer argument contains a null-check that throws (e.g. `arg ?? throw ...`) and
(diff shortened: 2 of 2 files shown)


siegfriedpammer:
Thanks for bringing attention to this issue and for taking the time to prepare a fix — much appreciated!

After digging into it, we concluded that this needs a more involved implementation at the ILAst level, as part of the null-coalescing transform. The main advantage of doing it there is the existing transform infrastructure and the precise semantic model: variables know all of their uses and the control flow is explicit, so the transform can reliably verify it only fires on the exact compiler-generated pattern. That information is reachable from the C# AST too, but only through an extra level of indirection, which makes such checks considerably more awkward there.

I'm therefore closing this PR as superseded by #3852, which implements the fix that way. Thanks again for the contribution!

Issue #1520, opened 2019-05-19
Ref local to struct not initialized as its real scope


sgkoishi:
[ILSpy 5.0 Preview 2](https://github.com/icsharpcode/ILSpy/releases/tag/v5.0-preview2) yields the following code for an assembly
```c#
ref float x; // error CS8174: A declaration of a by-reference variable must have an initializer
if (flag)
{
	x = ref value1;
	x += 1f;
}
else
{
	x = ref value2;
	x += 1f;
}
```
The ref local variable x was not declared in its corresponding scope, and cause the compile error CS8174.  
It happens when it's referencing field in a struct.

sgkoishi:
Cannot reproduce (7f2b3a)

sgkoishi:
[ScrPad.zip](https://github.com/icsharpcode/ILSpy/files/3268935/ScrPad.zip)
This binary reproduces the problem (with a little modification).

ILSpy's result:
```c#
	public static void Main(string[] args)
	{
		ref float field; // Compile error CS8174
		if (args.Length != 0)
		{
			field = (ref Struct.Field1);
			field += 1f;
			return;
		}
		field = (ref Struct.Field1);
		field += 1f;
		if ((double)Struct.Field1 > 1.5)
		{
			field = (ref Struct.Field1);
			field += 1f;
		}
	}
```
dnSpy's result:
```c#
	public static void Main(string[] args)
	{
		bool flag = args.Length != 0;
		if (flag)
		{
			ref float ptr = ref Program.Struct.Field1;
			ptr += 1f;
		}
		else
		{
			ref float ptr = ref Program.Struct.Field1;
			ptr += 1f;
			bool flag2 = (double)Program.Struct.Field1 > 1.5;
			if (flag2)
			{
				ptr = ref Program.Struct.Field1;
				ptr += 1f;
			}
		}
	}
```

dgrunwald:
This is a tricky case.
The control flow decompilation is independent of the local variables. It prefers using `if () { return; }` over `if-else` to keep the indentation level small.
But without the `else` block, it's impossible to declare the variable in a way that is legal C#.

sgkoishi:
> 
> 
> This is a tricky case.
> The control flow decompilation is independent of the local variables. It prefers using `if () { return; }` over `if-else` to keep the indentation level small.
> But without the `else` block, it's impossible to declare the variable in a way that is legal C#.

[...]

PR #3830, opened 2026-06-26
Fix #3827: keep the while-loop for a ref local used after the loop (avoid an uninitialized hoisted ref decl)

### LINKED ISSUE #1630: [ref local]CS8174: A declaration of a by-reference variable must have an initializer
Tested with 5.0.0.5059-preview4.

See that in Dictionary<TKey, TValue>.Remove(TKey) in netcoreapp2.2(3.0 use some different code, and not translated to `for` loop)

```cs
    public struct XXX
    {
        object data;
        int next;
        public static void Test()
        {
            XXX[] datas = new XXX[1];
            int idx = 0;
            while (idx >= 0)
            {
                ref XXX reference = ref datas[idx];
                Console.WriteLine(reference.data);
                idx = reference.next;
            }
        }
    }
```

Decompiled: (be converted to `for` loop but break the declaration of ref local)
```cs
        public static void Test1()
        {
            XXX[] datas = new XXX[1];
            ref XXX reference;//error CS8174: A declaration of a by-reference variable must have an initializer
            for (int idx = 0; idx >= 0; idx = reference.next)
            {
                reference = ref datas[idx];
                Console.WriteLine(reference.data);
            }
        }
```

### LINKED ISSUE #3827: Ref local used after a loop is emitted as an uninitialized 'ref T x;' (CS8174)
### Input code
The following compiles fine, but the decompiled output fails to recompile. The failing member is `Repro.Find`. The ref local is initialized before the loop, written through inside it, and used after it.

```c#
public sealed class Node
{
    public Node Next;
    public int Value;
}

public static class Repro
{
    private static readonly Node[] _buckets = new Node[4];

    public static Node Find(int h, Node newNode)
    {
        ref Node reference = ref _buckets[h & 3];
        while (reference != null)
        {
            if (reference.Value == h)
            {
                reference = newNode;
                break;
            }
            reference = ref reference.Next;
        }
        if (reference == null)
        {
            reference = newNode;
        }
        return reference;
    }
}
```

### Erroneous output
```c#
public static Node Find(int h, Node newNode)
{
    ref Node reference;
    for (reference = ref _buckets[h & 3]; reference != null; reference = ref reference.Next)
    {
        if (reference.Value == h)
        {
            reference = newNode;
            break;
        }
    }
    if (reference == null)
    {
        reference = newNode;
    }
    return reference;
}
```
This does not compile:
```
error CS8174: A declaration of a by-reference variable must have an initializer
```
The declaration `ref Node reference;` is split from its initializing ref-assignment (moved into the `for` initializer), but a ref local may not be declared without an initializer. The fix is to not form the `for`-loop in the first place: when a by-ref-like iteration variable is used after the loop, keep the `while`-loop so the initializer stays on the hoisted declaration (matching the original source):
```c#
ref Node reference = ref _buckets[h & 3];
while (reference != null)
{
    ...
    reference = ref reference.Next;
}
```

### Details
* Real-world occurrence: Enums.NET (net461) hash-bucket walk in `Enums.cs`.
* Root cause: in `DeclareVariables`, when a by-ref-like local is used in the for-iterator / after the loop, the insertion point resolves to the `ForStatement`, so `IsMatchingAssignment` returns `false` (a `ForStatement` is not an `AssignmentExpression`) and the code emits a `VariableDeclarationStatement` with a `null` initializer.
* Prior report: #1630 ("[ref local] CS8174 ...") reported the same symptom in 2019 but was closed by the stale-bot without a fix; #1520 is related. This issue provides a fresh minimal repro and root-cause analysis.

sailro:
Fixes #3827.

A `ref` local used after a `while`-loop has its declaration hoisted in front of the
loop. `PatternStatementTransform` then rewrote the `while` into a `for`, moving the
sole ref-assignment into the for-initializer — leaving `ref T x;` (no initializer)
in front of a headless `for (; cond; x = ref …)`, which does not compile
(CS8174: a by-reference variable must have an initializer).

### Fix
Per @siegfriedpammer's review, instead of emitting a headless for-loop,
`PatternStatementTransform.TransformFor` now leaves the loop as a `while` when the
iteration variable is a by-ref-like local used after the loop. The ref-assignment
stays on the hoisted declaration and the output matches the original source — both
simpler and free of the not-so-nice headless-for form.

**Before** (does not recompile — CS8174):
```c#
ref Node reference;
for (reference = ref _buckets[hash & 3]; reference != null; reference = ref reference.next)
{
    ...
}
```
**After:**
```c#
ref Node reference = ref _buckets[hash & 3];
while (reference != null)
{
    ...
    reference = ref reference.next;
}
```

### Test
`Pretty/RefLocalsAndReturns.RefLocalUsedAfterForLoop` round-trips the hoisted-decl +
while-loop form.

Real-world occurrence: Enums.NET (net461) hash-bucket walk. Same symptom as the
long-closed #1630 (stale-bot closed, no fix); the if/else ref-assignment variant
from #1520 is not addressed here.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/RefLocalsAndReturns.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/RefLocalsAndReturns.cs
@@ -168,17 +168,18 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 
 		public static Issue3827Node RefLocalUsedAfterForLoop(Issue3827Node[] buckets, int hash, Issue3827Node newNode)
 		{
-			// The ref local is used after the loop, so its declaration is hoisted in front of the
-			// for-statement; the for-initializer ref-assignment must stay on the declaration, because
-			// a ref local cannot be declared without an initializer.
+			// The ref local is used after the loop, so its declaration is hoisted in front of the loop.
+			// The loop is kept as a while-loop (rather than converted to a headless for-loop) so the
+			// ref-assignment stays on the declaration -- a ref local cannot be declared without an initializer.
 			ref Issue3827Node reference = ref buckets[hash & 3];
-			for (; reference != null; reference = ref reference.next)
+			while (reference != null)
 			{
 				if (reference.value == hash)
 				{
 					reference = newNode;
 					break;
 				}
+				reference = ref reference.next;
 			}
 			if (reference == null)
 			{
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/RefLocalsAndReturns.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/RefLocalsAndReturns.cs
@@ -45,6 +45,13 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 			}
 		}
 
+		public class Issue3827Node
+		{
+			public Issue3827Node next;
+
+			public int value;
+		}
+
 		public delegate ref T RefFunc<T>();
 		public delegate ref readonly T ReadOnlyRefFunc<T>();
 		public delegate ref TReturn RefFunc<T1, TReturn>(T1 param1);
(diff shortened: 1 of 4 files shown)


ICSharpCode.Decompiler.Tests/TestCases/Pretty/RefLocalsAndReturns.cs
```diff
@@ -159,6 +166,27 @@ public void RefReadonlyCallVirt(RefLocalsAndReturns provider)
 
 		private static int DefaultInt = 0;
 
+		public static Issue3827Node RefLocalUsedAfterForLoop(Issue3827Node[] buckets, int hash, Issue3827Node newNode)
+		{
+			// The ref local is used after the loop, so its declaration is hoisted in front of the
+			// for-statement; the for-initializer ref-assignment must stay on the declaration, because
+			// a ref local cannot be declared without an initializer.
+			ref Issue3827Node reference = ref buckets[hash & 3];
+			for (; reference != null; reference = ref reference.next)
```
siegfriedpammer: In this specific case I think it would make sense to also adapt the HighLevelLoopTransform or PatternStatementTransform (depending on what actually created the for-loop) to avoid for loops in cases where `ref` locals are used after the for loop container. What do you think? It's mainly stylistic preference, but headless for-loops don't look so nice imho.

sailro: Fully agree, my first pass was not optimal indeed. Implemented your suggestion in `PatternStatementTransform.TransformFor` and it's a lot simpler in the end.

siegfriedpammer (approved):

siegfriedpammer:

PR #3831, opened 2026-06-26
Fix #3825: reconstruct async iterators with [EnumeratorCancellation] and await in finally

### LINKED ISSUE #3825: async IAsyncEnumerable with [EnumeratorCancellation] and await in finally decompiles to a raw state machine
### Input code
The following async iterator compiles fine, but is not reconstructed when decompiled. The failing member is `ListAsync`.

```c#
using System.Threading;
using System.Threading.Tasks;

public static class Repro
{
    public static async IAsyncEnumerable<int> ListAsync([EnumeratorCancellation] CancellationToken cancellationToken)
    {
        int handle = await OpenAsync(cancellationToken).ConfigureAwait(false);
        try
        {
            yield return handle;
            await WorkAsync(cancellationToken).ConfigureAwait(false);
        }
        finally
        {
            await CloseAsync(handle, cancellationToken).ConfigureAwait(false);
        }
    }

    private static async Task<int> OpenAsync(CancellationToken ct) { await Task.Yield(); return 1; }
    private static async Task WorkAsync(CancellationToken ct) { await Task.Yield(); }
    private static async Task CloseAsync(int h, CancellationToken ct) { await Task.Yield(); }
}
```
Note: **removing `[EnumeratorCancellation]` makes the method decompile cleanly** (the `try`/`finally` is reconstructed correctly), which isolates the trigger to the combination of `[EnumeratorCancellation]` + an `await` in the `finally`.

### Erroneous output
ILSpy emits the raw compiler-generated state machine instead of reconstructing the iterator. The output references compiler-generated identifiers (`<>1__state`, `<handle>5__3`, `<ListAsync>d__0`) that don't parse, and contains:
```c#
catch (object obj)
{
    <>7__wrap3 = obj;
}
...
object obj2 = <>7__wrap3;
if (obj2 != null)
{
    ExceptionDispatchInfo.Capture((obj2 as Exception) ?? throw obj2).Throw();
}
```
The output does not compile: `catch (object obj)` is CS0155 (a caught type must derive from `System.Exception`), `throw obj` / `?? throw obj` is CS0155/CS0266, and the `<...>` names are invalid identifiers.

### Details
* Real-world occurrence: Renci.SshNet `SftpClient.ListDirectoryAsync` (net462).
* Root cause: `AwaitInFinallyTransform.MatchObjectStoreCatchHandler` only matches a catch entry point of exactly 2 or 3 instructions and returns `false` otherwise. With `[EnumeratorCancellation]` the catch body carries an extra instruction (e.g. a `stloc <>1__state` reset), the match fails, `AwaitInFinallyTransform` bails, the `TryCatch` survives, and `StatementBuilder.VisitTryCatch` emits `catch (object obj)` because the exception variable is loaded later (to rethrow).
* Not a duplicate of #3745 (that is the new runtime-async `RuntimeAsyncExceptionRewriteTransform`).

sailro:
Fixes #3825.

An `async IAsyncEnumerable<T>` iterator that has both an `[EnumeratorCancellation]`
cancellation-token parameter and an `await` inside a `finally` was not
reconstructed: the decompiler emitted the raw compiler-generated state machine
(`catch (object)`, `goto case`, `<>1__state`, ...), which does not compile.

### Root cause
For such an iterator, the hoisted-local cleanup (`stfld <>u__N(this, null)`) is
emitted before the combined `CancellationTokenSource` disposal in both the
set-result block and the catch block. `CheckSetResultReturnBlock` and
`ValidateCatchBlock` only consumed that cleanup after the disposal, so the
`pos + 2 == count` test missed the dispose pattern, the symbolic analysis failed,
and `AsyncAwaitDecompiler` left the raw state machine.

### Fix
Allow the hoisted-local cleanup to appear before the combined-tokens disposal as
well (one extra `MatchHoistedLocalCleanup` call in each of the two methods).

### Test
`Pretty/AsyncStreams.AwaitInFinallyWithCancellation` combines
`[EnumeratorCancellation]` with an `await` in a `finally`; it now round-trips.

Real-world occurrence: Renci.SshNet `SftpClient.ListDirectoryAsync` (net462).

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/AsyncStreams.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/AsyncStreams.cs
@@ -71,6 +71,20 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 				await Task.Delay(10).ConfigureAwait(continueOnCapturedContext: false);
 			}
 		}
+
+		public static async IAsyncEnumerable<int> AwaitInFinallyWithCancellation([EnumeratorCancellation] CancellationToken cancellationToken)
+		{
+			await Task.Yield();
+			try
+			{
+				yield return 1;
+				await Task.Delay(10, cancellationToken);
+			}
+			finally
+			{
+				await Task.Yield();
+			}
+		}
 	}
 
 	public struct TestStruct
--- a/ICSharpCode.Decompiler/IL/ControlFlow/AsyncAwaitDecompiler.cs
+++ b/ICSharpCode.Decompiler/IL/ControlFlow/AsyncAwaitDecompiler.cs
@@ -911,6 +911,11 @@ namespace ICSharpCode.Decompiler.IL.ControlFlow
 			finalStateKnown = true;
 			pos++;
 
+			// [optional] stfld <>u__N(ldloc this, ldnull)
+			// The hoisted-local cleanup may appear either before or after the combined-tokens disposal
+			// (the latter is present for async iterators with an [EnumeratorCancellation] token).
+			MatchHoistedLocalCleanup(block, ref pos);
+
 			if (pos + 2 == block.Instructions.Count && block.MatchIfAtEndOfBlock(out var condition, out var trueInst, out var falseInst))
 			{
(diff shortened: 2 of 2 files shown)


dgrunwald (approved):

siegfriedpammer:

PR #3832, opened 2026-06-27
Fix #3754: omit async stepping info for runtime-async methods

### LINKED ISSUE #3754: NullReferenceException generating portable PDB for runtime-async methods (.NET 11)
### Summary

Generating a portable PDB for an assembly that contains **runtime-async** methods (the new .NET 11 / "C# 15" runtime-async lowering) throws a `NullReferenceException` instead of producing a PDB. This reproduces against any .NET 11 preview shared-framework assembly that uses runtime async, e.g. `System.Net.Requests.dll`.

### Repro

```
ilspycmd -genpdb \
  "/usr/lib64/dotnet/shared/Microsoft.NETCore.App/11.0.0-preview.4.26230.115/System.Net.Requests.dll" \
  -o /tmp/pdbtest
```

(Same failure via the GUI: select the assembly -> "Generate portable PDB".)

### Stack trace

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Collections.Immutable.ImmutableArray`1.ThrowNullRefIfNotInitialized()
   at ICSharpCode.Decompiler.DebugInfo.AsyncDebugInfo.BuildBlob(MethodDefinitionHandle moveNext) in AsyncDebugInfo.cs:line 35
   at ICSharpCode.Decompiler.DebugInfo.PortablePdbWriter.WritePdb(PEFile file, CSharpDecompiler decompiler, DecompilerSettings settings, Stream targetStream) in PortablePdbWriter.cs:line 196
   at ICSharpCode.ILSpyCmd.ILSpyCmdProgram.GeneratePdbForAssembly(...) in IlspyCmdProgram.cs:line 616
```

### Root cause

`PortablePdbWriter.WritePdb` emits `MethodSteppingInformation` for every function where `ILFunction.IsAsync` is true:

```csharp
// PortablePdbWriter.cs
if (function.IsAsync)
{
    customMethodDebugInfo.Add((methodHandle,
        metadata.GetOrAddGuid(KnownGuids.MethodSteppingInformation),
        metadata.GetOrAddBlob(function.AsyncDebugInfo.BuildBlob(methodHandle))));
}
```

`ILFunction.IsAsync` is `AsyncReturnType != null`, i.e. it is set early from the method signature. But `ILFunction.AsyncDebugInfo` is only assigned at the very end of `AsyncAwaitDecompiler.Run` (the line that builds `awaitDebugInfos.ToImmutableArray()`). Several early-return paths in `Run` leave `AsyncDebugInfo` at its `default` value, whose `Awaits` is an **uninitialized** `ImmutableArray<Await>`:

- the **runtime-async** path: when the method is not a classic state-machine (`!MatchTaskCreationPattern && !MatchAsyncEnumeratorCreationPattern`) and `RuntimeAsync` is enabled, `Run` rewrites via the runtime-async transforms and returns early without setting `AsyncDebugInfo`. `RuntimeAsync` defaults to `true`, so this is the path .NET 11 runtime-async methods take.
- `catch (SymbolicAnalysisFailedException) { return; }`
- `AsyncAwait` setting disabled.

`AsyncDebugInfo.BuildBlob` then iterates that default array:

```csharp
foreach (var await in Awaits)   // NRE: Awaits is default(ImmutableArray<Await>)
```

### Suggested fix

Make the writer/`BuildBlob` robust to an uninitialized `Awaits` (e.g. guard with `Awaits.IsDefaultOrEmpty`, or only emit `MethodSteppingInformation` when `AsyncDebugInfo` was actually populated). Runtime-async methods have no yield/resume await offsets to record, so for them the stepping-information blob should simply be omitted rather than built from a default struct.

### Environment

- ILSpy `avalonia` branch (cross-platform), reproduced via `ilspycmd`.
- .NET SDK 11.0.100-preview.4; target assembly from `Microsoft.NETCore.App` 11.0.0-preview.4.26230.115.

siegfriedpammer:
Fixes #3754.

## Problem

Generating a portable PDB for an assembly containing .NET 11 **runtime-async** methods threw a `NullReferenceException` instead of producing a PDB:

```
System.NullReferenceException
   at ICSharpCode.Decompiler.DebugInfo.AsyncDebugInfo.BuildBlob(MethodDefinitionHandle moveNext)
   at ICSharpCode.Decompiler.DebugInfo.PortablePdbWriter.WritePdb(...)
```

## Root cause

`ILFunction.IsAsync` is derived from the method signature, so runtime-async methods (`MethodImplAsync` bit, no MoveNext state machine) report `IsAsync` without `AsyncAwaitDecompiler` ever populating `AsyncDebugInfo`. Its `Awaits` then stays an uninitialized `ImmutableArray<Await>`, and `PortablePdbWriter` threw building the `MethodSteppingInformation` blob from that default struct.

## Fix

Guard emission with `!function.AsyncDebugInfo.Awaits.IsDefault`. Runtime-async methods have no yield/resume offsets to record, so the stepping blob is omitted, matching the C# compiler (which emits no stepping information for them either). A genuinely zero-await classic state machine keeps an initialized empty `Awaits` and is unaffected (`IsDefault`, not `IsDefaultOrEmpty`).

## Test

Adds a `PdbGen/RuntimeAsync.cs` fixture and a `PdbGenerationTestRunner.RuntimeAsync` test that compiles a runtime-async method in-proc (new `runtimeAsync` flag on `Tester.CompileCSharpWithPdb`, mirroring the out-of-process `/features:runtime-async=on`), runs `PortablePdbWriter`, and asserts it does not throw and emits no `MethodSteppingInformation`. The test fails with the NRE before the fix and passes after; the full PdbGen fixture stays green.

Verified end-to-end with `ilspycmd -genpdb` on a minimal runtime-async assembly (valid PDB, no stepping info) and on the issue's `System.Net.Requests.dll` (past the NRE).

### DIFF
--- a/ICSharpCode.Decompiler.Tests/Helpers/Tester.cs
+++ b/ICSharpCode.Decompiler.Tests/Helpers/Tester.cs
@@ -847,9 +847,15 @@ namespace System.Runtime.CompilerServices
 			}
 		}
 
-		public static void CompileCSharpWithPdb(string assemblyName, Dictionary<string, string> sourceFiles)
+		public static void CompileCSharpWithPdb(string assemblyName, Dictionary<string, string> sourceFiles, CompilerOptions compilerOptions = CompilerOptions.None)
 		{
 			var parseOptions = new CSharpParseOptions(languageVersion: Microsoft.CodeAnalysis.CSharp.LanguageVersion.Latest);
+			if (compilerOptions.HasFlag(CompilerOptions.EnableRuntimeAsync))
+			{
+				// Mirror the out-of-process csc '/features:runtime-async=on' flag so the compiler lowers
+				// async methods to the .NET 11 runtime-async form (MethodImplAsync bit, no state machine).
+				parseOptions = parseOptions.WithFeatures(new[] { new KeyValuePair<string, string>("runtime-async", "on") });
+			}
 
 			List<EmbeddedText> embeddedTexts = new List<EmbeddedText>();
 			List<SyntaxTree> syntaxTrees = new List<SyntaxTree>();
--- a/ICSharpCode.Decompiler.Tests/TestCases/PdbGen/RuntimeAsync.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/PdbGen/RuntimeAsync.cs
@@ -1,0 +1,31 @@
+// Copyright (c) 2026 Siegfried Pammer
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this
+// software and associated documentation files (the "Software"), to deal in the Software
+// without restriction, including without limitation the rights to use, copy, modify, merge,
+// publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons
+// to whom the Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all copies or
+// substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+// INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+// PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+// FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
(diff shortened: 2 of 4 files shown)


ICSharpCode.Decompiler.Tests/Helpers/Tester.cs
```diff
@@ -847,9 +847,15 @@ internal static DecompilerSettings GetSettings(CompilerOptions cscOptions)
 			}
 		}
 
-		public static void CompileCSharpWithPdb(string assemblyName, Dictionary<string, string> sourceFiles)
+		public static void CompileCSharpWithPdb(string assemblyName, Dictionary<string, string> sourceFiles, bool runtimeAsync = false)
```
siegfriedpammer: maybe we should use CompilerOptions here instead of a bool flag

ICSharpCode.Decompiler.Tests/TestCases/PdbGen/RuntimeAsync.cs
```diff
@@ -0,0 +1,13 @@
+using System.Threading.Tasks;
```
siegfriedpammer, line 19: file header

PR #3837, opened 2026-06-27
Fix #3804: decompile foreach over inline array

### LINKED ISSUE #3804: Decompiling foreach over an inline array emits invalid '<PrivateImplementationDetails>.InlineArrayElementRef'
### Summary

Decompiling a `foreach` over an [inline array](https://learn.microsoft.com/dotnet/csharp/language-reference/proposals/csharp-12.0/inline-arrays) value produces code that does not compile: it references the compiler-internal `<PrivateImplementationDetails>.InlineArrayElementRef` helper, whose type name contains angle brackets and cannot be written in C#.

### Steps to reproduce

Compile this with an optimizing compiler:

```csharp

[InlineArray(4)]
public struct Buf4 { private int _element0; }

public static class C
{
    public static int Sum(Buf4 b)
    {
        int s = 0;
        foreach (int x in b) s += x;
        return s;
    }
}
```

Decompile type `C`.

### Actual output

```csharp
public static int Sum(Buf4 b)
{
    int num = 0;
    for (int i = 0; i < 4; i++)
    {
        int num2 = global::<PrivateImplementationDetails>.InlineArrayElementRef<Buf4, int>(ref b, i);
        num += num2;
    }
    return num;
}
```

This does not compile (`error CS1001: Identifier expected` at `global::<PrivateImplementationDetails>`). `<PrivateImplementationDetails>` is a compiler-generated type and `InlineArrayElementRef<TBuffer, TElement>` is a helper the compiler emits there to lower `foreach` over an inline-array value; the decompiler does not recognize the pattern and emits the raw call.

### Expected output

Valid C# equivalent to the source, e.g. reconstructing the loop via the inline-array indexer:

```csharp
public static int Sum(Buf4 b)
{
    int num = 0;
    for (int i = 0; i < 4; i++)
    {
        num += b[i];
    }
    return num;
}
```

(or the original `foreach`).

### Scope

- Specific to **`foreach` over an inline array**. Other inline-array patterns decompile correctly: direct indexing (`b[0]`), `for` with the indexer, index-from-end (`b[^1]`), range/slice (`b[1..4]`), implicit `Span<T>` conversion, ref-to-element, passing as a span, and inline-array fields.
- Not currently covered by `InlineArrayTests`.

### Version

`ICSharpCode.Decompiler: 11.0.0` (current master).

---
_This issue was filed by an AI agent (Claude, claude-opus-4-8, via Claude Code) on behalf of the repo owner._

siegfriedpammer:
Fixes #3804.

## Problem

Decompiling a `foreach` over an inline-array value emitted a `for` loop whose body called the compiler-internal `<PrivateImplementationDetails>.InlineArrayElementRef(ref b, i)` helper. That type name contains angle brackets and cannot be written in C#, so the output did not compile.

## Approach

The earlier version of this PR relaxed the IL matcher to rewrite `InlineArrayElementRef` to the inline-array indexer `b[i]` for any index. As @dgrunwald pointed out, that is unsound: `InlineArrayElementRef` is the compiler's **unchecked** element accessor, whereas the C# indexer `b[i]` (variable index) lowers to a **bounds-checked** span access, so the two only agree when the index is provably in-bounds.

This version reconstructs the `foreach` at the AST level instead, in `PatternStatementTransform`, alongside the existing `TransformForeachOnArray`. It matches the exact shape the compiler emits:

```
for (i = 0; i < N; i++) { item = <PrivateImplementationDetails>.InlineArrayElementRef(ref buffer, i); ... }
```

and only fires when the loop bound `N` equals `buffer`'s inline-array length. That equality is the in-bounds proof: `0 <= i < length`. The `InlineArrayElementRef` helper call is left untouched in the IL, so a loop that does not match this shape keeps the faithful (if unnameable) helper call rather than silently gaining a bounds check. This is sound in all cases, including hand-crafted IL.

Output:

```csharp
public int Sum(Byte16 b)
{
    int num = 0;
    foreach (byte b2 in b)
    {
        num += b2;
    }
    return num;
}
```

Verified that the checked span path (`b[i]` with a variable index), a partial-range loop (`i < 8`), and a genuine `for` loop reusing the index are all left unchanged.

## Test

Adds a `SumForeach` method to the `InlineArrayTests` Pretty fixture; it round-trips exactly (no `EXPECTED_OUTPUT` needed) across all `roslyn4OrNewerOptions` configs. Confirmed red without the transform (all 4 configs fail) and green with it; the full Pretty suite (1189 tests) stays green.

This PR was authored by an AI agent (Claude Code) operated by @siegfriedpammer and reviewed and manually verified by him.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/InlineArrayTests.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/InlineArrayTests.cs
@@ -99,6 +99,16 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 			return array[GetIndex()] = (array[GetIndex() + 1] = value);
 		}
 
+		public int SumForeach(Byte16 b)
+		{
+			int num = 0;
+			foreach (byte b2 in b)
+			{
+				num += b2;
+			}
+			return num;
+		}
+
 		public void OverloadResolution()
 		{
 			Receiver(GetByte16());
--- a/ICSharpCode.Decompiler/CSharp/Transforms/PatternStatementTransform.cs
+++ b/ICSharpCode.Decompiler/CSharp/Transforms/PatternStatementTransform.cs
@@ -95,6 +95,9 @@ namespace ICSharpCode.Decompiler.CSharp.Transforms
 			AstNode? result = TransformForeachOnArray(forStatement);
 			if (result != null)
 				return result;
+			result = TransformForeachOnInlineArray(forStatement);
+			if (result != null)
+				return result;
 			return base.VisitForStatement(forStatement);
 		}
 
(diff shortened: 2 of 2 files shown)


dgrunwald:
This changes the behavior with an out-of-bounds non-constant index.

This transform is only valid if the decompiler can prove the index is in-bounds.

siegfriedpammer:
Reworked to address the soundness concern. Rather than rewriting the unchecked `InlineArrayElementRef` helper to the bounds-checked indexer `b[i]`, the PR now reconstructs the `foreach` at the AST level and only fires when the loop bound equals the inline array's length -- which proves `0 <= i < length`. The helper call is left untouched in the IL, so any loop that doesn't match the exact compiler-emitted shape keeps the faithful (unnameable) call instead of silently gaining a bounds check. Force-pushed onto current master; the full Pretty suite stays green.

(Analysis and implementation by an AI agent (Claude Code) operated by @siegfriedpammer, reviewed and verified by him.)

PR #3851, opened 2026-07-03
Fix #3850: recover ReadOnlySpan<T> array literals from the legacy lazy cache

### LINKED ISSUE #3850: Decompiling a ReadOnlySpan<T> built from an array literal references the undeclared <PrivateImplementationDetails> cache (does not recompile)
### Summary

Decompiling a `ReadOnlySpan<T>` that is created from an array literal on a target framework **without** `RuntimeHelpers.CreateSpan` (e.g. .NET Framework, or netstandard2.0 + the System.Memory package) produces code that does not recompile: it references the compiler-synthesized `<PrivateImplementationDetails>` cache field, which is never declared and whose angle-bracket name is not expressible in C#.

### Input code

Compiled in **Release** targeting **.NET Framework 4.6.2** with the `System.Memory` package:

```csharp
using System;

public static class SpanCache
{
    public static ReadOnlySpan<char> NewLine => new char[] { '\r', '\n' };
}
```

On such a framework Roslyn caches the backing `char[]` in a lazy `<PrivateImplementationDetails>` static field (there is no `RuntimeHelpers.CreateSpan` to blit the data at run time).

### Erroneous output

```csharp
public static ReadOnlySpan<char> NewLine
{
    get
    {
        object obj = <PrivateImplementationDetails>.B7F5...7075_A1;
        if (obj == null)
        {
            obj = new char[2] { '\r', '\n' };
            <PrivateImplementationDetails>.B7F5...7075_A1 = (char[])obj;
        }
        return new ReadOnlySpan<char>((char[])obj);
    }
}
```

The synthesized `<PrivateImplementationDetails>` type is referenced but never declared, and its name is not expressible in C#. In whole-project output the name is escaped to `_003CPrivateImplementationDetails_003E`, which then fails to compile with **CS0400** (the type is undeclared).

### Expected output

The array literal should be recovered, so the `<PrivateImplementationDetails>` reference disappears — the same result the `RuntimeHelpers.CreateSpan` path (net7+) already produces:

```csharp
public static ReadOnlySpan<char> NewLine => new ReadOnlySpan<char>(new char[2] { '\r', '\n' });
```

### Details

* Version in use: master `fe50c00f` (11.0.0.9086); also reproduces on 10.1 (10.1.0.8386).
* Real-world occurrence: Markdig `CodeInlineParser` (net462).
* Root cause: ILSpy recovers the modern `RuntimeHelpers.CreateSpan` form (`TransformArrayInitializers.TransformRuntimeHelpersCreateSpanInitialization`) but not the legacy lazy-`char[]`-cache form, so the raw `<PrivateImplementationDetails>` field reference is left behind. This lazy cache has the same shape as the anonymous-method delegate cache already collapsed by `CachedDelegateInitialization`.

sailro:
Fixes #3850.

On target frameworks without `RuntimeHelpers.CreateSpan` (.NET Framework, or netstandard2.0 + System.Memory), Roslyn caches the backing array of a `ReadOnlySpan<T>` built from an array literal in a lazy `<PrivateImplementationDetails>` static field:

```csharp
object obj = <PrivateImplementationDetails>.cache;
if (obj == null) {
    obj = new char[] { CR, LF };
    <PrivateImplementationDetails>.cache = (char[])obj;
}
... new ReadOnlySpan<char>((char[])obj) ...
```

The decompiled output referenced the compiler-synthesized `<PrivateImplementationDetails>` type, which is never declared and whose name is not expressible in C#, so the output failed to recompile (CS0400 in whole-project output).

### Fix

A new `CachedReadOnlySpanInitialization` block transform collapses the lazy cache, mirroring `CachedDelegateInitialization` (which collapses the same lazy-static-field cache for anonymous-method delegates) and running right after it. Once the cache is collapsed, the existing array-initializer transforms recover the array literal, so the `<PrivateImplementationDetails>` reference disappears - the same result the modern `RuntimeHelpers.CreateSpan` path (net7+) already produces.

**Before** (does not recompile):
```csharp
public static ReadOnlySpan<char> NewLine
{
    get
    {
        object obj = <PrivateImplementationDetails>.B7F5...7075_A1;
        if (obj == null)
        {
            obj = new char[2] { CR, LF };
            <PrivateImplementationDetails>.B7F5...7075_A1 = (char[])obj;
        }
        return new ReadOnlySpan<char>((char[])obj);
    }
}
```
**After:**
```csharp
public static ReadOnlySpan<char> NewLine => new ReadOnlySpan<char>(new char[2] { CR, LF });
```

The transform only fires on a compiler-generated static cache field with this exact lazy-init shape, so it does not touch the existing `<PrivateImplementationDetails>` switch-on-string cache (covered by the existing `CS1xSwitch_Debug` test) or user code.

### Test

`ILPretty/CachedReadOnlySpanInitialization` exercises the legacy lazy-cache IL and checks the recovered output.

Real-world occurrence: Markdig `CodeInlineParser` (net462).

### DIFF
--- a/ICSharpCode.Decompiler.Tests/ILPrettyTestRunner.cs
+++ b/ICSharpCode.Decompiler.Tests/ILPrettyTestRunner.cs
@@ -274,6 +274,12 @@ namespace ICSharpCode.Decompiler.Tests
 		}
 
 		[Test]
+		public async Task CachedReadOnlySpanFromLazyCache()
+		{
+			await Run();
+		}
+
+		[Test]
 		public async Task ConstantBlobs()
 		{
 			await Run();
--- a/ICSharpCode.Decompiler/IL/Transforms/CachedReadOnlySpanInitialization.cs
+++ b/ICSharpCode.Decompiler/IL/Transforms/CachedReadOnlySpanInitialization.cs
@@ -1,0 +1,121 @@
+// Copyright (c) 2026 Sebastien Lebreton
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this
+// software and associated documentation files (the "Software"), to deal in the Software
+// without restriction, including without limitation the rights to use, copy, modify, merge,
+// publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons
+// to whom the Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all copies or
+// substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+// INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+// PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+// FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+using ICSharpCode.Decompiler.TypeSystem;
+
+namespace ICSharpCode.Decompiler.IL.Transforms
+{
+	/// <summary>
+	/// Collapses the compiler-generated lazy cache Roslyn emits for a <c>ReadOnlySpan&lt;T&gt;</c> that is
+	/// created from an array literal on target frameworks without <c>RuntimeHelpers.CreateSpan</c> (e.g.
+	/// .NET Framework or netstandard2.0 + System.Memory):
+	/// <code>
+	/// stloc V(ldobj T[](ldsflda &lt;PrivateImplementationDetails&gt;.cache))
+	/// if (V == null) {
+	///     stloc V(arrayInitializer)
+	///     stobj T[](ldsflda &lt;PrivateImplementationDetails&gt;.cache, ldloc V)
+	/// }
+	/// ... single usage of V, e.g. newobj ReadOnlySpan&lt;T&gt;(ldloc V) ...
+	/// </code>
+	/// is turned into:
+	/// <code>
+	/// stloc V(arrayInitializer)
+	/// ... single usage of V ...
+	/// </code>
+	/// Afterwards the existing array-initializer transforms recover the array literal, so the reference to
+	/// the compiler-synthesized <c>&lt;PrivateImplementationDetails&gt;</c> cache field disappears. Without
+	/// this transform the decompiled output references that field, whose escaped name
+	/// (<c>&lt;PrivateImplementationDetails&gt;</c>) is not expressible in C# and is never declared, so the
+	/// output does not recompile (CS0400).
+	///
+	/// This mirrors <see cref="CachedDelegateInitialization"/>, which collapses the analogous lazy cache for
+	/// anonymous-method delegates, and therefore runs right after it.
+	/// </summary>
+	public class CachedReadOnlySpanInitialization : IBlockTransform
+	{
+		public void Run(Block block, BlockTransformContext context)
+		{
+			if (!context.Settings.ArrayInitializers)
+				return;
+
+			// The store that loads the cache field precedes the if, at block.Instructions[i - 1],
+			// so there is nothing to match when i == 0.
+			for (int i = context.IndexOfFirstAlreadyTransformedInstruction - 1; i >= 1; i--)
+			{
+				if (block.Instructions[i] is IfInstruction inst && DoTransform(block, i, inst, context))
+				{
+					context.IndexOfFirstAlreadyTransformedInstruction = block.Instructions.Count;
+				}
+			}
+		}
+
+		/// <summary>
+		/// Matches
+		/// <code>
+		/// stloc V(ldobj(ldsflda cacheField))                       // block.Instructions[i - 1]
+		/// if (comp(ldloc V == ldnull)) {                           // block.Instructions[i]
+		///     stloc V(value)
+		///     stobj(ldsflda cacheField, ldloc V)
+		/// }
+		/// </code>
+		/// and replaces the load-from-cache with the initializer value, dropping the if:
+		/// <code>
+		/// stloc V(value)
+		/// </code>
+		/// </summary>
+		static bool DoTransform(Block block, int i, IfInstruction inst, BlockTransformContext context)
+		{
+			// storeBeforeIf: stloc V(ldobj(ldsflda cacheField)), cacheField a compiler-generated static field.
+			if (block.Instructions[i - 1] is not StLoc { Value: LdObj { Target: LdsFlda { Field: var cacheField } } } storeBeforeIf)
+				return false;
+
+			if (!cacheField.IsCompilerGeneratedOrIsInCompilerGeneratedClass())
+				return false;
+
+			// V is assigned exactly twice (before-if load + in-if init) and read exactly three times
+			// (null-check condition + cache write-back + one real downstream usage), with no address-of.
+			var v = storeBeforeIf.Variable;
+			if (v.StoreCount != 2 || v.LoadCount != 3 || v.AddressCount != 0)
+				return false;
+
+			// The if must be a simple `if (...) { ... }` (no else) with a two-instruction body.
+			if (!inst.FalseInst.MatchNop() || inst.TrueInst is not Block trueBlock || trueBlock.Instructions.Count != 2)
+				return false;
+
+			// condition: V == null (MatchCompEqualsNull also accepts null == V and the negated forms).
+			if (!inst.Condition.MatchCompEqualsNull(out var nullCheckArg) || !nullCheckArg.MatchLdLoc(v))
+				return false;
+
+			// trueBlock[0]: stloc V(value)
+			if (trueBlock.Instructions[0] is not StLoc storeValue || storeValue.Variable != v)
+				return false;
+
+			// trueBlock[1]: stobj(ldsflda cacheField, ldloc V) -> the write-back to the same cache field.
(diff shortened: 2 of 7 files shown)


ICSharpCode.Decompiler/IL/Transforms/CachedReadOnlySpanInitialization.cs
```diff
@@ -0,0 +1,126 @@
+// Copyright (c) 2024 ICSharpCode
```
siegfriedpammer:
```suggestion
// Copyright (c) 2026 Sebastien Lebreton
```

ICSharpCode.Decompiler/IL/Transforms/CachedReadOnlySpanInitialization.cs
```diff
@@ -0,0 +1,126 @@
...
+			context.Step("CachedReadOnlySpanInitialization", inst);
+			storeBeforeIf.Value = storeValue.Value;
+			block.Instructions.RemoveAt(i);
+			return true;
+		}
+
+		static bool MatchIsNull(ILInstruction condition, ILVariable v)
+		{
+			// comp(ldloc V == ldnull)
+			if (condition.MatchCompEquals(out var left, out var right) && right.MatchLdNull() && left.MatchLdLoc(v))
```
siegfriedpammer: the catch-all form to solve this would be a while (MatchLogicNot) loop that unpacks the negation. This is a proven pattern used throughout the codebase. After that use `MatchCompEqualsNull` to catch `left == null` and `null == right`  at the same time.

ICSharpCode.Decompiler/IL/Transforms/CachedReadOnlySpanInitialization.cs
```diff
@@ -0,0 +1,126 @@
...
+	/// This mirrors <see cref="CachedDelegateInitialization"/>, which collapses the analogous lazy cache for
+	/// anonymous-method delegates, and therefore runs right after it.
+	/// </summary>
+	public class CachedReadOnlySpanInitialization : IBlockTransform
+	{
+		public void Run(Block block, BlockTransformContext context)
+		{
+			if (!context.Settings.ArrayInitializers)
+				return;
+			for (int i = context.IndexOfFirstAlreadyTransformedInstruction - 1; i >= 1; i--)
```
siegfriedpammer, line 58: `i >= 1` made me raise an eyebrow until I read the pattern further down... might be worth a comment.

ICSharpCode.Decompiler/IL/Transforms/CachedReadOnlySpanInitialization.cs
```diff
@@ -0,0 +1,126 @@
...
+		/// </code>
+		/// and replaces the load-from-cache with the initializer value, dropping the if:
+		/// <code>
+		/// stloc V(value)
+		/// </code>
+		/// </summary>
+		static bool DoTransform(Block block, int i, IfInstruction inst, BlockTransformContext context)
+		{
+			// The if must be a simple `if (...) { ... }` (no else) with a two-instruction body.
+			if (!inst.FalseInst.MatchNop() || inst.TrueInst is not Block trueBlock || trueBlock.Instructions.Count != 2)
```
siegfriedpammer: nit: could use property patterns to make it shorter

ICSharpCode.Decompiler/IL/Transforms/CachedReadOnlySpanInitialization.cs
```diff
@@ -0,0 +1,126 @@
...
+				return false;
+			// storeBeforeIf: stloc V(ldobj(ldsflda cacheField)), cacheField a compiler-generated static field.
+			if (block.Instructions[i - 1] is not StLoc storeBeforeIf || storeBeforeIf.Value is not LdObj ldobj
+				|| !ldobj.Target.MatchLdsFlda(out var cacheField) || !cacheField.IsCompilerGeneratedOrIsInCompilerGeneratedClass())
```
siegfriedpammer: should probably be split into multiple sequential if statements, like we do in other transforms, for readability

ICSharpCode.Decompiler/IL/Transforms/CachedReadOnlySpanInitialization.cs
```diff
@@ -0,0 +1,126 @@
...
+				return false;
+			// trueBlock[1]: stobj(ldsflda cacheField, ldloc V) -> the write-back to the same cache field.
+			if (trueBlock.Instructions[1] is not StObj stobj || !stobj.Target.MatchLdsFlda(out var cacheField2)
+				|| !cacheField.Equals(cacheField2) || !stobj.Value.MatchLdLoc(v))
+			{
+				return false;
+			}
+			// V is assigned exactly twice (before-if load + in-if init) and read exactly three times
+			// (null-check condition + cache write-back + one real downstream usage), with no address-of.
+			if (v.StoreCount != 2 || v.LoadCount != 3 || v.AddressCount != 0)
```
siegfriedpammer: this check could be moved up, so we don't spend time on matching the full pattern, if this simple check fails at the end.

ICSharpCode.Decompiler.Tests/ILPrettyTestRunner.cs
```diff
@@ -285,6 +285,12 @@ public async Task Unsafe()
 			await Run(assemblerOptions: AssemblerOptions.Library | AssemblerOptions.UseLegacyAssembler);
 		}
 
+		[Test]
+		public async Task CachedReadOnlySpanInitialization()
```
siegfriedpammer: is there a reason why this is not a normal C# Pretty test? ILPretty tests have the disadvantage that they don't detect changes in Roslyn's pattern detection, when we update or add a new compiler version.

ICSharpCode.Decompiler.Tests/TestCases/ILPretty/CachedReadOnlySpanInitialization.il
```diff
@@ -0,0 +1,55 @@
+.assembly extern mscorlib
+{
```
siegfriedpammer: How did you generate this IL? It looks like it was generated by an LLM. Please use a C# program as starting point and "convert" it to IL using `ildasm`, `ilspycmd` or https://lab.razor.fyi. This way we ensure we actually test against the IL shape Roslyn emits (or emitted at a certain point in time) and the produced IL is generated by a decade old battle-tested tool instead of a "potentially hallucinating DWIM slot machine".

siegfriedpammer (changes requested):

ICSharpCode.Decompiler/IL/Transforms/CachedReadOnlySpanInitialization.cs
sailro: Done, updated to `Copyright (c) 2026 Sebastien Lebreton`.

ICSharpCode.Decompiler/IL/Transforms/CachedReadOnlySpanInitialization.cs
sailro: Done. I used `MatchCompEqualsNull`, which already implements exactly this: it calls `MatchCompEquals`, whose `while (MatchLogicNot(...) && arg is Comp)` loop unpacks the negation (flipping the equality/inequality parity), and it then accepts both `left == null` and `null == right`. So the condition is now `inst.Condition.MatchCompEqualsNull(out var nullCheckArg) && nullCheckArg.MatchLdLoc(v)`, and the custom `MatchIsNull` helper is gone.

ICSharpCode.Decompiler/IL/Transforms/CachedReadOnlySpanInitialization.cs
sailro: Added a comment explaining it: the cached store we match is at `block.Instructions[i - 1]`, so there is nothing to match when `i == 0`.

ICSharpCode.Decompiler/IL/Transforms/CachedReadOnlySpanInitialization.cs
sailro: Done — the preceding store is now matched with a property pattern:

```csharp
if (block.Instructions[i - 1] is not StLoc { Value: LdObj { Target: LdsFlda { Field: var cacheField } } } storeBeforeIf)
    return false;
```

ICSharpCode.Decompiler/IL/Transforms/CachedReadOnlySpanInitialization.cs
sailro: Done — split into sequential `if` guards (compiler-generated check, variable store/load/address counts, the if-shape, the null condition, then each body instruction), consistent with the other transforms.

ICSharpCode.Decompiler/IL/Transforms/CachedReadOnlySpanInitialization.cs
sailro: Done — the `StoreCount`/`LoadCount`/`AddressCount` check now runs right after we obtain the variable, before matching the rest of the pattern.

ICSharpCode.Decompiler.Tests/ILPrettyTestRunner.cs
sailro: No good reason — converted it to a proper Pretty test (`TestCases/Pretty/CachedReadOnlySpanInitialization.cs`). It compiles real C# with each supported Roslyn version, so it does exercise Roslyn's pattern detection: the expected output branches on `NET70`, keeping the `RuntimeHelpers.CreateSpan` form on modern targets and recovering the explicit `ReadOnlySpan<char>` constructor on Roslyn 2.10/3.11 (which lack `CreateSpan`). Verified across Roslyn 2.10/3.11/4.14/latest x optimize.

ICSharpCode.Decompiler.Tests/TestCases/ILPretty/CachedReadOnlySpanInitialization.il
sailro: You're right, and thanks for the pointers. I've removed the hand-written `.il` entirely — the new Pretty test lets Roslyn emit the pattern, so we test against the real compiler-produced IL shape rather than hand-authored IL.

siegfriedpammer:
@sailro I think I have to apologize to you, after some investigation I noticed that the legacy pattern with the cache field was in-fact the correct choice of test. I have re-added a test. Sorry again.

siegfriedpammer (approved):

PR #3854, opened 2026-07-03
Fix #3853: infer and restore `scoped` on ref and ref-struct locals

### LINKED ISSUE #3853: Decompiling a `scoped` ref local or ref-struct local drops the modifier (does not recompile)
### Summary

`scoped` on a **local variable** — a `scoped ref T` local, or a ref-struct local such as `scoped Span<T>` — is a compile-time ref-safety annotation with no runtime effect. Unlike `scoped` on a **parameter** (which is encoded as `ScopedRefAttribute` and correctly recovered by ILSpy), `scoped` on a local is not encoded in IL, metadata, or the PDB, so the decompiler drops it. When the local is later reassigned a value with a narrower escape scope, the decompiled output no longer recompiles: **CS8374** for `ref` locals, **CS8347** for ref-struct locals.

### Input code

Compiled in **Release** targeting **.NET 8.0** (`AllowUnsafeBlocks`, C# 11+):

```csharp
using System;

public class C
{
    // A 'scoped' ref local, ref-reassigned to the address of a method-local.
    public static int RefLocal(bool b, ref int x)
    {
        int y = 42;
        scoped ref int r = ref x;
        if (b)
        {
            r = ref y;
        }
        return r;
    }

    // A 'scoped' ref-struct local, reassigned to a call that captures a method-local.
    private static int f;
    private static Span<int> NoCapture(scoped ref int v) => new Span<int>(ref f);
    private static Span<int> Capture(ref int v) => new Span<int>(ref v);

    public static int RefStruct(bool b)
    {
        int value = 7;
        scoped Span<int> span = NoCapture(ref value);
        if (b)
        {
            span = Capture(ref value);
        }
        return span[0];
    }
}
```

### Erroneous output

The `scoped` modifier is dropped from both locals:

```csharp
public static int RefLocal(bool b, ref int x)
{
    int num = 42;
    ref int reference = ref x;          // <-- 'scoped' dropped
    if (b)
    {
        reference = ref num;
    }
    return reference;
}

public static int RefStruct(bool b)
{
    int v = 7;
    Span<int> span = NoCapture(ref v);  // <-- 'scoped' dropped
    if (b)
    {
        span = Capture(ref v);
    }
    return span[0];
}
```

```
error CS8374: Cannot ref-assign 'num' to 'reference' because 'num' has a narrower escape scope than 'reference'.
error CS8347: Cannot use a result of 'C.Capture(ref int)' in this context because it may expose variables referenced by parameter 'v' outside of their declaration scope
```

### Expected output

`scoped` is restored, so the output recompiles. Because `scoped` has no runtime effect, this is behavior-preserving:

```csharp
scoped ref int reference = ref x;
// ...
scoped Span<int> span = NoCapture(ref v);
```

### Details

* Version in use: master `6b2852ce` (11.x); also reproduces on 10.1 (10.1.0.8386).
* Root cause: `scoped` on a local is fully erased at compile time — identical IL, no attribute, and no PDB entry (verified: `ilspycmd -usepdb` recovers local *names* but not scoped-ness). ILSpy models `scoped` only on `IParameter` (via `ScopedRefAttribute`); `VariableDeclarationStatement` has no `scoped` concept. Recovering it therefore requires inferring, from the local's reassignments, that it must have been `scoped` for the original source to pass ref-safety.

sailro:
Fixes #3853.

`scoped` on a local (`scoped ref T`, or a ref-struct local such as `scoped Span<T>`) has no runtime effect and — unlike `scoped` on a parameter (`ScopedRefAttribute`) — is not encoded in IL, metadata, or the PDB, so the decompiler dropped it. The output then failed to recompile whenever the local was reassigned a narrower-scoped value: **CS8374** (ref locals) or **CS8347 / CS8352** (ref structs).

### Fix

`DeclareVariables` now infers `scoped` for the cases where it is *determined* by the local's use. A by-ref-like local (a `ref` local, or a ref struct — `IType.IsByRefLike`) that has more than one store and is assigned a value limited to a method-local's scope must have been declared `scoped` for the original source to pass ref-safety. Such a narrow value is:

- the **address of a method-local**, or a field thereof (`ldloca` / `ldflda` rooted in a `Local`/`StackSlot`), or
- a **call/newobj that passes such an address to a non-`scoped` by-ref parameter** — the returned ref struct captures it and inherits the narrow scope.

Passing a local's address to a `scoped` by-ref parameter does **not** capture it, so those calls correctly do not trigger the modifier (`!IParameter.Lifetime.ScopedRef`). The inference is gated on the existing `ScopedRef` setting and emits `Modifiers.Scoped` before the type.

Because `scoped` has no runtime effect and such a local provably cannot escape the method, restoring it is always behavior-preserving.

**Before** (does not recompile — CS8374 / CS8347):
```csharp
ref int reference = ref x;
reference = ref num;
// ...
Span<int> span = NoCapture(ref v);
span = Capture(ref v);
```
**After:**
```csharp
scoped ref int reference = ref x;
reference = ref num;
// ...
scoped Span<int> span = NoCapture(ref v);
span = Capture(ref v);
```

### Scope

Only the determined cases above (the combined `declaration = initializer` form) are inferred. A `scoped` local declared separately from its assignment, a direct `stackalloc` reassignment, and receiver-captured returns are not covered and decompile as before. The rule is conservative — it never adds `scoped` unless the assigned value is provably narrow — and the full `Pretty` suite confirms no over-application across the existing corpus.

### Test

`Pretty/RefFields` (roslyn4+): a new `LifetimeTests.ReassignScopedRefToLocal` (ref local) and the now-enabled `scoped Span<T>` case in `LifetimeTests.Calls` (which previously carried a `// -- would need scoped local, not yet implemented` note). Both fail without the fix (the modifier is dropped) and pass with it.


siegfriedpammer:
I think we are going to defer this until we have properly implemented `scoped`/`UnscopedRefAttribute` in the type system.

sailro:
Thank you for the guidance on the type-system prerequisite. I have split the replacement into two focused PRs:

- #3887 models `scoped`, `UnscopedRefAttribute`, and effective lifetime semantics in the type system without changing decompiler output.
- #3888 is the dependent draft that recovers erased `scoped` locals using that model and a late ILAst ref-safety analysis.

I am closing this PR as superseded by #3888.

PR #3863, opened 2026-07-05
Simplify hoisted null-guard fold to reference-type constructor chains

siegfriedpammer:
Follow-up to #3852.

A value-type constructor chains via `this = new TSelf(...)`, an ordinary body statement, so a hoisted argument null-guard in front of it is legal C# output as-is; folding it back only bought lifting the chain into a `this(...)` initializer. That cosmetic gain does not justify the extra stobj shape matching, so the guard now stays in the body for structs, and `GuardPrecedesChainedConstructorCall` (together with its two struct-shape matchers) is inlined away: the gate reduces to `ILInlining.IsInConstructorInitializer`, and the first-use search to a single `ILInlining.FindLoadInNext` call on the following statement, which also proves order-safe reachability before anything is mutated. On the bail-out path the operands stolen by the tentative `NullCoalescingInstruction` are restored to their primary positions, following the existing pattern in `TransformThrowExpressionValueTypes`.

The struct fixtures now pin the guard-in-body + `this = new TSelf(...)` form; `NullCheckedNullableStructThisChain` still folds via the independent `TransformThrowExpressionValueTypes` path. Full decompiler suite: 2308 passed, 0 failed.

### DIFF
--- a/ICSharpCode.Decompiler/IL/Transforms/NullCoalescingTransform.cs
+++ b/ICSharpCode.Decompiler/IL/Transforms/NullCoalescingTransform.cs
@@ -124,11 +124,10 @@ namespace ICSharpCode.Decompiler.IL.Transforms
 		/// redirecting the first following use of the parameter (the position the check was hoisted
 		/// from) to <c>temp</c>, and leave moving the coalescing expression into the chained call to
 		/// ILInlining, which does so only when that preserves the order of evaluation.
-		/// Reference-type chains (<c>base/this..ctor</c> calls) are identified by IL offset via
-		/// <see cref="ILInlining.IsInConstructorInitializer"/>; value types chain via
-		/// <c>this = new TSelf(...)</c>, i.e. <c>stobj(ldthis, newobj TSelf(...))</c>, which
-		/// <see cref="ILFunction.ChainedConstructorCallILOffset"/> does not report, so the stobj
-		/// shape is matched directly.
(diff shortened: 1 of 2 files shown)


ICSharpCode.Decompiler/IL/Transforms/NullCoalescingTransform.cs
```diff
@@ -145,26 +144,37 @@ bool TransformHoistedConstructorArgumentNullGuard(Block block, int pos, Statemen
...
+			var throwInstParent = throwInst.Parent;
+			var throwInstChildIndex = throwInst.ChildIndex;
+			var expressionWithThrow = new NullCoalescingInstruction(NullCoalescingKind.Ref, paramLoad, throwInst);
+			var result = ILInlining.FindLoadInNext(block.Instructions[pos + 1], paramLoad.Variable, expressionWithThrow,
+				InliningOptions.None);
+			if (result.Type != ILInlining.FindResultType.Found || result.LoadInst is not LdLoc firstUse)
 			{
-				firstUse = block.Instructions[i].Descendants.OfType<LdLoc>()
-					.FirstOrDefault(ld => ld.Variable == paramLoad.Variable);
+				// reset the primary positions (see remarks on ILInstruction.Parent)
```
dgrunwald, line 164: A simpler alternative might be to use the original `IfInstruction` as `inlinedExpression` for the FindLoadInNext call; then construct the new expression only if the transform is possible.

PR #3864, opened 2026-07-05
Fix #3860: Avoid 'out var' if the variable recurs in the argument list

### LINKED ISSUE #3860: Decompiled out variables fail to recompile when the same local is passed as multiple out arguments
### Input code
```c#
public static void f(out int x, out int y) {
	x = 0; y = 0;
}
public static void Demo() {
	int x;
	f(out x, out x);
}
```

### Erroneous output
With `OutVariables` enabled,
```c#
public static void Demo()
{
	f(out var x, out x);
}
```
Fails to recompile, with compiler error message:
```
error CS8196: Reference to an implicitly-typed variable 'x' is not permitted in this location.
```

### Details
ILSpy version 9.1.0.7988+03b7444943e720b3134d296c0c8dd3876f8ea4ce

Since the local in the caller is unused, I think it might as well decompile to `f(out _, out _);`, which recompiles to different IL but should have the same semantics? C# disallows the callee from attempting to read from out parameters, but IL does nothing to prevent the callee from `ldind` on the out parameter. ECMA-335 mentions that the CLI does not enforce the semantics of `[out]` but can use it for optimization.
Perhaps try to decompile to discards? Or perhaps gate it behind a setting flag?

siegfriedpammer:
Fixes #3860.

When the same local is passed as multiple out arguments of one call, DeclareVariables turned the first use into an implicitly-typed declaration, producing `f(out var x, out x)`, which fails to recompile with CS8196: referencing an implicitly-typed out variable in another argument of the declaring call is not permitted, because its type is only inferred once overload resolution of that call has completed. The explicitly-typed form `f(out int x, out x)` is valid C#, so the transform now falls back to the explicit type when the variable is referenced again within the declaring call (resolving identifiers through collision-merged variables).

New Pretty fixture methods in `OutVariables.cs` reproduce the issue (previously decompiled as `out var a, out a` in all compiler configurations); full Pretty/Ugly/ILPretty/VBPretty runs pass.

This PR was authored by an AI agent on behalf of @siegfriedpammer.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/OutVariables.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/OutVariables.cs
@@ -92,5 +92,24 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 			Console.WriteLine(value2);
 			Console.WriteLine(value2);
 		}
+
+		private static void GetTwo(out int a, out int b)
+		{
+			a = 1;
+			b = 2;
+		}
+
+		public static void SameVariableUsedForTwoOutParameters()
+		{
+			// The declaration must use the explicit type: referencing an implicitly-typed
+			// out variable in another argument of the declaring call is an error (CS8196).
+			GetTwo(out int a, out a);
+		}
+
+		public static int SameVariableUsedForTwoOutParametersAndRead()
+		{
+			GetTwo(out int a, out a);
+			return a;
+		}
 	}
 }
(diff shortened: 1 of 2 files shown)


dgrunwald:
Do we also need to worry about nested cases? `Foo(out var x, Bar(out x))`

siegfriedpammer:
I added a test case, seems Claude implemented support for it "by accident"... the weird first-child-next-sibling loop traverses all descendants of the individual arguments... I was thinking about changing the loop to actually use InvocationExpression.Arguments instead... what do you think?

PR #3866, opened 2026-07-05
Fix #3475: Emit 'true ? null : new { ... }' for null of anonymous type

### LINKED ISSUE #3475: `null` of anonymous type
### Input code
```csharp
using System;

static class Program
{
    static void Main() => Test(true ? null : new { A = 1, B = 2 });

    static void Test<T>(T t) => Console.WriteLine(typeof(T).FullName);
}
```
### Erroneous output
```csharp
using System;

internal static class Program
{
	private static void Main()
	{
		Test(null);
	}

	private static void Test<T>(T t)
	{
		Console.WriteLine(typeof(T).FullName);
	}
}
```
error CS0411: The type arguments for method 'Program.Test<T>(T)' cannot be inferred from the usage. Try specifying the type arguments explicitly.
### Details
* Product in use: ILSpy
* ILSpy version: 9.1.0.7988+03b7444943e720b3134d296c0c8dd3876f8ea4ce
* .NET version: 9.0.4+f57e6dc747158ab7ade4e62a75a6750d16b771e8

siegfriedpammer:
Fixes #3475.

**Problem**: `Test(true ? null : new { A = 1, B = 2 })` compiles to a bare `ldnull` passed to `Test<<>f__AnonymousType0<int, int>>`. The type argument cannot be written explicitly (anonymous types cannot be named), so ILSpy silently dropped the type argument list and emitted `Test(null)`, which no longer compiles (CS0411).

**Approach**: when type inference fails and the type arguments contain anonymous types, `CallBuilder` now rewrites null-literal arguments of anonymous type as `true ? null : new { A = default(int), B = default(int) }` -- a conditional expression whose never-taken branch creates an instance of the anonymous type. This is the minimal C# expression that produces a null value of an anonymous type, and it makes the type arguments inferable again. Roslyn folds the constant conditional back to `ldnull`, so the output round-trips.

The instance expression is not assembled by hand: a synthesized `newobj` instruction with `default.value` arguments (nested `newobj` for directly nested anonymous types) is run through the existing translation pipeline, which already renders anonymous-type creations as object-initializer syntax with exactly-typed values (`default(int)`, `(string)null`, ...). Exact types matter because anonymous-type property types are inferred from the initializer values. The original property values are not recoverable (the IL contains only `ldnull`) and are never evaluated.

Anonymous types that occur inside other constructed types (e.g. `List<T>` or arrays of an anonymous type) cannot be expressed without naming the type; those keep the previous output.

**Verification**: new round-trip cases in the `AnonymousTypes` pretty test, covering the plain, string-property, and nested cases (green in all 12 configurations runnable on Linux; mcs/legacy-csc configurations need Windows CI); full decompiler suite passes; the issue's exact reproduction decompiles to compilable code that prints the same anonymous type name at runtime as the original assembly.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/AnonymousTypes.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/AnonymousTypes.cs
@@ -133,5 +133,46 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 				Renamed = v.Minor
 			};
 		}
+
+		private void NullOfAnonymousType()
+		{
+			Identify(true ? null : new {
+				X = default(int),
+				Y = default(int)
+			});
+			Identify(true ? null : new {
+				N = default(int),
+				S = (string)null
+			});
+			Identify(true ? null : new {
+				Inner = new {
+					X = default(int)
+				}
+			});
+		}
+
+		private void NullOfAnonymousTypeNonGenericCall()
+		{
+#if OPT
+			MakeAction(new {
+				X = 0
+			})(null);
+#else
+			var action = MakeAction(new {
+				X = 0
+			});
+			action(null);
+#endif
+		}
+
+		private static void Identify<T>(T t)
+		{
+			Console.WriteLine(typeof(T).FullName);
+		}
+
+		private static Action<T> MakeAction<T>(T template)
+		{
+			return null;
+		}
 	}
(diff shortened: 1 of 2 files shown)


PetSerAl:
Little nitpicking:
Arrays of anonymous type could be expressed without naming the type: `new[] { new { A = 1, B = 2 } }`.

PR #3869, opened 2026-07-06
Add support for C# 15 closed hierarchies

christophwille:
## What this adds

C# 15 (shipping with .NET 11, currently preview 5) introduces [closed hierarchies](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-15#closed-hierarchies): a `closed` modifier on classes/records that restricts direct derivation to the declaring assembly, enabling exhaustive `switch` expressions without a default arm ([keyword reference](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/closed), [feature spec](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/closed-hierarchies)).

The preview 5 compiler encodes the modifier as (verified against the shipped Roslyn binary and the red-test diff; the older spec text still says `IsClosedTypeAttribute`, which is *not* what is emitted):

- `System.Runtime.CompilerServices.ClosedAttribute` on the type,
- `[CompilerFeatureRequired("ClosedClasses")]` on **every** constructor, including the synthesized copy constructor of records (no paired `[Obsolete]`, unlike required members),
- the `abstract` flag on the type (`closed` is implicitly abstract).

The BCL does not ship `ClosedAttribute` yet, so every assembly using `closed` declares its own copy (or references one from another assembly); the compiler matches it by full name.

Previously ILSpy showed the raw encoding (`[Closed] public abstract class ...` with `[CompilerFeatureRequired]` on constructors); now it prints `public closed class ...` again.

## Implementation

Mirrors the `required`-members pipeline end to end, gated on a new `DecompilerSettings.ClosedHierarchies` toggle (default on, disabled for language versions below C# 15):

| Area | Change |
|---|---|
| `KnownAttributes` | new `KnownAttribute.Closed` -> `S.R.CS.ClosedAttribute` |
| `Modifiers` | new `Modifiers.Closed`, printed after `sealed` in the modifier order |
| `CSharpDecompiler` | type level: remove `[Closed]`, set `Modifiers.Closed`, clear the implicit `Modifiers.Abstract`; constructor level: strip `CompilerFeatureRequired("ClosedClasses")` (composes with the existing `RequiredMembers` stripper when both features are used on one type) |
| `RecordDecompiler` | `IsAllowedAttribute` now accepts `CompilerFeatureRequired("ClosedClasses")`, so the synthesized copy constructor of a closed record is still recognized as compiler-generated and hidden |
| `DecompilerSettings` | new C# 15 gate in `SetLanguageVersion`/`GetMinimumRequiredVersion` |
| project export | `ProjectFileWriterDefault`/`ProjectFileWriterSdkStyle` emit `<LangVersion>preview</LangVersion>` for C# 15, because csc does not accept `-langversion:15.0` yet (and `LanguageVersion.CSharp15_0`/`Preview` share the enum value 1500, making `ToString()` ambiguous) |
| test harness | `RemoveEmbeddedAttributes` also strips a source-declared `S.R.CS.ClosedAttribute` polyfill class from decompiled output |
| ILSpy UI | resource string for the new setting |

## Tests

Two new Pretty fixtures on `roslyn5OrNewerOptions` + `CompilerOptions.Preview` (the pinned Roslyn `5.8.0-1.26302.115` / net11 preview 5 ref pack; same shape as `ExtensionEverything`):

- **`ClosedHierarchies.cs`** - minimum covering set for the definition side: plain `closed class` with multiple constructors (attribute stripping on each), `closed record` with positional record descendants, an **internal** descendant of a public closed record, an **internal** closed class, a nested closed class, a generic `closed class Tree<T>` with `Leaf<U> : Tree<U>` and `ArrayLeaf<V> : Tree<V[]>`, and a closed class with a `required` member so both `CompilerFeatureRequired` values land on one constructor. The `ClosedAttribute` polyfill is declared under `#if !EXPECTED_OUTPUT`.
- **`ClosedHierarchiesCrossAssembly.cs` + `.dep.cs`** - cross-assembly dimension via the `// #dependency` mechanism: the attribute *and* a closed hierarchy live in the referenced assembly; the decompiled assembly declares its own closed record using the dep-provided attribute (the shape everything will have once the BCL ships it) and derives a record from the dep's non-closed descendant (the legal cross-assembly derivation).

Deliberately **not** covered: consuming-side exhaustive switch expressions. Exhaustiveness is compile-time knowledge only - the IL of a switch over a closed hierarchy is identical to one over an open hierarchy, so there is nothing for a decompiler test to observe.

The fixtures were written and shown red first (TDD): the red diff doubled as the authoritative probe of what preview 5 actually emits.

## Changes to existing decompilation behavior

- **`Issue3684.cs` expected output changed** from `((BaseClass)this).Convert<T>(input)` to `Convert(input)`. This falls out of the harness fix below: the fixture's dependency assembly is now resolvable while decompiling, so ILSpy can prove the simple-name call binds to the base method and no longer emits the defensive cast + explicit type arguments. The calls are semantically identical, and the point of that test (removal of compiler-generated interface accessor helpers) is unaffected.
- **Whole-project export now writes `<LangVersion>preview</LangVersion>`** whenever the effective minimum language version is C# 15 - which it is by default, since the new setting defaults to on (the same convention as every previous default-on language feature setting).

## Pre-existing issues surfaced by the new tests

- **`#dependency` assemblies were never resolvable at decompile time.** `Tester.CompileCSharp` compiled them to a random temp file (`CompilerResults.PathToAssembly` falls back to `GetTempFileName()`), so the `UniversalAssemblyResolver` could not find them next to the main test assembly. `Issue3684` never noticed because plain classes decompile fine with an unresolved base type. Dependencies now compile to `<depName><config-suffix>.dll` next to the main output.
- **Latent NRE in `RecordDecompiler`.** `DetectPrimaryConstructor` dereferenced `chainedCtor.DeclaringTypeDefinition!` (and the same for the ctor-chain check), which is null when a record's base type lives in an assembly that cannot be resolved - crashing decompilation of such records anywhere, not just in tests. Both checks now use null-safe comparisons and degrade gracefully.

## Verification

- New tests: 4/4 green (both fixtures, optimized and non-optimized).
- Full `PrettyTestRunner`: 1854 test cases green.
- Whole `ICSharpCode.Decompiler.Tests` project: 3666 passed, 0 failed (12 pre-existing skips: mono/known bugs/ILSpy-tests submodule not checked out).
- `ILSpy.Tests` + `ILSpy.Tests.Windows`: 1091 passed.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/Issue3684.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/Issue3684.cs
@@ -15,7 +15,7 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 		{
 			T IInterface.Convert<T>(T input)
 			{
-				return ((BaseClass)this).Convert<T>(input);
+				return Convert(input);
 			}
 		}
 	}
--- a/ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
+++ b/ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
@@ -1703,6 +1703,12 @@ namespace ICSharpCode.Decompiler.CSharp
 				{
 					RemoveAttribute(typeDecl, KnownAttribute.Required);
 				}
+				if (settings.ClosedHierarchies && RemoveAttribute(typeDecl, KnownAttribute.Closed))
+				{
+					// closed classes are implicitly abstract
+					typeDecl.Modifiers |= Modifiers.Closed;
+					typeDecl.Modifiers &= ~Modifiers.Abstract;
+				}
 				if (typeDecl.ClassType == ClassType.Enum)
 				{
 					Debug.Assert(typeDef.Kind == TypeKind.Enum);
@@ -2006,6 +2012,10 @@ namespace ICSharpCode.Decompiler.CSharp
 				if (method.IsConstructor && settings.RequiredMembers && RemoveCompilerFeatureRequiredAttribute(methodDecl, "RequiredMembers"))
 				{
 					RemoveObsoleteAttribute(methodDecl, "Constructors of types with required members are not supported in this version of your compiler.");
+				}
+				if (method.IsConstructor && settings.ClosedHierarchies)
+				{
+					RemoveCompilerFeatureRequiredAttribute(methodDecl, "ClosedClasses");
 				}
 				return methodDecl;
 
--- a/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/ProjectFileWriterSdkStyle.cs
+++ b/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/ProjectFileWriterSdkStyle.cs
@@ -190,7 +190,10 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 
 		static void WriteProjectInfo(XmlTextWriter xml, IProjectInfoProvider project)
 		{
-			xml.WriteElementString("LangVersion", project.LanguageVersion.ToString().Replace("CSharp", "").Replace('_', '.'));
+			// C# 15 is still in preview; the compiler only accepts -langversion:preview for it.
+			xml.WriteElementString("LangVersion", project.LanguageVersion >= LanguageVersion.CSharp15_0
+				? "preview"
+				: project.LanguageVersion.ToString().Replace("CSharp", "").Replace('_', '.'));
 			xml.WriteElementString("AllowUnsafeBlocks", TrueString);
 			xml.WriteElementString("CheckForOverflowUnderflow", project.CheckForOverflowUnderflow ? TrueString : FalseString);
 
--- a/ICSharpCode.Decompiler.Tests/Helpers/Tester.cs
+++ b/ICSharpCode.Decompiler.Tests/Helpers/Tester.cs
@@ -575,7 +575,16 @@ namespace System.Runtime.CompilerServices
 			{
 				string depSourcePath = Path.GetFullPath(Path.Combine(
 					Path.GetDirectoryName(sourceFileName), match.Groups[1].Value));
-				var depResults = await CompileCSharp(depSourcePath, flags | CompilerOptions.Library).ConfigureAwait(false);
+				// Compile the dependency next to the main output, so that the assembly resolver
+				// finds it when the main assembly is decompiled.
+				string depOutputFileName = null;
+				if (outputFileName != null)
+				{
+					depOutputFileName = Path.Combine(
+						Path.GetDirectoryName(Path.GetFullPath(outputFileName)),
+						Path.GetFileNameWithoutExtension(depSourcePath) + GetSuffix(flags | CompilerOptions.Library) + ".dll");
+				}
+				var depResults = await CompileCSharp(depSourcePath, flags | CompilerOptions.Library, depOutputFileName).ConfigureAwait(false);
 				dependencyAssemblies.Add(Path.GetFullPath(depResults.PathToAssembly));
 			}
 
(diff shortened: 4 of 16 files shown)


ICSharpCode.Decompiler.Tests/TestCases/Pretty/Issue3684.cs
```diff
@@ -15,7 +15,7 @@ public class DerivedClass : BaseClass, IInterface
 		{
 			T IInterface.Convert<T>(T input)
 			{
-				return ((BaseClass)this).Convert<T>(input);
+				return Convert(input);
```
siegfriedpammer, line 18: how is this related to the new language feature? A consequence of the Tester change?

ICSharpCode.Decompiler/CSharp/ProjectDecompiler/ProjectFileWriterSdkStyle.cs
```diff
@@ -190,7 +190,10 @@ static void WriteDesktopExtensions(XmlTextWriter xml, ProjectType projectType)
 
 		static void WriteProjectInfo(XmlTextWriter xml, IProjectInfoProvider project)
 		{
-			xml.WriteElementString("LangVersion", project.LanguageVersion.ToString().Replace("CSharp", "").Replace('_', '.'));
+			// C# 15 is still in preview; the compiler only accepts -langversion:preview for it.
+			xml.WriteElementString("LangVersion", project.LanguageVersion >= LanguageVersion.CSharp15_0
```
siegfriedpammer, line 194: this should use "Preview" instead of  the explicit version. Otherwise we will have to keep this updated. Preview and CSharp15_0 point to the same number as long as C# 15 is in preview.

siegfriedpammer:
Missing:

* CSharpHighlightingTokenWriter
* Type System should provide this info, so we can perform switch expression exhaustiveness checks and remove the default case.
* Test cases that actually make use of the feature in switch expressions, however, this is blocked because we don't support pattern matching in switch at all.

siegfriedpammer:
@dgrunwald given that this PR only implements the declaration part of the feature and switch-on-type/pattern is a bigger chunk, which I definitely wouldn't want vibe-coded without prior careful thought and design (aka "prompt engineering"), should we pause work now, or are you fine with declaration site only, just like we did with extension everything? There is a small difference, since extension method syntax is purely optional, but this effectively changes the meaning of switch.

ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
```diff
@@ -1703,6 +1703,12 @@ EntityDeclaration DoDecompile(ITypeDefinition typeDef, DecompileRun decompileRun
 				{
 					RemoveAttribute(typeDecl, KnownAttribute.Required);
 				}
+				if (settings.ClosedHierarchies && RemoveAttribute(typeDecl, KnownAttribute.Closed))
+				{
+					// closed classes are implicitly abstract
```
dgrunwald, line 1708: If C# closed classes are implicitly abstract then we should only do this change for abstract classes; not for other classes that happen to have the attribute (because someone manually added it).

siegfriedpammer:
```
✕ Do not use 'System.Runtime.CompilerServices.IsClosedTypeAttribute'. This is reserved for compiler usage.[CS8335](https://msdn.microsoft.com/query/roslyn.query?appId=roslyn&k=k(CS8335))
```

siegfriedpammer: but yes, the pattern is: IsClosedTypeAttribute implies abstract

dgrunwald: CS8335 is only a C# compiler error; other languages can use that attribute however they like.

PR #3887, opened 2026-07-19
Model scoped and UnscopedRef lifetimes in the type system

sailro:
Related to #3853. This is the type-system prerequisite requested during review of #3854.

tl;dr: ILSpy currently treats `ScopedRefAttribute` as a single boolean and has no semantic
model for `UnscopedRefAttribute`, implicit scopes, or the module's ref-safety rules. This PR
aligns the type system with Roslyn without changing decompiler output. A separate follow-up
will use this model to recover erased `scoped` local modifiers.

### Problem

`LifetimeAnnotation.ScopedRef` only reports whether a parameter has
`ScopedRefAttribute`. That is sufficient to print the explicit `scoped` keyword, but not to
reason about effective lifetimes:

- `ScopedRefAttribute` represents `ScopedRef` on a by-ref parameter and `ScopedValue` on a
  by-value ref-like parameter.
- `out` is implicitly `ScopedRef` only for modules using
  `[module: RefSafetyRules(11)]`.
- ref-like `params` collections are implicitly `ScopedValue`, while their compiler-emitted
  `ScopedRefAttribute` is redundant and must not be printed.
- `UnscopedRefAttribute` can remove an implicit scope.
- a struct receiver is implicitly `ScopedRef`; `UnscopedRefAttribute` may be placed on the
  method, property, or accessor.

Without these distinctions, a local-lifetime analysis cannot tell whether an argument may be
captured by a call.

### Solution

- Add `ScopedKind` with `None`, `ScopedRef`, and `ScopedValue`.
- Extend `LifetimeAnnotation` with the explicit attribute-encoded scope and
  `HasUnscopedRefAttribute`, while preserving the current `ScopedRef` API.
  `ScopedKind` replaces the obsolete preview-era `RefScoped` and `ValueScoped` fields.
- Recognize `System.Diagnostics.CodeAnalysis.UnscopedRefAttribute` and
  `System.Runtime.CompilerServices.RefSafetyRulesAttribute`.
- Decode `ScopedRefAttribute` according to parameter ref-ness and ref-like type semantics.
  Conflicting or malformed annotations degrade to no explicit scope.
- Read exactly `RefSafetyRules(11)` from the module, matching Roslyn's old/new escape-rule
  selection.
- Add internal helpers for effective parameter scope and implicit struct-receiver scope,
  including method/property/accessor `UnscopedRefAttribute`.

The display path continues to use only the explicit scope. Implicit `out`, `this`, and
ref-like `params` scopes therefore do not introduce redundant `scoped` keywords, and
`UnscopedRefAttribute` remains visible as a normal attribute.

### Tests

- explicit `scoped ref` and `scoped` ref-struct parameters
- implicit new-rules `out`
- `[UnscopedRef] out`
- legacy `out` from a module without `RefSafetyRules(11)`
- struct, class, static, method, property, and accessor receiver scopes
- ref-like `params` effective scope
- `ScopedRef` option disabled, including preservation of the raw attribute
- existing `RefFields` and `ParamsCollections` Pretty matrices remain byte-identical

Validation:

- Debug and Release solution builds: 0 warnings, 0 errors
- full Pretty suite: 1906 passed, 5 existing skips
- full solution test suite: 4887 passed, 15 existing skips

### Review focus

The main review points are the public `LifetimeAnnotation` shape and the exact mapping of
Roslyn's declared versus effective scope rules.

### DIFF
--- a/ICSharpCode.Decompiler/TypeSystem/IParameter.cs
+++ b/ICSharpCode.Decompiler/TypeSystem/IParameter.cs
@@ -18,7 +18,6 @@
 
 #nullable enable
 
-using System;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
 
@@ -52,23 +51,7 @@ namespace ICSharpCode.Decompiler.TypeSystem
 		/// Gets or sets the scope explicitly encoded by <c>ScopedRefAttribute</c>.
 		/// Implicit and redundant scopedness is excluded.
 		/// </summary>
-		public ScopedKind DeclaredScope {
-			get {
-#pragma warning disable 618
-				if (ValueScoped)
-					return ScopedKind.ScopedValue;
-				if (RefScoped)
-					return ScopedKind.ScopedRef;
-#pragma warning restore 618
-				return ScopedKind.None;
-			}
-			set {
-#pragma warning disable 618
-				RefScoped = value != ScopedKind.None;
-				ValueScoped = value == ScopedKind.ScopedValue;
-#pragma warning restore 618
-			}
-		}
+		public ScopedKind DeclaredScope { get; set; }
 
 		/// <summary>
 		/// C# 11 scoped annotation: "scoped ref" (ScopedRefAttribute)
@@ -77,12 +60,6 @@ namespace ICSharpCode.Decompiler.TypeSystem
 			get { return DeclaredScope != ScopedKind.None; }
 			set { DeclaredScope = value ? ScopedKind.ScopedRef : ScopedKind.None; }
 		}
-
-		[Obsolete("Use ScopedRef property instead of directly accessing this field")]
-		public bool RefScoped;
-
-		[Obsolete("C# 11 preview: \"ref scoped\" no longer supported")]
-		public bool ValueScoped;
 
 		/// <summary>
 		/// Gets or sets whether <c>UnscopedRefAttribute</c> is present.
--- a/ICSharpCode.Decompiler/TypeSystem/IParameter.cs
+++ b/ICSharpCode.Decompiler/TypeSystem/IParameter.cs
@@ -36,16 +36,46 @@ namespace ICSharpCode.Decompiler.TypeSystem
 		RefReadOnly,
 	}
 
+	/// <summary>
+	/// Describes which part of a parameter or local is restricted to the current scope.
+	/// </summary>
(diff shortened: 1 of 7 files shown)


siegfriedpammer:
@dgrunwald this needs a design/review session

ICSharpCode.Decompiler/TypeSystem/Implementation/KnownAttributes.cs
```diff
@@ -200,6 +204,9 @@ public static class KnownAttributes
 			new TopLevelTypeName("System.Runtime.CompilerServices", "InlineArrayAttribute"),
 			// C# 14 attributes:
 			new TopLevelTypeName("System.Runtime.CompilerServices", "ExtensionMarkerAttribute"),
+			// Lifetime attributes:
```
siegfriedpammer: nit: normally we would group/sort the attributes by language version (C# 11 in this case)

ICSharpCode.Decompiler/TypeSystem/IParameter.cs
```diff
@@ -36,23 +36,58 @@ public enum ReferenceKind : byte
...
+		/// <summary>
+		/// C# 11 scoped annotation: "scoped ref" (ScopedRefAttribute)
+		/// </summary>
+		public bool ScopedRef {
+			get { return DeclaredScope != ScopedKind.None; }
+			set { DeclaredScope = value ? ScopedKind.ScopedRef : ScopedKind.None; }
 		}
 
 		[Obsolete("Use ScopedRef property instead of directly accessing this field")]
 		public bool RefScoped;
```
siegfriedpammer: this should be removed now

sailro: Removed it, along with the equally obsolete preview-era `ValueScoped` field. `DeclaredScope` is now the single source of truth, while the current `ScopedRef` property remains available.

ICSharpCode.Decompiler/TypeSystem/Implementation/KnownAttributes.cs
sailro: Thanks, moved both new entries into the C# 11 section in the enum and the parallel type-name table.

dgrunwald (approved):

PR #3888, opened 2026-07-19
Fix #3853: infer scoped locals from ref safety

### LINKED ISSUE #3853: Decompiling a `scoped` ref local or ref-struct local drops the modifier (does not recompile)

sailro:
Fixes #3853. Supersedes #3854. Depends on #3887.

tl;dr: `scoped` on a local is erased from IL and PDBs, so type-system support cannot recover
the modifier directly. This PR adds a late ILAst ref-safety analysis that emits `scoped` only
when a later assignment or mutation is strictly narrower than the declaration initializer.

### Problem

A local's `scoped` modifier has no metadata representation. Dropping it produces invalid C#
when a wide initializer is followed by a narrower operation:

- ref reassignment to method-local storage, CS8374
- a ref-struct result that captures a local/ref/out argument, CS8347/CS8352/CS9077
- `stackalloc`, CS8353
- propagation through nested calls or other ref-struct locals
- storing a narrow value into a ref-struct field
- passing a narrow value to a method that may capture it into a writable ref-struct receiver

A shape-only check is not sufficient. It can both miss required modifiers and add redundant
ones when the initializer is already narrow.

### Solution

Add `IntroduceScopedModifierOnLocals`, a late `IILTransform` that mirrors Roslyn's ref-safety
model:

- track separate ref-escape and value-escape contexts
- use the ordered contexts current method, return-only, and calling method
- infer each local's contexts from its declaration initializer
- iterate to a fixpoint for local-to-local propagation
- apply Roslyn's updated and legacy invocation argument filtering
- account for scoped and unscoped parameters, `out`, struct receivers, ref-to-ref-struct
  returns, fields, conditionals, null coalescing, and unknown signatures
- account for later `StLoc` assignments, field stores, and captures into writable instance or
  classic ref extension receivers
- handle bare declarations separately from declarations combined with an initializer
- treat unknown shapes conservatively, so they cannot cause a false `scoped`

The transform records the result on `ILVariable.IsScoped`. `DeclareVariables` only renders the
modifier, following the existing `IntroduceRefReadOnlyModifierOnLocals` pattern.

### Tests

The `Pretty/RefFields` Roslyn 4+ matrix covers:

- ref-local and ref-struct positive cases
- local, ref-parameter, out-parameter, receiver, stackalloc, nested-call, and local-copy
  propagation
- scoped-ref ref-struct value propagation
- field stores and captures into instance/ref-extension receivers
- class, readonly, in/ref-readonly, readonly-ref-struct, and plain-struct receivers that must
  not cause `scoped`
- narrow initializer followed by narrow or wider stores, which must not over-apply

Validation:

- Debug and Release solution builds: 0 warnings, 0 errors
- full Pretty suite: 1906 passed, 5 existing skips
- full solution test suite: 4887 passed, 15 existing skips

### Review focus

The main review area is the transfer logic in `IntroduceScopedModifierOnLocals`, especially
invocation escape and receiver-capture rules.

### DIFF
--- a/ICSharpCode.Decompiler/IL/Transforms/IntroduceScopedModifierOnLocals.cs
+++ b/ICSharpCode.Decompiler/IL/Transforms/IntroduceScopedModifierOnLocals.cs
@@ -1,0 +1,670 @@
+// Copyright (c) 2026 Sebastien Lebreton
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this
+// software and associated documentation files (the "Software"), to deal in the Software
+// without restriction, including without limitation the rights to use, copy, modify, merge,
+// publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons
+// to whom the Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all copies or
+// substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+// INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+// PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+// FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+#nullable enable
+
+using System.Collections.Generic;
+using System.Linq;
+
+using ICSharpCode.Decompiler.IL.Transforms;
+using ICSharpCode.Decompiler.TypeSystem;
+
+namespace ICSharpCode.Decompiler.IL
+{
+	/// <summary>
+	/// Infers the C# <c>scoped</c> modifier on locals whose initializer has a wider
+	/// ref-safety context than a later assignment.
+	/// </summary>
+	/// <remarks>
+	/// The modifier is erased from IL and PDBs, so recovery compares the ref-safe-context
+	/// (for ref locals) or safe-context (for ref-struct locals) of the declaration initializer
+	/// with each later store. The three contexts are ordered from narrowest to widest:
+	/// current method, return-only, calling method. A narrower later store proves that the
+	/// original declaration required <c>scoped</c>.
+	/// </remarks>
+	class IntroduceScopedModifierOnLocals : IILTransform
+	{
+		public void Run(ILFunction function, ILTransformContext context)
+		{
+			if (!context.Settings.ScopedRef)
+				return;
+
+			foreach (var nestedFunction in function.Descendants.OfType<ILFunction>())
+			{
+				if (nestedFunction.Method is not {
+					ParentModule: MetadataModule { UseUpdatedEscapeRules: true }
+				} method)
+				{
+					continue;
+				}
+
+				new Analyzer(nestedFunction, method, context).Run();
+			}
+		}
+
+		enum SafeContext : byte
+		{
+			CurrentMethod,
+			ReturnOnly,
+			CallingMethod,
+		}
+
+		readonly struct LocalScopes(
+			SafeContext? refEscape,
+			SafeContext? valEscape,
+			bool isScoped,
+			bool isScopedWithoutInitializer)
+		{
+			public SafeContext? RefEscape { get; } = refEscape;
+			public SafeContext? ValEscape { get; } = valEscape;
+			public bool IsScoped { get; } = isScoped;
+			public bool IsScopedWithoutInitializer { get; } = isScopedWithoutInitializer;
+		}
+
+		sealed class Analyzer(ILFunction function, IMethod method, ILTransformContext context)
+		{
+			readonly Dictionary<ILVariable, StLoc[]> _stores = [];
+			readonly Dictionary<ILVariable, List<StObj>> _fieldStores = [];
+			readonly Dictionary<ILVariable, List<CallInstruction>> _receiverCalls = [];
+			Dictionary<ILVariable, LocalScopes> _scopes = [];
+
+			public void Run()
+			{
+				foreach (var variable in function.Variables)
+				{
+					variable.IsScoped = false;
+					variable.IsScopedWithoutInitializer = false;
+					if (variable.Kind is not (VariableKind.Local or VariableKind.StackSlot)
+						|| !variable.Type.IsRefLikeOrAllowsRefLikeType()
+						|| variable.UsesInitialValue)
+					{
+						continue;
+					}
+
+					var stores = variable.StoreInstructions.OfType<StLoc>().ToArray();
+					if (stores.Length == 0 || stores.Any(store => store.Parent == null))
+						continue;
+
+					_stores.Add(variable, stores);
+					_scopes.Add(variable, default);
+				}
+
+				foreach (var store in function.Descendants.OfType<StObj>())
+				{
+					if (!TryGetStorageRoot(store.Target, out var variable)
+						|| !_stores.ContainsKey(variable))
+					{
+						continue;
+					}
+
+					if (!_fieldStores.TryGetValue(variable, out var stores))
+					{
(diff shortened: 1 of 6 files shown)


ICSharpCode.Decompiler/IL/Transforms/IntroduceScopedModifierOnLocals.cs
```diff
@@ -0,0 +1,670 @@
...
+					variable.IsScopedWithoutInitializer = false;
+					if (variable.Kind is not (VariableKind.Local or VariableKind.StackSlot)
+						|| !variable.Type.IsRefLikeOrAllowsRefLikeType()
+						|| variable.UsesInitialValue)
+					{
+						continue;
+					}
+
+					var stores = variable.StoreInstructions.OfType<StLoc>().ToArray();
+					if (stores.Length == 0 || stores.Any(store => store.Parent == null))
```
dgrunwald: How is store.Parent==null possible?
Any StLoc part of the main tree (detached old nodes are removed from StoreInstructions) will have a parent. Maybe turn this into an assertion?

ICSharpCode.Decompiler/IL/Transforms/IntroduceScopedModifierOnLocals.cs
```diff
@@ -0,0 +1,670 @@
...
+					}
+					stores.Add(store);
+				}
+
+				foreach (var call in function.Descendants.OfType<CallInstruction>())
+				{
+					bool isRefStructReceiver = call.IsInstanceCall
+						&& call.Method.DeclaringType.IsRefLikeOrAllowsRefLikeType()
+						&& !call.Method.ThisIsRefReadOnly;
+					if (!(isRefStructReceiver || IsRefStructExtensionReceiver(call))
```
dgrunwald: Why is this restricted to `this` parameters of instance/extension methods?
Isn't it also possible to have a non-extension method that updates a ref-param in a way that impacts lifetime analysis?

ICSharpCode.Decompiler/IL/Transforms/IntroduceScopedModifierOnLocals.cs
```diff
@@ -0,0 +1,670 @@
...
+			{
+				if (!call.Method.IsExtensionMethod || call.Arguments.Count == 0)
+					return false;
+
+				IParameter? receiver = call.GetParameter(0);
+				return receiver is { ReferenceKind: ReferenceKind.Ref }
+					&& GetValueType(receiver.Type).IsRefLikeOrAllowsRefLikeType();
+			}
+
+			static bool TryGetStorageRoot(ILInstruction target, out ILVariable variable)
```
dgrunwald: Not sure if this is exhaustive for all kinds of lifetime relevant stuff that can be stored within a struct.
Maybe accesses to inline arrays need to be considered?

ICSharpCode.Decompiler/IL/Transforms/IntroduceScopedModifierOnLocals.cs
```diff
@@ -0,0 +1,670 @@
...
+					continue;
+				}
+
+				new Analyzer(nestedFunction, method, context).Run();
+			}
+		}
+
+		enum SafeContext : byte
+		{
+			CurrentMethod,
```
dgrunwald: Please add comments matching these to the specification terms like `*function-member*`

ICSharpCode.Decompiler/IL/Transforms/IntroduceScopedModifierOnLocals.cs
dgrunwald: It's probably also possible to have stuff like `Identity(ref var).Field = ref ...;`?

ICSharpCode.Decompiler/IL/Transforms/IntroduceScopedModifierOnLocals.cs
sailro: Hi Daniel an thank you for the review. 

Agreed. I’ll replace this runtime skip with a `Debug.Assert` and retain only the empty-store guard.

ICSharpCode.Decompiler/IL/Transforms/IntroduceScopedModifierOnLocals.cs
sailro: You’re right. I reproduced this with `Capture(ref Holder, Span<int>)` : without  `scoped` , Roslyn reports `CS8350`/`CS8352`; with `scoped`, it compiles. I’ll generalize capture tracking to writable ref and out ref-struct arguments, not just instance and extension receivers, and add regression coverage.

ICSharpCode.Decompiler/IL/Transforms/IntroduceScopedModifierOnLocals.cs
sailro: So I checked this against Roslyn. An inline array whose element can carry ref-safety state is currently unsupported (`CS9184` / `CS0306`), while references obtained from ordinary inline arrays are already handled by the  `LdElemaInlineArray` escape cases. I don’t see a currently expressible missing storage case here.

ICSharpCode.Decompiler/IL/Transforms/IntroduceScopedModifierOnLocals.cs
sailro: Got it, I’ll document the mappings explicitly.

ICSharpCode.Decompiler/IL/Transforms/IntroduceScopedModifierOnLocals.cs
sailro: Indeed. `Identity(ref holder).Inner = inner` requires `scoped`, and the current root walker stops at the ref-returning call. I’ll add a regression test and trace ref-return calls back to their compatible contributing ref arguments.

PR #3893, opened 2026-07-19
Fix #3892: preserve unsafe for pointer constructors

### LINKED ISSUE #3892: Field initializer calling a pointer-parameter constructor loses unsafe (CS0214)
### Summary

When a field initializer invokes a constructor that has a pointer parameter, ILSpy can drop the
required `unsafe` modifier if the argument itself does not contain pointer syntax. The decompiled
output then fails to recompile with CS0214.

### Input code

Compiled in Release targeting .NET 8.0 with C# 12 and `AllowUnsafeBlocks` enabled:

```csharp
public unsafe struct NativeBuffer
{
    public NativeBuffer(void* pointer)
    {
    }
}

public static class C
{
    private static readonly unsafe NativeBuffer Empty = new(null);
}
```

### Erroneous output

```csharp
public static class C
{
    private static readonly NativeBuffer Empty = new NativeBuffer(null);
}
```

```text
error CS0214: Pointers and fixed size buffers may only be used in an unsafe context
error CS0214: Pointers and fixed size buffers may only be used in an unsafe context
```

### Expected output

The object creation must remain in an unsafe context, for example:

```csharp
private static readonly unsafe NativeBuffer Empty = new NativeBuffer(null);
```

### Details

* Version in use: master `7554c1b0` (`11.0.0.9223`)
* The repro was verified in an isolated compile, decompile, recompile round trip with the complete reference set.
* `IntroduceUnsafeModifier` checks invocation resolve results for pointer parameters, but object creation expressions do not receive the equivalent check.

sailro:
Fixes #3892.

### Problem

`IntroduceUnsafeModifier` detects pointer parameters on ordinary invocation expressions, but not
on object creation expressions.

If the constructor argument itself contains no pointer syntax, such as `new NativeBuffer(null)`,
the required unsafe context is therefore lost. A field that originally had an `unsafe` modifier
is emitted without one, and the output fails to recompile with CS0214.

### Solution

Apply the existing `HasUnsafeResolveResult` check to `ObjectCreateExpression`, matching the
handling already used for `InvocationExpression`.

This uses the selected constructor's resolved parameter types, so it remains general and does
not depend on a particular struct or argument shape.

### Tests

Added a Pretty regression case where a static field constructs a value through a `void*`
constructor while passing `null`.

Validation:

- `UnsafeCode` Pretty matrix: 20 passed
- Full solution test suite: 4,885 passed, 15 existing skips

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/UnsafeCode.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/UnsafeCode.cs
@@ -24,6 +24,14 @@ using System.Runtime.InteropServices;
 
 namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 {
+	[StructLayout(LayoutKind.Sequential, Size = 1)]
+	internal struct PointerConstructor
+	{
+		public unsafe PointerConstructor(void* pointer)
+		{
+		}
+	}
+
 	internal class SizeofTest
 	{
 		private struct StructWithStaticField
@@ -682,5 +690,10 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 				*ptr = null;
 			}
 		}
+	}
+
+	internal static class UnsafeObjectCreation
+	{
+		private unsafe static readonly PointerConstructor pointerConstructor = new PointerConstructor(null);
 	}
 }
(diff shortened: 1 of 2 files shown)


siegfriedpammer:

PR #3896, opened 2026-07-19
Fix #3895: keep byref-like ToString in string concatenation

### LINKED ISSUE #3895: String concatenation drops Span<char>.ToString() and does not recompile (CS0019)
### Summary

ILSpy removes an explicit `ToString()` call from a `Span<char>` operand when reconstructing
string concatenation. Unlike ordinary value types, a byref-like type cannot be converted to
`object` for the built-in string concatenation operator, so the resulting C# does not compile.

### Input code

Compiled in Release targeting .NET 8.0 with C# 12:

```csharp
using System;

public static class C
{
    public static string M(Span<char> span) =>
        "prefix" + span.ToString();
}
```

### Erroneous output

```csharp
using System;

public static class C
{
    public static string M(Span<char> span)
    {
        return "prefix" + span;
    }
}
```

```text
error CS0019: Operator '+' cannot be applied to operands of type 'string' and 'Span<char>'
```

### Expected output

The explicit call must be retained:

```csharp
return "prefix" + span.ToString();
```

### Details

* Version in use: master `7554c1b0` (`11.0.0.9223`)
* With string-concatenation reconstruction disabled, ILSpy correctly emits `string.Concat("prefix", span.ToString())`.
* `RemoveRedundantToStringInConcat` assumes that the built-in operator can reproduce the removed call, but that is not true for byref-like values.

sailro:
Fixes #3895.

### Problem

String-concatenation reconstruction removes explicit `ToString()` calls when the C# compiler
would normally reproduce an equivalent call through the built-in `+` operator.

That assumption does not hold for byref-like values. `Span<T>` and `ReadOnlySpan<T>` cannot be
converted to `object`, so removing the call from `"prefix" + span.ToString()` produces
`"prefix" + span`, which fails with CS0019.

### Solution

Keep the explicit `ToString()` call when its target is byref-like.

The surrounding `string.Concat` reconstruction still proceeds, so the result remains the
natural and compilable `"prefix" + span.ToString()` form.

### Tests

Added a Roslyn 3+ Pretty fixture covering both `Span<char>` and `ReadOnlySpan<char>`.

Validation:

- `StringConcatenation` Pretty matrix: 6 passed
- Full solution test suite: 4,891 passed, 15 existing skips

### DIFF
--- a/ICSharpCode.Decompiler.Tests/PrettyTestRunner.cs
+++ b/ICSharpCode.Decompiler.Tests/PrettyTestRunner.cs
@@ -339,6 +339,12 @@ namespace ICSharpCode.Decompiler.Tests
 		}
 
 		[Test]
+		public async Task StringConcatenation([ValueSource(nameof(roslyn3OrNewerOptions))] CompilerOptions cscOptions)
+		{
+			await RunForLibrary(cscOptions: cscOptions);
+		}
+
+		[Test]
 		public async Task Async([ValueSource(nameof(defaultOptions))] CompilerOptions cscOptions)
 		{
 			await RunForLibrary(cscOptions: cscOptions);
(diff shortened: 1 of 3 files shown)


siegfriedpammer:

PR #3903, opened 2026-07-22
Provide a WholeProjectDecompiler constructor with no default parameter values

ds5678:
### Problem

I'm trying to trim types from my application, and I'm having issues with the NativeAOT compiler not realizing that my parameters are never null.

https://github.com/icsharpcode/ILSpy/blob/9898adde0d96a123850f0317858dfdcb30c628e6/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs#L130

https://github.com/icsharpcode/ILSpy/blob/9898adde0d96a123850f0317858dfdcb30c628e6/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs#L132

### Solution

* I created an interface `IAssemblyReferenceClassifier` and replaced most uses of `AssemblyReferenceClassifier` with it.
* I moved default initialization from the most specific `WholeProjectDecompiler` constructor to the second most one.
* I added `ArgumentNullException` checks to the most specific `WholeProjectDecompiler` constructor.

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
+++ b/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
@@ -41,6 +41,8 @@ using ICSharpCode.Decompiler.Util;
 
 using static ICSharpCode.Decompiler.Metadata.MetadataExtensions;
 
+#nullable enable
+
 namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 {
 	/// <summary>
@@ -75,7 +77,7 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 
 		public IAssemblyReferenceClassifier AssemblyReferenceClassifier { get; }
 
-		public IDebugInfoProvider DebugInfoProvider { get; }
+		public IDebugInfoProvider? DebugInfoProvider { get; }
 
 		/// <summary>
 		/// The MSBuild ProjectGuid to use for the new project.
@@ -89,17 +91,17 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 		/// This property is set by DecompileProject() and protected so that overridden protected members
 		/// can access it.
 		/// </remarks>
-		public string TargetDirectory { get; protected set; }
+		public string TargetDirectory { get; protected set; } = string.Empty;
 
 		/// <summary>
 		/// Path to the snk file to use for signing.
 		/// <c>null</c> to not sign.
 		/// </summary>
-		public string StrongNameKeyFile { get; set; }
+		public string? StrongNameKeyFile { get; set; }
 
 		public int MaxDegreeOfParallelism { get; set; } = Environment.ProcessorCount;
 
-		public IProgress<DecompilationProgress> ProgressIndicator { get; set; }
+		public IProgress<DecompilationProgress>? ProgressIndicator { get; set; }
 		#endregion
 
 		public WholeProjectDecompiler(IAssemblyResolver assemblyResolver)
@@ -110,9 +112,9 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 		public WholeProjectDecompiler(
 			DecompilerSettings settings,
 			IAssemblyResolver assemblyResolver,
-			IProjectFileWriter projectWriter,
-			IAssemblyReferenceClassifier assemblyReferenceClassifier,
-			IDebugInfoProvider debugInfoProvider)
+			IProjectFileWriter? projectWriter,
+			IAssemblyReferenceClassifier? assemblyReferenceClassifier,
+			IDebugInfoProvider? debugInfoProvider)
 			: this(settings, Guid.NewGuid(), assemblyResolver, projectWriter ?? IProjectFileWriter.FromSettings(settings), assemblyReferenceClassifier ?? new AssemblyReferenceClassifier(), debugInfoProvider)
 		{
 		}
@@ -123,7 +125,7 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 			IAssemblyResolver assemblyResolver,
 			IProjectFileWriter projectWriter,
 			IAssemblyReferenceClassifier assemblyReferenceClassifier,
-			IDebugInfoProvider debugInfoProvider)
+			IDebugInfoProvider? debugInfoProvider)
 		{
 			Settings = settings ?? throw new ArgumentNullException(nameof(settings));
 			ProjectGuid = projectGuid;
@@ -355,7 +357,7 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 		{
 			foreach (var r in module.Resources.Where(r => r.ResourceType == ResourceType.Embedded))
 			{
-				Stream stream = r.TryOpenStream();
+				Stream? stream = r.TryOpenStream();
 				if (stream == null)
 					continue;
 
@@ -373,12 +375,12 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 							foreach (var (name, value) in resourcesFile)
 							{
 								string fileName = SanitizeFileName(name);
-								string dirName = Path.GetDirectoryName(fileName);
+								string? dirName = Path.GetDirectoryName(fileName);
 								if (!string.IsNullOrEmpty(dirName) && directories.Add(dirName))
 								{
 									CreateDirectory(Path.Combine(TargetDirectory, dirName));
 								}
-								Stream entryStream = (Stream)value;
+								Stream entryStream = (Stream)value!;
 								entryStream.Position = 0;
 								individualResources.AddRange(
 									WriteResourceToFile(fileName, name, entryStream));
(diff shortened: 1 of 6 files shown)


ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
siegfriedpammer, line 122: isn't this a compile-time silent breaking change, since now anyone deriving from `WholeProjectDecompiler` using the `protected` ctor, will get exceptions in their face, if they relied on the default initialization of `AssemblyReferenceClassifier` and `projectWriter`?

ds5678: I'm not really sure how nullable annotations work for the repository (because of the reference assembly annotation package), but ideally it would go from maybe-null to not-null, so any callers who provide a maybe-null argument will get a warning at compile time.

siegfriedpammer: the reference annotator package just copies all the annotations from a given reference framework to the target framework of the decompiler (netstandard2.0) at build-time so Roslyn sees the annotated variants when compiling the decompiler project.

The only issue is that we haven't really invested in annotating everything and there's multiple reasons:
- It's a ton of work
- The old AST model used a null object pattern, making NRT somewhat useless
- Most transforms use a dataflow pattern, that's not really NRT friendly: fields are set in Run and are effectively non-null for the entire duration of the transform (and therefore safe), but this cannot be expressed nicely just by sprinkling `?` and some extra attributes.

Given that the new AST model actively uses NRT, it might be worth asking Claude to try to annotate everything.

But as it stands now this is a somewhat ugly breaking change, also third-party users have no way of instantiating the project writer and classifier defaults, afaik...

ds5678: > But as it stands now this is a somewhat ugly breaking change

I tried to make it as minimally breaking as possible, but I suspected that this might be the answer.

> third-party users have no way of instantiating the project writer and classifier defaults, afaik...

The classifier is public, but yes the project writers are internal. Is it intended for them to be an implementation detail?

siegfriedpammer: I think we could make the project writer API public too... git history of the whole feature shows that all the other parts have been made public already. So... are we fine with just public or do we unseal at the same time?

ds5678: > are we fine with just public or do we unseal at the same time?

I'm leaning towards not unsealing because that involves decisions about what the API should be. However, one of my implementations involved a fair bit of copy-pasting. Do you think it would benefit from inheritance?

https://github.com/AssetRipper/AssetRipper/blob/0730de9853007bf255363cb68eadb05e11cf10a3/Source/AssetRipper.Export.PrimaryContent/Scripts/ProjectFileWriter.cs

I have another implementation for `IProjectFileWriter`. Do you think it should be included upstream?

https://github.com/AssetRipper/AssetRipper/blob/0730de9853007bf255363cb68eadb05e11cf10a3/Source/AssetRipper.Export/Scripts/NullProjectFileWriter.cs

siegfriedpammer: `WriteFrameworkInfo` sounds like a strong candidate for an extensibility point...

Ideas:
- `virtual void WriteCustomPropertyGroup(XmlTextWriter xml, ...);` which allows adding one more `PropertyGroup`. And we'd probably need the same a custom `ItemGroup`?
- maybe model properties and items using an object model that is then serialized to XML

I think the template method approach might be sufficient... might be a bit weird, if we want to avoid `<PropertyGroup></PropertyGroup>` if the method is not overridden.

Do you have a better idea perhaps?

ds5678: What are your thoughts on this approach?

https://github.com/ds5678/ILSpy/tree/publicize-project-file-writer-implementations

If you like the direction it's heading, I can upgrade it to a full pull request and rebase this pull request after it's merged.

siegfriedpammer: In general I like the direction, just two things:

1) why use `event` instead of a virtual template method?
2) if we keep the event, we should probably avoid TOCTOU at the use-sites.

siegfriedpammer: Addendum: I just feel like the "multicast" nature of events/delegates, might be a source of bugs here...

ds5678: > 1. why use `event` instead of a virtual template method?

I couldn't think of a clean way to do virtual template methods. I only chose events because they were concise.

> 2. if we keep the event, we should probably avoid TOCTOU at the use-sites.

I added a commit which prevents this.

> Addendum: I just feel like the "multicast" nature of events/delegates, might be a source of bugs here...

I'm not sure exactly what you're referring to, but removing all previously added methods returns the event to null.

https://lab.razor.fyi/#41rAxMUVUJSfXpSYq5dcLDSRqbQ4My9dIbiyuCQ115qLq7gksSQzWSE5J7G4WAGqkKuaS0FBQQEqlVqWmlei4JhckpmfZ6_gao0sV5afmaLgm5iZp6EJFoZoBAFXBW1bBV-IYjjfD5mviyaviyLvnJ9XnJ-TqhdelFmS6pOZl6rhqpBZrJBXmpOjYK-gBKKVFKwUlPLyS8CCSpoQvbWYztPQVKjGFPaDCNdyeTEXleZFcXKceHGg80u6AEsCYwUjAA

https://lab.razor.fyi/#ZY49CsJAEIXxp9CtxBM8UiWIOYAiIpJGVAQLwcolLmEhzsruJiriAbyBvSfQs9krSRQCvmKG-ea9YdizythCq0jznR-a9r2aGEkRlidjxa7PWBhzY_B1sDMDAGO5lSFEKshiFFqpaIigX96lSm4x45JcL8dFMFPKNfYYgMThd9b1imymAJ0B9v7sj8zLpPvnyUnJM1ZkVCz8lZZWTCUJN4A0oCSOMYSTdQc9OKRsDp3vD5e8Fu-7Hs6leV7MFzap6YTWzcbt8b6-olZ9UzlWPg

siegfriedpammer: > I'm not sure exactly what you're referring to, but removing all previously added methods returns the event to null.

Nothing major, I am just wondering what the behavior of two registered handlers would be? Would both modify the same PropertyGroup container? Is that what we expect? Sorry for asking all those questions :sweat_smile:

ds5678: > I am just wondering what the behavior of two registered handlers would be? Would both modify the same PropertyGroup container?  Is that what we expect?

It feels intuitive that an event would create only one custom group, but if multiple groups are preferred, we could easily call `GetInvocationList` (which I didn't know about before now) and document the behavior.

Alternatively, we could go back to your virtual method suggestion. I'm not beholden to any particular approach.

> Sorry for asking all those questions 😅

Don't apologize. I want the best design, too.

Also, I can be a stickler when reviewing pull requests, so I have no room to complain when it happens to me. 😅

siegfriedpammer: hmmm... maybe the `event` solution is not that bad... I mean, template method would be ideal, but then you always get an empty property group element, if it's not overwritten.

Another idea would be going the `protected virtual IEnumerable<(string Name, string Value)> GetCustomProperties()` route... what do you think - not flexible enough?

siegfriedpammer: so we have 4 approaches:

1) delegate field
2) event
3) template method writing to the XmlWriter directly
4) template method returning (string, string) tuples

siegfriedpammer: I think 4 would be best, if it's not too rigid.

ds5678: > template method returning (string, string) tuples

I like this proposal. However, it only handles PropertyGroup. Are we okay with not having an API for a custom ItemGroup?

siegfriedpammer: do you have a use case for that? Feels very specifc and I would like to avoid YAGNI... also then the API would have to be a method returning `IEnumerable<XElement>`?

ds5678: > do you have a use case for that?

I currently have no need for injecting a custom ItemGroup. My only ItemGroup-related need is to change how references are emitted. I put virtual methods in my branch to handle that common use-case.

siegfriedpammer (approved):

siegfriedpammer:

PR #3910, opened 2026-07-24
Fix #3909: emit nullable override disambiguators

### LINKED ISSUE #3909: Generic override loses where T : default and does not recompile
### Summary

ILSpy omits the `where T : default` constraint from a generic override whose signature uses
`T?`. Without that source-only constraint, C# interprets `T?` as `Nullable<T>` instead of a
nullable annotation on an unconstrained type parameter. The emitted override no longer matches its
base member and fails to recompile with CS0115 and CS0453.

### Input code

Compiled in Release targeting .NET 10.0 with nullable annotations enabled:

```csharp
#nullable enable

public abstract class Base
{
    public virtual T? RoundTrip<T>(T? value) => value;
}

public sealed class Derived : Base
{
    public override T? RoundTrip<T>(T? value)
        where T : default => value;
}
```

### Erroneous output

```csharp
public sealed class Derived : Base
{
    public override T? RoundTrip<T>(T? value)
    {
        return value;
    }
}
```

```text
error CS0115: 'Derived.RoundTrip<T>(T?)': no suitable method found to override
error CS0453: The type 'T' must be a non-nullable value type in order to use it
as parameter 'T' in the generic type or method 'Nullable<T>'
```

### Expected output

The override must retain the discriminator required by C#:

```csharp
public override T? RoundTrip<T>(T? value)
    where T : default
{
    return value;
}
```

### Details

* Version in use: master `753f1d24` (`11.0.0.9247-preview1`)
* `where T : default` is legal only on overrides and explicit interface implementations; it distinguishes nullable annotations from `Nullable<T>` when the inherited type parameter has neither a reference-type nor value-type constraint.
* ILSpy currently has no source-emission path for the `default` constraint and normally suppresses type-parameter constraints on overrides.

sailro:
Fixes #3909.

### Problem

C# normally inherits generic constraints on overrides and explicit interface implementations, so
ILSpy correctly avoids restating them. Nullable annotations introduce one exception: when a
method type parameter appears as `T?`, C# requires `class`, `struct`, or `default` to distinguish
a nullable annotation from `Nullable<T>`.

Omitting that discriminator can change the signature and produce CS0115 / CS0453.

### Solution

For overrides and explicit interface implementations:

- collect method type parameters that appear with a nullable annotation anywhere in the return or
  parameter types;
- skip value-type-constrained parameters, where `T?` already means `Nullable<T>`;
- emit `class` for an inherited reference-type constraint, otherwise emit `default`.

Roslyn records the inherited reference/value constraint flags on the implementation method's own
type parameters, so this remains exact even when the base assembly cannot be resolved. Other
inherited constraints remain omitted, as required by C#.

### Tests

Added a Roslyn 3+ Pretty fixture covering:

- unconstrained and `notnull` parameters requiring `default`;
- `class` and `class?` contracts requiring plain `class`;
- `struct` and non-nullable-signature controls;
- return-only and nested-array nullable annotations;
- an explicit interface implementation.

Validation:

- `Issue3909` Pretty matrix: 6 passed
- related generic / nullable Pretty matrices: 26 passed
- full Pretty suite: 1,919 passed, 5 existing skips
- type-system unit tests: 174 passed
- full solution test suite: 4,927 passed, 15 existing skips

### DIFF
--- a/ICSharpCode.Decompiler.Tests/PrettyTestRunner.cs
+++ b/ICSharpCode.Decompiler.Tests/PrettyTestRunner.cs
@@ -664,6 +664,12 @@ namespace ICSharpCode.Decompiler.Tests
 		}
 
 		[Test]
+		public async Task Issue3909([ValueSource(nameof(roslyn3OrNewerOptions))] CompilerOptions cscOptions)
+		{
+			await RunForLibrary(cscOptions: cscOptions | CompilerOptions.NullableEnable);
+		}
+
+		[Test]
 		public async Task Issue3452([ValueSource(nameof(roslyn4OrNewerOptions))] CompilerOptions cscOptions)
 		{
 			await RunForLibrary(cscOptions: cscOptions | CompilerOptions.NullableEnable);
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/Issue3909.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/Issue3909.cs
@@ -1,0 +1,105 @@
+#nullable enable
+namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
+{
+	internal class Issue3909
+	{
+		public abstract class Base
+		{
+			public virtual T? Unconstrained<T>(T? value)
+			{
+				return value;
+			}
+
+			public virtual T? ReferenceType<T>(T? value) where T : class
+			{
+				return value;
+			}
+
+			public virtual T? ReferenceTypeNullable<T>(T? value) where T : class?
+			{
+				return value;
+			}
+
+			public virtual T? NotNull<T>(T? value) where T : notnull
+			{
+				return value;
+			}
+
+			public virtual T? ValueType<T>(T? value) where T : struct
+			{
+				return value;
+			}
+
+			public virtual U? ReturnOnly<U>()
+			{
+				return default(U);
+			}
+
+			public virtual T?[] Nested<T>(T?[] values)
+			{
+				return values;
+			}
+
+			public virtual T Identity<T>(T value)
+			{
+				return value;
+			}
+		}
+
+		public sealed class Derived : Base
+		{
+			public override T? Unconstrained<T>(T? value) where T : default
+			{
+				return value;
+			}
+
+			public override T? ReferenceType<T>(T? value) where T : class
+			{
+				return value;
+			}
+
+			public override T? ReferenceTypeNullable<T>(T? value) where T : class
+			{
+				return value;
+			}
+
+			public override T? NotNull<T>(T? value) where T : default
+			{
+				return value;
+			}
+
+			public override T? ValueType<T>(T? value)
+			{
+				return value;
+			}
+
+			public override U? ReturnOnly<U>() where U : default
+			{
+				return default(U);
+			}
+
+			public override T?[] Nested<T>(T?[] values) where T : default
+			{
+				return values;
+			}
+
+			public override T Identity<T>(T value)
+			{
+				return value;
+			}
+		}
+
+		public interface IRoundTrip
+		{
+			T? RoundTrip<T>(T? value);
+		}
+
+		public class ExplicitImpl : IRoundTrip
+		{
+			T? IRoundTrip.RoundTrip<T>(T? value) where T : default
+			{
+				return value;
+			}
+		}
+	}
+}
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/Issue3909.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/Issue3909.cs
@@ -207,6 +207,32 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 			}
 		}
 
+		public abstract class MethodChainBase
+		{
+			public virtual T? DependentOnClassConstrained<T, U>(T? value, U other) where T : U where U : class
+			{
+				return value;
+			}
+
+			public virtual T? DependentOnClassType<T, U>(T? value, U other) where T : U where U : Node
+			{
+				return value;
+			}
+		}
+
+		public sealed class MethodChainDerived : MethodChainBase
+		{
+			public override T? DependentOnClassConstrained<T, U>(T? value, U other) where T : default
+			{
+				return value;
+			}
+
+			public override T? DependentOnClassType<T, U>(T? value, U other) where T : class
+			{
+				return value;
+			}
+		}
+
 		public class ContainerBase<TOuter> where TOuter : class
 		{
 			public virtual TOuter? Pick<TItem>(TItem item)
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/Issue3909.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/Issue3909.cs
@@ -1,3 +1,6 @@
+using System;
+using System.Collections.Generic;
+
 #nullable enable
 namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 {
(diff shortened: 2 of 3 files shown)


siegfriedpammer:
Which version of c# was this feature introduced? Is there a specification?

sailro:
C# 9.0. There is a specification, though the situation is worth spelling out, because the standard has not caught up with it.

### Version

The compiler is explicit about it. Compiling the fixture with `LangVersion 8.0`:

```text
error CS8400: Feature 'default type parameter constraints' is not available in C# 8.0. Please use language version 9.0 or greater.
error CS8627: A nullable type parameter must be known to be a value type or non-nullable reference type unless language version '9.0' or greater is used.
```

The second one matters here: `T?` on a type parameter that is not constrained to `class` or `struct` is itself a C# 9 feature, so the annotation and the constraint always appear together. C# 8 already allowed restating `class` / `struct` on an override for this purpose; C# 9 only added `default`.

### Specification

The normative text is the csharplang speclet [Unconstrained type parameter annotations](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/unconstrained-type-parameter-annotations.md) (champion issue dotnet/csharplang#3297, design meetings LDM 2019-11-25 and LDM 2020-06-17). Its "`default` constraint" section gives the two rules, both enforced by Roslyn:

* it is an error to use `default` other than on a method override or explicit interface implementation - CS8823
* it is an error to use `default` when the corresponding type parameter of the overridden or interface method is constrained to a reference type or a value type - CS8822 (with CS8665 as the converse guard for `class`: "... is not a reference type")

ECMA-334 does not cover the feature. The published 7th edition (December 2023) standardizes C# 7, and even the committee's C# 9 draft (`dotnet/csharpstandard`, branch `draft-v9`) has not incorporated it yet: in `standard/classes.md`, the 15.2.5 grammar is still

```text
primary_constraint
    : class_type nullable_type_annotation?
    | 'class' nullable_type_annotation?
    | 'struct'
    | 'notnull'
    | 'unmanaged'
    ;
```

and 15.6.1 still reads "Such declarations may only have *type_parameter_constraints_clause*s containing the *primary_constraint*s `class` and `struct`". So for now the speclet and the Roslyn implementation are the only authorities. The user documentation is [where (generic type constraint)](https://learn.microsoft.com/dotnet/csharp/language-reference/keywords/where-generic-type-constraint).

### A defect in this PR, found while re-deriving the rule from the speclet

Please hold off reviewing the current head - answering your question made me check the patch against the speclet rather than against my own test cases, and it is not correct for one class of constraints. I would rather report that myself than have you find it.

Measuring what the compiler actually requires, for a base method `public virtual void M<T>(T? v)` with the given constraint and an override with the given disambiguator:

| base constraint | required on the override |
|---|---|
| none, `notnull`, `Enum`, an interface, `T : U` (U `class`-constrained or unconstrained) | `where T : default` |
| `class`, `class?`, a class **type** (e.g. `Stream`), `Delegate`, `MulticastDelegate`, `T : U` (U : `Stream`) | `where T : class` |
| `struct`, `unmanaged` | nothing |

The patch derives the discriminator from `ITypeParameter.HasReferenceTypeConstraint`, which reflects only the metadata `ReferenceTypeConstraint` flag, that is, the `class` / `class?` keyword. For a class-type constraint it therefore emits `default`, which is CS8822, and the output still does not recompile. The restated `class` leaves no trace to read back either - for a base declaring `where T : Stream`, the override's type parameter is `flags=None, constraints=[Stream]` - so the discriminator has to be derived, not looked up.

ILSpy already has the right predicate: the tri-state `IType.IsReferenceType`, whose `AbstractTypeParameter` implementation carries the Object / ValueType / Enum carve-out and chains through `T : U` via `EffectiveBaseClass`.

This mirrors the wording of the compiler's own diagnostics and reproduces every row of the table above. I will push that shortly, together with fixture cases for a class-type constraint, `Delegate`, `Enum`, an interface constraint and both `T : U` directions, plus a nested-position case (the requirement applies to any occurrence of `T?` in the signature - `T?[]`, `List<T?>`, a tuple element, `ref` / `out`, or the return type). I will also key the collector on type-parameter identity rather than owner kind plus index, so that a `SpecializedMethod` signature carrying a foreign type parameter of the same index cannot produce a spurious clause.

One residual I would leave unless you prefer it hardened: for constraint shapes only IL can express (a concrete delegate, array, struct or enum constraint - CS0701 / CS0706 in C#), `IsReferenceType` can differ from Roslyn's classification.

### Language version

The clause is only emitted where a `T?` annotation on a non-value-type-constrained parameter is already being rendered, which is itself C# 9 and only happens when `NullableReferenceTypes` is enabled. Gating the constraint alone would therefore trade CS8400 for CS8627 / CS0453 and gain nothing. If you would like the setting to report honestly, the coherent change is for `NullableReferenceTypes` to imply C# 9 in `GetMinimumRequiredVersion()` - it maps to C# 8 today, so ILSpy can already emit C# 9-only `T?`. Happy to do that here or in a separate PR, whichever you prefer, and there is no urgency on my side.

sailro:
Pushed the correction.

**What changed.** The discriminator is now derived from the tri-state `ITypeParameter.IsReferenceType` instead of the `HasReferenceTypeConstraint` / `HasValueTypeConstraint` metadata flags.

`AbstractTypeParameter.IsReferenceType` already implements the classification this needs, including the Object / ValueType / Enum carve-out and the chain through a type-parameter constraint via `EffectiveBaseClass`, so the three outcomes line up with the wording of the compiler's own diagnostics ("is not a reference type" in CS8665, "is constrained to a reference type or a value type" in CS8822).

The collector now matches on type-parameter identity rather than owner kind plus index, so a specialized signature carrying a foreign type parameter that shares an index with one of the method's own cannot produce a spurious clause.

**Test coverage.** The `Issue3909` fixture gained a class-type constraint, `Delegate`, `Enum`, an interface constraint, both directions of a `T : TOuter` chain (`TOuter : Node` requiring `class`, `TOuter : class` requiring `default`), a `List<T?>` nested annotation, a two-type-parameter method where only one is annotated, and a control where the annotated parameter belongs to the containing type while the method has its own parameter at the same index.

Checked as a red/green pair: with the previous implementation the fixture fails on exactly the three class-type rows, for every compiler in the matrix:

```text
-  public override T? ClassType<T>(T? value) where T : class
+  public override T? ClassType<T>(T? value) where T : default
-  public override T? DelegateType<T>(T? value) where T : class
+  public override T? DelegateType<T>(T? value) where T : default
-  public override T? Chained<T>(T? value) where T : class
+  public override T? Chained<T>(T? value) where T : default
```

**Validation**

* `Issue3909` Pretty matrix: 6 passed
* full Pretty suite: 1,919 passed, 5 existing skips
* full solution test suite: 4,927 passed, 15 existing skips, 0 failed
* Release solution build: 0 warnings, 0 errors

siegfriedpammer:
Checked how this interacts with C# 13 `allows ref struct` (verified against csc and the compiled metadata):

- `allows ref struct` is inherited implicitly on overrides/EIIs; restating it is CS0460, both alone and as `where T : default, allows ref struct`. The only legal form on a `T?` override is plain `where T : default` — which this PR emits, since such a type parameter has `IsReferenceType == null`. The `class` arm can never collide (`class` + `allows ref struct` is CS9243).
- The correctness is one refactor away from breaking, though: Roslyn re-emits the byreflike flag on the override's *own* type parameter (`Derived::M<byreflike T>` in IL), and the general constraint printer emits `allows ref struct` from exactly that flag. The new path is safe only because it hand-builds the disambiguator constraint instead of going through `ConvertTypeParameterConstraint`.

Could you add a case to the fixture pinning this? E.g. under `#if CS130` (the roslyn3OrNewerOptions set has no net40 configs, so that gating suffices):

```csharp
public class BaseWithAllowsRefStruct
{
	public virtual void Annotated<T>(T? value) where T : allows ref struct
	{
	}

	public virtual void Plain<T>(T value) where T : allows ref struct
	{
	}
}

public class DerivedWithAllowsRefStruct : BaseWithAllowsRefStruct
{
	public override void Annotated<T>(T? value) where T : default
	{
	}

	public override void Plain<T>(T value)
	{
	}
}
```

*This comment was written by an AI agent (Claude Code) operated by @siegfriedpammer.*

sailro:
Thanks, that is a good catch on the refactor hazard, and I reproduced your findings before acting on them.

Verified with csc (.NET 10 SDK, `LangVersion latest`) against a base declaring `public virtual void M<T>(T? v) where T : allows ref struct`:

| override clause | result |
|---|---|
| (none) | CS0115, CS0453 |
| `where T : default` | OK |
| `where T : allows ref struct` | CS0460 (+ CS0115, CS0453) |
| `where T : default, allows ref struct` | CS0460 |
| `where T : class` | CS8665 |
| `where T : class, allows ref struct` | CS0460 (+ CS8665) |

And `class` together with `allows ref struct` on a declaration is CS9243 ("Cannot allow ref structs for a type parameter known from other constraints to be a class"), so the `class` arm indeed cannot collide.

Your metadata point holds too. Reflecting over the compiled assembly, the flag is on the override just as it is on the base:

```text
B.Annotated<T> flags=AllowByRefLike (raw=0x20)
D.Annotated<T> flags=AllowByRefLike (raw=0x20)
```

and `ConvertTypeParameterConstraint` prints `allows ref struct` from exactly that flag (`tp.AllowsRefLikeType`), so routing the disambiguator through it would emit CS0460.

Added your fixture case verbatim, under `#if CS130` as you suggested. I checked that the block is genuinely exercised rather than passing vacuously: temporarily making the new path append `allows ref struct` when the flag is set turns the fixture red on exactly the four Roslyn 4.14 / latest configurations, with the diff

```text
+ public override void Annotated<T>(T? value) where T : default, allows ref struct
```

while the two Roslyn 3.11 configurations stay green because CS130 is not defined there. That is the CS0460 form, so the case pins what you asked it to pin.

I also recorded the invariant next to the code, since the fixture alone does not explain why the clause is hand-built:

```csharp
// The clause is built here rather than through ConvertTypeParameterConstraint, which also
// prints 'allows ref struct' from the byreflike flag. That flag is re-emitted on the
// override's own type parameter as well, and restating it is CS0460.
```

Pushed as a separate commit (`ee2d747`). Validation on the new head: full solution suite 4,927 passed, 15 existing skips, 0 failed; Debug and Release builds 0 warnings, 0 errors; formatting verified with the pinned `dotnet-format` from `BuildTools/pre-commit`.

PR #3911, opened 2026-07-24
Fix #3908: use the async iterator element type for yield

### LINKED ISSUE #3908: Constrained generic async iterator yield is emitted as object (CS0266)
### Summary

An async iterator that yields a type parameter through one of its interface constraints can be
decompiled with an unnecessary cast to `object`. The resulting `yield return` does not match the
iterator element type and fails to recompile with CS0266.

The equivalent synchronous iterator decompiles correctly.

### Input code

Compiled in Release targeting .NET 10.0 with C# 14:

```csharp

public interface IEntry
{
}

public static class C
{
    public static async IAsyncEnumerable<IEntry> CastAsync<TEntry>(
        IAsyncEnumerable<TEntry> entries)
        where TEntry : IEntry
    {
        await foreach (var entry in entries)
            yield return entry;
    }
}
```

### Erroneous output

```csharp
public static async IAsyncEnumerable<IEntry> CastAsync<TEntry>(
    IAsyncEnumerable<TEntry> entries)
    where TEntry : IEntry
{
    await foreach (TEntry entry in entries)
    {
        yield return (object)entry;
    }
}
```

```text
error CS0266: Cannot implicitly convert type 'object' to 'IEntry'.
An explicit conversion exists (are you missing a cast?)
```

### Expected output

The constraint already provides the required implicit conversion:

```csharp
yield return entry;
```

### Details

* Version in use: master `753f1d24` (`11.0.0.9247-preview1`)
* A synchronous `IEnumerable<IEntry>` iterator with the same `TEntry : IEntry` constraint decompiles without the `object` cast.
* The issue appears specific to async-iterator current-value reconstruction.

sailro:
Fixes #3908.

### Problem

`StatementBuilder.VisitYieldReturn` derives the target element type from `IEnumerable<T>` /
`IEnumerator<T>`. For an async iterator, that lookup returns an unknown type even though
`AsyncAwaitDecompiler` has already recovered the element type.

Without the target type, the compiler-generated `box T` used to store a constrained generic value
in the async iterator's current field remains visible as `(object)value`. The resulting
`yield return` fails with CS0266 when the iterator returns an interface constraint.

### Solution

Use `ILFunction.AsyncReturnType` as the element type for async iterators, falling back to the
existing synchronous enumerable lookup.

The existing expression conversion logic can then remove compiler boxing when an implicit
conversion to the recovered element type exists. No async-state-machine-specific cast rewrite is
needed.

### Tests

Extended the `AsyncStreams` Pretty fixture with:

- a constrained generic async iterator returning its interface constraint;
- synchronous and asynchronous object iterators yielding boxed `bool`, `char`, and enum values,
  ensuring their source values remain intact.

Validation:

- `AsyncStreams` Pretty matrix: 6 passed
- runtime-async `AsyncStreams` matrix: 2 passed
- full Pretty suite: 1,913 passed, 5 existing skips
- full solution test suite: 4,921 passed, 15 existing skips

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/AsyncStreams.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/AsyncStreams.cs
@@ -6,6 +6,12 @@ using System.Threading.Tasks;
 
 namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 {
+	public enum AsyncStreamColor
+	{
+		Red,
+		Green
+	}
+
 	public class AsyncStreams
 	{
 		public static async IAsyncEnumerable<int> CountTo(int until)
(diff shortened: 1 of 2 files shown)


siegfriedpammer:

PR #3913, opened 2026-07-26
Emit C# 7.1 default literals

siegfriedpammer:
Implements target-typed `default` output (C# 7.1), building on the conversion-test groundwork from #3912.

**Semantics layer**
- `Conversion.DefaultLiteralConversion` and `DefaultLiteralResolveResult` implement the standard's default literal conversions (10.2.16), resolving the long-standing TODO in `CSharpConversions.ImplicitConversion`. The resolve result is a compile-time constant, like the null literal.
- Unit tests pin overload resolution, operator resolution and type inference behavior for default-literal arguments against compiled csc probes: better-conversion-target tie-breaks, CS0121-style ambiguity, no participation in type inference (CS0411) but riding along when another argument fixes the type parameter, and binary operators with one default operand. No production changes were needed there — the existing resolution stack already matches csc.

**Syntax + emission**
- `DefaultValueExpression.Type` is now optional; without a type the output visitor prints the bare `default` literal.
- New `DecompilerSettings.DefaultLiterals` (C# 7.1, on by default) and `IntroduceDefaultLiterals` transform: `default(T)` becomes `default` in contexts where the target type is explicit in the surrounding syntax and identical to T — declaration/field initializers, simple assignment right-hand sides, and return statements (including async `Task<T>`). An identity conversion is required, because e.g. `object o = default(SomeStruct);` boxes a non-null struct while the bare literal would be null.
- Call arguments follow the implicitly-typed-out-variable pattern: the overload resolution recheck receives untyped `DefaultLiteralResolveResult`s and falls back to the typed form if the expected member is no longer the unique winner. Validated arguments are annotated rather than rewritten, because later transforms can move an argument into an operator operand where the untyped literal would be invalid; `IntroduceDefaultLiterals` honors the annotation only if the expression is still in an argument position. Operator method calls never participate.

**Tests**
- New `DefaultLiteral` pretty fixture covering the positive contexts and the must-stay-typed ones (ambiguous overloads, boxing targets, indexer arguments, user-defined operator operands, `== default` comparisons).
- New `OverloadResolution` pretty fixture pinning the disambiguation syntax CallBuilder emits (numeric/object casts, `(int?)`, params expanded vs. array form, explicit type arguments, typed vs. bare default), plus a CS71 section in the Correctness fixture executing the same shapes so the recompiled output must pick the same overloads at run time.
- Fixture updates across the suite are gated with `#if CS71` so pre-7.1 compiler configurations (legacy csc, mcs, Roslyn 1.3.2 — filtered out on non-Windows platforms) keep compiling the typed form.

This pull request was authored by an AI agent (Claude Code) operated by @siegfriedpammer.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/PrettyTestRunner.cs
+++ b/ICSharpCode.Decompiler.Tests/PrettyTestRunner.cs
@@ -489,6 +489,12 @@ namespace ICSharpCode.Decompiler.Tests
 		}
 
 		[Test]
+		public async Task OverloadResolution([ValueSource(nameof(roslyn2OrNewerOptions))] CompilerOptions cscOptions)
+		{
+			await RunForLibrary(cscOptions: cscOptions);
+		}
+
+		[Test]
 		public async Task PatternMatching([ValueSource(nameof(roslyn2OrNewerOptions))] CompilerOptions cscOptions)
 		{
 			await RunForLibrary(cscOptions: cscOptions);
--- a/ICSharpCode.Decompiler.Tests/PrettyTestRunner.cs
+++ b/ICSharpCode.Decompiler.Tests/PrettyTestRunner.cs
@@ -495,6 +495,12 @@ namespace ICSharpCode.Decompiler.Tests
 		}
 
 		[Test]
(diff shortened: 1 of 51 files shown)


dgrunwald:
Closing in favor of the approach in #4014.

PR #3920, opened 2026-07-27
Fix #1142: recover enum members for out-of-range constant operands

### LINKED ISSUE #1142: Enum type inference could use improvement
C#
`var value = 0xfffffe0b;`

IL
`IL_0000: ldc.i4 -501`

siegfriedpammer:
Fixes #1142.

**Commit 1 -- the decompiler fix.** Addition/subtraction on an enum whose constant operand's numeric value does not fit the underlying type (an `int` constant standing for a high `uint` member: `ldc.i4 -501` for `0xfffffe0b`) failed to resolve as enum arithmetic and fell back to integer arithmetic with casts. The same source expression could even render two different ways in one method, depending on context:

```csharp
// before
if ((uint)((int)myEnum - -501) <= 4u)
{
    Console.WriteLine((uint)myEnum - 4294966795u);
}

// after
if (myEnum - MyEnum.Val1 <= 4)
{
    Console.WriteLine(myEnum - MyEnum.Val1);
}
```

Mechanics: when the plain binary-operator resolution fails its validity check, `HandleBinaryNumeric` retries once with constant operands reinterpreted in the enum type, keeping the adjusted form only if it passes the same check; `AdjustConstantToType` gains an unchecked-cast fallback for enum targets, valid because the reinterpretation is lossless whenever the constant's stack type matches the enum's underlying stack type (the IL constant *is* the member's bit pattern). Deliberately a fallback stage and not a pre-adjustment: an eager variant rewrote valid `data - 1` (enum minus underlying, yielding the enum -- see `EnumDiffNumber` in `TypeAnalysisTests`) into nonsense like `data - StringComparison.CurrentCultureIgnoreCase`. The new fixture methods were verified red before the fix (producing exactly the broken output above).

**Commit 2 -- the IL-view nicety from the issue discussion.** Negative `ldc.i4`/`ldc.i8` operands now carry their hexadecimal form as a comment, `ldc.i4 -501 // 0xfffffe0b` (dotPeek parity). Short forms (`ldc.i4.s`) stay bare -- their range is readable. Because the disassembler round-trip comparer strips comments, the new `Disassembler/Pretty/NegativeConstants` test pins the rendering with explicit content asserts.

Not addressed (separable style question): printing large unsigned constants as hex outside bitwise contexts (`return 4294966795u;` vs `0xFFFFFE0B`) -- the existing `ShouldDisplayAsHex` heuristic is bitwise-only and widening it is a taste decision.

Full decompiler suite green on Linux (2507 passed / 0 failed / 38 skipped).

This PR was prepared by an AI agent (Claude) on behalf of @siegfriedpammer.

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
@@ -1709,6 +1709,26 @@ namespace ICSharpCode.Decompiler.CSharp
 			}
 
 			var rr = resolverWithOverflowCheck.ResolveBinaryOperator(op, left.ResolveResult, right.ResolveResult);
+			if ((rr.IsError || NullableType.GetUnderlyingType(rr.Type).GetStackType() != inst.UnderlyingResultType
+				|| !IsCompatibleWithSign(rr.Type, inst.Sign))
+				&& op is BinaryOperatorType.Add or BinaryOperatorType.Subtract
+				&& (left.Type.Kind == TypeKind.Enum || right.Type.Kind == TypeKind.Enum))
+			{
+				// enum +/- constant did not resolve (e.g. an int constant whose numeric value
+				// does not fit the unsigned underlying type, issue #1142); the IL constant is
+				// the enum's bit pattern, so retry with it reinterpreted in the enum type
+				// before falling back to integer arithmetic with casts.
+				var adjustedLeft = AdjustConstantExpressionToType(left, right.Type);
+				var adjustedRight = AdjustConstantExpressionToType(right, left.Type);
+				var adjustedRR = resolverWithOverflowCheck.ResolveBinaryOperator(op, adjustedLeft.ResolveResult, adjustedRight.ResolveResult);
+				if (!(adjustedRR.IsError || NullableType.GetUnderlyingType(adjustedRR.Type).GetStackType() != inst.UnderlyingResultType
+					|| !IsCompatibleWithSign(adjustedRR.Type, inst.Sign)))
+				{
+					left = adjustedLeft;
+					right = adjustedRight;
+					rr = adjustedRR;
+				}
+			}
 			if (rr.IsError || NullableType.GetUnderlyingType(rr.Type).GetStackType() != inst.UnderlyingResultType
 				|| !IsCompatibleWithSign(rr.Type, inst.Sign))
 			{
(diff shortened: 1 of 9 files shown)


ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -3890,6 +3910,14 @@ private ResolveResult AdjustConstantToType(ResolveResult rr, IType typeHint)
 			else if (typeHint.Kind == TypeKind.Enum || typeHint.IsKnownType(KnownTypeCode.Char) || typeHint.IsCSharpSmallIntegerType())
 			{
 				var castRR = resolver.WithCheckForOverflow(true).ResolveCast(typeHint, rr);
+				if ((!castRR.IsCompileTimeConstant || castRR.IsError) && typeHint.Kind == TypeKind.Enum
```
siegfriedpammer: @dgrunwald I am not entirely sure about this change... wouldn't it make more sense to split enum and char/small int into two?

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -3890,6 +3910,14 @@ private ResolveResult AdjustConstantToType(ResolveResult rr, IType typeHint)
...
+					&& rr.Type.GetStackType() == typeHint.GetEnumUnderlyingType().GetStackType())
+				{
+					// An IL constant is a bit pattern: when the stack types match, reinterpreting
+					// it in the enum's underlying type is lossless even if the numeric value
+					// changes sign (e.g. int -501 for a uint-based enum member 0xfffffe0b).
+					castRR = resolver.WithCheckForOverflow(false).ResolveCast(typeHint, rr);
```
dgrunwald: While sign-changes are lossless; conversions to 16-bit enum types could be lossy!

dgrunwald (changes requested):

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
dgrunwald: I think it makes sense to split the logic: one code path for enums >= 32 bits; another for small enums +other small types. Or maybe all we really need is to compare the type sizes (check overflow only when narrowing).

dgrunwald (approved):

Issue #3714, opened 2026-04-11
Local function issue


CreateAndInject:
See ViewIsoForFab.GetDrawings, search `P_`, you can see there're many things like `P_10.endPointRelation.GetRelatedEndPointOrNull(startPoint)`, it should be `endPointRelation.GetRelatedEndPointOrNull(startPoint)`, and `endPointRelation` should be a closure in the context

CreateAndInject:
I'm sorry, I uploaded incorrect file earlier, see this:
[Arin.RevitCommandForFab.zip](https://github.com/user-attachments/files/26654576/Arin.RevitCommandForFab.zip)

Issue #3202, opened 2024-04-28
ILSpy generated code reference DisplayClass.


CreateAndInject:
[file.zip](https://github.com/icsharpcode/ILSpy/files/15143321/file.zip)

Check DefaultCheckCopyed:
![image](https://github.com/icsharpcode/ILSpy/assets/2529999/e98a2fd3-5010-476b-bf9e-42fb0511259d)

Another error in Class148.GetItem:
![image](https://github.com/icsharpcode/ILSpy/assets/2529999/13bbf33e-ceaf-4baa-9edf-8cdfde9bdac7)

ElektroKill:
Issue in method `Class148.DefaultCheckCopyed` occurs because when TDCU executes it expects the first instruction of the init block of the display class to be the first assignment to the display class. This is not the case for this file for an unknown to me reason (i suspect its a result of obfuscation or deobfuscation though).

Displayclass init block:
```
Block IL_0020 (incoming: 1) {
	if (comp.i4(callvirt get_CopyedOption(ldloc option) == ldc.i4 1)) leave IL_0020 (ldc.i8 0)
	stloc has(ldc.i4 0)
	// Diplay class init start
	stobj System.Int64(delayex.ldflda iid(ldloca I_0), ldc.i8 0)
	// display class init end
	stobj System.Collections.Generic.IEnumerable`1[[System.String]](delayex.ldflda hist(ldloca I_0), nullable.rewrap(call Where(call Select(nullable.unwrap.o(await(call GetByOuterId(ldloc user, ldloc siteId, ldloc oid))), ILFunction <DefaultCheckCopyed>b__29_1[System.Func`2[[Class218],[System.String]]] {
		param _class218_0 : Class218(Index=0, LoadCount=1, AddressCount=0, StoreCount=1)

		BlockContainer {
			Block IL_0000 (incoming: 1) {
				leave IL_0000 (callvirt get_rst_link(ldloc _class218_0))
			}

		}
	}
	), ILFunction <DefaultCheckCopyed>b__29_2[System.Func`2[[System.String],[System.Boolean]]] {
		param _string_1 : System.String(Index=0, LoadCount=1, AddressCount=0, StoreCount=1)

		BlockContainer {
			Block IL_0000 (incoming: 1) {
				leave IL_0000 (comp.i4(call IsNullOrWhiteSpace(ldloc _string_1) == ldc.i4 0))
			}

		}
	}
	)))
[...]

PR #3929, opened 2026-07-28
Fix #2032: avoid propagating mutated SROA fields

### LINKED ISSUE #2032: Follow up on #1981
Currently aggressive TDCU only works correctly, if no local variable is assigned in the aggregate as well as the body of the ILFunction. See `Test9` of the SROA test suite.

There are two options we already have discussed:
`original` and `copy` are `ILVariable`s
1) ensure that `original.LoadCount == 1 && original.IsSingeDefinition && stloc copy(ldloc original)  dominates all other uses of copy`
2) or neither original nor copy is mutated (after initialization), so all uses see the same reference: `original.IsSingleDefinition && copy.IsSingleDefinition && copy.StoreInstructions.Single().MatchLdLoc(original)`

Find a way to make sure either condition holds before and after TDCU or find another applicable condition.

siegfriedpammer:
## Summary
- Keep aggressive SROA from propagating a display-class field to its source variable when the field is assigned again after initialization.
- Re-enable the existing AggressiveScalarReplacementOfAggregates Test9 regression case.

Fixes #2032

## Testing
- OPENSSL_ENABLE_SHA1_SIGNATURES=1 dotnet test --project ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj --filter 'FullyQualifiedName~UglyTestRunner.AggressiveScalarReplacementOfAggregates' --report-trx

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Ugly/AggressiveScalarReplacementOfAggregates.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Ugly/AggressiveScalarReplacementOfAggregates.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Threading;
 
 namespace ICSharpCode.Decompiler.Tests.TestCases.Ugly
 {
@@ -180,15 +181,26 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Ugly
 			Console.WriteLine("{0} {1}", displayClass.field1, displayClass.field2);
 		}
 
-		//		public void Test9()
-		//		{
-		//			DisplayClass displayClass = new DisplayClass {
-		//				thisField = this,
-		//				field1 = 1,
-		//				field2 = "Hello World!"
-		//			};
-		//			displayClass.thisField = new Program();
-		//			Console.WriteLine("{0} {1}", this, displayClass.thisField);
-		//		}
+		public void Test9()
+		{
+			DisplayClass displayClass = new DisplayClass {
+				thisField = this,
+				field1 = 1,
+				field2 = "Hello World!"
+			};
+			displayClass.thisField = new Program();
+			Console.WriteLine("{0} {1}", this, displayClass.thisField);
+		}
+
+		public void Test10()
+		{
+			DisplayClass displayClass = new DisplayClass {
+				thisField = this,
+				field1 = 1,
+				field2 = "Hello World!"
+			};
+			Interlocked.Exchange(ref displayClass.thisField, new Program());
+			Console.WriteLine("{0} {1}", this, displayClass.thisField);
+		}
 	}
 }
--- a/ICSharpCode.Decompiler.Tests/TestCases/Ugly/AggressiveScalarReplacementOfAggregates.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Ugly/AggressiveScalarReplacementOfAggregates.cs
@@ -212,5 +212,38 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Ugly
 			};
 			Console.WriteLine("{0} {1} {2}", displayClass.thisField, displayClass.field1, displayClass.field2);
 		}
+
+		// The one case where mutation and propagation still coexist: the field is initialized
+		// from a parameter that is loaded exactly once, so the store can be redirected to it.
+		public void Test12(Program other)
+		{
+			DisplayClass displayClass = new DisplayClass {
+				thisField = other,
+				field1 = 1,
+				field2 = "Hello World!"
+			};
+			displayClass.thisField = new Program();
+			Console.WriteLine("{0} {1}", this, displayClass.thisField);
+		}
+
+		// The same mutation from inside a lambda: capturing the display class keeps it
+		// materialized, so propagation never arises here.
+		public void Test13(Program other)
+		{
+			DisplayClass displayClass = new DisplayClass {
+				thisField = other,
+				field1 = 1,
+				field2 = "Hello World!"
+			};
+			Invoke(delegate {
+				displayClass.thisField = new Program();
+			});
+			Console.WriteLine("{0} {1}", this, displayClass.thisField);
+		}
+
+		private static void Invoke(Action action)
+		{
+			action();
+		}
 	}
 }
--- a/ICSharpCode.Decompiler.Tests/TestCases/Ugly/AggressiveScalarReplacementOfAggregates.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Ugly/AggressiveScalarReplacementOfAggregates.cs
@@ -202,5 +202,15 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Ugly
 			Interlocked.Exchange(ref displayClass.thisField, new Program());
 			Console.WriteLine("{0} {1}", this, displayClass.thisField);
 		}
+
+		public void Test11()
+		{
+			DisplayClass displayClass = new DisplayClass {
+				thisField = this,
+				field1 = 1,
+				field2 = "Hello World!"
+			};
+			Console.WriteLine("{0} {1} {2}", displayClass.thisField, displayClass.field1, displayClass.field2);
+		}
 	}
 }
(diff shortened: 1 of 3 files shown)


ICSharpCode.Decompiler/IL/Transforms/TransformDisplayClassUsage.cs
```diff
@@ -227,6 +227,19 @@ bool ValidateUse(DisplayClass container, ILInstruction use)
...
+						{
+							// The field is mutated after initialization (stored again, or its address
+							// escapes). Propagation is only sound if the store can be redirected to
+							// the propagated variable: possible for parameters (their remaining uses
+							// are restricted in ResolveVariableToPropagate), but not for 'this' and
+							// not for a variable that is itself a scalar-replaced display class.
+							var propagatedVariable = variable.GetOrDeclare();
+							if (propagatedVariable.IsThis() || displayClasses.ContainsKey(propagatedVariable))
+							{
+								variable.Propagate(null);
```
christophwille, line 240: Cancelling propagation *during* the validation loop extends an existing order sensitivity: consider `dc2.f = dc1;` (dc1 itself a display class, so the field is propagatable) followed by a later mutation of `dc2.f`. If dc1 is validated before dc2, its store-into-field use is accepted via the `StObj` case on the strength of `declaredVar.CanPropagate == true`; dc2's validation then cancels that propagation here, leaving dc1 scalar-replaced while the surviving `dc2.f = dc1` store still references it. The pre-existing post-pass `Propagate(null)` had the same theoretical hazard, but this guard triggers in more situations. Aggressive-SROA-only and needs contrived input, so a follow-up (e.g. re-validating display classes referenced by a cancelled field's initializer, or iterating validation to a fixpoint) seems fine - just flagging it as a conscious decision.

ICSharpCode.Decompiler/IL/Transforms/TransformDisplayClassUsage.cs
```diff
@@ -245,6 +258,23 @@ bool ValidateUse(DisplayClass container, ILInstruction use)
 			}
 		}
 
+		/// <summary>
+		/// True when the given field access is a plain read or one of the recorded
+		/// initializer stores; false for any other store or for an escaping address.
+		/// </summary>
+		static bool IsReadOnlyOrInitializerUse(LdFlda ldflda, VariableToDeclare variable)
```
christophwille: Nit: `IsReadOnlyOrInitializerUse` could be misread as relating to C# `readonly`. Something like `IsPlainReadOrInitializerStore` would be self-describing. Purely cosmetic, feel free to ignore.

ICSharpCode.Decompiler.Tests/TestCases/Ugly/AggressiveScalarReplacementOfAggregates.cs
```diff
@@ -180,15 +181,36 @@ public void Test8b(int i)
...
+		public void Test10()
+		{
+			DisplayClass displayClass = new DisplayClass {
+				thisField = this,
+				field1 = 1,
+				field2 = "Hello World!"
+			};
+			Interlocked.Exchange(ref displayClass.thisField, new Program());
+			Console.WriteLine("{0} {1}", this, displayClass.thisField);
+		}
```
christophwille, line 204: Suggestion: add a fixture for the one propagate-despite-mutation path that still survives - a field initialized from a *non-this parameter* (single load) and then reassigned. That is now the only case where mutation and propagation coexist, and the soundness argument in the new `ValidateUse` comment rests on the `LoadCount == 1` restriction in `ResolveVariableToPropagate`, so it would be good to pin it down. Optionally also a variant where the mutation happens inside a nested lambda (should already be caught since `LoadInstructions` spans nested functions, but it is the classic display-class shape and cheap to lock in).

christophwille (commented):
Reviewed both commits end to end, including a local run of the test suite.

**Summary**

1. The `ValidateUse` guard cancels propagation when a propagatable field is mutated after initialization (second store, or address escape like `Interlocked.Exchange`), but only when the propagation source is `this` or another display class. Non-this parameters keep propagating, which is sound because `ResolveVariableToPropagate` only allows them with `LoadCount == 1`, so the mutation store can be redirected to the parameter.
2. The `VisitStLoc` initializer-inlining change correctly skips initializer stores of still-propagated fields (previously they would re-emit as a self-assignment; for `this` that is invalid C#), and the `insertionIndex` bookkeeping keeps inserted stores contiguous. `context.EndStep(null)` is reachable when all stores are skipped, but `EndStep` is null-safe, so that is fine.

Details I checked: `IsReadOnlyOrInitializerUse` classifies all use shapes conservatively (nested `LdFlda`, `ldflda` as store value, call arguments all count as mutation); `Initializers` is non-empty whenever `CanPropagate` is true (both set together in `AddVariable`); `HandleInitBlock` breaks on duplicate fields so a second store in the init region is treated as mutation; the guard composes symmetrically with the existing post-pass in `AnalyzeFunction` that restricts surviving propagation to parameters and display classes; mutations inside nested lambdas are caught because `LoadInstructions` spans nested functions.

**Verification**

Ran `UglyTestRunner.AggressiveScalarReplacementOfAggregates` on this branch locally: all 4 `UseRoslyn4_14_0` configs pass. The 4 `UseRoslynLatest` configs failed in my environment, but an unrelated control test (`NoNewOfT`) fails identically (`csc` cannot load `System.Runtime.CompilerServices.Unsafe 6.0.3.0`), so that is a local toolset issue with the Roslyn 5.9.0 transport-feed compiler, not this PR.

**Verdict**

Looks good to me. The fix is conservative in the right places and well-commented. Inline comments cover one theoretical validation-order edge case and two optional test suggestions; none are blockers.

*This review was AI-assisted (Claude Code, claude-fable-5), driven and checked by a human.*

ICSharpCode.Decompiler/IL/Transforms/TransformDisplayClassUsage.cs
siegfriedpammer: Tried hard to reproduce this and could not; sharing the evidence so the flag has a disposition.

1. **Targeted construction.** Built the exact scenario the comment describes (dc1 referenced from a recorded initializer store of dc2, later mutation of the field, second store to defeat the stloc rescue path). Tracing shows why it stays safe: the *container* is consistently validated first, so the cancellation lands before the referenced class is validated - it is then simply rejected and stays materialized, output correct.
2. **In-transform oracle over a corpus.** Instrumented the guard to report the exact hazardous sequence (cancelling propagation to a display class that already passed validation) and decompiled System.Private.CoreLib 8.0.29, System.Linq, System.Collections.Immutable, plus the repro assemblies from #3714 and #3202 (the latter obfuscated), under both SROA settings: zero firings across ~250k methods.
3. **Mechanism-guided inversion.** The protective order comes from initializer folding: a display class whose initializer folds into an object-initializer block is detected via a late-appended InitializerTarget temp, so the referenced class is enumerated after its container. A deliberately inverted construction (referenced class unfolded/detected early, container folded/detected late) still did not fire the oracle.

Not proven unreachable - the enumeration order is emergent, not enforced - so agreed on treating a fixpoint/re-validation as the follow-up shape if a trigger ever surfaces. No defensive code added now, deliberately: no failing fixture, nothing to exercise it.

_(Reply written by an AI agent on the author's behalf.)_

ICSharpCode.Decompiler/IL/Transforms/TransformDisplayClassUsage.cs
siegfriedpammer: Renamed to `IsPlainReadOrInitializerStore` in 5a1115019.

_(Reply written by an AI agent on the author's behalf.)_

ICSharpCode.Decompiler.Tests/TestCases/Ugly/AggressiveScalarReplacementOfAggregates.cs
siegfriedpammer: Added in 5a1115019: Test12 pins the surviving parameter path (output shows the store redirected to the single-load parameter), Test13 records the lambda variant - with a comment noting it does not exercise the guard, since capturing the display class keeps it materialized and propagation never arises there. Test13 routes the delegate through a helper because the Ugly runner defines no OPT symbol, so the Debug/Release lambda-inlining split cannot be handled with #if in these fixtures.

_(Reply written by an AI agent on the author's behalf.)_

Issue #829, opened 2017-08-27
newcompiler C# language support status


dgrunwald:
# [C# 2](https://msdn.microsoft.com/en-us/library/7cz8t42e(v=vs.80).aspx) (VS 2005)
- [X] Generics
- [ ] ~~Partial types~~
- [X] Anonymous methods
- [X] Iterators
- [ ] Nullable types
- [X] Getter/setter separate accessibility
- [ ] Method group conversions (delegates)
- [X] Co- and Contra-variance for delegates and interfaces
- [X] Static classes
- [ ] Delegate inference

# [C# 3](https://msdn.microsoft.com/en-us/library/bb308966.aspx) (VS 2008)
- [X] Implicitly typed local variables
- [X] Object and collection initializers
- [X] Auto-Implemented properties
- [X] Anonymous types
- [ ] Extension methods
- [ ] Query expressions
- [X] Lambda expression
- [ ] Expression trees
- [ ] ~~Partial methods~~

# [C# 4](https://msdn.microsoft.com/en-us/magazine/ff796223.aspx) (VS 2010)
- [ ] Dynamic binding
- [ ] Named and optional arguments
- [X] Generic co- and contravariance
- [ ] Embedded interop types ("NoPIA")

# [C# 5](https://blogs.msdn.microsoft.com/mvpawardprogram/2012/03/26/an-introduction-to-new-features-in-c-5-0/) (VS 2012)
- [ ] Asynchronous methods
- [ ] ~~Caller info attributes~~

# [C# 6](https://github.com/dotnet/roslyn/wiki/New-Language-Features-in-C%23-6) (VS 2015)
- [X] Compiler-as-a-service (Roslyn)
- [ ] ~~Import of static type members into namespace~~
- [ ] Exception filters
- [ ] Await in catch/finally blocks
- [X] Auto property initializers
- [X] Default values for getter-only properties
- [ ] Expression-bodied members
- [ ] Null propagator (null-conditional operator, succinct null checking)
- [ ] String interpolation
- [ ] ~~nameof operator~~
- [ ] Dictionary initializer

# [C# 7.0](https://blogs.msdn.microsoft.com/dotnet/2017/03/09/new-features-in-c-7-0/) (Visual Studio 2017)
- [ ] [Out variables](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.0/out-var.md)
[...]

PR #3930, opened 2026-07-29
Decompile C# 14 implicit span conversions (first-class span types)

siegfriedpammer:
Part of #829 (does not close it).

Implements decompilation of C# 14 implicit span conversions ([first-class span types](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-14.0/first-class-span-types.md)) and aligns the resolver with that spec. This PR started as a tests-only PR: the `FirstClassSpanConversions` fixture landed as an `Assert.Ignore`'d desired-output spec. The implementation followed on the same branch; the ignore is gone and that fixture is now the acceptance test.

## What the decompiler now does

The C# 14 compiler lowers implicit span conversions to helper calls - `MemoryExtensions.AsSpan(string)`, `ReadOnlySpan<T>.CastUp<TDerived>`, and the span `op_Implicit` operators - so decompiled code showed the lowered form: `s.AsSpan()`, `ReadOnlySpan<Base>.CastUp(s)`, `(ReadOnlySpan<int>)a`, `(Base[]?)a`. `CallBuilder` now folds these back into conversions, riding the existing conversion machinery:

- the fold always builds an explicit cast carrying a `ConversionResolveResult`;
- consumption sites (arguments, returns, initializers, extension receivers, implicit-`in` arguments) make the cast implicit where the context allows;
- the overload-resolution recheck re-adds a cast when the bare argument would bind to a different overload. A deliberate `str.AsSpan()` overload disambiguation therefore canonicalizes to the equivalent explicit span cast `(ReadOnlySpan<char>)str`.

Gated by the existing `DecompilerSettings.FirstClassSpanTypes` (C# 14.0), so older language targets keep the helper calls.

## Resolver alignment with the spec

Auditing the pre-existing span groundwork against the proposal found three deviations. Each is fixed and pinned by unit tests whose expectations were established by compiling probe programs with the C# 14 compiler:

- **Type inference**: lower-bound inference into a `Span<T>` target now makes an *exact* element inference (`Span<T>` is invariant) - `M<T>(Span<T>, T)` called with `(Span<string>, object)` fails inference (CS0411) instead of unifying to `T = object`.
- **Betterness**: `ReadOnlySpan<E1>` vs `ReadOnlySpan<E2>` is decided by implicit convertibility between the *span types*, not the element types - overloads on `ReadOnlySpan<int>`/`ReadOnlySpan<long>` are ambiguous (CS0121), not resolvable.
- **Explicit span conversion**: newly modeled, together with the spec rule that user-defined conversions are not considered between span-convertible types - `Span<object> s = stringArray;` is CS0266, not an implicit conversion via `op_Implicit(object[])` plus array covariance. Span conversions are also excluded from extension receivers during method-group conversion (CS0123), while invocations keep them.

## Tests

- `FirstClassSpanTypes` (pretty, green from the start): pins the C# 14 overload-resolution winners under span betterness.
- `FirstClassSpanConversions` (pretty, the acceptance test): every conversion shape - string to `ReadOnlySpan<char>` in argument, return, local-initializer and extension-receiver position; span variance via `CastUp`; covariant arrays; implicit-`in` arguments; and a by-value/`in` overload pair where the `in` keyword must survive decompilation to keep binding the same overload.
- 17 resolver unit tests in `ICSharpCode.Decompiler.Tests/Semantics/` (inference, betterness, conversion classification, method-group receivers, and the `in`-binding matrix), each annotated with the Roslyn diagnostic it mirrors.

Full decompiler suite on Linux: 3099 passed / 0 failed / 44 skipped, and a clean `-t:Rebuild` of `ILSpy.XPlat.slnf`.

This PR was prepared by an AI agent (Claude) on behalf of @siegfriedpammer.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/Semantics/OverloadResolutionTests.cs
+++ b/ICSharpCode.Decompiler.Tests/Semantics/OverloadResolutionTests.cs
@@ -365,16 +365,6 @@ namespace ICSharpCode.Decompiler.Tests.Semantics
 			}
 		}
 
-		// The legacy reference mscorlib predates Span<T>, so span tests resolve against a
-		// .NET ref assembly; the test assembly itself provides the operator fixture above.
-		static readonly Lazy<ICompilation> spanCompilation = new Lazy<ICompilation>(
-			delegate {
-				string path = Path.Combine(
-					Helpers.Tester.RefAssembliesToolset.GetPath(".NETCoreApp,Version=v5.0"), "System.Runtime.dll");
-				return new SimpleCompilation(TypeSystemLoaderTests.TestAssembly,
-					new PEFile(path, new FileStream(path, FileMode.Open, FileAccess.Read)));
-			});
-
 		static IMethod MakeMethodIn(ICompilation c, params Type[] parameterTypes)
 		{
 			var m = new FakeMethod(c, SymbolKind.Method);
@@ -393,7 +383,7 @@ namespace ICSharpCode.Decompiler.Tests.Semantics
 			// to ReadOnlySpan<E2> - the span types, not the element types. No span conversion
 			// relates ReadOnlySpan<int> and ReadOnlySpan<long>, so neither target is better
 			// and the call is ambiguous; Roslyn reports CS0121.
-			var c = spanCompilation.Value;
+			var c = RefAssemblyCompilation.Instance;
 			var r = new OverloadResolution(c, new[] {
 				new ResolveResult(c.FindType(typeof(ConvertibleToBothReadOnlySpans)))
 			});
@@ -422,7 +412,7 @@ namespace ICSharpCode.Decompiler.Tests.Semantics
 			// Roslyn: OnlyIn(arr) compiles - a value argument may bind to an 'in' parameter
 			// through the implicit span conversion (a temporary is created). OnlyIn(in arr)
 			// is CS1503: an explicit 'in' argument must have the parameter's own type.
-			var c = spanCompilation.Value;
+			var c = RefAssemblyCompilation.Instance;
 			var arrayArg = new ResolveResult(new ArrayType(c, c.FindType(KnownTypeCode.Int32)));
 
 			var implicitIn = new OverloadResolution(c, new[] { arrayArg });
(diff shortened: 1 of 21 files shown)


ICSharpCode.Decompiler.Tests/Semantics/OverloadResolutionTests.cs
```diff
@@ -347,5 +349,188 @@ public static void Main(string[] args)
...
+
+			public static implicit operator ReadOnlySpan<long>(ConvertibleToBothReadOnlySpans c)
+			{
+				return default;
+			}
+		}
+
+		// The legacy reference mscorlib predates Span<T>, so span tests resolve against a
+		// .NET ref assembly; the test assembly itself provides the operator fixture above.
+		static readonly Lazy<ICompilation> spanCompilation = new Lazy<ICompilation>(
```
dgrunwald: We have this code block in multiple test files now.
Can we deduplicate this code; and make it so that the compilation is loaded only once overall?

ICSharpCode.Decompiler/CSharp/CallBuilder.cs
```diff
@@ -1535,7 +1583,12 @@ private ExpressionWithResolveResult HandleImplicitConversion(IMethod method, Tra
 			var conversions = CSharpConversions.Get(expressionBuilder.compilation);
 			IType targetType = method.ReturnType;
 			var conv = conversions.ImplicitConversion(argument.Type, targetType);
-			if (!(conv.IsUserDefined && conv.IsValid && conv.Method.Equals(method, NormalizeTypeVisitor.TypeErasure)))
+			// An implicit span conversion subsumes the conversion helper the compiler emitted:
+			// the language defines the conversion in terms of exactly these helpers.
+			bool directlyConvertible = conv.IsValid
+				&& (conv.IsImplicitSpanConversion
+					|| (conv.IsUserDefined && conv.Method.Equals(method, NormalizeTypeVisitor.TypeErasure)));
```
dgrunwald: We shouldn't accept implicit span conversions if the caller expects us to check a user-defined conversion.

dgrunwald (changes requested):

dgrunwald (approved):

PR #3933, opened 2026-07-29
Convert float/double += 1 into increment and decrement expressions

siegfriedpammer:
Revives the 2020 `test-cases-fp-types` fixture work: ~770 lines of compound-assignment / increment / decrement coverage for `float`, `double`, and `decimal` in `CompoundAssignmentTest.cs`, mirroring the existing integer sections (fields, properties, params, locals, ref returns, custom-struct receivers).

The revived fixtures exposed an asymmetry: post-increment/decrement on float/double round-tripped as `x++`/`x--`, but the pre forms decompiled to `x += 1f`, because the increment detection in `PrettifyAssignments` only accepted integer constants. Since C# defines `++`/`--` on floating-point types as adding/subtracting exactly 1, a constant-1 float/double operand now also converts. Decimal needed no change (`op_Increment`/`op_Decrement` path).

Verified: `CompoundAssignmentTest` green in all 24 configs, full decompiler suite green on Linux (3082 passed / 0 failed / 44 skipped).

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
@@ -697,6 +697,13 @@ namespace ICSharpCode.Decompiler.CSharp
 				return new CastExpression(ConvertType(type), expr.WithRR(crr))
 					.WithRR(new ConversionResolveResult(type, crr, Conversion.NullLiteralConversion));
 			}
+			else if (type.IsKnownType(KnownTypeCode.Decimal))
+			{
+				expr = new PrimitiveExpression(0m);
+				constantType = type;
+				constantValue = 0m;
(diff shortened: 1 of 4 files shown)


ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -697,6 +697,16 @@ internal ExpressionWithResolveResult GetDefaultValueExpression(IType type)
 				return new CastExpression(ConvertType(type), expr.WithRR(crr))
 					.WithRR(new ConversionResolveResult(type, crr, Conversion.NullLiteralConversion));
 			}
+			else if (type.IsKnownType(KnownTypeCode.Decimal))
+			{
+				// decimal has no IL literal, so a zero-valued decimal reaches the decompiler as a
```
siegfriedpammer: no comment needed here

PR #3936, opened 2026-07-29
Add "Open from running process" dialog

christophwille:
Adds `File > Open from running process...`: a dialog that lists the running .NET processes, shows the assemblies loaded in the selected one, and opens the ones you pick. This is feature parity with dotPeek's "Explore running processes", but built on the runtime's diagnostics endpoint rather than OS module enumeration, so it works on Windows, Linux and macOS alike.

Selection is driven by buttons (**Refresh**, **Add Entry Assembly**, **Add Selected**, **Cancel**), modelled on the "Open from NuGet feed" dialog; there is no context menu.

## Why this matters more with .NET 10, not less

With a modern .NET app the executable you see in the process list is a native apphost. `ILSpy.exe` contains no IL at all - open it and you get "PE file does not contain any managed metadata". The assembly you actually want is `ILSpy.dll` beside it, and neither the process list nor the command line tells you that.

Verified against a real running ILSpy instance from this branch:

```
name=ILSpy kind=CoreClr runtime=10.0.10 arch=x64
commandLine="Z:\...\ILSpy\bin\Debug\net10.0\ILSpy.exe"      <- what the OS shows: no IL in it
entryAssemblyName=ILSpy
entryPath=Z:\...\ILSpy\bin\Debug\net10.0\ILSpy.dll          <- what you actually want to open
modules=118                                                  incl. ILSpy.ReadyToRun.Plugin.dll
```

The indirection gets worse, not better, in the cases users care about: single-file bundles (assemblies packed inside the apphost), plugins loaded at runtime, `Assembly.Load(byte[])`, and custom `AssemblyLoadContext`s. A process explorer is the only thing that answers "which managed assemblies is this process actually running, and where are they?" - and as a bonus it shows which plugins are genuinely loaded, not merely present on disk.

## Options explored

- **Diagnostics IPC (chosen).** Since .NET Core 3.0 every CoreCLR process serves an endpoint - named pipe `dotnet-diagnostic-{pid}` on Windows, unix socket `dotnet-diagnostic-{pid}-*-socket` in `$TMPDIR` elsewhere. Its existence *is* the ".NET process" detector, and `ProcessInfo2` returns the managed entry assembly, command line, architecture and runtime version. Same code path on all three OSes; no debugger attach, no ptrace, no privileges beyond same-user. This is how `dotnet-trace ps` works.
- **EventPipe rundown for the loaded-assembly list (chosen).** A short session with `requestRundown` makes the runtime emit one event per loaded module and assembly, with real paths, from inside the process. Covers dynamic and bundled assemblies that no file-backed listing can see.
- **OS module enumeration - Toolhelp32 / PEB / `Process.Modules` (rejected for CoreCLR).** It answers a different question: its list is dominated by native libraries, and it never sees bundled or in-memory assemblies. It is, however, exactly right for .NET Framework (see below).
- **`Microsoft.Diagnostics.NETCore.Client` (rejected).** It would enumerate processes and start sessions, but it does not parse the resulting events, so `TraceEvent` would follow - a large dependency for a small need. Both the IPC protocol and a scoped nettrace reader are implemented here instead, so **no new package reference**; `Directory.Packages.props` and every `packages.lock.json` are untouched.
- **ClrMD / `Microsoft.Diagnostics.Runtime` (rejected for now).** It could dump the raw PE bytes of an in-memory assembly out of a target - the one case this PR cannot open. Its macOS support has historically been the weakest of the three platforms, so it is the wrong primary mechanism; it remains the natural route if in-memory assemblies are wanted later.
- **`ICorPublish` / `mscoree` for .NET Framework (rejected).** COM interop for a Windows-only legacy path, when the desktop CLR's own module list already yields the answer.

## Implementation overview

- `ILSpy/Processes/` holds all the non-UI logic behind `IProcessExplorer`, so the dialog's view model is driven by a fake in headless tests. Placed in the app rather than `ICSharpCode.ILSpyX` to keep that shipped package's public API unchanged; it is easy to promote later if `ilspycmd` wants it.
- `DiagnosticsIpcMessage` implements the wire format (20-byte header, UTF-16 length-prefixed strings); `DiagnosticsIpcClient` speaks `ProcessInfo2` with a `ProcessInfo` fallback for pre-.NET 6 runtimes, and runs the EventPipe session.
- `NettraceRundownReader` is a deliberately scoped nettrace parser: it walks the FastSerialization object stream far enough to decode loader rundown events and steps over everything else (stacks, sequence points, other providers) by size. It guards the container version and fails with a clear message on anything newer.
- Two findings that only surfaced by running against live processes:
  - The CLR's own providers ship **empty event names** in EventPipe metadata, so events must be matched by provider *plus* numeric id - and the ids collide between providers (152 is `ModuleLoad` for the runtime provider but `DomainModuleDCEnd`, a different payload shape, for the rundown provider).
  - The runtime's **default rundown keyword** (`0x80020139`) also collects the JIT and IL-to-native-map rundown. In a long-running process that buries the module list and overruns the session buffer: an 11.8 MB trace arrived with zero module events in it. Requesting the loader keyword alone via `CollectTracing4` (falling back to `CollectTracing2` on older runtimes) turns the same query into 124 KB with all modules present.
- Menu entry sits between "Open from NuGet feed" and "Reload", enabled on every OS.

## Platform differences

| | Windows | Linux | macOS |
|---|---|---|---|
| Transport | named pipe | unix socket in `$TMPDIR` | unix socket in `$TMPDIR` |
| CoreCLR processes + assemblies | yes | yes | yes |
| .NET Framework processes | yes | n/a | n/a |

The CoreCLR path is genuinely identical everywhere; only the transport differs, and that is isolated in `DiagnosticsPortScanner.Windows.cs` / `.Unix.cs` (partial class, both always compiled, dispatched by `OperatingSystem.IsWindows()` - the existing `ShellHelper` pattern). On macOS the diagnostics socket is not merely convenient but effectively the only option, since native process introspection (`task_for_pid`) requires entitlements or root under SIP.

.NET Framework is Windows-only by nature and is the one place where behaviour differs: those processes predate the diagnostics endpoint, so they are detected by `clr.dll`/`mscorwks.dll` in their OS module list and their assemblies are read from that same list, filtered to files carrying a CLI header. All of it lives in one `[SupportedOSPlatform("windows")]` file.

## Limitations

- **Same-user processes only**, and processes started with `DOTNET_EnableDiagnostics=0` expose no endpoint and are invisible. Both are inherent to the mechanism and affect `dotnet-trace` equally. The dialog states this in a hint line rather than pretending the list is exhaustive.
- **Assemblies that exist only in memory cannot be opened.** `Assembly.Load(byte[])` and dynamic assemblies have no file anywhere; they are listed and their rows say so, but adding them is not offered. Reading the bytes out of the target would need a ClrMD-style route.
- **Elevated processes** are not visible to a non-elevated ILSpy.
- **Containerized processes** may place their socket in a namespace that is not reachable.
- **.NET Framework fidelity is lower**: only assemblies the desktop loader registered are listed, so byte-array loads are missing there too, and processes that deny module access are skipped silently. Those processes also mostly report **NGen native images** (`mscorlib.ni.dll` out of `C:\Windows\assembly\NativeImages_v4.0.30319_64\...`) rather than the IL assemblies they were compiled from: the images carry metadata and are genuinely what the process loaded, so they are listed truthfully, but their IL lives in an assembly the module list does not name, which makes opening one of limited use. Mapping native images back to their source assemblies would be guesswork against the GAC, so the list stays honest about what is loaded.
- A **suspended runtime** (`DOTNET_DefaultDiagnosticPortSuspend`) is listed but its assembly query will hit the timeout rather than answer.
- Collecting the assembly list takes a moment per process; it runs on selection, cancels when the selection moves on, and is bounded by a timeout.

## What could have been simpler

The scope choice worth revisiting is the loaded-assembly list. **Without it**, the feature would be roughly a third of the code: process discovery plus one `ProcessInfo2` call gives the entry assembly, which is the headline use case ("show me the dll behind this exe"), and ILSpy resolves the rest of an app's references from that assembly's directory anyway. The EventPipe session, the nettrace reader, and the two live-process findings above all exist purely to enumerate assemblies.

It earns its place for plugins, dynamically loaded assemblies and single-file bundles - where the loaded set genuinely cannot be derived from the entry assembly - but a first cut without it would have been a legitimate, much smaller PR.

Smaller simplifications also available: dropping .NET Framework support (Windows-only path, one extra file), and dropping the assembly-name resolution that gives in-memory modules a meaningful label instead of a generic one.

## Testing

Full `ILSpy.Tests` suite green, 55 of the tests new.

The protocol layers are tested against the real thing rather than mocked: the test host is itself a .NET process with a live diagnostics port, so the tests self-inspect on whatever OS CI runs - port scan finds the current pid, `ProcessInfo2` reports the test host's managed entry assembly, and a real rundown is collected and parsed. The fixture also emits a dynamic assembly before collecting that rundown, so the "listed but has no file" classification is pinned against a real one. Timeouts are pinned against an endpoint that accepts and then says nothing, and the refused-socket path against a bound-but-not-listening socket planted under a live pid (Linux/macOS, where that transport is used). The view model and dialog are covered headlessly against a fake explorer, and `MainMenuTests` pins the menu item's position, icon and enabled state on every OS.

The .NET Framework path has its own live tests in the OS-gated `ILSpy.Tests.Windows` project, using Windows PowerShell 5.1 as a deterministic desktop-CLR process: detection and description, the merge contract with the diagnostics scan, the managed-only module filter, and `ProcessExplorer`'s routing of a Framework process to the module scan.


ILSpy/Processes/ProcessExplorer.cs
```diff
@@ -0,0 +1,141 @@
...
+			string? processName = TryGetProcessName(pid);
+			if (processName == null)
+				return null; // Exited between the port scan and now.
+			try
+			{
+				var info = await DiagnosticsIpcClient.GetProcessInfoAsync(pid, cancellationToken).ConfigureAwait(false);
+				return new RunningDotNetProcess(pid, processName, RuntimeKind.CoreClr,
+					info.ClrVersion, info.Architecture, info.CommandLine, info.EntryAssemblyName);
+			}
+			catch (Exception ex) when (ex is IOException or TimeoutException or UnauthorizedAccessException)
```
siegfriedpammer: **A single unreachable process aborts the whole listing on Linux/macOS.**

`SocketException` is not an `IOException` (it derives from `Win32Exception`), so a refused unix-socket connect escapes this filter, propagates through the `Task.WhenAll` in `EnumerateProcesses`, and fails the entire enumeration. The view model catch-all then clears the list and shows the raw message, so the user gets an empty process list with "Connection refused" instead of one skipped process - the opposite of what `IProcessExplorer` documents ("Unreachable processes (exited mid-scan, access denied, hung) are silently skipped").

Reproduced: planting a bound-but-not-listening `dotnet-diagnostic-<live-pid>-1234-socket` in `$TMPDIR` and calling `GetProcessesAsync` gives `System.Net.Sockets.SocketException: Connection refused`, `is IOException? False`.

Triggers in normal use: a .NET process exiting between the port scan and the connect (MSBuild workers and VBCSCompiler come and go constantly), a runtime whose diagnostics server is already torn down, or a stale socket file plus PID reuse. Windows is unaffected - a missing pipe surfaces as `TimeoutException`, which is caught.

_(Review comment written by an AI agent.)_

ILSpy/ViewModels/OpenFromProcessDialogViewModel.cs
```diff
@@ -0,0 +1,218 @@
...
+			finally
+			{
+				if (!cts.IsCancellationRequested)
+					IsLoadingProcesses = false;
+			}
+		}
+
+		void ApplyFilter()
+		{
+			Processes.Clear();
```
siegfriedpammer: **Every filter keystroke drops the selection and empties the assemblies pane.**

`ApplyFilter()` rebuilds the bound collection on each `FilterText` change. Clearing it makes the DataGrid drop `SelectedItem`, and since that binding is TwoWay, `null` is written back into `SelectedProcess` - clearing the modules list, the selected modules and the entry-assembly path, **even when the selected process still matches the new filter**.

Confirmed headlessly against the real dialog: select the pid-1 row (1 module listed), then set `FilterText` to a value that still matches it. Result: `Processes=1` (it is still in the list), but `SelectedProcess=(null)` and `Modules=0`.

The existing view-model tests don't catch this because they assert on `vm.Processes` with no grid attached.

_(Review comment written by an AI agent.)_

ILSpy/ViewModels/OpenFromProcessDialogViewModel.cs
```diff
@@ -0,0 +1,218 @@
...
+		async Task LoadModulesAsync(ProcessRowViewModel? process)
+		{
+			modulesCts?.Cancel();
+			var cts = modulesCts = new CancellationTokenSource();
+			int generation = ++modulesGeneration;
+
+			Modules.Clear();
+			SelectedModules.Clear();
+			SetEntryAssemblyPath(null);
+			if (process == null)
```
siegfriedpammer, line 227: **The assemblies progress bar can spin forever.**

This early return happens before the `try`/`finally` below, so the `process == null` path never touches `IsLoadingModules`. The superseded in-flight load's `finally` is generation-gated (`if (generation == modulesGeneration)`), and that check now fails - so nothing resets the flag, and `ModulesLoadingBar` animates indefinitely over an empty pane until another process is selected.

Reached by clicking Refresh while a module query is in flight (`RefreshAsync` sets `SelectedProcess = null`), which is realistic since a query can run up to the 30 s rundown timeout on an unresponsive process. The filter behaviour noted separately reaches it too.

Why the suite stays green: with `FakeProcessExplorer` the parked query is cancelled *inline* inside `CancellationTokenSource.Cancel()`, so the reset happens to land before the generation bump. Substituting an explorer whose completion arrives asynchronously - as the real `Task.Run` + socket-I/O one does - leaves `IsLoadingModules == true`.

_(Review comment written by an AI agent.)_

ILSpy/ViewModels/OpenFromProcessDialogViewModel.cs
```diff
@@ -0,0 +1,218 @@
...
+
+		public OpenFromProcessDialogViewModel(IProcessExplorer explorer)
+		{
+			this.explorer = explorer;
+			SelectedModules.CollectionChanged += (_, _) => AddSelectedModulesCommand.NotifyCanExecuteChanged();
+		}
+
+		partial void OnFilterTextChanged(string value) => ApplyFilter();
+
+		partial void OnSelectedProcessChanged(ProcessRowViewModel? value) => _ = LoadModulesAsync(value);
```
siegfriedpammer: **Fire-and-forget convention.**

`_ = SomethingAsync()` is the pattern `FireAndForgetExtensions.HandleExceptions` exists to replace, and its doc comment names it explicitly. Same on line 96 (`Refresh`). The new menu command in `FileCommands.cs` gets this right with `ShowAsync(owner).HandleExceptions()`.

Concretely: the `finally` blocks here set observable properties, so a throwing `PropertyChanged` subscriber faults a discarded task that is only surfaced later, on the finalizer thread.

_(Review comment written by an AI agent.)_

ILSpy/ViewModels/OpenFromProcessDialogViewModel.cs
```diff
@@ -0,0 +1,218 @@
...
+				allProcesses.AddRange(processes.Select(p => new ProcessRowViewModel(p)));
+				ApplyFilter();
+			}
+			catch (OperationCanceledException)
+			{
+				// Superseded by another refresh, or the dialog closed.
+			}
+			catch (Exception ex)
+			{
+				ErrorMessage = ex.Message;
```
siegfriedpammer, line 163: **Asymmetric error path leaves a stale selection.**

The success path nulls `SelectedProcess` (line 139) before repopulating; this catch path clears `allProcesses`/`Processes` but leaves `SelectedProcess` and `entryAssemblyPath` pointing at a row that is no longer in the list, so `AddEntryAssemblyCommand` stays enabled against a process list that has just been emptied. In the running UI the grid heals this by pushing `null` back - which is the same path that leaves the modules spinner stuck.

_(Review comment written by an AI agent.)_

ILSpy/ViewModels/OpenFromProcessDialogViewModel.cs
```diff
@@ -0,0 +1,218 @@
...
+		{
+			Processes.Clear();
+			foreach (var process in allProcesses.Where(p => p.Matches(FilterText)))
+				Processes.Add(process);
+		}
+
+		async Task LoadModulesAsync(ProcessRowViewModel? process)
+		{
+			modulesCts?.Cancel();
+			var cts = modulesCts = new CancellationTokenSource();
```
siegfriedpammer, line 221: **`CancellationTokenSource` instances are never disposed.**

Each refresh/selection replaces `refreshCts`/`modulesCts` without disposing the predecessor, and `CancelAllOperations` only cancels. No timers are attached so the practical impact is small, but the registrations accumulate for the dialog's lifetime.

_(Review comment written by an AI agent.)_

ILSpy/Processes/DiagnosticsIpcClient.cs
```diff
@@ -0,0 +1,256 @@
...
+
+		const string RuntimeProviderName = "Microsoft-Windows-DotNETRuntime";
+		const ulong LoaderKeyword = 0x8;
+		const uint InformationalLevel = 4;
+		const uint NetTraceFormat = 1;
+		// The session only ever carries loader events plus the rundown, so a small buffer is
+		// ample; oversizing it would make the target runtime reserve memory for nothing.
+		const uint CircularBufferMB = 16;
+
+		static readonly TimeSpan CommandTimeout = TimeSpan.FromSeconds(2);
```
siegfriedpammer: **2 s is the one tight timing constant in the PR** (everything else budgets 30-60 s), and it degrades silently.

When it fires, `ProcessExplorer.DescribeCoreClrProcessAsync` deliberately returns the process with **null** `EntryAssemblyName`/`RuntimeVersion`. `ProcessExplorerTests` asserts on exactly those fields, so on a CPU-starved CI runner the failure reads "expected ILSpy.Tests but found <null>" rather than as a timeout. In `DiagnosticsIpcClientTests` the same timeout surfaces as an unhandled `OperationCanceledException`, since the `GetProcessInfoAsync` fallback only catches `IOException`/`EndOfStreamException`.

_(Review comment written by an AI agent.)_

ILSpy.Tests/Processes/ProcessExplorerTests.cs
```diff
@@ -0,0 +1,147 @@
...
+	public async Task The_Current_Process_Is_Listed_As_A_CoreClr_Process()
+	{
+		using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(60));
+
+		var processes = await Explorer.GetProcessesAsync(cts.Token);
+
+		var self = processes.Should().ContainSingle(p => p.Pid == Environment.ProcessId).Subject;
+		self.Kind.Should().Be(RuntimeKind.CoreClr);
+		self.ProcessName.Should().NotBeNullOrWhiteSpace();
+		self.EntryAssemblyName.Should().Be(Assembly.GetEntryAssembly()!.GetName().Name);
```
siegfriedpammer: These two assertions are the ones that would go red if the 2 s `CommandTimeout` in `DiagnosticsIpcClient` is exceeded under CI load: the timeout is caught deliberately and yields a process with null `EntryAssemblyName`/`RuntimeVersion`, so the failure surfaces here as a confusing null mismatch rather than as a timeout. (See the comment on that constant.)

_(Review comment written by an AI agent.)_

ILSpy/Processes/NetFrameworkProcesses.Windows.cs
```diff
@@ -0,0 +1,123 @@
...
+	/// Finds and inspects .NET Framework processes, which exist on Windows only and predate
+	/// the diagnostics endpoint every CoreCLR process exposes. Both answers come from the OS
+	/// module list: the desktop CLR is present in it as clr.dll (or mscorwks.dll before .NET
+	/// 4), and - unlike CoreCLR - the desktop loader registers the assemblies it loads from
+	/// disk there too, so filtering that list to files carrying a CLI header yields the
+	/// managed set. Assemblies loaded from a byte array have no file anywhere and are
+	/// therefore invisible on this path.
+	/// </summary>
+	[SupportedOSPlatform("windows")]
+	static class NetFrameworkProcesses
```
siegfriedpammer, line 39: **The .NET Framework path has no test coverage at all** - enumeration, the native-module scan, the `IsManagedAssembly` filtering integration, and the `RuntimeKind.NetFramework` branch of `ProcessExplorer.EnumerateModules`.

The repo already has an OS-gated `ILSpy.Tests.Windows` project that could host a live test (or at least unit-level coverage of the filtering).

_(Review comment written by an AI agent.)_

ILSpy/Processes/NettraceRundownReader.cs
```diff
@@ -0,0 +1,435 @@
...
+				}
+
+				// No file behind it: a byte-array load, a dynamic assembly, or a module
+				// whose file the current user cannot see. It can be listed, not opened.
+				assemblyNames.TryGetValue(module.AssemblyId, out string? assemblyName);
+				string name = assemblyName
+					?? NonEmpty(module.IlPath)
+					?? NonEmpty(module.NativePath)
+					?? "(dynamic module)";
+				if (seen.Add("\0" + name))
```
siegfriedpammer, line 415: **The in-memory classification branch is never exercised by a test.**

The live test host has no guaranteed byte-array or dynamic assembly, and there is no synthetic nettrace fixture, so nothing pins this path - yet the dialog's whole "listed but cannot be opened" behaviour depends on it being right.

It does work in practice: against a live ILSpy instance the rundown yields 116 modules with exactly one classified in-memory ("Anonymously Hosted DynamicMethods Assembly"). Worth locking in.

_(Review comment written by an AI agent.)_

siegfriedpammer (commented):
Review of the process-explorer feature, posted as individual comments on each finding. Everything below was verified by running the code on Linux (KDE/Wayland, .NET 10), not by reading alone.

**Overall: strong, well-engineered feature.** The diagnostics-IPC + EventPipe-rundown approach is the right call and it demonstrably works: driving the real explorer against a live ILSpy instance discovered 10 processes, collected and parsed 116 modules in 142 ms, correctly classified the one dynamic assembly as in-memory, and resolved the headline case - native apphost to actual IL - to `ILSpy.dll`. The `CollectTracing4` payload layout is right, the loader-keyword finding in the description is real, and file headers, ASCII-only, en-US and full string localization all match the repo conventions. All 32 process tests passed on three consecutive runs; dialog and menu tests 24/24.

Three defects look worth fixing before merge - one functional (the whole process list fails when any one process is unreachable) and two UX (filter keystrokes drop the selection; the assemblies progress bar can stick). Each has a reproduction in its comment. The rest are test-coverage gaps and nits.

One hypothesis I checked and could **not** substantiate, recorded so nobody re-litigates it: I suspected the rundown's background copy task could fault unobserved when a collection is cancelled, which `GlobalExceptionHandler` would turn into a spurious crash dialog. Six cancellations at varying points produced zero unobserved exceptions - the task ends Canceled, not Faulted. Not an issue.

_(This review was produced by an AI agent on Siegfried's behalf.)_

ILSpy/Processes/NettraceRundownReader.cs
```diff
@@ -0,0 +1,435 @@
...
+			RuntimeProviderName => meta.EventId == RuntimeAssemblyLoad,
+			_ => false,
+		};
+
+		static string ReadUtf16NullTerminated(BinaryReader reader)
+		{
+			var builder = new StringBuilder();
+			while (true)
+			{
+				ushort c = reader.ReadUInt16();
```
siegfriedpammer, line 384: **Unbounded string read inside an event payload.**

`ReadBlock` checks that a payload does not run past its block, but this loop has no bound of its own: on a truncated or unexpected payload layout it reads until it finds a zero word, potentially past the payload and the block, and ends in an `EndOfStreamException` that `ReadModules` does not catch. Hardening only - the payload offsets are correct for the events actually matched.

_(Review comment written by an AI agent.)_

ILSpy.Tests/Processes/NettraceRundownReaderTests.cs
```diff
@@ -0,0 +1,135 @@
...
+		var modules = NettraceRundownReader.ReadModules(Rundown());
+
+		// The native side of the same process is dominated by libraries the decompiler has no
+		// use for - the runtime host itself among them, loaded by every CoreCLR process.
+		// Asking the runtime yields exactly the managed set, which is why the feature reads
+		// the rundown rather than enumerating native modules.
+		File.Exists(NativeRuntimeHost.FullPath).Should().BeTrue(
+			"the runtime host of the current process is the native module to look for");
+
+		modules.Select(m => m.Name).Should().NotContain(
```
siegfriedpammer: **This assertion is a four-prefix blacklist**, so a reader change that starts admitting *other* native modules - hostpolicy, clrjit, libSystem.Native.so, ntdll - passes unnoticed.

The positive property is directly checkable and much stronger: every listed module that has a path satisfies `ProcessExplorer.IsManagedAssembly`.

_(Review comment written by an AI agent.)_

ILSpy/ViewModels/ProcessRowViewModel.cs
```diff
@@ -0,0 +1,79 @@
...
+
+		public int Pid => Process.Pid;
+
+		public string ProcessName => Process.ProcessName;
+
+		public string? Architecture => Process.Architecture;
+
+		public string? EntryAssembly => Process.EntryAssemblyName;
+
+		public string? CommandLine => Process.CommandLine;
```
siegfriedpammer: Nit: `CommandLine` is not bound by anything in `OpenFromProcessDialog.axaml` - dead public surface, unless a column for it is planned.

_(Review comment written by an AI agent.)_

ILSpy/Commands/FileCommands.cs
```diff
@@ -153,6 +154,30 @@ async Task ShowAsync(global::Avalonia.Controls.Window owner)
 		}
 	}
 
+	[ExportMainMenuCommand(ParentMenuID = nameof(Resources._File), Header = nameof(Resources.OpenFrom_RunningProcess), MenuCategory = nameof(Resources.Open), MenuOrder = 1.7)]
```
siegfriedpammer: Nit: this menu entry has no `MenuIcon`, while its immediate siblings (`OpenFrom_GAC`, `OpenFrom_NuGetFeed`) both set one.

_(Review comment written by an AI agent.)_

ILSpy/Processes/DiagnosticsIpcClient.cs
```diff
@@ -0,0 +1,256 @@
...
+		}
+
+		static async Task<DotNetProcessInfo> QueryProcessInfoAsync(int pid, byte commandId, CancellationToken cancellationToken)
+		{
+			using var timeout = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+			timeout.CancelAfter(CommandTimeout);
+
+			Stream stream = await ConnectAsync(pid, timeout.Token).ConfigureAwait(false);
+			await using (stream.ConfigureAwait(false))
+			{
```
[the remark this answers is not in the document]
christophwille: This compiles as written, so there is nothing to fix here.

The extension is `TaskAsyncEnumerableExtensions.ConfigureAwait(IAsyncDisposable, bool)`, which lives in `System.Threading.Tasks` - already imported at the top of this file. `System.Runtime.CompilerServices` is not involved. The branch builds clean on Windows, Linux and macOS.

It is not redundant either: `await using` awaits `DisposeAsync()`, and closing a named pipe or a unix socket has no reason to marshal its continuation back to the captured context. It is consistent with the `.ConfigureAwait(false)` on every other await in the file. No change.

ILSpy/Processes/DiagnosticsIpcMessage.cs
```diff
@@ -0,0 +1,115 @@
...
+		}
+
+		public static string? ReadString(BinaryReader reader)
+		{
+			uint length = reader.ReadUInt32();
+			if (length == 0)
+				return null;
+			if (length > MaxStringLength)
+				throw new IOException($"Diagnostics IPC string length {length} exceeds the sanity limit.");
+			var chars = new char[length];
```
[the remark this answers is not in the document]
christophwille: This compiles as written. An array-creation expression accepts `int`, `uint`, `long` or `ulong` as the length - `new char[someUint]` is legal C# and needs no cast. The branch builds clean on Windows, Linux and macOS.

The bound that actually matters is already two lines up: `if (length > MaxStringLength)` rejects anything over 1 MB before the allocation, so the length can never approach `int.MaxValue`. No change.

ILSpy/Processes/ProcessExplorer.cs
christophwille: Fixed. You are right about the inheritance and about the consequence: the queries run concurrently, so this classification decides between losing one row and losing the listing.

The filter is now a named predicate, `ProcessExplorer.IsUnreachable`, and covers `Win32Exception` - `SocketException`'s base, so both the refused-socket and the access-denied variants land in it. It is documented as the place where "that process did not answer" is told apart from "this code is wrong", because everything not in it must keep propagating.

Red first, twice:

- `Every_Way_An_Endpoint_Can_Fail_To_Answer_Counts_As_Unreachable` - runs everywhere, walks each way an endpoint can fail to answer, and asserts that `InvalidDataException` (a reader defect) is *not* absorbed. It failed on exactly the `SocketException` line.
- `A_Socket_That_Refuses_The_Connection_Costs_One_Row_Not_The_Listing` - your repro, as a test. It starts a live non-.NET child process, plants a bound-but-not-listening socket under its pid in `$TMPDIR`, asserts that `ConnectAsync` really does raise `SocketException` (so the test cannot pass vacuously if the plant fails some other way), and only then asserts that the scan still lists the reachable processes and lists the unreachable one with null metadata. Gated off Windows, so it runs in the Linux and macOS CI jobs.

One related removal: the `catch (OperationCanceledException) when (!cancellationToken.IsCancellationRequested)` arm below is gone, because the client no longer produces that for an expired budget - see the reply on `CommandTimeout`.

ILSpy/ViewModels/OpenFromProcessDialogViewModel.cs
christophwille: Fixed, and reproduced exactly as you described it before the fix.

`ApplyFilter` now brings the bound collection in line in place - remove what the filter no longer admits, insert what is missing at its index - instead of clearing and refilling. A row that still matches is never removed, so the grid never drops it and never writes the null back.

Pinned by `Filtering_Keeps_A_Selected_Process_That_Still_Matches` in `OpenFromProcessDialogStructureTests`, i.e. against the real dialog with the grid attached; your point about why the existing view-model tests cannot see this is the reason it lives there and not next to them. Before the fix: `Processes=1`, `SelectedProcess=null`, `Modules=0`.

Writing it turned up the mirror case, which behaves differently from what one would guess: filtering the selected row *out* did not clear the selection either, because Avalonia's `DataGrid` holds on to a removed item rather than nulling `SelectedItem`. So `ApplyFilter` now drops a selection the filter no longer admits explicitly rather than relying on the grid to do it - otherwise the assembly pane goes on describing a process that is no longer listed. Second test: `Filtering_Out_The_Selected_Process_Clears_Its_Assemblies`.

ILSpy/ViewModels/OpenFromProcessDialogViewModel.cs
christophwille: Fixed: the `process == null` path now sets `IsLoadingModules = false` before returning, with a comment saying why it has to - the query it just cancelled can no longer say so on its way out, because the generation guard below is exactly what stops it.

Your account of why the suite stayed green was precisely right, and it was the interesting part. `FakeProcessExplorer` parked on `ModulesGate.Task.WaitAsync(cancellationToken)`, so `Cancel()` completed the parked query inline on the cancelling thread and the reset landed before the generation bump. The gate now awaits without the token, which is also the truer model of the real explorer - socket I/O finishing on a thread-pool thread, never inline. With that in place, `Refreshing_While_Assemblies_Are_Loading_Stops_The_Progress_Bar` reproduced the stuck indicator before the fix and passes after it.

That test drives the realistic route you named: Refresh while a query is in flight, which nulls the selection.

ILSpy/ViewModels/OpenFromProcessDialogViewModel.cs
christophwille: Fixed - both sites now use `.HandleExceptions()`: `OnSelectedProcessChanged` and `Refresh`. You are right that the menu command already got this right and these did not; there was no reason for the inconsistency.

ILSpy/ViewModels/OpenFromProcessDialogViewModel.cs
christophwille: Fixed. Both paths now go through one `ReplaceProcesses`, which clears the selection before swapping the list - so the asymmetry cannot come back by editing one path and not the other. Its doc comment states the reason: the rows the selection points into are about to be gone, and a selection left behind keeps the entry-assembly button armed against a process nobody can see.

Test `A_Failed_Scan_Leaves_Nothing_Selected_To_Act_On`, red before the fix on exactly that: `SelectedProcess` still set and `AddEntryAssemblyCommand` still enabled over an emptied list.

Your closing observation - that the running UI heals this via the same path that leaves the spinner stuck - is no longer load-bearing either way: the grid no longer has to push anything back, and the spinner path is fixed separately.

ILSpy/ViewModels/OpenFromProcessDialogViewModel.cs
christophwille: Fixed. There is now a `CancelAndDispose(ref CancellationTokenSource?)` used at every replacement and by `CancelAllOperations`, so no source outlives the query it belongs to.

The ordering is deliberate and documented at the helper: cancel first, so the registrations have already fired by the time it is disposed, and the method that owns the source only reads `IsCancellationRequested` after its await - which stays valid on a disposed source, unlike `Token`. Agreed on the impact being small; it is cheap to be correct about.

ILSpy/Processes/DiagnosticsIpcClient.cs
christophwille: Both halves of this are addressed - the tightness and the silence.

**The budget is now 10 s.** The only thing that ever reaches the far end of it is a runtime that is never going to answer: one suspended on a startup diagnostic port, or wedged. A healthy runtime answers a status query in milliseconds, so a tight budget buys nothing and only misreports a loaded machine. The constant carries that reasoning now instead of just a number.

**It no longer degrades silently.** A new `WithBudgetAsync` wraps each command and converts an expired budget into `TimeoutException("Process {pid} did not answer its diagnostics endpoint within N seconds.")`. Because `ProcessExplorer` already classified `TimeoutException` as unreachable, the listing behaviour is unchanged - but the client's own contract now separates "the target went quiet" from "the caller cancelled", which is the distinction whose absence made the `DiagnosticsIpcClientTests` failure unreadable. It also improves what the user sees: the dialog's error bar now names the process instead of showing a bare cancellation. The same wrapper covers the rundown budget.

Two new tests, both cheap:

- `An_Endpoint_That_Accepts_But_Never_Answers_Is_Reported_As_A_Timeout` - a real endpoint (named pipe on Windows, unix socket elsewhere) that accepts the connection and then says nothing, which is how a suspended runtime behaves. It asserts the message names the pid. A new internal overload takes the budget so the test injects 250 ms rather than waiting out the production one.
- `Caller_Cancellation_Is_Not_Disguised_As_A_Timeout` - the other direction, so the conversion cannot swallow a genuine cancellation.

The first was red before the fix, throwing `OperationCanceledException`.

ILSpy.Tests/Processes/ProcessExplorerTests.cs
christophwille: Addressed from both ends.

Here: both assertions now carry `because "the runtime answered the process-info query"`, with a comment above them stating that a null in either field means the query failed or ran out its budget - it is not an "unknown entry assembly", because a runtime always knows its own.

And at the constant: the expired budget is now a `TimeoutException` naming the process rather than a silent null, and the budget is 10 s instead of 2 s. So under CI load this reads as a timeout rather than as a null mismatch. Details in the reply on `CommandTimeout`.

ILSpy/Processes/NetFrameworkProcesses.Windows.cs
christophwille: Agreed, and the gap is closed: `ILSpy.Tests.Windows/Processes/NetFrameworkProcessesTests.cs`, six live tests. The fixture is Windows PowerShell 5.1 - a .NET Framework 4.x application present on every Windows installation including the CI runner, so there is a deterministic desktop-CLR process to inspect rather than whatever happens to be running.

Covered: detection and description (kind, CLR version, architecture from the Framework directory, entry assembly), the `alreadyListed` merge contract with the diagnostics scan, the native-module scan with the managed-only property asserted positively per module, `ProcessExplorer`'s `RuntimeKind.NetFramework` routing end to end through `GetProcessesAsync`/`GetModulesAsync`, and a process that has exited yielding nothing rather than an error.

Writing it turned up something worth having on the record. The desktop loader mostly reports **NGen native images** - `mscorlib.ni.dll` out of `C:\Windows\assembly\NativeImages_v4.0.30319_64\...` - rather than the IL assemblies. They carry metadata, so they are genuinely managed modules and genuinely what the process loaded, and the list is truthful in saying so. But their IL lives in the assembly the image was compiled from, and the module list does not name it, so opening a native image is of limited use. `The_Core_Library_Is_Listed_However_The_Runtime_Loaded_It` pins that behaviour and explains it, and I have added it to the PR description's Limitations as part of the .NET Framework fidelity gap. Mapping native images back to their source assemblies would be guesswork against the GAC, so I have not tried.

ILSpy/Processes/NettraceRundownReader.cs
christophwille: Pinned, and it needed no synthetic fixture after all. The `[OneTimeSetUp]` now emits a `Run`-access dynamic assembly named `ILSpy.Tests.InMemoryRundownFixture` before collecting the rundown, and holds the reference for the fixture's lifetime so the runtime still reports it. `An_Assembly_With_No_File_Is_Listed_But_Marked_Unopenable` then asserts the reader brings it back with `IsInMemory == true` and `Path == null` - the exact pair the dialog's "listed but cannot be opened" behaviour rests on.

Deterministic on every OS, and it rides along on the rundown the fixture already collects, so it costs nothing in run time. Thanks for the note that it does work in practice - matching your 116-module observation against a live ILSpy gave me the confidence that emitting one would show up rather than needing bytes hand-rolled into a fake stream.

ILSpy/Processes/NettraceRundownReader.cs
christophwille: Fixed. `ReadBlock` now hands each event the stream position where its payload ends rather than the payload size - it already computed that bound to validate the payload against the block, so the information was there and simply was not being passed down. `ReadUtf16NullTerminated(reader, payloadEnd)` refuses to read past it and throws `InvalidDataException("A Nettrace event payload ends inside an unterminated string.")`, which is the reader's documented failure mode and what the view model turns into an error message, instead of an `EndOfStreamException` from somewhere further down the stream.

Two tests: `A_Payload_That_Ends_Inside_A_String_Is_Rejected`, and `A_Terminated_String_Inside_Its_Payload_Reads_Back`, which also checks that the terminator is consumed and the bytes after it are left alone - so the bound cannot be "fixed" by stopping one word early.

To be straight about process: this one I wrote fix and test together rather than red first, since a genuine red would have meant threading an unused parameter through first. You called it hardening and I agree - the offsets are right for the events actually matched.

ILSpy.Tests/Processes/NettraceRundownReaderTests.cs
christophwille: Agreed - the blacklist was weaker than the property it was standing in for, and it is now stated as the property: every listed module that has a path must satisfy `ProcessExplorer.IsManagedAssembly`. Asserted per module, so a failure names the offender rather than just saying the collection is wrong.

The runtime host is still asserted absent, but by its concrete full path via `NativeRuntimeHost` instead of a name prefix, which keeps the "excludes native" intent anchored to a real file that demonstrably exists. hostpolicy, clrjit, libSystem.Native and ntdll are now covered by the property rather than by being listed.

The same positive property is asserted on the .NET Framework module list in the new `ILSpy.Tests.Windows` fixture, since that path filters the OS module list and has the same failure mode.

ILSpy/ViewModels/ProcessRowViewModel.cs
christophwille: Removed. `RunningDotNetProcess.CommandLine` stays, since it is what the entry-assembly fallback resolves from when the module list is unavailable, but the row no longer re-exposes it.

No column is planned either: the Entry assembly column already answers the question one would consult a command line for - which of these three `dotnet` processes is the one I want - and it answers it with the name the runtime reports rather than with a path the user has to read.

ILSpy/Commands/FileCommands.cs
christophwille: Fixed. Christoph supplied `ILSpy/Assets/Icons/Process.svg` (16x16, same construction as the other icons in that folder); it is registered as `Images.Process` and wired up as `MenuIcon = "Images/Process"`, so the entry now matches its siblings in the open-from group.

`MainMenuTests` asserts the item carries a non-null icon, so this cannot regress back to blank unnoticed.

christophwille:
Thanks - the review was unusually actionable, and every one of the three defects reproduced exactly as written. All fourteen points are addressed in a follow-up commit; details are in the reply under each comment.

**The three defects**, each red first, then fixed:

- **`SocketException` escaping the unreachable-process filter.** The classification is now a named `ProcessExplorer.IsUnreachable` covering `Win32Exception`, so both transports are handled. Two tests: one that walks every way an endpoint can fail to answer (and checks a reader defect is *not* absorbed), and your own repro - a bound-but-not-listening socket planted under a live non-.NET child pid, gated to Linux/macOS.
- **Filter keystrokes dropping the selection.** `ApplyFilter` now syncs the bound collection in place instead of clearing it, so a row that still matches is never removed and the grid never writes null back. Pinned against the real dialog, since the view model alone cannot see it. This also turned up the mirror case: filtering the selected row *out* did not clear the selection either, because Avalonia's `DataGrid` keeps a removed item - so that is now explicit rather than left to the grid.
- **The assemblies progress bar sticking.** The `process == null` path resets the flag. Your diagnosis of why the suite stayed green was the key: the fake explorer's gate completed inline on cancellation. It now awaits without the token, which is also the truer model of the real explorer, and the new test reproduced the stuck bar before the fix.

**The `CommandTimeout` constant** is 10 s instead of 2 s, and no longer degrades silently: an expired budget becomes `TimeoutException("Process {pid} did not answer...")`. `ProcessExplorer` already classified `TimeoutException` as unreachable, so the listing is unchanged - but the client now distinguishes "the target went quiet" from "the caller cancelled", which is what made that test failure unreadable, and the dialog's error bar now names the process. Pinned by an endpoint that accepts and then says nothing, with the budget injected so the test costs 250 ms.

**Coverage gaps closed.** The in-memory branch is pinned by emitting a real dynamic assembly before the shared rundown is collected. The four-prefix blacklist is replaced by the property itself - every path-backed module must satisfy `IsManagedAssembly`, asserted per module - and that same property now guards the .NET Framework list. The .NET Framework path gets six live tests in `ILSpy.Tests.Windows`, with Windows PowerShell 5.1 as a deterministic desktop-CLR fixture.

**One thing writing those tests taught me**, now in the PR's Limitations: a .NET Framework process mostly reports NGen **native images** (`mscorlib.ni.dll` from the NGen cache), not the IL assemblies. They carry metadata and are genuinely what the process loaded, so the list is truthful - but their IL lives in an assembly the module list does not name, which makes opening one of limited use. Mapping images back to their sources would be guesswork against the GAC, so the list stays honest instead of clever.

Both nits are done: `ProcessRowViewModel.CommandLine` is gone, and the menu entry has an icon (`Images/Process`, from an SVG Christoph supplied), with `MainMenuTests` asserting it is there.

Also noted and not re-litigated: your unobserved-exception hypothesis about the rundown's copy task. Recording that it was checked and found not to be an issue is genuinely useful.

The two Copilot findings are both incorrect - `new char[uint]` is legal C# and the `ConfigureAwait(IAsyncDisposable, bool)` extension is in `System.Threading.Tasks`, which is imported. Answered individually under each.

christophwille:
ai;dr a screenshot (Claude insists on removing it from the main description)

image

Now with ListBox instead:

image

ILSpy/Processes/DiagnosticsIpcClient.cs
```diff
@@ -0,0 +1,286 @@
...
+					// runtime writes the rundown into the same connection, and a full buffer
+					// would block the stop from completing.
+					var trace = new MemoryStream();
+					Task drain = session.CopyToAsync(trace, token);
+					try
+					{
+						await StopTracingAsync(pid, sessionId, token).ConfigureAwait(false);
+						await drain.ConfigureAwait(false);
+					}
+					catch
```
siegfriedpammer, line 203: **The drain task faults unobserved when StopTracing fails with a non-cancellation error, and the app turns that into a spurious crash dialog.**

Last round's recorded check cleared the *cancellation* path: cancelled at any point, the copy task ends Canceled, which never counts as unobserved. This `catch` has a second way in, though: `StopTracingAsync` failing with a real error - most plausibly the target exiting mid-collection, which resets the stop connection. Then `trace` is disposed, the exception rethrows, the `await using` tears down `session` - and `drain` is still parked on a read of it. It faults with `IOException(SocketException 125)`, nobody ever awaits it, and at the next GC `TaskScheduler.UnobservedTaskException` fires, which `GlobalExceptionHandler.OnUnobservedTask` routes into `Report(...)`: a crash dialog, minutes after the dialog's error bar already explained the failure correctly. The `CollectTracing2` fallback retries the same path, so one dying process produces two such faults.

Reproduced deterministically on this branch with a fake endpoint that grants a session id, holds the trace connection open, and resets the stop connection: the client fails with `IOException: Connection reset by peer` (which the error bar shows - correct), and the finalizer then surfaces `AggregateException(IOException(SocketException 125: Operation canceled))` out of `Stream.CopyToAsync`.

Fix verified against the same repro (the unobserved exception is gone; the primary `IOException` still propagates unchanged): in this `catch`, dispose `session` first - that is what unblocks the drain - then await `drain` swallowing its failure, then dispose `trace`. The later `await using` disposal of `session` is idempotent.

```csharp
catch
{
	await session.DisposeAsync().ConfigureAwait(false);
	try
	{
		await drain.ConfigureAwait(false);
	}
	catch
	{
		// The primary failure is propagating; the drain ended on the torn-down
		// connection and its own failure is only a symptom of that.
	}
	await trace.DisposeAsync().ConfigureAwait(false);
	throw;
}
```

<details>
<summary>Repro test (self-contained, Linux/macOS; asserts the unobserved exception fires, so it goes red once the fix is in)</summary>

```csharp
[TestFixture]
[Platform(Exclude = "Win")]
public class UnobservedDrainReproTests
{
	const int FakePid = 0x7FFE_0000;

	[Test]
	public async Task Drain_Task_Faults_Unobserved_When_Stop_Fails_NonCancellationally()
	{
		string socketPath = Path.Combine(Path.GetTempPath(), $"dotnet-diagnostic-{FakePid}-1-socket");
		File.Delete(socketPath);
		using var listener = new Socket(AddressFamily.Unix, SocketType.Stream, ProtocolType.Unspecified);
		listener.Bind(new UnixDomainSocketEndPoint(socketPath));
		listener.Listen(4);

		var unobserved = new TaskCompletionSource<Exception>(TaskCreationOptions.RunContinuationsAsynchronously);
		EventHandler<UnobservedTaskExceptionEventArgs> handler = (_, e) => unobserved.TrySetResult(e.Exception!);
		TaskScheduler.UnobservedTaskException += handler;
		try
		{
			var serverTask = RunServerAsync(listener);

			Exception? caught = null;
			try
			{
				using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(30));
				using var trace = await DiagnosticsIpcClient.CollectModuleRundownAsync(FakePid, cts.Token);
			}
			catch (Exception ex)
			{
				caught = ex;
			}
			caught.Should().NotBeNull("the fake stop connection dies, so the collection must fail");
			caught.Should().NotBeOfType<OperationCanceledException>(
				"the repro needs the non-cancellation failure path");

			// Give the drain task time to fault, then finalize the abandoned task.
			await Task.Delay(500);
			GC.Collect();
			GC.WaitForPendingFinalizers();
			GC.Collect();

			var fired = await Task.WhenAny(unobserved.Task, Task.Delay(3000));
			fired.Should().Be(unobserved.Task, "the drain task faulted with nobody awaiting it");
		}
		finally
		{
			TaskScheduler.UnobservedTaskException -= handler;
			File.Delete(socketPath);
		}
	}

	static async Task RunServerAsync(Socket listener)
	{
		// Connections alternate: collect (odd) -> success + session id, held open;
		// stop (even) -> closed without an answer.
		for (int i = 1; i <= 4; i++)
		{
			Socket client;
			try
			{
				client = await listener.AcceptAsync();
			}
			catch (Exception)
			{
				return;
			}
			if (i % 2 == 1)
				_ = HandleCollectAsync(client);
			else
				client.Dispose();
		}
	}

	static async Task HandleCollectAsync(Socket client)
	{
		try
		{
			var stream = new NetworkStream(client, ownsSocket: false);
			var buffer = new byte[512];
#pragma warning disable CA2022 // one partial read is fine; the request content is irrelevant
			await stream.ReadAsync(buffer);
#pragma warning restore CA2022

			// Success response: magic, size 28, command set 0xFF, id 0x00, reserved, 8-byte session id.
			var response = new byte[28];
			System.Text.Encoding.ASCII.GetBytes("DOTNET_IPC_V1\0", response);
			BinaryPrimitives.WriteUInt16LittleEndian(response.AsSpan(14), 28);
			response[16] = 0xFF;
			response[17] = 0x00;
			BinaryPrimitives.WriteUInt64LittleEndian(response.AsSpan(20), 0xDEADBEEF);
			await stream.WriteAsync(response);
			// Held open: the client's drain parks on a read that only ends when the
			// client tears the connection down.
		}
		catch (Exception)
		{
			// The client tearing the connection down is expected.
		}
	}
}
```

</details>

_(Review comment written by an AI agent.)_

siegfriedpammer (commented):
Follow-up review of f574c59d0, verified by checking out the PR head and running it on Linux (KDE, .NET 10) - not by reading alone.

**All 14 points from the previous round are fixed as described in the replies.** Each claimed mechanism is present and does what the reply says: `ProcessExplorer.IsUnreachable` covers `Win32Exception` (and the refused-socket test pins your repro end to end), `ApplyFilter` syncs the bound collection in place and explicitly drops a filtered-out selection, the `process == null` path resets `IsLoadingModules`, both fire-and-forget sites use `HandleExceptions()`, `ReplaceProcesses` unifies the success and failure paths, `CancelAndDispose` covers every replacement site, the 10 s budget expires into a `TimeoutException` naming the pid, the in-memory branch is pinned by a real dynamic assembly, the managed-only assertion is the positive property per module, the .NET Framework path has six live tests, `CommandLine` is gone from the row VM, and the menu icon is asserted in `MainMenuTests`. Full `ILSpy.Tests` suite on this branch: **1166 passed, 0 failed, 3 skipped**. The `ILSpy.Tests.Windows` fixtures could not be run here (Windows-only by design) and were reviewed by reading; Windows PowerShell 5.1 as a deterministic desktop-CLR fixture is a sound choice.

**One new defect, with a repro and a verified fix** - see the inline comment on `DiagnosticsIpcClient.cs`. The rundown's drain task faults unobserved when `StopTracingAsync` fails with a *non-cancellation* error, and `GlobalExceptionHandler.OnUnobservedTask` turns that into a spurious crash dialog at the next GC. This is the sibling of the hypothesis recorded last round: the cancellation path is indeed clean (the task ends Canceled, six runs out of six), but the non-cancellation branch of the same `catch` is not, and it reproduces deterministically.

**Checked and fine**, recorded so nobody re-litigates:

- `ApplyFilter`'s in-place merge is correct: after the removal pass, `Processes` is always a subsequence of `matching` (both preserve `allProcesses` order, and a refresh replaces every row object so stale rows are always evicted), so inserting at the first mismatch index is sound.
- `HungDiagnosticsEndpoint`'s discarded accept tasks cannot fault unobserved in the current tests: the client connects in every test that uses the endpoint, so `WaitForConnectionAsync`/`AcceptAsync` complete successfully before `Dispose` runs.
- `CancelAndDispose`'s dispose-straight-after-cancel is safe as its comment claims: `IsCancellationRequested` is the only member read after disposal, and it stays valid on a disposed source.

_(This review was produced by an AI agent on Siegfried's behalf.)_

ILSpy/Processes/DiagnosticsIpcClient.cs
christophwille: Confirmed, and fixed in a7920b1. The analysis is right on every step: the non-cancellation path through this `catch` never touches `drain`, `drain` is parked on a read of `session`, the `await using` on the way out is what fails that read, and `GlobalExceptionHandler.OnUnobservedTask` turns the finalizer's report into `Report(...)`. Two of them per dying target, because the `CollectTracing2` fallback walks the same path.

Worth adding to the account: on that path `trace` is disposed *while the drain is still copying into it*, so the drain can equally well fault on a write to a disposed `MemoryStream`. The ordering in the suggested fix removes that too, which is why I took it as given.

What landed is the same three steps, in a named method so the ordering constraint has somewhere to live, and with each `await` on the cleanup path unable to skip the one after it (a throwing `DisposeAsync` must not be what costs us the `await drain`):

```csharp
internal static async Task AbandonCollectionAsync(Stream session, Task drain, MemoryStream trace)
{
    await SuppressFailureAsync(session.DisposeAsync().AsTask()).ConfigureAwait(false);
    await SuppressFailureAsync(drain).ConfigureAwait(false);
    await trace.DisposeAsync().ConfigureAwait(false);
}
```

On the test, I went a different way than the suggested repro, and it is worth saying why. I first built the end-to-end version of it as a cross-platform fake endpoint (named pipe and unix socket, dispatching on the command id in the request header) and it passed on Windows before the fix - vacuously: the fake never received the stop connection at all, the client sat out the full 30 s rundown budget, and a `TimeoutException` plus a *cancelled* drain is not an unobserved fault. The Windows named-pipe transport cannot express this symptom in the first place, since an aborted overlapped read ends the copy cancelled rather than faulted, and cancelled tasks are never unobserved. So an end-to-end test of it would have been unix-only, and I have no unix box here to see it fail on - a regression guard I cannot watch go red is not one I want to add.

Instead the invariant is pinned where it holds on all three OSes - the drain must not still be running once the failure path is done with it:

```csharp
using var session = new BlockedConnection();   // reads complete only when it is torn down, then fail
var trace = new MemoryStream();
Task drain = session.CopyToAsync(trace);
drain.IsCompleted.Should().BeFalse(...);

await DiagnosticsIpcClient.AbandonCollectionAsync(session, drain, trace);

drain.IsCompleted.Should().BeTrue(
    "a drain still running once the failure path is done with it is a drain nobody will ever await");
trace.CanRead.Should().BeFalse("the half-copied trace of a failed collection is dropped");
```

Red against the previous body of that method (`Expected drain.IsCompleted to be True ... but found False`), green after, and it fails on Windows too if someone drops the `await drain` later - which the unix-only version would not have.

christophwille:
## Why the dialog's two grids became ListBoxes

Trackpad scrolling in this dialog was unusable, and the cause turned out to be two
independent defects in ProDataGrid 12.0.4 (the newest release; nothing newer to upgrade to).
Recording this here so the decision is not re-litigated from scratch, and so anyone who
switches back knows what to answer for.

### Repro

1. Open the dialog on a machine with enough .NET processes to overflow the list (~60 here).
2. Two-finger scroll the process list downwards with a trackpad, past the last row.
3. Keep pushing, then scroll back up.

A mouse wheel or dragging the scroll bar does **not** reproduce it: both move in whole rows
or set an absolute position. A trackpad sends a stream of sub-row events - measured on
macOS, deltas of 0.02-0.12 (roughly 0.3-3px), each with a small horizontal component
alongside the vertical one. That is what walks the control into the broken state.

### Exhibited behavior

* A horizontal scroll bar appears and disappears every few frames, taking its height off
  the rows area each time, so the grid flickers and the rows jump. Frame-stepping a screen
  recording showed it toggling on 62 of 120 sampled frames.
* The rows slide sideways by up to 10px during a purely vertical scroll.
* The end of the list runs away: the scroll thumb shrinks to its minimum and you can keep
  scrolling long past the last row, which drifts in and out of view.

### Measurements

Instrumenting the running dialog (logging extent, viewport, offset, column widths and
scroll-bar state on every layout pass, then scrolling by hand) gave 776 samples:

| observation | value |
| --- | --- |
| Reported content height while scrolled to the top | 288, 1354, 1554, 1950, **8735** px |
| ...as multiples of the 24px row | 12, 56, 65, 81, **364** rows |
| Actual list height | ~1500px |
| Relationship that held throughout | `extentHeight ~= offsetY + 168` |
| Largest vertical offset reached | 8567px |
| Horizontal offset during vertical scrolling | up to 10.24px |
| Samples where the columns were 824px wide in an 806px viewport | 594 / 776 |

### Root causes

**The vertical runaway.** `DataGrid.EdgedRowsHeightCalculated` returns
`_verticalOffset - NegVerticalOffset + <heights of the displayed rows>` whenever the last
row is realized, and separately writes an offset-derived value back into the row-height
estimator (`RowHeightEstimate = num4 / num6`). So an offset that is briefly too large
inflates the per-row estimate, which inflates the extent, which permits a larger offset -
a positive feedback loop that ratchets up and does not recover, as the "back at the top"
numbers above show. Both statements are unconditional: no public property gates either.

**The sideways drift.** The star-sized last column is measured against the full grid width
(824px) rather than the cells viewport (806px), so the grid always believes it has one
vertical scroll bar's worth of content to reach. `DataGridRowsPresenter.SyncOffset` pushes
the grid's own `HorizontalOffset` straight into the presenter, clamped only by that phantom
overflow - which is why the horizontal component of a trackpad scroll shifts the rows.

### Fixes tried, and why each was not enough

| Attempt | Result |
| --- | --- |
| `HorizontalScrollBarVisibility="Disabled"` | Removes the flickering scroll bar - confirmed in the running app, 0 of 521 recorded frames. Does **not** stop the sideways drift: it hides the scroll bar without removing the phantom overflow that `HorizontalOffset` is clamped to. |
| Per-grid `DefaultRowHeightEstimator` instead of the default `AdvancedRowHeightEstimator` | Makes the estimate exact in a headless test (1.1x -> 1.0x), but cannot break a loop whose input is the scroll offset. The runaway persisted in the app. |
| `RowHeight="24"` (pinning a uniform row height) | No effect on the reported extent. |
| `VerticalScrollBarVisibility="Visible"` | No effect; the phantom 18px is not caused by the scroll bar appearing. |
| `ScrollViewer.AllowAutoHide` | No effect. |
| Fixed-width or `Auto` last column instead of star | Trades the flicker for a permanently visible horizontal scroll bar. |
| Disabling scrolling on the template's inner `PART_ScrollViewer` | No effect - the offset does not come from the ScrollViewer. |

### What replaced it

Both lists are now plain `ListBox`es with a column-aligned `ItemTemplate` and a header row.
A ListBox's virtualizing panel derives its extent from the items alone, so the feedback
loop has nothing to feed on. Regression tests assert, over 900 trackpad-sized wheel events,
that the horizontal offset stays exactly 0, the content height is single-valued and equal to
`rowCount * rowHeight`, the maximum offset lands exactly at `extent - viewport`, and the last
row comes to rest fully inside the viewport - the four things the DataGrid could not do.

The cost: the column widths now live in two places (the header row and the item template)
and have to be kept in step, and the sortable, resizable column headers are gone.

Worth reporting upstream to wieslawsoltes/ProDataGrid; the same two defects will affect any
DataGrid here with a star-sized last column (`SearchPane`, `CompareView`,
`OpenFromGacDialog`), since the estimator default is process-wide.

PR #3942, opened 2026-07-31
Fix #3938: nested-type members must not inherit ExtensionInfo

### LINKED ISSUE #3938: `System.Xml.XPath.XDocumentExtensions+XDocumentNavigable.CreateNavigator` fails to decompile.
### Input code
[ilspy-dll.zip](https://github.com/user-attachments/files/30534407/ilspy-dll.zip)

You can also use stock `System.Xml.XPath.XDocument.dll` from the .NET shared framework (tested with .NET 10 and 11).

The exception reproduces when viewing any individual member of the `XDocumentNavigable` class in the decompiler. It does _not_ reproduce when decompiling the whole class.

### Erroneous output
```
Error decompiling @06000003 System.Xml.XPath.XDocumentExtensions.XDocumentNavigable.CreateNavigator
in assembly "C:\Program Files\dotnet\shared\Microsoft.NETCore.App\10.0.10\System.Xml.XPath.XDocument.dll"
 ---> System.InvalidOperationException: Nullable object must have a value.
   at System.ThrowHelper.ThrowInvalidOperationException_InvalidOperation_NoValue() in offset 10
   at System.Nullable`1.get_Value() in offset 8
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun decompileRun, ITypeResolveContext decompilationContext, ExtensionInfo extensionInfo) in CSharpDecompiler.cs:line 2036
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun decompileRun, ITypeResolveContext decompilationContext, ExtensionInfo extensionInfo) in CSharpDecompiler.cs:line 2113
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DoDecompile(IMethod method, DecompileRun decompileRun, ITypeResolveContext decompilationContext, ExtensionInfo extensionInfo) in CSharpDecompiler.cs:line 1953
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.Decompile(IEnumerable`1 definitions) in CSharpDecompiler.cs:line 1259
   at ICSharpCode.ILSpy.CSharpLanguage.DecompileMethod(IMethod method, ITextOutput output, DecompilationOptions options) in offset 0
   at ICSharpCode.ILSpy.TextView.DecompilerTextView.<>c__DisplayClass58_0.<DecompileAsync>b__0() in offset 35
```

### Details
* Product in use: ILSpy
* Version in use: 10.1.1.8388+1377eb6e7351b21112858c8c1df39848f40181ec
* It throws only when setting language version to C# 14 and above.

siegfriedpammer:
Fixes #3938.

Decompiling a single member of an ordinary nested type inside a static `[Extension]` class (e.g. `System.Xml.XPath.XDocumentExtensions.XDocumentNavigable.CreateNavigator` in the .NET 10 BCL) threw `InvalidOperationException: Nullable object must have a value` with language version C# 14+.

Root cause, in two layers:

- `MetadataTypeDefinition.ExtensionInfo` was non-null for every static class carrying `[Extension]` — including classic `this`-parameter extension classes with no C# 14 extension blocks at all. The group scan found nothing, so the object was empty, but consumers took non-null to mean "extension container". It now returns null when there are no extension groups. (Empty extension blocks like `extension(int) {}` still produce marker groups and are unaffected.)
- `ResolveExtensionInfo` handed the container's info to members of *any* nested type, so `DecompileBody` dereferenced a missing `InfoOfExtensionMember` mapping. It now returns the info only for members that actually belong to one of the container's extension blocks — which also covers containers that mix real extension blocks with ordinary nested types.

`ExtensionInfo` only ever models C# 14 extension blocks (its maps are populated solely from the `<>E__`/`<G>$` encodings); classic extension methods are resolved via `HasExtensions`/`IsExtensionMethod`, which are unchanged.

Both new tests in `IsolatedMethodDecompilationTests` (classic-only and mixed-container fixtures) reproduced the reported exception before the fix; the stock .NET 10 `System.Xml.XPath.XDocument.dll` member from the issue now decompiles cleanly.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/IsolatedMethodDecompilationTests.cs
+++ b/ICSharpCode.Decompiler.Tests/IsolatedMethodDecompilationTests.cs
@@ -69,5 +69,53 @@ namespace ICSharpCode.Decompiler.Tests
 			Assert.That(code, Does.Contain("Number = 42"));
 			Assert.That(code, Does.Contain("Text = \"hello\""));
 		}
+
+		[Test]
+		public async Task NestedTypeMemberInClassicExtensionClass()
+		{
+			// #3938: [Extension] on the container alone must not make members of an ordinary
+			// nested type extension members; here the container has no extension blocks at all.
+			await DecompileNestedGetMethod("ClassicExtensions").ConfigureAwait(false);
+		}
+
+		[Test]
+		public async Task NestedTypeMemberInExtensionBlockContainer()
+		{
+			// Same as above, but the container also has a real C# 14 extension block whose
+			// ExtensionInfo must not be attributed to the nested type's members.
+			await DecompileNestedGetMethod("BlockExtensions").ConfigureAwait(false);
+		}
+
+		static async Task DecompileNestedGetMethod(string containerName)
+		{
+			var csFile = Path.Combine(TestCasePath, "ExtensionContainerNestedType.cs");
+			var compilation = await Tester.CompileCSharp(csFile, CompilerOptions.UseRoslynLatest | CompilerOptions.Optimize | CompilerOptions.Library).ConfigureAwait(false);
+			try
+			{
+				var settings = new DecompilerSettings();
+				using var file = new FileStream(compilation.PathToAssembly, FileMode.Open, FileAccess.Read);
+				var module = new PEFile(compilation.PathToAssembly, file, PEStreamOptions.PrefetchEntireImage);
+				var targetFramework = module.Metadata.DetectTargetFrameworkId();
+				var resolver = new UniversalAssemblyResolver(compilation.PathToAssembly, false, targetFramework, null, PEStreamOptions.PrefetchMetadata);
+				resolver.AddSearchDirectory(Tester.RefAssembliesToolset.GetPath(targetFramework));
+				var typeSystem = new DecompilerTypeSystem(module, resolver, settings);
+				var decompiler = new CSharpDecompiler(typeSystem, settings);
+
+				var getMethod = typeSystem.MainModule.TypeDefinitions
+					.Single(t => t.Name == containerName)
+					.NestedTypes.Single(t => t.Name == "Nested")
+					.Methods.Single(m => m.Name == "Get");
+
+				var code = decompiler.Decompile(getMethod.MetadataToken).ToString();
+
+				Assert.That(code, Does.Contain("return 42;"));
+				// The nested type's method must not be rendered inside an extension block.
+				Assert.That(code, Does.Not.Contain("extension("));
+			}
+			finally
+			{
+				compilation.DeleteTempFiles();
+			}
+		}
 	}
 }
--- a/ICSharpCode.Decompiler/TypeSystem/TypeSystemExtensions.cs
+++ b/ICSharpCode.Decompiler/TypeSystem/TypeSystemExtensions.cs
@@ -890,6 +890,11 @@ namespace ICSharpCode.Decompiler.TypeSystem
 			return ns;
 		}
 
+		/// <summary>
+		/// Returns the <see cref="ExtensionInfo"/> of the container type the member's extension
+		/// block belongs to, or null if the member is not part of an extension block. Fields and
+		/// events always return null: extension blocks cannot declare them.
+		/// </summary>
 		public static ExtensionInfo? ResolveExtensionInfo(this IMember member)
 		{
 			if (member is null)
@@ -898,7 +903,19 @@ namespace ICSharpCode.Decompiler.TypeSystem
 			}
 			var td = member.DeclaringTypeDefinition;
 			Debug.Assert(td != null, "IMember.DeclaringTypeDefinition should never be null");
-			return td.DeclaringTypeDefinition?.ExtensionInfo ?? td.DeclaringTypeDefinition?.DeclaringTypeDefinition?.ExtensionInfo;
+			var info = td.DeclaringTypeDefinition?.ExtensionInfo ?? td.DeclaringTypeDefinition?.DeclaringTypeDefinition?.ExtensionInfo;
+			if (info == null)
+				return null;
+			// The container's info only applies to members declared in one of its extension
+			// blocks; members of other nested types must not pick it up.
+			IMethod? method = member switch {
+				IMethod m => m,
+				IProperty p => p.Getter ?? p.Setter,
+				_ => null
+			};
+			if (method == null || info.InfoOfExtensionMember((IMethod)method.MemberDefinition) == null)
+				return null;
+			return info;
 		}
 	}
 }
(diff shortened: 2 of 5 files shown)


ICSharpCode.Decompiler/TypeSystem/TypeSystemExtensions.cs
```diff
@@ -898,7 +898,15 @@ public static IType AsParameterizedType(this ITypeDefinition td)
...
-			return td.DeclaringTypeDefinition?.ExtensionInfo ?? td.DeclaringTypeDefinition?.DeclaringTypeDefinition?.ExtensionInfo;
+			var info = td.DeclaringTypeDefinition?.ExtensionInfo ?? td.DeclaringTypeDefinition?.DeclaringTypeDefinition?.ExtensionInfo;
+			if (info == null)
+				return null;
+			// The container's info only applies to members declared in one of its extension
+			// blocks; members of other nested types must not pick it up.
+			IMethod? method = member as IMethod ?? (member as IProperty)?.Getter ?? (member as IProperty)?.Setter;
+			if (method == null || info.InfoOfExtensionMember((IMethod)method.MemberDefinition) == null)
+				return null;
+			return info;
```
christophwille, line 918: Two minor points on the new check:

1. **`ResolveExtensionInfo` deserves a doc comment now.** It is public API and its semantics changed meaningfully: it returns null for members that are not extension members, including all `IField`/`IEvent` inputs (previously it returned the container's info for any member of any nested type). No current caller is affected, but a one-line `/// <summary>` stating "returns the container's info only if the member belongs to one of its extension blocks" would pin the contract.
2. **Style nit:** `member as IProperty` is evaluated twice, and the `(IMethod)method.MemberDefinition` cast sits after a null check that already proved `method` non-null. A switch expression reads slightly cleaner:

```csharp
IMethod? method = member switch {
	IMethod m => m,
	IProperty p => p.Getter ?? p.Setter,
	_ => null
};
if (method == null || info.InfoOfExtensionMember((IMethod)method.MemberDefinition) == null)
	return null;
```

ICSharpCode.Decompiler.Tests/IsolatedMethodDecompilationTests.cs
```diff
@@ -69,5 +69,51 @@ public async Task StaticConstructorKeepsFieldInitializers()
...
+				var decompiler = new CSharpDecompiler(typeSystem, settings);
+
+				var getMethod = typeSystem.MainModule.TypeDefinitions
+					.Single(t => t.Name == containerName)
+					.NestedTypes.Single(t => t.Name == "Nested")
+					.Methods.Single(m => m.Name == "Get");
+
+				var code = decompiler.Decompile(getMethod.MetadataToken).ToString();
+
+				Assert.That(code, Does.Contain("return 42;"));
```
christophwille, line 111: `Does.Contain("return 42;")` is sufficient to catch the regression (the bug threw before producing any output), but adding `Assert.That(code, Does.Not.Contain("extension("));` would additionally guard against the nested member being mis-rendered inside an extension block in the `BlockExtensions` case.

christophwille (commented):
**Verdict: LGTM.** The fix addresses the root cause rather than the crash site, and comes with targeted regression tests for both failure layers. Detailed notes below; inline suggestions are all minor and non-blocking.

## Correctness (traced against all consumers)

- **The crash path is closed at its root.** `DecompileBody` only receives a non-null `extensionInfo` from the isolated-member entry points -- whole-type decompilation passes `null` down (`CSharpDecompiler.cs:1888/1896`) and uses the container info only to *skip* implementation members. Gating `ResolveExtensionInfo` on `InfoOfExtensionMember` membership therefore guarantees the `!.Value` at `CSharpDecompiler.cs:2150` is only reached with a valid mapping.
- The `Debug.Assert(extensionInfo != null)` at lines 1398/1410 stays valid: those paths are reached only from extension-member UI nodes, whose members are always in the map.
- **The null-when-empty change fixes a latent second bug for free.** The two-level fallback `td.DeclaringTypeDefinition?.ExtensionInfo ?? ...DeclaringTypeDefinition?.ExtensionInfo` (in `ResolveExtensionInfo`, `RequiredNamespaceCollector.cs:88`, `CSharpAmbience.cs:248`, `CSharpDecompiler.cs:1646/1666`) previously could stop at a non-null-but-empty info from an intermediate nested type and never fall through to the real container. Now it falls through correctly.
- The claim about empty extension blocks holds in both encodings: a bare `extension(int) {}` still yields a marker (`<>E__` group with `<Extension>$`, or a `<M>$` marker type), so `ExtensionGroups.Count > 0`.
- All other consumers handle the new null (`TypeTreeNode.cs:152` pattern-matches, `CSharpDecompiler.cs:1700` uses `?.` + `?? []`; `MinimalCorlib`/`SyntheticWpfModule` already returned null, so null was always part of the contract).
- Removing the loop-carried `parentExtensionInfo` in `Decompile(definitions)` is a good cleanliness fix -- the old variable persisted stale values across iterations.
- Caching still works: the empty `ExtensionInfo` is stored in the field, so repeated getter calls only repeat the cheap `Count > 0` check, not the nested-type scan.

## Non-blocking follow-up (outside this diff)

- `MethodTreeNode.cs:54` and `PropertyTreeNode.cs:58` still do `ResolveExtensionInfo()?.InfoOfExtensionMember(...)` themselves, so the membership test now runs twice. Harmless; they could be simplified to a null check on `ResolveExtensionInfo()` in a follow-up.
- External ILSpyX/plugin consumers of `ITypeDefinition.ExtensionInfo` will now see null for classic-only `[Extension]` classes -- a correct tightening, but worth a mention if release notes track API behavior.

Test coverage, conventions (self-contained comments, headerless fixture matching `TestCases/`, ASCII-only), performance, and security all check out.

ICSharpCode.Decompiler.Tests/IsolatedMethodDecompilationTests.cs
```diff
@@ -69,5 +69,51 @@ public async Task StaticConstructorKeepsFieldInitializers()
...
+		{
+			// Same as above, but the container also has a real C# 14 extension block whose
+			// ExtensionInfo must not be attributed to the nested type's members.
+			await DecompileNestedGetMethod("BlockExtensions").ConfigureAwait(false);
+		}
+
+		static async Task DecompileNestedGetMethod(string containerName)
+		{
+			var csFile = Path.Combine(TestCasePath, "ExtensionContainerNestedType.cs");
+			var compilation = await Tester.CompileCSharp(csFile, CompilerOptions.UseRoslynLatest | CompilerOptions.Optimize | CompilerOptions.Library).ConfigureAwait(false);
```
christophwille, line 92: Each test compiles the same fixture separately, so the Roslyn invocation runs twice. Fine at this scale; if this fixture grows, a shared `[OneTimeSetUp]` compile would avoid the duplication.

PR #3946, opened 2026-08-01
Annotate ICSharpCode.Decompiler for AOT compatibility

ds5678:
### Problem

Currently, the `ICSharpCode.Decompiler` package is not annotated for AOT compatibility, so downstream consumers get build warnings when trying to compile with NativeAOT.

### Solution

* I added a .NET 10 target.
* I added `UnconditionalSuppressMessage` attributes to suppress build warnings where necessary. Of course, all suppressions are annotated with justifications for why they're actually safe.
  * Unfortunately, `UnconditionalSuppressMessage` is only available on .NET 5 and later.
* I added obsolete attributes to several serialization-related methods to align with .NET 10 (and fix the compile errors it was giving me).
* I fixed some code that was not compatible with AOT to make it compatible.

### DIFF
--- a/ICSharpCode.Decompiler/ICSharpCode.Decompiler.csproj
+++ b/ICSharpCode.Decompiler/ICSharpCode.Decompiler.csproj
@@ -2,7 +2,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net10.0</TargetFrameworks>
     <!-- Carry every distribution RID in the lock file so a RID-specific `dotnet publish -r <rid>` of a
          consuming app (ILSpy) restores in locked mode (RestoreLockedMode below) on any host OS, with no
          per-publish restore escape hatch. Must match ILSpy.csproj's RuntimeIdentifiers. -->
@@ -80,6 +80,10 @@
     </GetPackageVersionDependsOn>
   </PropertyGroup>
 
+  <PropertyGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))">
+    <IsAotCompatible>true</IsAotCompatible>
+  </PropertyGroup>
+
   <PropertyGroup>
     <ManagePackageVersionsCentrally>false</ManagePackageVersionsCentrally>
   </PropertyGroup>
@@ -101,7 +105,7 @@
     </PackageReference>
   </ItemGroup>
 
-  <ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="TunnelVisionLabs.ReferenceAssemblyAnnotator" Version="1.0.0-alpha.160" PrivateAssets="all" />
 
     <!-- Specifies the version of Microsoft.NETCore.App.Ref to obtain nullability information from. -->
--- a/ICSharpCode.Decompiler/Metadata/UniversalAssemblyResolver.cs
+++ b/ICSharpCode.Decompiler/Metadata/UniversalAssemblyResolver.cs
@@ -418,6 +418,16 @@ namespace ICSharpCode.Decompiler.Metadata
 			return path ?? version?.ToString() ?? ".";
 		}
 
+#if NET5_0_OR_GREATER
+		[UnconditionalSuppressMessage("SingleFile", "IL3002:Avoid calling members marked with 'RequiresAssemblyFilesAttribute' when publishing as a single-file", Justification = "This method returns null when the module has no file path.")]
+#endif
+		static string? GetSystemObjectModuleFullyQualifiedName()
+		{
+			// FullyQualifiedName returns "<Unknown>" for modules with no file path.
+			string? path = typeof(object).Module.FullyQualifiedName;
+			return path is "<Unknown>" ? null : path;
+		}
+
 		string? ResolveInternal(IAssemblyReference name)
 		{
 			if (name == null)
@@ -428,12 +438,14 @@ namespace ICSharpCode.Decompiler.Metadata
 				return assembly;
 
 			string[] framework_dirs;
-			string framework_dir;
+			string? framework_dir;
 
 			switch (decompilerRuntime)
 			{
 				case DecompilerRuntime.Mono:
-					framework_dir = Path.GetDirectoryName(typeof(object).Module.FullyQualifiedName)!;
+					framework_dir = Path.GetDirectoryName(GetSystemObjectModuleFullyQualifiedName());
+					if (string.IsNullOrEmpty(framework_dir))
+						goto NotFound;
 					framework_dirs = new[] { framework_dir, Path.Combine(framework_dir, "Facades") };
 					break;
 				case DecompilerRuntime.NETCoreApp:
@@ -454,7 +466,9 @@ namespace ICSharpCode.Decompiler.Metadata
 					framework_dirs = new[] { framework_dir };
 					break;
 				default:
-					framework_dir = Path.GetDirectoryName(typeof(object).Module.FullyQualifiedName)!;
+					framework_dir = Path.GetDirectoryName(GetSystemObjectModuleFullyQualifiedName());
+					if (string.IsNullOrEmpty(framework_dir))
+						goto NotFound;
 					framework_dirs = new[] { framework_dir };
 					break;
 			}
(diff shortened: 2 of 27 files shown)


christophwille:
Hi Jeremy, you should have asked before embarking on this endeavor. Because ics.d is intentionally netstandard only and will remain netstandard for the forseeeable future, and multi-targetting is absolutely off-limits (we had that discussion internally more than once).

Reason why no multi-targetting: then all our projects (tests, ILSpy) would start using the net10 tfm compiled variant and basically the netstandard target would go untested. And netstandard is our ticket for being included in Roslyn.

If you can come up with a different approach from multi-targetting, we can talk.

ds5678:
> Reason why no multi-targetting: then all our projects (tests, ILSpy) would start using the net10 tfm compiled variant and basically the netstandard target would go untested. And netstandard is our ticket for being included in Roslyn.

Is my latest commit an acceptable solution to this concern?

christophwille:
@siegfriedpammer Would having the SetTargetFramework attribute on icsd.tests suffice, and the rest of ILSpy use the net10 target? That way both TFMs should be properly exercised in our repository.

ICSharpCode.Decompiler/ICSharpCode.Decompiler.csproj
```diff
@@ -80,6 +80,10 @@
     </GetPackageVersionDependsOn>
   </PropertyGroup>
 
+  <PropertyGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))">
```
christophwille, line 83: If we target net10, why would we compare to net80?

christophwille:
> [!NOTE]
> This comment was entirely AI generated (Claude Fable 5 via Claude Code) on behalf of @christophwille.

Here is an assessment of the `SetTargetFramework` approach and the alternatives, given the constraint that netstandard2.0 must remain the tested target.

## TL;DR

`SetTargetFramework` does technically satisfy the constraint (everything in-repo compiles against and tests the netstandard2.0 build), but it is the fragile version of the right idea:

- It is scattered across ~12 csproj files and must be remembered on every future `ProjectReference`. Forgetting it fails *silently* — we would start testing the net10 build without noticing.
- NuGet restore is documented to **not** honor it ([NuGet/Home#12436](https://github.com/NuGet/Home/issues/12436)) — the restore graph still nearest-matches, so a net10/net11 consumer's lock file and transitive-dependency view follow the net10 target while compilation uses the netstandard asset. With `RestoreLockedMode` on the core libraries and CI keyed off `packages.lock.json`, that mismatch is exactly the kind of thing this repo is sensitive to.
- There is also a new gap this PR doesn't address: the shipped net10.0 asset would be executed by **zero** tests — we would be publishing an entirely untested binary as the one AOT consumers actually load.

## Alternatives, ranked for the constraint

### 1. Stay netstandard2.0-only and polyfill the annotation attributes (recommended)

The trimmer/ILC match `UnconditionalSuppressMessage`, `RequiresUnreferencedCode`, `RequiresDynamicCode`, `DynamicallyAccessedMembers` etc. **by fully-qualified name, not by assembly** — internal copies defined in the netstandard2.0 build are honored when a downstream app NativeAOT-publishes against it. The repo already does exactly this for nullability attributes in `ICSharpCode.Decompiler/NRTAttributes.cs`, so it is an established in-repo pattern (PolySharp exists for the same reason ecosystem-wide). The `IsAotCompatible`/`IsTrimmable` markers are just assembly-level `AssemblyMetadata` attributes that the SDK normally injects for net7+; they can be added by hand to the netstandard target.

Result: one TFM, one binary, the tested binary *is* the shipped binary, nothing about the Roslyn-inclusion story changes, and none of the csproj churn in consumers is needed. Most of the genuinely valuable content of this PR — the actual AOT code fixes, obsoletions, and suppressions with justifications — carries over unchanged.

The one real gap: the SDK's trim/AOT **analyzers don't run** on a netstandard2.0 TFM, so annotations can't be verified at build time. That can be fixed with a non-packed validation build — either a CI-only sibling csproj compiling the same sources under net10.0 with `<EnableAotAnalyzer>`/`<EnableTrimAnalyzer>` on, or an inner multi-target whose net10 compile is excluded from the package and whose only job is analysis. That keeps the annotations honest without shipping a second asset.

### 2. If shipping two TFMs is acceptable: centralize the pin and test both

Instead of ~12 scattered `SetTargetFramework` attributes, put one

```xml
<ProjectReference Update="@(ProjectReference)"
                  Condition="'%(FileName)' == 'ICSharpCode.Decompiler'"
                  SetTargetFramework="TargetFramework=netstandard2.0" />
```

in the root `Directory.Build.targets` — one place, future references automatically pinned, no silent-regression risk. Then make the pin a property (`SetTargetFramework="TargetFramework=$(DecompilerTestTfm)"`) and add a CI matrix leg that flips it to net10.0 for the decompiler test suite, so *both* shipped assets get test coverage instead of neither problem existing.

The NuGet-restore caveat above still applies and needs an `updatedeps.ps1` run plus a check that consumer lock files are stable. In practice the runtime risk here is low because the netstandard build's package deps (System.Memory, System.Reflection.Metadata, System.Collections.Immutable) are all inbox on net10, but it should be verified, not assumed.

### 3. Conditional TFM (net10 only when packing) — reject

`<TargetFrameworks>` that differ between dev and pack builds would keep consumer csproj files untouched, but it makes restore graphs unstable between local and CI, which is incompatible with the repo's locked-mode restore, and the net10 asset would be even less exercised.

## Bottom line

If "netstandard is the product and must be the tested thing" is a hard rule, then multi-targeting solves the problem in the wrong place. Option 1 gives downstream NativeAOT consumers exactly what this PR set out to deliver (no AOT warnings, real suppressions with justifications, the `IsAotCompatible` marker) while keeping the package single-TFM and keeping the tested and shipped binary identical.

Suggested direction: keep the code fixes and suppressions, replace the multi-target + `SetTargetFramework` mechanics with polyfilled attributes, and add an analyzer-only net10 validation build so the annotations can't rot. If there is a concrete AOT scenario that name-matched attributes in a netstandard asset *can't* fix, that would be the argument for option 2 — but then the centralized pin and a net10 test leg should be conditions of acceptance.

ICSharpCode.Decompiler/ICSharpCode.Decompiler.csproj
ds5678: That's the earliest runtime that `IsAotCompatible` is valid. A warning is issued if the property is applied on an earlier framework.

```
NETSDK1210

IsAotCompatible and EnableAotAnalyzer are not supported for the target framework. Consider multi-targeting to a supported framework to enable ahead-of-time compilation analysis, and set IsAotCompatible only for the supported frameworks. For example:
<IsAotCompatible Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))">true</IsAotCompatible>
```

ICSharpCode.Decompiler/Metadata/UniversalAssemblyResolver.cs
```diff
@@ -494,12 +508,15 @@ string FindClosestVersionDirectory(string basePath, Version? version)
 				// regardless of the requested version (the runtime itself rolls forward in the
 				// same way). Without this, e.g. the type-forwards of a netstandard facade
 				// (pointing to a versioned System.Runtime) are unresolvable on such hosts.
-				string runtimeDir = Path.GetDirectoryName(typeof(object).Module.FullyQualifiedName)!;
+				string? runtimeDir = Path.GetDirectoryName(GetSystemObjectModuleFullyQualifiedName());
+				if (string.IsNullOrEmpty(runtimeDir))
+					goto NotFound;
```
siegfriedpammer, line 513: Why not use a static local function?

siegfriedpammer:
@ds5678 would polyfilling the attributes work for you?

ds5678:
> @ds5678 would polyfilling the attributes work for you?

AOT warnings only appear on versions of .NET that have been annotated, which was the original reason for multi-targeting .NET 10. Without the compiler issuing warnings, code changes can silently break AOT compatibility.

christophwille:
But Claude also suggested "helper" constructs to verify the warnings to the side.

ds5678:
I assume that this is the ideal solution?

https://github.com/tunnelvisionlabs/ReferenceAssemblyAnnotator/pull/95

siegfriedpammer:
Isn't ReferenceAssemblyAnnotator pretty much dead?

christophwille:
@ds5678 Did you have a chance to look into option 1 (without touching ReferenceAssemblyAnnotator). We do want to support your AOT needs.

ds5678:
> Did you have a chance to look into option 1 (without touching ReferenceAssemblyAnnotator). We do want to support your AOT needs.

I was waiting to see if there was any movement on my ReferenceAssemblyAnnotator pull request because I think that is the best solution. If another week goes by with no news, I'll work on the option you suggested.

PR #3949, opened 2026-08-01
Fix #3803: Crash on nested deconstruction of custom structs

### LINKED ISSUE #3388: Exception while decompiling nested tuple deconstruction
### Input code
```cs
class DeconstructErrorTest
{
    struct Data
    {
        public void Deconstruct(out object o1, out object o2)
        {
            o1 = default;
            o2 = default;
        }
    }

    static void Test(KeyValuePair<object, Data> pair)
    {
        var (key, (value, _)) = pair;

        Empty(key);
        Empty(value);
    }

    static void Empty(object o) => _ = o;
}
```

### Erroneous output
```
Error decompiling @0600009B DeconstructErrorTest.Test
 ---> System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
   at System.ThrowHelper.ThrowArgumentOutOfRange_IndexMustBeLessException() in offset 0
   at System.Collections.Generic.List`1.get_Item(Int32 index) in offset 9
   at ICSharpCode.Decompiler.IL.InstructionCollection`1.get_Item(Int32 index) in InstructionCollection.cs:line 46
   at ICSharpCode.Decompiler.CSharp.ExpressionBuilder.<VisitDeconstructInstruction>g__ConstructTuple|160_0(MatchInstruction matchInstruction, <>c__DisplayClass160_0&) in ExpressionBuilder.cs:line 4601
   at ICSharpCode.Decompiler.CSharp.ExpressionBuilder.VisitDeconstructInstruction(DeconstructInstruction inst, TranslationContext context) in ExpressionBuilder.cs:line 4583
   at ICSharpCode.Decompiler.IL.DeconstructInstruction.AcceptVisitor[C,T](ILVisitor`2 visitor, C context) in Instructions.cs:line 6729
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.Default(ILInstruction inst) in StatementBuilder.cs:line 91
   at ICSharpCode.Decompiler.IL.ILVisitor`1.VisitDeconstructInstruction(DeconstructInstruction inst) in Instructions.cs:line 7614
   at ICSharpCode.Decompiler.IL.DeconstructInstruction.AcceptVisitor[T](ILVisitor`1 visitor) in Instructions.cs:line 6725
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.Convert(ILInstruction inst) in StatementBuilder.cs:line 80
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertBlockContainer(BlockStatement blockStatement, BlockContainer container, IEnumerable`1 blocks, Boolean isLoop) in StatementBuilder.cs:line 1443
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertBlockContainer(BlockContainer container, Boolean isLoop) in StatementBuilder.cs:line 1347
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.VisitBlockContainer(BlockContainer container) in StatementBuilder.cs:line 1233
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.Convert(ILInstruction inst) in StatementBuilder.cs:line 80
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertAsBlock(ILInstruction inst) in StatementBuilder.cs:line 85
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun decompileRun, ITypeResolveContext decompilationContext) in CSharpDecompiler.cs:line 1780
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun decompileRun, ITypeResolveContext decompilationContext) in CSharpDecompiler.cs:line 1795
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DoDecompile(IMethod method, DecompileRun decompileRun, ITypeResolveContext decompilationContext) in CSharpDecompiler.cs:line 1644
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.<>c__DisplayClass75_0.<DoDecompile>g__DoDecompileMember|0(IEntity entity, RecordDecompiler recordDecompiler, PartialTypeInfo partialType) in CSharpDecompiler.cs:line 1519
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DoDecompile(ITypeDefinition typeDef, DecompileRun decompileRun, ITypeResolveContext decompilationContext) in CSharpDecompiler.cs:line 1382
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.Decompile(IEnumerable`1 definitions) in CSharpDecompiler.cs:line 1038
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.Decompile(EntityHandle[] definitions) in CSharpDecompiler.cs:line 1010
   at ICSharpCode.ILSpy.CSharpLanguage.DecompileType(ITypeDefinition type, ITextOutput output, DecompilationOptions options) in offset 65
   at ICSharpCode.ILSpy.TreeNodes.TypeTreeNode.Decompile(Language language, ITextOutput output, DecompilationOptions options) in offset 0
   at ICSharpCode.ILSpy.TextView.DecompilerTextView.DecompileNodes(DecompilationContext context, ITextOutput textOutput) in offset 103
   at ICSharpCode.ILSpy.TextView.DecompilerTextView.<>c__DisplayClass58_0.<DecompileAsync>b__0() in offset 18
```

### Details
* Product in use: ILSpy 9.0 VS extension

### LINKED ISSUE #3803: System.ArgumentOutOfRangeException: Index was out of range. (nested deconstruction causing crashes)
### Steps to reproduce
1. `ilspycmd -t Microsoft.CodeAnalysis.LanguageServer.LspWorkspaceManager Microsoft.CodeAnalysis.LanguageServer.Protocol.dll`, OR 
2. `ilspycmd -t CForEach M.dll` OR
3. `ilspycmd -t CAssign M.dll`

### Error message shown

```
ilspycmd -t Microsoft.CodeAnalysis.LanguageServer.LspWorkspaceManager \                                                                                                                  ✔ 
    "/Users/noahpeterson/.vscode/extensions2/ms-dotnettools.csharp-2.120.3-darwin-arm64/.roslyn/Microsoft.CodeAnalysis.LanguageServer.Protocol.dll"
Error decompiling @06000E51 Microsoft.CodeAnalysis.LanguageServer.LspWorkspaceManager.GetLspSolutionsAsync
in assembly "/Users/noahpeterson/.vscode/extensions2/ms-dotnettools.csharp-2.120.3-darwin-arm64/.roslyn/Microsoft.CodeAnalysis.LanguageServer.Protocol.dll"
 ---> System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
   at System.ThrowHelper.ThrowArgumentOutOfRange_IndexMustBeLessException() in offset 0
   at System.Collections.Generic.List`1.get_Item(Int32 index) in offset 9
   at ICSharpCode.Decompiler.IL.InstructionCollection`1.get_Item(Int32 index) in InstructionCollection.cs:line 46
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.<TranslateDeconstructionDesignation>g__ConstructDesignation|43_0(MatchInstruction matchInstruction, <>c__DisplayClass43_0&) in StatementBuilder.cs:line 824
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.TranslateDeconstructionDesignation(DeconstructInstruction inst, Boolean isForeach) in StatementBuilder.cs:line 812
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.TransformToForeach(UsingInstruction inst, Expression resource) in StatementBuilder.cs:line 742
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.VisitUsingInstruction(UsingInstruction inst) in StatementBuilder.cs:line 530
   at ICSharpCode.Decompiler.IL.UsingInstruction.AcceptVisitor[T](ILVisitor`1 visitor) in Instructions.cs:line 2076
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.Convert(ILInstruction inst) in StatementBuilder.cs:line 79
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertBlockContainer(BlockStatement blockStatement, BlockContainer container, IEnumerable`1 blocks, Boolean isLoop) in StatementBuilder.cs:line 1450
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertBlockContainer(BlockContainer container, Boolean isLoop) in StatementBuilder.cs:line 1344
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.VisitBlockContainer(BlockContainer container) in StatementBuilder.cs:line 1230
   at ICSharpCode.Decompiler.IL.BlockContainer.AcceptVisitor[T](ILVisitor`1 visitor) in Instructions.cs:line 854
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.Convert(ILInstruction inst) in StatementBuilder.cs:line 79
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertAsBlock(ILInstruction inst) in StatementBuilder.cs:line 84
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.<DeclareLocalFunctions>g__TranslateFunction|61_0(ILFunction function) in StatementBuilder.cs:line 1410
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.DeclareLocalFunctions(ILFunction currentFunction, BlockContainer container, BlockStatement blockStatement) in StatementBuilder.cs:line 1381
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertBlockContainer(BlockContainer container, Boolean isLoop) in StatementBuilder.cs:line 1345
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.VisitBlockContainer(BlockContainer container) in StatementBuilder.cs:line 1230
   at ICSharpCode.Decompiler.IL.BlockContainer.AcceptVisitor[T](ILVisitor`1 visitor) in Instructions.cs:line 854
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.Convert(ILInstruction inst) in StatementBuilder.cs:line 79
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertAsBlock(ILInstruction inst) in StatementBuilder.cs:line 84
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun decompileRun, ITypeResolveContext decompilationContext, ExtensionInfo extensionInfo) in CSharpDecompiler.cs:line 2098
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun decompileRun, ITypeResolveContext decompilationContext, ExtensionInfo extensionInfo) in CSharpDecompiler.cs:line 2113
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DoDecompile(IMethod method, DecompileRun decompileRun, ITypeResolveContext decompilationContext, ExtensionInfo extensionInfo) in CSharpDecompiler.cs:line 1953
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.<DoDecompile>g__DoDecompileMember|76_0(IEntity entity, RecordDecompiler recordDecompiler, PartialTypeInfo partialType, ExtensionInfo extensionInfo, <>c__DisplayClass76_0&) in CSharpDecompiler.cs:line 1789
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DoDecompile(ITypeDefinition typeDef, DecompileRun decompileRun, ITypeResolveContext decompilationContext, Boolean asExtension) in CSharpDecompiler.cs:line 1641
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DoDecompileTypes(IEnumerable`1 types, DecompileRun decompileRun, ITypeResolveContext decompilationContext, SyntaxTree syntaxTree) in CSharpDecompiler.cs:line 805
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DecompileType(FullTypeName fullTypeName) in CSharpDecompiler.cs:line 1195
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DecompileTypeAsString(FullTypeName fullTypeName) in CSharpDecompiler.cs:line 1208
   at ICSharpCode.ILSpyCmd.ILSpyCmdProgram.Decompile(String assemblyFileName, TextWriter output, String typeName) in IlspyCmdProgram.cs:line 595
   at ICSharpCode.ILSpyCmd.ILSpyCmdProgram.<OnExecuteAsync>g__PerformPerFileAction|99_0(String fileName, <>c__DisplayClass99_0&, <>c__DisplayClass99_1&) in IlspyCmdProgram.cs:line 351
   at ICSharpCode.ILSpyCmd.ILSpyCmdProgram.<OnExecuteAsync>d__99.MoveNext() in IlspyCmdProgram.cs:line 258
```

### Details
* Product in use: `ilspycmd`
* Version in use: `ilspycmd: 11.0.0.0`, `ICSharpCode.Decompiler: 11.0.0.0`

Minimum viable reproduction:

M.cs
```

  struct Inner { public void Deconstruct(out int a, out int b) { a = 0; b = 0; } }
  struct Outer { public void Deconstruct(out int x, out Inner inner) { x = 0; inner = default; } }

  class CForEach   // -> crashes ConstructDesignation (StatementBuilder.cs:824)
  {
      public int M(IEnumerable<Outer> items)
      {
          int n = 0;
          foreach (var (x, (a, b)) in items) n += x + a + b;
          return n;
      }
  }

  class CAssign    // -> crashes ConstructTuple (ExpressionBuilder.cs:4747)
  {
      public int M(Outer o)
      {
          var (x, (a, b)) = o;
          return x + a + b;
      }
  }
```
  M.csproj: `net8.0`, `<Optimize>true</Optimize>`, `OutputType=Library`.

siegfriedpammer:
Fixes #3803. Fixes #3388.

Nested deconstruction (`var (x, (a, b)) = o;`, also the `foreach` form) crashed the decompiler — Debug builds assert in `DeconstructInstruction.CheckInvariant`, Release builds throw the reported `ArgumentOutOfRangeException`. This hits real-world assemblies, e.g. `Microsoft.CodeAnalysis.LanguageServer.Protocol.dll` shipped with the VS Code C# extension.

**Root cause.** In optimized code a deconstruction element that is used only once after the deconstruction is not stored to a temporary. `DeconstructionTransform.MatchAssignments` compensated for that, but only for *trailing* elements. A nested deconstruction copies the inner element to a temporary (the defensive copy for the inner `Deconstruct` receiver), which counts as the only matched assignment — leaving the *leading* elements without one, with their single load outside the new `DeconstructInstruction`. That violates the instruction's invariant that all pattern-variable loads are descendants of it, and the statement/expression builders crash on the malformed pattern.

**Fix.** The forwarding fixup now covers all unassigned elements (guarded by an `IsDescendantOf` check), and inserts the forwarding assignments in pattern order, since `StatementBuilder.TranslateDeconstructionDesignation` and `ExpressionBuilder.ConstructTuple` pair pattern variables with assignments positionally.

**Second bug (pre-existing, surfaced by the fixture).** Deconstructing a struct with a custom `Deconstruct` passes the receiver by reference; `VisitDeconstructInstruction` emitted the address operand as an invalid cast — `var (x, y) = (S)(ref s);` — on released 10.1.1 even without nesting. Now unwrapped like `TranslateTarget` does for member access targets.

**Output shape note.** The inner deconstruction currently decompiles as an explicit `inner.Deconstruct(out var a, out var b);` call rather than nested designation sugar — `DeconstructionTransform` never builds nested sub-patterns (the IL node and the statement/expression builders already support them, so reconstructing the sugar is a self-contained future feature in the transform).

**Tests.** Nested assignment, `foreach`, and the #3388 KeyValuePair-with-discard shapes added to the Correctness fixture: red before (all compiler-matrix variants crash at decompile), green after, with runtime output compared against the original — so element evaluation order is verified, not just compilability. Full decompiler suite: 3332 tests, 0 failures.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Correctness/DeconstructionTests.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Correctness/DeconstructionTests.cs
@@ -6,6 +6,15 @@ using System.Threading.Tasks;
 
 namespace ICSharpCode.Decompiler.Tests.TestCases.Correctness
 {
+	static class KeyValuePairExtensions
+	{
+		public static void Deconstruct<TKey, TValue>(this KeyValuePair<TKey, TValue> pair, out TKey key, out TValue value)
+		{
+			key = pair.Key;
+			value = pair.Value;
+		}
+	}
+
 	class DeconstructionTests
 	{
 		public static void Main()
@@ -137,6 +146,81 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Correctness
 			NullReferenceException_RefLocalReferencesArrayElement_Deconstruction(out _, null);
 			DeconstructTupleSameVar(("a", "b"));
 			DeconstructTupleListForEachSameVar(new List<(string, string)> { ("a", "b") });
+			StructDeconstruction_Assignment(new NestedInner { Value = 7 });
+			NestedDeconstruction_Assignment(new NestedOuter { Value = 42 });
+			NestedDeconstruction_ForEach(new List<NestedOuter> {
+				new NestedOuter { Value = 1 },
+				new NestedOuter { Value = 2 }
+			});
+			NestedDeconstruction_DiscardedElement(new KeyValuePair<object, DiscardData>("key", default(DiscardData)));
+		}
+
+		public struct DiscardData
+		{
+			public void Deconstruct(out object o1, out object o2)
+			{
+				Console.WriteLine("DiscardData.Deconstruct");
+				o1 = 1;
+				o2 = 2;
+			}
+		}
+
+		public void NestedDeconstruction_DiscardedElement(KeyValuePair<object, DiscardData> pair)
+		{
+			Console.WriteLine("NestedDeconstruction_DiscardedElement:");
+			var (key, (value, _)) = pair;
+			Console.WriteLine(key);
+			Console.WriteLine(value);
+		}
+
+		public struct NestedInner
+		{
+			public int Value;
+
+			public void Deconstruct(out int a, out int b)
+			{
+				Console.WriteLine("NestedInner.Deconstruct");
+				a = Value + 1;
+				b = Value + 2;
+			}
+		}
+
+		public struct NestedOuter
+		{
+			public int Value;
+
+			public void Deconstruct(out int x, out NestedInner inner)
+			{
+				Console.WriteLine("NestedOuter.Deconstruct");
+				x = Value;
+				inner = new NestedInner { Value = Value * 10 };
+			}
+		}
+
+		public void StructDeconstruction_Assignment(NestedInner s)
+		{
+			Console.WriteLine("StructDeconstruction_Assignment:");
+			var (a, b) = s;
+			Console.WriteLine(a);
+			Console.WriteLine(b);
+		}
+
+		public void NestedDeconstruction_Assignment(NestedOuter o)
+		{
+			Console.WriteLine("NestedDeconstruction_Assignment:");
+			var (x, (a, b)) = o;
+			Console.WriteLine(x);
+			Console.WriteLine(a);
+			Console.WriteLine(b);
+		}
+
+		public void NestedDeconstruction_ForEach(IEnumerable<NestedOuter> items)
+		{
+			Console.WriteLine("NestedDeconstruction_ForEach:");
+			foreach (var (x, (a, b)) in items)
+			{
+				Console.WriteLine(x + a + b);
+			}
 		}
 
 		public void Property_NoDeconstruction_SwappedAssignments()
--- a/ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
+++ b/ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
@@ -19,6 +19,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Diagnostics;
 using System.Linq;
 using System.Resources;
 
@@ -390,26 +391,63 @@ namespace ICSharpCode.Decompiler.IL.Transforms
 
 			if (deconstructionResults != null)
 			{
-				int i = previousIndex + 1;
-				while (i < deconstructionResults.Length)
+				foreach (var v in deconstructionResults)
 				{
-					var v = deconstructionResults[i];
-					// this should only happen in release mode, where usually the last deconstruction element
-					// is not stored to a temporary, if it is used directly (and only once!)
-					// after the deconstruction.
-					if (v?.LoadCount == 1)
-					{
-						delayedActions += (DeconstructInstruction deconstructInst) => {
-							var freshVar = context.Function.RegisterVariable(VariableKind.StackSlot, v.Type);
-							deconstructInst.Assignments.Instructions.Add(new StLoc(freshVar, new LdLoc(v)));
-							v.LoadInstructions[0].Variable = freshVar;
-						};
-					}
-					i++;
+					// In optimized code a deconstruction element is not stored to a temporary,
+					// if it is used directly (and only once!) after the deconstruction. This
+					// happens for trailing elements, but also for leading elements, e.g., when
+					// a nested deconstruction copies the inner element to a temporary before
+					// the elements preceding it are used. Forward such elements through a fresh
+					// variable assigned inside the deconstruction, so that every pattern
+					// variable's load is a descendant of the deconstruct instruction.
+					// The assignment is inserted in pattern order, because StatementBuilder and
+					// ExpressionBuilder pair pattern variables with assignments positionally.
+					// LoadCount must be read eagerly, at match time: for a tuple deconstruction
+					// the elements are the fresh "E_i" variables created in FindIndex, whose
+					// loads only materialize when the delayed ReplaceWith actions run, so
+					// LoadCount is still 0 here and forwarding never fires on that path. That
+					// is load-bearing, not incidental: the fresh variables are never registered
+					// in deconstructionResultsLookup, so GetAssignmentIndex could not position
+					// a forwarding assignment among a tuple's assignments.
+					if (v?.LoadCount != 1)
+						continue;
+					delayedActions += (DeconstructInstruction deconstructInst) => {
+						var load = v.LoadInstructions[0];
+						if (load.IsDescendantOf(deconstructInst))
+							return;
+						// MatchDeconstruction registered every deconstruction result in the
+						// lookup, and the tuple path never gets here (see above); a miss would
+						// leave the load outside the deconstruct instruction, i.e. a malformed
+						// pattern, because the transform is already committed at this point.
+						bool isDeconstructionResult = deconstructionResultsLookup.TryGetValue(v, out int index);
+						Debug.Assert(isDeconstructionResult);
+						var freshVar = context.Function.RegisterVariable(VariableKind.StackSlot, v.Type);
+						var instructions = deconstructInst.Assignments.Instructions;
+						int insertPos = 0;
+						while (insertPos < instructions.Count && GetAssignmentIndex(instructions[insertPos]) < index)
+							insertPos++;
+						instructions.Insert(insertPos, new StLoc(freshVar, new LdLoc(v)));
+						load.Variable = freshVar;
+					};
 				}
 			}
 
 			return startPos != pos;
+
+			int GetAssignmentIndex(ILInstruction inst)
+			{
+				if (DeconstructInstruction.IsAssignment(inst, context.TypeSystem, out _, out var value)
+					&& value.MatchLdLoc(out var inputVariable))
+				{
+					if (deconstructionResultsLookup.TryGetValue(inputVariable, out int index))
+						return index;
+					// Forwarding assignments produced for conversions load a fresh variable;
+					// their pattern index is that of the conversion output they store to.
+					if (inst is StLoc stLoc && deconstructionResultsLookup.TryGetValue(stLoc.Variable, out index))
+						return index;
+				}
+				return int.MaxValue;
+			}
 
 			void AddMissingAssignmentsForConversions(int index, ref Action<DeconstructInstruction> delayedActions)
 			{
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/DeconstructionTests.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/DeconstructionTests.cs
@@ -70,6 +70,17 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 			}
 		}
 
+		public struct StructDeconstructionSource<T, T2>
+		{
+			public int Dummy { get; set; }
+
+			public void Deconstruct(out T a, out T2 b)
+			{
+				a = default(T);
+				b = default(T2);
+			}
+		}
+
 		private class AssignmentTargets
 		{
 			public int IntField;
@@ -126,6 +137,11 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 		private DeconstructionSource<T, T2, T3> GetSource<T, T2, T3>()
 		{
 			return null;
+		}
+
+		private StructDeconstructionSource<T, T2> GetStructSource<T, T2>()
+		{
+			return default(StructDeconstructionSource<T, T2>);
 		}
 
 		private ref T GetRef<T>()
@@ -297,6 +313,13 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 			};
 			Console.WriteLine(myInt3);
 			Console.WriteLine(myInt4);
+		}
+
+		public void LocalVariable_NoConversion_Struct_Custom()
+		{
+			var (value, value2) = GetStructSource<string, int>();
+			Console.WriteLine(value);
+			Console.WriteLine(value2);
 		}
 
 		public void Property_NoConversion_Custom()
--- a/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
@@ -4885,6 +4885,12 @@ namespace ICSharpCode.Decompiler.CSharp
 		{
 			IType rhsType = inst.Pattern.Variable.Type;
 			var rhs = Translate(inst.Pattern.TestedOperand, rhsType);
+			if (rhs.Expression is DirectionExpression dirExpr)
+			{
+				// Deconstructing a value type takes the address of the deconstructed value:
+				// (ref x) => x
+				rhs = rhs.UnwrapChild(dirExpr.Expression);
+			}
 			rhs = rhs.ConvertTo(rhsType, this); // TODO allowImplicitConversion
 			var assignments = inst.Assignments.Instructions;
 			int assignmentPos = 0;


christophwille:
## Code review

### What this does

Two independent fixes plus regression coverage:

1. **`DeconstructionTransform.MatchAssignments`** - the "element used once, never stored to a temporary" fixup only ran for *trailing* elements (`i = previousIndex + 1 ..`). A nested deconstruction stores the inner element to a temporary (the defensive copy for the inner `Deconstruct` receiver), which consumes the match, so *leading* elements were left with their single load outside the new `DeconstructInstruction` - violating the invariant that `DeconstructInstruction.CheckInvariant` -> `ValidatePattern` asserts. The fixup now covers every unassigned element, guarded by `IsDescendantOf`, and inserts in pattern order.
2. **`ExpressionBuilder.VisitDeconstructInstruction`** - the tested operand's address was emitted as `(S)(ref s)`. Now unwrapped, consistent with the existing `DirectionExpression` unwrap sites in `TranslateTarget` / the `LdObj` path (which also match on the syntax shape alone, so the new site follows precedent).

### Verification

Beyond reading the diff, I reproduced the before/after independently (macOS, `net11.0`, Debug):

* Copied the new fixture onto the parent commit (`6d213669a`) and rebuilt: 6 `DeconstructionTests` variants fail there with `Debug.Fail` in `ValidatePattern` at `LoadInstructions[0].IsDescendantOf(this)`, plus `DecompilerException: Error decompiling ... NestedDeconstruction_DiscardedElement`. The fixture is genuinely red before the fix, not just an addition.
* The same variants pass on the branch.
* Full suite run on both trees with the failure sets diffed: **0 new failures, 6 fixed** - exactly the crashing variants. (The remaining failures are macOS environment limits - net40 targets, legacy Roslyn reference assemblies, Windows-only ILAsm - and are identical on both sides.)

Ordering logic checks out: a forwarding assignment is itself `stloc fresh(ldloc v)` with `v` in `deconstructionResultsLookup`, so it resolves through the first branch of `GetAssignmentIndex` and later insertions position correctly against earlier ones and against the `AddMissingAssignmentsForConversions` fixups. Inserting in pattern order also incidentally corrects a latent mis-ordering the old append-at-end path could produce when a conversion-only element preceded a forwarded trailing element.

The delayed-action chain is safe: the forwarding actions are appended last in `MatchAssignments`, so by the time they run, `addAssignment` / tuple-access / conversion fixups have already reparented their instructions into `deconstructInst.Assignments`, which is what makes the `IsDescendantOf` guard meaningful. Element values are stable (`StoreCount == 0`), so moving a load earlier in time cannot change what it reads.

### Findings

All minor - nothing blocking. Comments left at the relevant lines.

1. `DeconstructionTransform.cs:410` - the `TryGetValue` miss returns silently *after* the transform has been committed, so it cannot abort; it would re-create exactly the malformed instruction this PR fixes. Currently unreachable, so an assert would document that better than a silent return.
2. `DeconstructionTransform.cs:404` / `:425` - tuple-element variables are never registered in `deconstructionResultsLookup`, so `GetAssignmentIndex` returns `int.MaxValue` for every assignment of a tuple deconstruction. Harmless only because the eager `LoadCount` check keeps forwarding from firing on that path - an undocumented coupling worth a line of comment.
3. `Correctness/DeconstructionTests.cs:202` - the `ExpressionBuilder` fix has no coverage independent of the nesting fix, and no `Pretty` test pins the emitted shape.

### Conventions

Commit message, `Assisted-by:` trailer, ASCII-only, en-US, no new files needing license headers, comments self-contained - all fine. The build is warning-clean.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
```diff
@@ -390,27 +390,53 @@ bool MatchAssignments(Block block, ref int pos,
...
+					// The assignment is inserted in pattern order, because StatementBuilder and
+					// ExpressionBuilder pair pattern variables with assignments positionally.
+					if (v?.LoadCount != 1)
+						continue;
+					delayedActions += (DeconstructInstruction deconstructInst) => {
+						var load = v.LoadInstructions[0];
+						if (load.IsDescendantOf(deconstructInst))
+							return;
+						if (!deconstructionResultsLookup.TryGetValue(v, out int index))
+							return;
```
christophwille: Minor: this bail-out cannot actually protect anything. Delayed actions run from `TransformDeconstruction` *after* `context.Step`, once `replacement` is already built and about to be written into the block - so returning here leaves the pattern variable's load outside the `DeconstructInstruction`, i.e. exactly the malformed shape this PR fixes, just without the transform having declined it.

In practice it is unreachable: `MatchDeconstruction` registers every element of `deconstructionResults` in the lookup before returning, and the tuple path is filtered out earlier by the `LoadCount` check. Given that, a `Debug.Assert` (or dropping the guard entirely) states the invariant instead of quietly swallowing a violation of it.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
```diff
@@ -390,27 +390,53 @@ bool MatchAssignments(Block block, ref int pos,
...
+					// if it is used directly (and only once!) after the deconstruction. This
+					// happens for trailing elements, but also for leading elements, e.g., when
+					// a nested deconstruction copies the inner element to a temporary before
+					// the elements preceding it are used. Forward such elements through a fresh
+					// variable assigned inside the deconstruction, so that every pattern
+					// variable's load is a descendant of the deconstruct instruction.
+					// The assignment is inserted in pattern order, because StatementBuilder and
+					// ExpressionBuilder pair pattern variables with assignments positionally.
+					if (v?.LoadCount != 1)
+						continue;
```
christophwille, line 413: Worth a line of comment here, because two non-obvious things have to line up for the tuple case to stay correct.

`LoadCount` is read eagerly, at match time, while the decision is applied later from the delayed action. For a tuple deconstruction the elements are the fresh `E_i` variables created in `FindIndex`, and their loads only come into existence when the delayed `inst.ReplaceWith(new LdLoc(...))` runs - so `LoadCount` is still 0 here and forwarding never fires on that path. Same as before this PR, so no behaviour change.

That matters because `E_i` is never added to `deconstructionResultsLookup` (only `MatchDeconstruction` and `MatchConversions` populate it). If forwarding ever did fire for a tuple, `GetAssignmentIndex` would return `int.MaxValue` for *every* assignment in the block, `insertPos` would stay 0, and the forwarding assignment would land at the front - mispairing designators with assignments, since both builders consume them positionally. So the eager read is load-bearing, not incidental, and moving the check into the delayed action (where it reads more naturally) would silently break tuples.

ICSharpCode.Decompiler.Tests/TestCases/Correctness/DeconstructionTests.cs
```diff
@@ -137,6 +149,72 @@ public void Test()
...
+				Console.WriteLine("NestedOuter.Deconstruct");
+				x = Value;
+				inner = new NestedInner { Value = Value * 10 };
+			}
+		}
+
+		public void NestedDeconstruction_Assignment(NestedOuter o)
+		{
+			Console.WriteLine("NestedDeconstruction_Assignment:");
+			var (x, (a, b)) = o;
```
christophwille, line 211: Coverage suggestion: this is currently the only thing exercising the `ExpressionBuilder` half of the PR, and only as a side effect of the nested case. A plain non-nested `var (a, b) = someStruct;` would pin that fix on its own - useful since the two fixes are independent and the struct one reproduces without nesting.

It would also explain why the bug survived this long. `DeconstructionSource` in both fixtures is a `class`, and every assignment-form deconstruction in `Pretty/DeconstructionTests.cs` goes through it or through tuples - so nothing took the address of a struct receiver. `DeconstructDictionaryForEach` does deconstruct a `KeyValuePair`, but the `foreach` shape is built by `StatementBuilder.TranslateDeconstructionDesignation`, which never translates `Pattern.TestedOperand` and so never reaches `VisitDeconstructInstruction`.

A `Pretty` case would additionally lock in the emitted shape (including the `inner.Deconstruct(out var a, out var b);` form noted in the description), so a future change in the transform that starts reconstructing nested designations shows up as an expected-output diff rather than silently.

ICSharpCode.Decompiler.Tests/TestCases/Correctness/DeconstructionTests.cs
```diff
@@ -6,6 +6,18 @@
 
 namespace ICSharpCode.Decompiler.Tests.TestCases.Correctness
 {
+	static class KeyValuePairExtensions
+	{
+		// KeyValuePair<K, V>.Deconstruct does not exist on all target frameworks of the
```
siegfriedpammer: Drop this comment completely

PR #3950, opened 2026-08-01
Reconstruct nested deconstruction designations

siegfriedpammer:
Builds on #3949 (merged), rebased onto current master.

After #3949, nested deconstruction decompiled without crashing but unsugared: `var (x, (a, b)) = o;` came out as a flat deconstruction followed by an explicit `inner.Deconstruct(out var a, out var b);` call. The IL pattern node (`MatchInstruction` with `IsDeconstructCall` sub-patterns), its invariants, and the statement/expression builders all already support nested patterns — only `DeconstructionTransform` never built them.

**Change** (two commits — the chain-matching mechanism, then the rule changes that let chains nest): in `DeconstructionTransform`, `MatchDeconstruction` now consumes chained `Deconstruct` calls — including the defensive copy Roslyn emits for struct elements — into nested match patterns, recursively. Leaves get flat indices in depth-first order, which is exactly the order `StatementBuilder`/`ExpressionBuilder` pair assignments with designators, so conversion/assignment matching runs unchanged on top. Three matching-rule adjustments follow from the chain being consumed: a call pattern no longer needs a matched assignment (single-use leaves are covered by the existing forwarding fixup); an unrelated assignment ends a call pattern instead of rejecting it (tuple patterns still reject — their element list is discovered from the assignments, ending early would misread a suffix); and a pattern is not rooted on an element of an enclosing deconstruction (blocks are processed back to front — the inner call is visited first and would otherwise sugar piecemeal, starving the outer call).

The unrelated-assignment rule is keyed on the pattern being call-rooted, not on nesting, so it also fixes *flat* custom deconstructions: `var (a, b) = GetSource(...);` followed by any unrelated assignment previously decompiled as an explicit `Deconstruct` call and now sugars (pinned by the `LocalVariable_NoConversion_Custom_UnrelatedAssignmentAfter` fixture).

One production change lands outside the transform, as its own commit carrying its two pointer-target Pretty fixtures: `DeconstructInstruction.IsAssignment` falls back to `stobj.Type` when a pointer target was materialized through a stack slot (whose type erases to unknown during target evaluation). This independently enables deconstruction into pointer targets — `(*p, value) = tuple;` — including plain tuples with no nesting at all.

**Nested tuples too**: the second half of the series extends the nesting to pure tuple chains (fixtures, then the temporary-consuming mechanism, then the deferral guard) — `var (x, (a, b)) = GetTuple<int, (int, int)>();`, which Roslyn lowers to one temporary per nested designation (parents before children) plus element reads in depth-first leaf order, previously decompiled as a flat deconstruction plus separate element statements. The matcher now consumes the temporaries into a tuple-node tree with flat depth-first leaf indices (the same design as the call nesting). An element variable that escapes the deconstruction (used after the statement) demotes back to a designator leaf via a blacklist-and-retry, and a dry-run guard — the tuple analogue of the call path's — defers inner flat matches to the enclosing attempt. Two IL realities drove the matcher details: earlier transforms rewrite non-escaping element reads from `ldloca` to `ldloc`, and stack-slot temporaries carry imprecise types (the container's element type is authoritative; the match variable is retyped so the `IsDeconstructTuple` invariant holds). One production change again lands outside the transform, in its own commit: `TupleType.FromUnderlyingType` threw `NullReferenceException` on non-tuple input instead of returning null as documented.

**Scope**: call-under-call and tuple-in-tuple nesting, any depth, assignment + foreach forms, discards, conversions on leaves. Mixed containers (a call pattern absorbing tuple elements, or a custom-deconstructed element inside a tuple pattern) intentionally stay flat, as do by-ref extension receivers (`Deconstruct(this in S, ...)`).

**Tests**: each feature's fixtures land ahead of its implementation and are red until it. For the call nesting: 20 Pretty fixtures (the flat unrelated-assignment form, struct/class inners, both-elements-nested, depth 3, inner discard, nested/typed/nullable conversions, `System.Tuple` sources, the two bail-out shapes `ElementDeconstructedAfterBarrier`/`OuterElementUsedTwice`, two foreach forms incl. `KeyValuePair` extension Deconstruct; the two pointer-target forms ride in the pointer commit) and 18 Correctness cases with printed `Deconstruct` calls pinning evaluation order at runtime, including member hiding (`NestedDeconstruction_HiddenDeconstructMethod` pins the `BindsOnElementType` gate), side-effecting LHS targets, checked/nullable conversions, `in`-parameter, conditional and generic-constraint sources, and the tuple-with-custom-element corruption repro. For the tuple nesting: 6 Pretty fixtures (depth 2 and 3, both elements nested, conversions on inner leaves, the escaping-element case pinned per config, foreach) and 5 Correctness cases with runtime-value pins. Every commit in the series builds, and at every point the only failing tests are the not-yet-implemented spec fixtures (verified per intermediate). Full decompiler suite at the head: 3332 tests, 0 failures.

Output polish for #3803/#3388-adjacent cases; closes nothing by itself.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Correctness/DeconstructionTests.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Correctness/DeconstructionTests.cs
@@ -182,6 +182,51 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Correctness
 			NestedDeconstruction_ForEachDictionary_Conversions(new Dictionary<string, NestedInner> {
 				{ "k1", new NestedInner { Value = 19 } }
 			});
+			NestedTupleDeconstruction_Values((20, (21, 22)));
+			NestedTupleDeconstruction_Depth3((23, (24, (25, 26))));
+			NestedTupleDeconstruction_Conversions((27, (28, 29)));
+			NestedTupleDeconstruction_ElementUsedOutside((30, (31, 32)));
+			NestedTupleDeconstruction_ForEach(new List<(int, (int, int))> {
+				(33, (34, 35)),
+				(36, (37, 38))
+			});
+		}
+
+		public void NestedTupleDeconstruction_Values((int, (int, int)) t)
+		{
+			Console.WriteLine("NestedTupleDeconstruction_Values:");
+			var (x, (a, b)) = t;
+			Console.WriteLine(x + " " + a + " " + b);
+		}
+
+		public void NestedTupleDeconstruction_Depth3((int, (int, (int, int))) t)
+		{
+			Console.WriteLine("NestedTupleDeconstruction_Depth3:");
+			var (x, (a, (b, c))) = t;
+			Console.WriteLine(x + " " + a + " " + b + " " + c);
+		}
+
+		public void NestedTupleDeconstruction_Conversions((int, (int, int)) t)
+		{
+			Console.WriteLine("NestedTupleDeconstruction_Conversions:");
+			(long x, (long a, long b)) = t;
+			Console.WriteLine(x + " " + a + " " + b);
+		}
+
+		public void NestedTupleDeconstruction_ElementUsedOutside((int, (int, int)) t)
+		{
+			Console.WriteLine("NestedTupleDeconstruction_ElementUsedOutside:");
+			var (x, inner) = t;
+			Console.WriteLine(x + " " + inner.Item1 + " " + inner.Item2);
+		}
+
+		public void NestedTupleDeconstruction_ForEach(List<(int, (int, int))> list)
+		{
+			Console.WriteLine("NestedTupleDeconstruction_ForEach:");
+			foreach (var (x, (a, b)) in list)
+			{
+				Console.WriteLine(x + " " + a + " " + b);
+			}
 		}
 
 		public class ConstrainedSource
--- a/ICSharpCode.Decompiler.Tests/TestCases/Correctness/DeconstructionTests.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Correctness/DeconstructionTests.cs
@@ -15,6 +15,15 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Correctness
 		}
 	}
 
+	static class TupleClassExtensions
+	{
+		public static void Deconstruct<T1, T2>(this Tuple<T1, T2> tuple, out T1 item1, out T2 item2)
+		{
+			item1 = tuple.Item1;
+			item2 = tuple.Item2;
+		}
+	}
+
 	class DeconstructionTests
 	{
 		public static void Main()
@@ -153,6 +162,206 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Correctness
 				new NestedOuter { Value = 2 }
 			});
 			NestedDeconstruction_DiscardedElement(new KeyValuePair<object, DiscardData>("key", default(DiscardData)));
+			NestedDeconstruction_ClassInner(new ClassInnerOuter { Value = 7 });
+			NestedDeconstruction_Depth3(new DeepOuter { Value = 3 });
+			NestedDeconstruction_LhsSideEffects_DeconstructionOrder_Assignments();
+			NestedDeconstruction_Conversions_AfterAllDeconstructCalls();
+			NestedDeconstruction_TypedDeclaration_Conversions(new NestedOuter { Value = 5 });
+			NestedDeconstruction_DiscardWithSideEffectTargets();
+			NestedDeconstruction_SystemTupleSource(Tuple.Create(8, new NestedInner { Value = 4 }));
+			NestedDeconstruction_CheckedConversions(new NestedOuter { Value = 9 });
+			NestedDeconstruction_GenericConstraintSource(new ConstrainedSource { Value = 11 });
+			NestedDeconstruction_InParameterSource(new NestedOuter { Value = 12 });
+			NestedDeconstruction_ConditionalSource(c: true, new NestedOuter { Value = 13 }, new NestedOuter { Value = 14 });
+			NestedDeconstruction_TupleOuterConversions((15, new NestedInner { Value = 6 }));
+			NestedDeconstruction_TypedConversions_UnrelatedCallAfter(new NestedOuter { Value = 16 });
+			NestedDeconstruction_NullableConversions(new NestedOuter { Value = 17 });
+			NestedDeconstruction_MyIntConversionOnNestedLeaves(new NestedOuter { Value = 18 });
+			NestedDeconstruction_ForEachDictionary_Conversions(new Dictionary<string, NestedInner> {
+				{ "k1", new NestedInner { Value = 19 } }
+			});
+		}
+
+		public class ConstrainedSource
+		{
+			public int Value;
+
+			public void Deconstruct(out int a, out NestedInner inner)
+			{
+				Console.WriteLine("ConstrainedSource.Deconstruct");
+				a = Value;
+				inner = new NestedInner { Value = Value * 10 };
+			}
+		}
+
+		public void NestedDeconstruction_SystemTupleSource(Tuple<int, NestedInner> tup)
+		{
+			Console.WriteLine("NestedDeconstruction_SystemTupleSource:");
+			(long x, (long a, long b)) = tup;
+			int z = Side();
+			Console.WriteLine(x + " " + a + " " + b + " " + z);
+		}
+
+		public void NestedDeconstruction_CheckedConversions(NestedOuter o)
+		{
+			Console.WriteLine("NestedDeconstruction_CheckedConversions:");
+			checked
+			{
+				(long x, (long a, long b)) = o;
+				Console.WriteLine(x + " " + a + " " + b);
+			}
+		}
+
+		public void NestedDeconstruction_GenericConstraintSource<T>(T o) where T : ConstrainedSource
+		{
+			Console.WriteLine("NestedDeconstruction_GenericConstraintSource:");
+			var (a, (c, d)) = o;
+			Console.WriteLine(a + " " + c + " " + d);
+		}
+
+		public void NestedDeconstruction_InParameterSource(in NestedOuter o)
+		{
+			Console.WriteLine("NestedDeconstruction_InParameterSource:");
+			(long x, (long a, long b)) = o;
+			Console.WriteLine(x + " " + a + " " + b);
+		}
+
+		public void NestedDeconstruction_ConditionalSource(bool c, NestedOuter o1, NestedOuter o2)
+		{
+			Console.WriteLine("NestedDeconstruction_ConditionalSource:");
+			(long x, (int a, int b)) = c ? o1 : o2;
+			int z = Side();
+			Console.WriteLine(x + " " + a + " " + b + " " + z);
+		}
+
+		public void NestedDeconstruction_TupleOuterConversions((int, NestedInner) tup)
+		{
+			Console.WriteLine("NestedDeconstruction_TupleOuterConversions:");
+			(long x, (long a, long b)) = tup;
+			Console.WriteLine(x + " " + a + " " + b);
+		}
+
+		public void NestedDeconstruction_TypedConversions_UnrelatedCallAfter(NestedOuter o)
+		{
+			Console.WriteLine("NestedDeconstruction_TypedConversions_UnrelatedCallAfter:");
+			(long x, (long a, long b)) = o;
+			int z = Side();
+			Console.WriteLine(x + " " + a + " " + b + " " + z);
+		}
+
+		public void NestedDeconstruction_NullableConversions(NestedOuter o)
+		{
+			Console.WriteLine("NestedDeconstruction_NullableConversions:");
+			(long? x, (long? a, int? b)) = o;
+			Console.WriteLine(x + " " + a + " " + b);
+		}
+
+		public void NestedDeconstruction_MyIntConversionOnNestedLeaves(NestedOuter o)
+		{
+			Console.WriteLine("NestedDeconstruction_MyIntConversionOnNestedLeaves:");
+			(MyInt x, (MyInt a, long b)) = o;
+			Console.WriteLine(x + " " + a + " " + b);
+		}
+
+		public void NestedDeconstruction_ForEachDictionary_Conversions(Dictionary<string, NestedInner> d)
+		{
+			Console.WriteLine("NestedDeconstruction_ForEachDictionary_Conversions:");
+			foreach ((string k, (long a, long b)) in d)
+			{
+				Console.WriteLine(k + " " + a + " " + b);
+			}
+		}
+
+		// The evaluation order of a deconstruction-assignment is: (1) all side-effects of
+		// the left-hand-side targets, (2) all Deconstruct invocations, (3) conversions,
+		// (4) assignments. Get(i), the Deconstruct methods, MyInt's implicit conversions,
+		// and the property setters all print, so any phase reordering breaks the output diff.
+		public void NestedDeconstruction_LhsSideEffects_DeconstructionOrder_Assignments()
+		{
+			Console.WriteLine("NestedDeconstruction_LhsSideEffects_DeconstructionOrder_Assignments:");
+			(Get(0).IntProperty, (Get(1).IntProperty, Get(2).IntProperty)) = new NestedOuter { Value = 11 };
+		}
+
+		public void NestedDeconstruction_Conversions_AfterAllDeconstructCalls()
+		{
+			Console.WriteLine("NestedDeconstruction_Conversions_AfterAllDeconstructCalls:");
+			(Get(0).My, (Get(1).IntProperty, Get(2).My)) = new NestedOuter { Value = 21 };
+		}
+
+		public void NestedDeconstruction_TypedDeclaration_Conversions(NestedOuter o)
+		{
+			Console.WriteLine("NestedDeconstruction_TypedDeclaration_Conversions:");
+			(MyInt x, (long a, MyInt b)) = o;
+			Console.WriteLine(x);
+			Console.WriteLine(a);
+			Console.WriteLine(b);
+		}
+
+		public void NestedDeconstruction_DiscardWithSideEffectTargets()
+		{
+			Console.WriteLine("NestedDeconstruction_DiscardWithSideEffectTargets:");
+			(Get(0).IntProperty, (_, Get(1).My)) = new NestedOuter { Value = 31 };
+		}
+
+		public int Side()
+		{
+			Console.WriteLine("Side()");
+			return 5;
+		}
+
+		public class NestedClassInner
+		{
+			public int Value;
+
+			public void Deconstruct(out int a, out int b)
+			{
+				Console.WriteLine("NestedClassInner.Deconstruct");
+				a = Value + 1;
+				b = Value + 2;
+			}
+		}
+
+		public struct ClassInnerOuter
+		{
+			public int Value;
+
+			public void Deconstruct(out int x, out NestedClassInner inner)
+			{
+				Console.WriteLine("ClassInnerOuter.Deconstruct");
+				x = Value;
+				inner = new NestedClassInner { Value = Value * 10 };
+			}
+		}
+
+		public void NestedDeconstruction_ClassInner(ClassInnerOuter o)
+		{
+			Console.WriteLine("NestedDeconstruction_ClassInner:");
+			var (x, (a, b)) = o;
+			Console.WriteLine(x);
+			Console.WriteLine(a);
+			Console.WriteLine(b);
+		}
+
+		public struct DeepOuter
+		{
+			public int Value;
+
+			public void Deconstruct(out int x, out ClassInnerOuter mid)
+			{
+				Console.WriteLine("DeepOuter.Deconstruct");
+				x = Value;
+				mid = new ClassInnerOuter { Value = Value * 100 };
+			}
+		}
+
+		public void NestedDeconstruction_Depth3(DeepOuter o)
+		{
+			Console.WriteLine("NestedDeconstruction_Depth3:");
+			var (x, (y, (a, b))) = o;
+			Console.WriteLine(x);
+			Console.WriteLine(y);
+			Console.WriteLine(a);
+			Console.WriteLine(b);
 		}
 
 		public struct DiscardData
--- a/ICSharpCode.Decompiler.Tests/TestCases/Correctness/DeconstructionTests.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Correctness/DeconstructionTests.cs
@@ -169,6 +169,7 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Correctness
 			NestedDeconstruction_TypedDeclaration_Conversions(new NestedOuter { Value = 5 });
 			NestedDeconstruction_DiscardWithSideEffectTargets();
 			NestedDeconstruction_HiddenDeconstructMethod(default(HidingOuter));
+			NestedDeconstruction_TupleWithCustomElement((7, new NestedInner { Value = 3 }));
 			NestedDeconstruction_SystemTupleSource(Tuple.Create(8, new NestedInner { Value = 4 }));
 			NestedDeconstruction_CheckedConversions(new NestedOuter { Value = 9 });
 			NestedDeconstruction_GenericConstraintSource(new ConstrainedSource { Value = 11 });
@@ -353,6 +354,17 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Correctness
 		{
 			Console.WriteLine("Side()");
 			return 5;
+		}
+
+		// A tuple deconstruction whose element is custom-deconstructed, followed by an
+		// unrelated assignment: the tuple part must not be consumed into a pattern rooted
+		// in the element's Deconstruct call.
+		public void NestedDeconstruction_TupleWithCustomElement((int, NestedInner) tup)
+		{
+			Console.WriteLine("NestedDeconstruction_TupleWithCustomElement:");
+			var (x, (_, _)) = tup;
+			int z = Side();
+			Console.WriteLine(x * x + z);
 		}
 
 		public class NestedClassInner
--- a/ICSharpCode.Decompiler.Tests/TestCases/Correctness/DeconstructionTests.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Correctness/DeconstructionTests.cs
@@ -168,6 +168,7 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Correctness
 			NestedDeconstruction_Conversions_AfterAllDeconstructCalls();
 			NestedDeconstruction_TypedDeclaration_Conversions(new NestedOuter { Value = 5 });
 			NestedDeconstruction_DiscardWithSideEffectTargets();
+			NestedDeconstruction_HiddenDeconstructMethod(default(HidingOuter));
 			NestedDeconstruction_SystemTupleSource(Tuple.Create(8, new NestedInner { Value = 4 }));
 			NestedDeconstruction_CheckedConversions(new NestedOuter { Value = 9 });
 			NestedDeconstruction_GenericConstraintSource(new ConstrainedSource { Value = 11 });
@@ -301,6 +302,51 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Correctness
 		{
 			Console.WriteLine("NestedDeconstruction_DiscardWithSideEffectTargets:");
 			(Get(0).IntProperty, (_, Get(1).My)) = new NestedOuter { Value = 31 };
+		}
+
+		public class HidingBase
+		{
+			public int Value;
+
+			public void Deconstruct(out string a, out double b)
+			{
+				Console.WriteLine("HidingBase.Deconstruct");
+				a = "base" + Value;
+				b = 0.5;
+			}
+		}
+
+		public class HidingDerived : HidingBase
+		{
+			public new void Deconstruct(out string a, out double b)
+			{
+				Console.WriteLine("HidingDerived.Deconstruct");
+				a = "derived";
+				b = 99.5;
+			}
+		}
+
+		public struct HidingOuter
+		{
+			public void Deconstruct(out int x, out HidingDerived d)
+			{
+				Console.WriteLine("HidingOuter.Deconstruct");
+				x = 1;
+				d = new HidingDerived { Value = 5 };
+			}
+		}
+
+		// The base-typed view forces the call to bind to HidingBase.Deconstruct; a nested
+		// designation cannot express that, because it rebinds on the element's static type,
+		// where the hiding method wins.
+		public void NestedDeconstruction_HiddenDeconstructMethod(HidingOuter o)
+		{
+			Console.WriteLine("NestedDeconstruction_HiddenDeconstructMethod:");
+			var (_, d) = o;
+			HidingBase b = d;
+			var (a, c) = b;
+			Console.WriteLine(a);
+			Console.WriteLine(c);
 		}
 
 		public int Side()
--- a/ICSharpCode.Decompiler/IL/Instructions/DeconstructInstruction.cs
+++ b/ICSharpCode.Decompiler/IL/Instructions/DeconstructInstruction.cs
@@ -259,12 +259,41 @@ namespace ICSharpCode.Decompiler.IL
 					if (stobj.Target.InferType(typeSystem) is ByReferenceType brt)
 						expectedType = brt.ElementType;
 					else
-						expectedType = SpecialType.UnknownType;
+					{
+						// Pointer targets do not infer a ByReferenceType. stobj.Type cannot stand
+						// in for the element type: it comes from the store opcode, which is
+						// sign-agnostic, so a uint* and an int* both report int32 and a byte*
+						// reports sbyte. Recover the declared type instead, and only fall back to
+						// the store where the target is not a pointer at all.
+						expectedType = GetPointerElementType(stobj.Target, typeSystem) ?? stobj.Type;
+					}
 					value = stobj.Value;
 					return true;
 				default:
 					return false;
 			}
+		}
+
+		/// <summary>
+		/// The element type of a pointer-typed target. A pointer passing through a stack slot
+		/// is typed IntPtr there, so the declared type has to be taken from the definition the
+		/// slot was filled from. Returns null if the target is not a pointer.
+		/// </summary>
+		static IType GetPointerElementType(ILInstruction target, ICompilation typeSystem)
+		{
+			// Bounded because a definition chain could be cyclic in invalid IL.
+			for (int step = 0; step < 4; step++)
+			{
+				if (target.InferType(typeSystem) is PointerType pointerType)
+					return pointerType.ElementType;
+				if (!target.MatchLdLoc(out var v) || !v.IsSingleDefinition
+					|| v.StoreInstructions.Count != 1 || !(v.StoreInstructions[0] is StLoc store))
+				{
+					return null;
+				}
+				target = store.Value;
+			}
+			return null;
 		}
 
 		internal override void CheckInvariant(ILPhase phase)
--- a/ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
+++ b/ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
@@ -20,11 +20,9 @@
 
 using System;
 using System.Collections.Generic;
-using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
-using System.Resources;
 
 using ICSharpCode.Decompiler.CSharp.Resolver;
 using ICSharpCode.Decompiler.TypeSystem;
@@ -33,8 +31,47 @@ using ICSharpCode.Decompiler.Util;
 namespace ICSharpCode.Decompiler.IL.Transforms
 {
 	/// <summary>
-	/// 
+	/// Detects that a run of statements is a lowered deconstruction assignment - rooted in a
+	/// Deconstruct call or in tuple element reads, including nested designations - and folds
+	/// it into a single DeconstructInstruction.
 	/// </summary>
+	/*
+		stloc tuple(call MakeIntIntTuple(ldloc this))
+	----
+		stloc myInt(call op_Implicit(ldfld Item2(ldloca tuple)))
+		stloc a(ldfld Item1(ldloca tuple))
+		stloc b(ldloc myInt)
+	==>
+		deconstruct {
+			init:
+				<empty>
+			deconstruct:
+				match.deconstruct(temp = ldloca tuple) {
+					match(result0 = deconstruct.result 0(temp)),
+					match(result1 = deconstruct.result 1(temp))
+				}
+			conversions: {
+				stloc conv2(call op_Implicit(ldloc result1))
+			}
+			assignments: {
+				stloc a(ldloc result0)
+				stloc b(ldloc conv2)
+			}
+		}
+
+		A nested designation over Deconstruct calls (var (x, (a, b)) = o;) chains the calls,
+		with a defensive copy for struct elements:
+			call Deconstruct(ldloc o, ldloca x', ldloca inner)
+			call Deconstruct(ldloca inner, ldloca a', ldloca b')
+			...conversions/assignments over the leaves x', a', b'...
+
+		A nested designation over tuples (var (x, (a, b)) = t;) is lowered to one temporary
+		per nested designation, followed by element reads in depth-first leaf order:
+			stloc inner(ldobj(ldflda Item2(ldloca t)))
+			stloc x(ldobj(ldflda Item1(ldloca t)))
+			stloc a(ldobj(ldflda Item1(ldloca inner)))
+			stloc b(ldobj(ldflda Item2(ldloca inner)))
+	 * */
 	class DeconstructionTransform : IStatementTransform
 	{
 		StatementTransformContext context = null!;
@@ -43,30 +80,6 @@ namespace ICSharpCode.Decompiler.IL.Transforms
 		ILVariable? tupleVariable;
 		TupleType? tupleType;
 
-		/*
-			stloc tuple(call MakeIntIntTuple(ldloc this))
-		----
-			stloc myInt(call op_Implicit(ldfld Item2(ldloca tuple)))
-			stloc a(ldfld Item1(ldloca tuple))
-			stloc b(ldloc myInt)
-		==>
-			deconstruct {
-				init:
-					<empty>
-				deconstruct:
-					match.deconstruct(temp = ldloca tuple) {
-						match(result0 = deconstruct.result 0(temp)),
-						match(result1 = deconstruct.result 1(temp))
-					}
-				conversions: {
-					stloc conv2(call op_Implicit(ldloc result1))
-				}
-				assignments: {
-					stloc a(ldloc result0)
-					stloc b(ldloc conv2)
-				}
-			}
-		 * */
 		void IStatementTransform.Run(Block block, int pos, StatementTransformContext context)
 		{
 			if (!context.Settings.Deconstruction)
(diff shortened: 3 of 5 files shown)


christophwille:
## Code review

### What this does

`DeconstructionTransform` now reconstructs nested designations. `MatchDeconstruction` consumes chained `Deconstruct` calls (including Roslyn's defensive copy for struct elements) into a recursive `DeconstructionCall` tree, `BuildPatternMatch` turns that into nested `MatchInstruction` sub-patterns, and leaves get flat depth-first indices so the existing conversion/assignment matching keeps working unchanged. Three matching rules relax as a consequence, plus `IsConsumableByEnclosingDeconstruction` defers an inner call to its enclosing one (blocks are walked back to front, so the inner call is reached first).

The IL-side design is sound: depth-first leaf indexing is exactly the order `ConstructTuple` / `ConstructDesignation` consume assignments; nested receiver variables end up with `LoadCount == SubPatterns.Count`, so `MatchInstruction.HasDesignator` stays false and `ValidatePattern` is satisfied; and `DeconstructResultInstruction` correctly keeps per-call indices while the lookup keeps flat ones - two separate index spaces that never mix.

### Verification

Beyond reading the diff (macOS, `net11.0`, Debug):

* **TDD claim holds.** Built the test-only commit (`f8694ae63`) on its own: 6 `DeconstructionTests` variants fail there, all 6 green on the implementation commit. Red-then-green, as described.
* **No regressions.** Full suite on `f96283358` and on the base `555e364d8`, failure sets diffed: **identical**, zero new failures, on 3332 tests. (The environmental failures are macOS-only - net40 targets, legacy Roslyn reference assemblies, Windows ILAsm - and match on both sides.)
* **Riskiest new path checked.** The relaxed "unrelated assignment ends the pattern" rule can leave a later element without an assignment, so the #3949 forwarding fixup pulls its load into the deconstruct instruction. Fed the transform an interleaved case (`x = a; z = GetInt(); y = b;`) and the output is correct - the element is forwarded through a fresh variable and statement order is preserved.
* **The `IsAssignment` hunk is load-bearing.** Reverted it in isolation: the 6 runnable variants fail, with `Pointer_NoConversion_Tuple` and `Pointer_Nested_Custom` regressing to unsugared form.

Test coverage is considerably better than the description advertises ("9 Pretty fixtures ... plus class-inner and depth-3 Correctness cases"): there are ~18 of each, including member hiding pinning `BindsOnElementType`, evaluation-order pinning with side-effecting LHS targets, checked/nullable conversions, `in` parameter and conditional sources, and the two bail-out paths (`ElementDeconstructedAfterBarrier`, `OuterElementUsedTwice`). Worth updating the description - the extra cases are the strongest argument for the PR.

### Findings

Nothing blocking; no correctness defect found. Comments left at the relevant lines.

1. **`DeconstructionTransform.cs:591` - the biggest user-visible win in this PR is untested and unadvertised.** `allowUnrelatedAssignments` is keyed on `rootCall != null`, not on nesting, so it also fixes flat deconstructions. Measured on the base branch: `var (a, b) = GetSource<MyInt?, MyInt>(); int v = GetInt();` decompiles as an explicit `Deconstruct` call today, and sugars correctly with this PR. Every fixture for the rule is nested, so a later narrowing would silently regress the flat form.
2. **`DeconstructInstruction.cs:266` - second production change, outside the stated scope.** The description says the change is "all in `DeconstructionTransform`". This hunk independently enables deconstruction into pointer targets, including for plain tuples with no nesting at all.
3. **`DeconstructionTransform.cs:345` - the bail-out comment overstates what the dry run proves.** It establishes that the chain is consumable, not that the enclosing attempt will succeed; `MatchConversions`/`MatchAssignments` can still reject there, and the back-to-front walk gives no second chance. I could not construct a case that actually loses sugar, so this is about the comment, not a defect.

Minor: `MatchDeconstruction` (line 311) returns `bool` but the result is discarded at its only call site, which tests `rootCall != null` instead - `void` would read more honestly.

### Conventions

Commit messages, `Assisted-by:` trailer, ASCII-only, en-US, self-contained comments, no new files needing license headers, warning-clean build. The two-commit split (spec, then implementation) matches the repo's TDD rule and genuinely verifies.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
```diff
@@ -349,14 +577,27 @@ bool MatchConversion(ILInstruction inst, out ILInstruction inputInstruction,
...
 				int index = FindIndex(valueInst, out var tupleAccessAdjustment);
+				if (index < 0 && allowUnrelatedAssignments)
+				{
+					// For a Deconstruct call the element list is fixed by the call's
+					// out-arguments, so an assignment whose value is unrelated to the
+					// deconstruction just ends the pattern and stays after the deconstruct
+					// instruction. (For tuples the elements are discovered from the
+					// assignments, so ending early would misread a suffix as the pattern:
+					// keep rejecting there.)
+					break;
```
christophwille, line 818: This rule is keyed on `rootCall != null`, not on whether anything nested was consumed - so it applies to *every* call-rooted deconstruction, flat ones included. That turns out to be the largest user-visible improvement in the PR, and it has no test.

Measured it on the base branch by adding this fixture:

```csharp
public void Flat_UnrelatedAssignmentAfter()
{
    var (myInt3, myInt4) = GetSource<MyInt?, MyInt>();
    int value = GetInt();
    Console.WriteLine(myInt3);
    Console.WriteLine(myInt4);
    Console.WriteLine(value);
}
```

On `555e364d8` that decompiles to:

```csharp
GetSource<MyInt?, MyInt>().Deconstruct(out var a, out var b);
MyInt? myInt = a;
MyInt myInt2 = b;
int @int = GetInt();
```

With this PR it comes out as the deconstruction. So today any deconstruction followed by a local declaration loses its sugar - a very common shape, and the fix is not mentioned in the title or the description (which frames the relaxation purely as a consequence of consuming the chain).

Every fixture exercising this rule is nested (`LocalVariable_Nested_TypedConversions_UnrelatedCallAfter`, `NestedDeconstruction_TypedConversions_UnrelatedCallAfter`, ...). If someone later narrows the condition to something like `rootCall.NestedCalls.Any(c => c != null)` - which reads like a safe tightening - the flat improvement disappears with the whole suite still green. A flat Pretty fixture would pin it.

On the rule itself: I fed the transform an interleaved case (`x = a; z = GetInt(); y = b;`) where breaking early leaves the trailing element without an assignment and the #3949 forwarding fixup has to pull its load in. Output was correct - element forwarded through a fresh variable, statement order preserved.

ICSharpCode.Decompiler/IL/Instructions/DeconstructInstruction.cs
```diff
@@ -259,7 +259,12 @@ internal static bool IsAssignment(ILInstruction inst, ICompilation typeSystem, o
 					if (stobj.Target.InferType(typeSystem) is ByReferenceType brt)
 						expectedType = brt.ElementType;
 					else
-						expectedType = SpecialType.UnknownType;
+					{
+						// Pointer-typed targets (or targets whose pointer type got erased in a
+						// stack slot) do not infer a ByReferenceType; the type of the store
+						// itself is just as precise.
+						expectedType = stobj.Type;
+					}
```
christophwille, line 269: This is a second production change, and the description says the change is "all in `DeconstructionTransform`" - easy for a reviewer to skip past.

It is load-bearing and correct: reverting just this hunk on top of the rest of the branch makes all 6 runnable `DeconstructionTests` variants fail, with `Pointer_NoConversion_Tuple` and `Pointer_Nested_Custom` falling back to unsugared output. But note what that first fixture is - a plain **tuple** deconstruction with no nesting anywhere:

```csharp
public unsafe void Pointer_NoConversion_Tuple(int* p)
{
    int value;
    (*p, value) = GetTuple<int, int>();
```

So this hunk independently lifts a pre-existing limitation (deconstruction into a pointer target) that has nothing to do with nested designations. Given the repo's convention that the commit body carries the *why*, this looks like it wants its own commit and a line in the description.

Blast radius is smaller than it first appears, which is worth stating explicitly somewhere: of the three `IsAssignment` call sites, only `MatchAssignment` reads `expectedType` (feeding `IsCompatibleImplicitConversion`); `CheckInvariant` and `GetAssignmentIndex` both discard it. So the change can only widen what `DeconstructionTransform` accepts - it cannot weaken the invariant check.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
```diff
@@ -263,43 +294,240 @@ bool TransformDeconstruction(Block block, int pos)
...
+			if (!TryFindEnclosingDeconstructionCall(block, pos, out int enclosingPos))
+				return false;
+			var enclosingCall = MatchDeconstructionCall(block.Instructions[enclosingPos], out _);
+			if (enclosingCall == null)
+				return false;
+			// Dry run of the nested matching: only if it consumes the instructions up to and
+			// including the call at pos will the attempt at the enclosing position absorb it.
+			int scanPos = enclosingPos + 1;
+			MatchNestedDeconstructions(block, ref scanPos, enclosingCall);
+			return scanPos > pos;
```
christophwille: The comment at the call site says the check "must be precise" and lists the ways the enclosing call can fail to match - but the dry run only establishes that `MatchNestedDeconstructions` consumes the chain. The enclosing attempt can still bail out afterwards in `MatchConversions` or `MatchAssignments`, and since `StatementTransform.Run` walks positions high-to-low and nothing requests a rerun on failure, the inner call has already been passed over by then. The whole chain would stay unsugared.

I tried to build a case that actually loses sugar relative to the base and could not: the shapes that break the outer's conversion/assignment matching (out-of-order or unrelated assignments interleaved with the leaves) break the inner's matching too, so the inner would not have sugared standalone either. The two call-level bail-outs *are* covered - `LocalVariable_Nested_ElementDeconstructedAfterBarrier` and `LocalVariable_Nested_OuterElementUsedTwice`, both of which correctly fall through to `MatchDeconstructionCall` returning null.

So this is about the comment rather than the code: it promises a precision the dry run does not provide. Something like "the enclosing attempt may still fail later, in which case the chain stays unsugared" would set the right expectation for whoever touches this next.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
siegfriedpammer: The flat form is now pinned: `LocalVariable_NoConversion_Custom_UnrelatedAssignmentAfter` in the Pretty fixtures is exactly this shape, and the description calls out the flat improvement explicitly.

ICSharpCode.Decompiler/IL/Instructions/DeconstructInstruction.cs
siegfriedpammer: Split out as its own commit ("Support deconstruction into pointer targets", currently 1116f5e26), carrying its two Pretty fixtures (`Pointer_NoConversion_Tuple`, `Pointer_Nested_Custom`) and the call-site blast radius in the commit body; the description names the change as well.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
siegfriedpammer: Reworded: the call-site comment now states the dry run only proves the enclosing attempt consumes the chain, that `MatchConversions`/`MatchAssignments` can still reject it with no second chance from the back-to-front walk, and that such a rejection only costs sugar, never correctness.

ICSharpCode.Decompiler.Tests/TestCases/Correctness/DeconstructionTests.cs
```diff
@@ -15,6 +15,18 @@ public static void Deconstruct<TKey, TValue>(this KeyValuePair<TKey, TValue> pai
 		}
 	}
 
+	static class TupleClassExtensions
+	{
+		// TupleExtensions.Deconstruct is not available on all target frameworks of the
```
siegfriedpammer: comment is not needed

siegfriedpammer: Removed; squashed back into "Add nested deconstruction runtime fixtures".

ICSharpCode.Decompiler/IL/Instructions/DeconstructInstruction.cs
```diff
@@ -259,7 +259,12 @@ internal static bool IsAssignment(ILInstruction inst, ICompilation typeSystem, o
 					if (stobj.Target.InferType(typeSystem) is ByReferenceType brt)
 						expectedType = brt.ElementType;
 					else
-						expectedType = SpecialType.UnknownType;
+					{
+						// Pointer-typed targets (or targets whose pointer type got erased in a
+						// stack slot) do not infer a ByReferenceType; the type of the store
+						// itself is just as precise.
+						expectedType = stobj.Type;
```
christophwille: **Correctness (confirmed):** `IsAssignment`'s stobj fallback replaces the `UnknownType` sentinel that rejected stores whose target type cannot be inferred as `ByReferenceType`. `StObj.Type` comes from the IL store opcode and is sign-agnostic relative to the true pointer element type, so a sign-mismatched pointer store now passes the implicit-conversion check that previously rejected the transform.

**Failure scenario:** For `var (el, y) = t; *p = (uint)el;` with `(int, int) t` and `uint* p`, the int-to-uint cast is a no-op in IL (`stind.i4`, so `StObj.Type` = int32). `InferType` of the pointer target yields `PointerType`, not `ByReferenceType`, so `expectedType` becomes int32 and `IsCompatibleImplicitConversion` sees an identity conversion. The transform sugars this into `(*p, y) = t;`, which does not recompile (CS0266: cannot implicitly convert int to uint). On master, `expectedType` was `UnknownType`, the conversion check failed, and correct unsugared code was emitted.

*Found by high-effort multi-agent review; independently verified.*

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
```diff
@@ -187,126 +283,437 @@ bool InlineDeconstructionInitializer(Block block, int pos)
...
+				// its elements are evaluated before the parent's remaining ones
+				pendingElements.Push((nested, 0));
+			}
+
+			static bool IsReceiverReference(ILInstruction target, ILVariable receiver)
+			{
+				return MatchLdLocOrLdLoca(target, out var v) && v == receiver;
+			}
+
+			static bool BindsOnElementType(IMethod method, IType elementType)
```
christophwille, line 567: **Correctness (confirmed):** `BindsOnElementType` only guards against instance-method hiding; it returns true (`method.IsStatic`) for any extension-method `Deconstruct` as long as no instance method of the same arity exists on the element's type chain. It therefore does not detect a more-specific competing extension method that a recompiled nested designation would bind to instead of the one the IL actually calls.

**Failure scenario:** IL calls extension `E1.Deconstruct(this Base, out int, out int)` on an element whose static type is `Derived`, while `E2.Deconstruct(this Derived, out int, out int)` is also in scope. The transform folds the explicit call into `var (x, (a, b)) = o;`, and recompiling that source resolves to E2 instead of E1 -- the decompiled program has different runtime behavior than the original assembly.

*Found by high-effort multi-agent review; independently verified.*

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
```diff
@@ -187,126 +283,437 @@ bool InlineDeconstructionInitializer(Block block, int pos)
...
+				if (!TupleType.GetTupleElementTypes(temp.Type).IsDefaultOrEmpty
+					&& !NormalizeTypeVisitor.TypeErasure.EquivalentTypes(elementType, temp.Type))
+				{
+					break;
+				}
+				var node = new TupleNode(temp, tempType);
+				containerNode.NestedElements[index] = node;
+				// The temporary's store reads one element of the container.
+				containerNode.MatchedAccessCount++;
+				this.tupleNodes.Add(temp, node);
```
christophwille, line 662: **Correctness (confirmed):** `MatchNestedTupleDesignations` calls `tupleNodes.Add(temp, node)` where `temp` can already be present as the container node registered by `ResolveTupleContainer`, throwing `ArgumentException` (duplicate key) instead of bailing out. Reachable via a self-referential element store when a reference type named ``System.ValueTuple`2`` (which `IsTupleCompatible` accepts by name) is instantiated recursively, e.g. `ValueTuple<int, C>` where `C` contains itself, so the elementType/temp.Type equivalence guard passes.

**Failure scenario:** Decompiling an assembly that ships its own class `System.ValueTuple` (polyfill or obfuscated code) containing `t = t.Item2;` makes `DeconstructionTransform` throw `ArgumentException` "An item with the same key has already been added", so the whole method fails to decompile with an error stub instead of falling back to the flat assignment output.

*Found by high-effort multi-agent review; independently verified.*

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
```diff
@@ -81,70 +98,151 @@ void IStatementTransform.Run(Block block, int pos, StatementTransformContext con
...
 				}
-				if (this.tupleType.Cardinality < 2)
-					return -1;
-				if (v != tupleVariable || !this.tupleType.Equals(tupleType))
-					return -1;
-				if (this.deconstructionResults[index] == null)
+				// Without any assignment the statement is a plain Deconstruct call, unless a nested
+				// deconstruction was consumed: then all leaves are single-use elements handled by
+				// the forwarding fixup in MatchAssignments.
+				if (!anyAssignments && !(rootCall != null && rootCall.NestedCalls.Any(c => c != null)))
```
christophwille, line 211: **Correctness (confirmed):** The removed `return startPos != pos` invariant (a deconstruction requires at least one matched assignment) is exempted for call-rooted patterns containing a nested call, so a chain of explicit source-level `Deconstruct()` call statements whose out-results are never consumed is now folded into a deconstruction declaration.

**Failure scenario:** Source that literally writes `o.Deconstruct(out var i, out var inner); inner.Deconstruct(out var a, out var b);` with all results unused decompiles as a discard deconstruction like `(_, (_, _)) = o;` instead of the explicit call statements. Master kept the calls verbatim because zero matched assignments rejected the pattern; the decompiled output now misrepresents the shape of the original code.

*Found by high-effort multi-agent review; independently verified.*

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
```diff
@@ -187,126 +283,437 @@ bool InlineDeconstructionInitializer(Block block, int pos)
...
+			if (value is Conv conv)
+				value = conv.Argument;
+			if (!MatchTupleElementRead(value, out var container, out _, out _))
+				return false;
+			if (!(container.StoreInstructions is [StLoc store]) || store.Parent != block)
+				return false;
+			if (!MatchTupleElementStore(store, out _, out _, out _, out _))
+				return false;
+			enclosingPos = store.ChildIndex;
+			while (enclosingPos > 0 && MatchTupleElementStore(block.Instructions[enclosingPos - 1], out _, out _, out _, out _))
```
christophwille: **Correctness (confirmed):** `TryFindEnclosingTupleDesignation`'s backward walk steps over *any* `MatchTupleElementStore` match regardless of container, so it crosses into an adjacent preceding deconstruction's element stores. The dry run then starts at the wrong position, fails, and the nested designation is dismembered instead of reconstructed.

**Failure scenario:** Two back-to-back deconstructions with no intervening init statement (sources already materialized, e.g. parameters): `var (a, inner1) = t1; var (b, (c, d)) = t2;`. At the `stloc c` position, the walk starts from inner2's store and keeps stepping back over `stloc inner1` and `stloc a`, landing `enclosingPos` at the *first* deconstruction's start. The dry run from there fails, so `IsConsumableByEnclosingDeconstruction` returns false and `(c, d) = inner2` is transformed standalone. When the walk later reaches inner2's store, its uses are no longer tuple element reads, so nesting is impossible: output is `var inner2 = t2.Item2; (var b, _) = t2; var (c, d) = inner2;` instead of `var (b, (c, d)) = t2;` -- the nested reconstruction silently fails whenever two deconstruction statement runs are adjacent. The walk should only step back over stores belonging to the same container tree (e.g. same root container), not any tuple element store.

*Found by high-effort multi-agent review; independently verified.*

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
```diff
@@ -81,70 +98,151 @@ void IStatementTransform.Run(Block block, int pos, StatementTransformContext con
...
-					return -1;
-				if (this.deconstructionResults[index] == null)
+				// Without any assignment the statement is a plain Deconstruct call, unless a nested
+				// deconstruction was consumed: then all leaves are single-use elements handled by
+				// the forwarding fixup in MatchAssignments.
+				if (!anyAssignments && !(rootCall != null && rootCall.NestedCalls.Any(c => c != null)))
+					return false;
+				// first tuple element may not be discarded,
+				// otherwise we would run this transform on a suffix of the actual pattern.
+				if (deconstructionResults[0] == null)
```
christophwille: **Quality regression (confirmed):** The `deconstructionResults[0] == null` check runs before the escaped-tuple-node retry, so a wrongly-nested first element aborts the match instead of being demoted via `doNotNest` and retried -- losing sugar master already produces.

**Failure scenario:** For `var (inner, x) = GetTuple<(int, int), int>(); Console.WriteLine(inner.Item1 + inner.Item2); Console.WriteLine(x);`, `MatchNestedTupleDesignations` consumes `stloc inner` as a nested-designation temporary (all of inner's uses are tuple element reads, satisfying `AllUsesAreTupleElementReads`), but the reads inside `WriteLine` are never consumed by the pattern, so leaves 0..1 stay unassigned and this check returns false before `EscapedTupleNodes()` can add `inner` to `doNotNest` and retry. The retry would succeed as a flat match. Master sugars this as `var (inner, x) = t;`; the PR emits raw Item1/Item2 element stores. Moving the escape check before the `[0] == null` check restores the match.

*Found by high-effort multi-agent review; independently verified.*

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
```diff
@@ -187,126 +283,437 @@ bool InlineDeconstructionInitializer(Block block, int pos)
...
+				this.tupleNodes.Add(temp, node);
+				InitializeFlatLeafIndices();
+				pos++;
+			}
+
+			static bool AllUsesAreTupleElementReads(ILVariable temp)
+			{
+				foreach (var use in temp.AddressInstructions.Concat<ILInstruction>(temp.LoadInstructions))
+				{
+					if (!(use.Parent is LdFlda elementAccess && elementAccess.Parent is LdObj))
```
christophwille, line 671: **Correctness (confirmed):** `AllUsesAreTupleElementReads` only accepts the shallow `LdObj(LdFlda(use))` shape, so Rest-chained element reads of long tuples (cardinality >= 8) disqualify a nested designation temporary.

**Failure scenario:** A nested tuple designation whose inner tuple has 8 or more elements, e.g. `var (a, (e1, e2, e3, e4, e5, e6, e7, e8)) = t;`. The inner temporary's reads of elements 8+ are lowered as `ldobj(ldflda Item1(ldflda Rest(ldloca temp)))`: the use's parent is the Rest `LdFlda` whose own parent is another `LdFlda`, not `LdObj`, so `AllUsesAreTupleElementReads` returns false and `MatchNestedTupleDesignations` refuses the nesting -- even though `MatchTupleFieldAccess` (used by `FindIndex`/`MatchTupleElementRead`, whose doc comment explicitly says Rest chains are flattened) would consume those reads fine. The outer match then falls apart into the flat two-statement form instead of the nested designation. The helper should walk the `LdFlda` chain (or reuse `MatchTupleElementRead` on the enclosing `LdObj`) instead of checking only the direct parent pair.

*Found by high-effort multi-agent review; independently verified.*

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
```diff
@@ -81,70 +98,151 @@ void IStatementTransform.Run(Block block, int pos, StatementTransformContext con
...
+		bool TransformDeconstruction(Block block, int pos)
 		{
-			public IType inputType;
-			public Conv conv;
+			int startPos = pos;
+			// Blocks are processed back to front, so the inner parts of a nested deconstruction
+			// are visited before the position its matching starts at; matching them on their own
+			// would consume the pattern piecemeal. Defer to the enclosing attempt where one
+			// exists (see the guard for the precision guarantees).
+			if (IsConsumableByEnclosingDeconstruction(block, pos))
```
christophwille, line 130: **Correctness (plausible):** With greedy per-position matching removed, an inner deconstruction position is skipped when a dry run says an enclosing match will consume it -- but the enclosing attempt runs only after all intermediate positions are visited by the whole transform pipeline, which can rewrite the block so the enclosing match no longer fires, and the skipped position gets no second chance. (The doc comment on `IsConsumableByEnclosingDeconstruction` itself concedes the dry run cannot promise the enclosing attempt still matches.)

**Failure scenario:** A nested shape where an intermediate statement between the inner and outer pattern is rewritten by another statement transform (e.g. inlining a single-use temp) after the dry run succeeds: the outer `MatchDeconstructionSequence` fails when finally attempted, and the inner deconstruction was already skipped, so the decompiled output regresses to raw `Deconstruct` calls / tuple element reads with compiler temporaries, where master at least produced the flat `(a, b) = temp;` sugar for the inner part.

*Found by high-effort multi-agent review; verifier judged the mechanism real but could not construct a concrete trigger.*

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
```diff
@@ -187,126 +283,437 @@ bool InlineDeconstructionInitializer(Block block, int pos)
...
 			{
-				pos++;
-			}
-			if (!MatchConversions(block, ref pos, out var conversions, out var conversionStLocs, ref delayedActions))
 				return false;
+			}
+			// The dry run leaves the matcher state behind, which is safe because it runs before
+			// the attempt at this position, and both that attempt and Run reset it. It does not
+			// modify the block: all rewrites are delayed actions.
+			return MatchDeconstructionSequence(block, enclosingPos, out int endPos, out _, out _, out _, out _)
```
christophwille: **Efficiency (confirmed):** `IsConsumableByEnclosingDeconstruction` re-runs the full `MatchDeconstructionSequence` dry run for every inner statement of the same pattern, matching each nested deconstruction O(k) times.

**Details:** A nested deconstruction with k inner statements (nested Deconstruct calls, defensive copies, designation temp stores) triggers k separate full dry runs from the same `enclosingPos` -- each allocating the conversions Dictionary, conversionStLocs List, DeconstructionCall arrays, and delayedActions closures that are then discarded -- plus the final real match: k+1 full matches of one pattern, and each dry run can itself loop on escape retries. On methods dense with tuple/deconstruction code this is quadratic wasted work. Cheaper: since the back-to-front walk visits the k inner positions consecutively, memoize the last dry-run result (block, enclosingPos, endPos) and reuse it instead of re-matching.

*Found by high-effort multi-agent review; independently verified.*

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
```diff
@@ -330,11 +737,17 @@ bool MatchConversions(Block block, ref int pos,
 			return true;
```
christophwille, line 763: **Correctness (confirmed):** (Anchored here because the affected loop -- `if (index <= previousIndex) return false;` at ~line 726, just above -- is unchanged context the diff cannot comment on.) `MatchConversions` still hard-rejects a call-rooted pattern when an unrelated conversion-shaped statement follows the `Deconstruct` calls -- inconsistent with the new `allowUnrelatedAssignments` handling one phase later.

**Failure scenario:** A nested Deconstruct-call chain whose leaves are all single-use (the new `anyAssignments == false` + `NestedCalls` path), followed by an unrelated widening assignment such as `long l = someInt;`: `MatchConversions` matches the stloc-of-Conv shape, `FindIndex(ldloc someInt)` returns -1, and `if (index <= previousIndex) return false;` rejects the *entire* sequence instead of ending the conversion run -- so no deconstruction is built at all and the output shows raw explicit `Deconstruct` calls. The same trailing statement without a conv (e.g. `int l = GetInt();`) is handled gracefully by the new break in `MatchAssignments`. `MatchConversions` needs the same "index < 0 for a call-rooted pattern ends the run" treatment (stop consuming, leave the statement after the deconstruction) rather than returning false.

*Found by high-effort multi-agent review; independently verified.*

christophwille:
## Code review summary (high-effort multi-agent review)

Reviewed `17e6649ba...33c0e986d` (5 changed files). 18 candidate findings were pooled from four independent finder passes; each was adversarially verified by a separate agent against the PR head. 17 survived, 1 was refuted; the 10 most severe are posted as inline comments at their code locations. All are in `DeconstructionTransform.cs` unless noted.

### Correctness (recompilation / behavior)

1. **`DeconstructInstruction.cs:266` -- stobj fallback accepts sign-mismatched pointer stores.** The fallback replaces the `UnknownType` sentinel; `StObj.Type` is sign-agnostic (from the store opcode), so `(*p, y) = t;` can be emitted where the original required an explicit `(uint)` cast -- output no longer recompiles (CS0266).
2. **`:544` -- `BindsOnElementType` misses competing extension methods.** It guards only against instance-method hiding; a more-specific extension `Deconstruct` in scope means the folded nested designation rebinds to a different method than the IL calls -- changed runtime behavior.
3. **`:636` -- reachable `ArgumentException` (duplicate `tupleNodes` key).** A self-referential element store on a by-name-matched `System.ValueTuple` polyfill/obfuscated class crashes the whole method's decompilation instead of falling back.
4. **`:211` -- unused explicit `Deconstruct()` call chains are folded into discard designations** (`(_, (_, _)) = o;`), misrepresenting source that master reproduced verbatim.

### Sugar-loss regressions (output quality)

5. **`:381` -- the backward walk in `TryFindEnclosingTupleDesignation` crosses into an adjacent preceding deconstruction's element stores**, so nested reconstruction silently fails whenever two deconstruction runs are adjacent. Walk should stay within the same container tree.
6. **`:215` -- `deconstructionResults[0] == null` check runs before the escaped-tuple-node retry**, so a wrongly-nested first element aborts instead of being demoted and retried -- a regression vs. master's flat sugar.
7. **`:726` (anchored at 737) -- `MatchConversions` hard-rejects the whole call-rooted pattern on a trailing unrelated conversion**, inconsistent with the new `allowUnrelatedAssignments` break in `MatchAssignments`.
8. **`:645` -- `AllUsesAreTupleElementReads` misses Rest-chained reads**, so nested designations with inner tuples of 8+ elements are refused.
9. **`:130` (plausible, no concrete trigger constructed) -- a skipped inner position gets no second chance** if intermediate transforms rewrite the block so the enclosing match no longer fires; the dry-run doc comment itself concedes this window.

### Efficiency

10. **`:318` -- `IsConsumableByEnclosingDeconstruction` re-runs the full dry run per inner statement** (k+1 full matches of one pattern, with per-run allocations). Memoizing the last dry-run result over the consecutive back-to-front walk would remove the quadratic rework.

**Refuted during verification:** the concern that break-and-commit on unrelated assignments could leave unconsumed Deconstruct results ill-formed -- the `LoadCount <= 1` constraint in `MatchDeconstructionCall` plus the forwarding fixup covers all cases.

Seven further low-severity confirmed findings (minor duplication and micro-inefficiencies) were dropped under the report cap.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
```diff
@@ -187,126 +283,437 @@ bool InlineDeconstructionInitializer(Block block, int pos)
...
+				var inst = block.Instructions.ElementAtOrDefault(pos);
+				if (inst != null && inst.MatchStLoc(out var copy, out var copiedValue)
+					&& copiedValue.MatchLdLoc(result)
+					&& copy.StoreCount == 1
+					&& copy.LoadCount + copy.AddressCount == 1)
+				{
+					receiver = copy;
+					pos++;
+					inst = block.Instructions.ElementAtOrDefault(pos);
+				}
```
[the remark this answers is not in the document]
siegfriedpammer: Done, in "Index the block directly when reading the statement at a position". `InstructionCollection` is indexable and knows its count, so all five call sites now go through a small `InstructionAt(block, pos)` helper instead of enumerating.

ICSharpCode.Decompiler/IL/Instructions/DeconstructInstruction.cs
siegfriedpammer: Confirmed and fixed; the fix is folded into "Support deconstruction into pointer targets".

Reproducing it needed a different shape than the one described: `var (el, y) = t; *p = (uint)el;` leaves the tuple on the stack (`dup`/`ldfld`), so the matcher never runs. It only occurs in the assignment form, which C# cannot express for a sign-mismatched pointer, so I built the IL by compiling with `int*` and retyping the parameter to `uint32*`. ILSpy then emitted `(*p, value) = GetTuple<int, int>();`, which fails to recompile with CS0266.

The stated cause is not quite right either: `InferType` does not yield a `PointerType`. The target is an anonymous stack slot typed `System.IntPtr` - every pointer element type erases that way - so `stobj.Type` really is the only type present there. The fix follows the slot's single definition, which still carries the declared pointer type.

The same wrong type also has a symptom that is reachable from ordinary C#, in the opposite direction: `stind.i4` reports `int` for a `uint*` target and `stind.i1` reports `sbyte` for a `byte*` one, so **every unsigned pointer target silently lost its deconstruction**. `(*p, v) = GetTuple<uint, int>()` with `uint* p` is legal C# and decompiled to `*p = tuple.Item1;`. Pinned by `Pointer_NoConversion_Tuple_UInt` and `_Byte`.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
siegfriedpammer: Confirmed and fixed; folded into "Keep the Deconstruct call explicit when the element type hides it".

Verified by running both binaries rather than by reading: with an extension `Deconstruct` on a base type forced by a cast, and a more specific one on the derived type in scope, the original assembly prints `1, 1, 2` and the recompiled decompiler output prints `1, 3, 4`.

Which extensions are in scope where the output is compiled is not knowable in an IL transform, so the binding is only certain when the element type is the receiver type of the called method itself. Anything else keeps the explicit call, and with it the cast that selected the method. Pinned by `Nested_CompetingExtensionDeconstruct`.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
siegfriedpammer: Fixed, but at the layer above: "Require System.ValueTuple to be a struct to be tuple compatible".

`TupleType.IsTupleCompatible` accepted `TypeKind.Class` for a type named `System.ValueTuple`. C# requires that type to be a struct - Roslyn rejects a class outright with CS8182 - so a class of that name is an unrelated type that was already being rendered with tuple syntax. It was also the only way a tuple could appear to contain itself, which is what registered the same variable as a node of the tree twice and threw.

With the type-system check corrected the duplicate key is unreachable, so no guard was added in the transform. Worth noting the check has accepted classes since tuples were added to the type system, alongside a name comparison against `"ValueType"` corrected later, so it looks like it was never deliberate.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
siegfriedpammer: Reproduces, but leaving it as is.

`o.Deconstruct(out var i, out var inner); inner.Deconstruct(out var a, out var b);` with every result unused does decompile to `(_, (_, _)) = o;`. That calls the same methods in the same order, discards the same results, and compiles - so it is a difference in how the shape is described, not in what the code does, and the sugared form is arguably the better rendering of it.

Happy to reconsider if the verbatim calls matter for a use case I am not seeing.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
siegfriedpammer: Could not reproduce, and the code is gone regardless.

Two adjacent deconstructions with materialized sources: the walk does step back over the preceding run's stores exactly as described, but the dry run started from that earlier position still matches, covers the position, and both runs are reconstructed. It even sugars two plain element reads that master leaves alone.

The backward walk existed only to give the dry run a start position. "Decide deferral to an enclosing deconstruction in O(1)" removes the dry run, so the walk went with it. Both shapes I tried are pinned by `LocalVariable_Nested_TupleInner_AfterAdjacentDeconstruction` and `_AfterAdjacentElementReads`.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
siegfriedpammer: Could not reproduce as a regression.

An escaping tuple as the **first** element reconstructs correctly (`var (tuple2, value) = ...`). It is the **second** element that stays unsugared - and master produces byte-identical output for that shape, so it is a pre-existing limitation rather than something this branch introduced.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
siegfriedpammer: Confirmed; tracked separately as #3962 rather than fixed here.

The diagnosis needs a correction for whoever picks it up: extending `AllUsesAreTupleElementReads` to step over `Rest` accesses is not sufficient. I made exactly that change and the match still bails at that same guard, so the use shape differs from what the code reads like and the cause needs establishing first.

The index arithmetic is already in place - `TupleTransform.MatchTupleFieldAccess` walks the `Rest` chain and folds it back into a flat position, and `TupleType` counts cardinality flattened. Only sugar is lost; the current output is correct and recompiles.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
siegfriedpammer: Reproduced and fixed, in "Decide deferral to an enclosing deconstruction in O(1)".

The trigger is a statement that is not part of the designation sitting between the nested temporary and the reads of it:

```csharp
(int, int) item = source.Item2;
Console.WriteLine(source.Item1);   // barrier
var (value, value2) = item;        // master sugars this; the branch did not
```

The inner position defers to an enclosing pattern that cannot reach it, the enclosing attempt fails, and the walk does not come back. Confirmed against master, which reconstructs both shapes.

The fix is the same rule the call chains use: only defer when every statement between the enclosing store and this position is itself an element store of the designation. Pinned by `LocalVariable_Nested_TupleInner_BarrierBeforeInnerReads` and `_BarrierAfterTemporary`.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
siegfriedpammer: Fixed in "Decide deferral to an enclosing deconstruction in O(1)": the dry run is gone entirely, not memoized.

The decisions it was making are available from the shape of the block. A nested `Deconstruct` call can only be consumed by an enclosing one that is the immediately preceding statement, looking through the defensive copy; a designation's temporaries and reads are stored back to back, so anything else in between is a barrier and the deconstruction is matched at that position instead. What remained was an enclosing call that is adjacent but cannot match anyway, which the constraint `MatchDeconstructionCall` already places on its out-parameters decides.

Measuring which fixtures each variant broke was what located the last case: dropping the dry run alone regressed only the two barrier fixtures, and requiring adjacency for tuple designations too regressed only the tuple ones.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
siegfriedpammer: Could not reproduce.

A conversion-shaped statement unrelated to the pattern after the calls - `var (x, (a, b)) = o; long unrelated = GetInt();` - reconstructs fine, so `MatchConversions` does not reject the call-rooted pattern there.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
christophwille: Re-tested against the current head, side by side with an ilspycmd built from the merge-base: the regression is still there. It needs the escaping element to sit in the **first** position *and* to be read only through its elements -- which is the shape the original failure scenario described, but my wording buried both conditions.

```csharp
public static ((int, int), int) GetTuple()
{
	return ((1, 2), 3);
}

public static void EscapeFirstElement()
{
	var (inner, x) = GetTuple();
	Console.WriteLine(inner.Item1 + inner.Item2);
	Console.WriteLine(x);
}
```

merge-base output (Debug and Release alike):

```csharp
var (tuple2, value) = GetTuple();
Console.WriteLine(tuple2.Item1 + tuple2.Item2);
Console.WriteLine(value);
```

this branch:

```csharp
((int, int), int) tuple = GetTuple();
(int, int) item = tuple.Item1;
int item2 = tuple.Item2;
Console.WriteLine(item.Item1 + item.Item2);
Console.WriteLine(item2);
```

Why the shapes you tried behave differently:

- Whole-variable read (`Console.WriteLine(tuple2)`): fails `AllUsesAreTupleElementReads`, so the temporary never becomes a nested node in the first place; the flat match works with nothing to retry. (Verified: reconstructs on both builds.)
- Second element escaping via element reads: leaf 0 is assigned, so control reaches `EscapedTupleNodes()` and the `doNotNest` retry restores the flat match. (Also verified fine.)
- First element escaping via element reads: every leaf of the wrongly-nested node precedes any assigned leaf, so `deconstructionResults[0] == null` returns false *before* the escape check runs, and the retry that would fix it is unreachable.

Checking `EscapedTupleNodes()` before the `deconstructionResults[0] == null` bail (retry when something escaped, fail as before when nothing did) should restore the flat sugar; this shape would make a good first-element twin to `LocalVariable_TupleInner_ElementUsedOutside`.

_AI re-review (Claude) posted on Christoph's behalf._

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
christophwille: Withdrawn -- I could not reproduce it either against the current head. With the leaf copies present in IL, the trailing `stloc l(conv ...)` is only ever seen by `MatchAssignments`, where the `allowUnrelatedAssignments` break ends the pattern cleanly. And in the no-copies form (leaves consumed in a single expression, Release build), the branch still reconstructs `var (x, (a, b)) = o;` with a following `long l = GetInt(); Console.WriteLine(l + l);` left intact. Every shape I could produce from C# sugars correctly where the merge-base leaves raw calls, so the `index <= previousIndex` rejection in `MatchConversions` does not appear reachable in practice.

_AI re-review (Claude) posted on Christoph's behalf._

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
christophwille: Agreed -- leaving it as is makes sense. Same methods called in the same order, same results discarded, and the output compiles; the discard designation communicates "results unused" at least as clearly as the verbatim calls. No change requested.

_AI re-review (Claude) posted on Christoph's behalf._

ICSharpCode.Decompiler/IL/Instructions/DeconstructInstruction.cs
christophwille: Fix verified against the current head, and thanks for the correction on the mechanism -- the IntPtr-typed stack slot, not a `PointerType` inference, is indeed what the scenario hinged on; the definition-chain walk in `GetPointerElementType` covers it. The unsigned-target repro decompiles to `(*p, value) = GetTuple<uint, int>();` now.

One boundary probed while verifying: pointer-arithmetic targets (`(*(p + 1), v) = ...`, `(p[2], v) = ...`) spill the address to a pointer-typed local first, stay unsugared, and are byte-identical to the merge-base output -- a pre-existing un-lifted case, not a regression, just noting it for completeness.

_AI re-review (Claude) posted on Christoph's behalf._

christophwille:
## Re-review of the updated branch

Follow-up to the review from this morning, checked against the current head (00075660f). Verification: built `ilspycmd` from this branch and from the merge-base side by side, decompiled targeted repros for the disputed findings (Debug and Release), and ran the full `DeconstructionTests` matrix locally -- Pretty, Correctness and Ugly runners, all 24 configurations green.

### Verified fixed

- **stobj pointer element type** (`DeconstructInstruction.IsAssignment`): the definition-chain recovery works, `(*p, value) = GetTuple<uint, int>();` reconstructs, and the `_UInt`/`_Byte` fixtures pin the sign-agnostic-opcode symptom. Pointer-arithmetic targets stay unsugared but byte-identical to the merge-base (pre-existing un-lifted case, noted in the thread).
- **Competing extension `Deconstruct`** (`BindsOnElementType`): requiring the element type to be the receiver type itself is the right conservative rule; the residual risk of a same-receiver-type extension in a closer scope is exactly the exposure the flat `(x, y) = o;` sugar has always had, so nesting is now at parity or stricter.
- **Duplicate tuple node** -> fixed one layer down in `TupleType.IsTupleCompatible`: requiring a struct matches C# (Roslyn rejects a `System.ValueTuple` class with CS8182), and it removes the only route to a self-containing "tuple". The `FromUnderlyingType` `IsDefaultOrEmpty` guard correctly handles the default-array case its new callers can hit.
- **Deferral precision and cost** (former dry run): the O(1) adjacency/shape rules in `IsConsumableByEnclosingDeconstruction` check out, the "costs sugar, never correctness" argument holds (the deferral mutates nothing), and the barrier fixtures pin the shapes the dry run used to get wrong. This also resolves the quadratic re-matching finding.
- **`ElementAtOrDefault`** replaced by the indexed `InstructionAt` helper at all call sites.
- The eager-`LoadCount` reasoning documented at the forwarding fixup (fresh tuple `E_i` variables never reach the lookup because their loads only materialize when the delayed actions run) was checked and holds, including the `Debug.Assert` claim.

### Accepted as argued

- **Unused explicit `Deconstruct` chains rendered as discard deconstruction**: same calls, same order, same discards, compiles -- a rendering choice, agreed to leave as is.
- **Backward-walk over adjacent designations**: moot, the walk no longer exists; the `_AfterAdjacent*` fixtures pin both shapes.
- **Long-tuple `Rest` chains**: tracked as #3962, agreed.

### Withdrawn

- **`MatchConversions` rejecting a trailing unrelated conversion**: could not be reproduced from any C#-reachable shape against the current head; details in the thread.

### Still open (1)

- **Escape-retry unreachable for a first-position element**: the regression from the morning review still reproduces on the current head -- `var (inner, x) = t;` with `inner` read only via `inner.Item1`/`inner.Item2` loses the flat sugar the merge-base produces, because `deconstructionResults[0] == null` returns before `EscapedTupleNodes()` can trigger the `doNotNest` retry. Repro, explanation of why the earlier probes did not trigger it, and a suggested reorder are in the thread reply.

_This is an AI-generated re-review (Claude), posted on Christoph's behalf._

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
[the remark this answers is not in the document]
siegfriedpammer: Correcting my earlier reply: I have reverted this, because the premise does not hold.

`Enumerable.ElementAtOrDefault` does not enumerate for this type. It forwards to `TryGetElementAt`, which starts with

```csharp
if (source is IList<TSource> list)
{
    if (!(found = (uint)index < (uint)list.Count)) return default(TSource);
    return list[index];
}
```

and `InstructionCollection<T>` is declared `IList<T>, IReadOnlyList<T>`, so the call was already a bounds check plus an indexer access. Replacing it changed nothing but added a helper, so the commit is dropped and the call sites are back to `ElementAtOrDefault`.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
siegfriedpammer: You are right and my earlier probe was wrong. Reproduced exactly as described, and fixed in "Demote an escaping first element instead of giving up on the pattern" (f7eeb4518).

Your diagnosis of why my shapes missed it is what I got wrong: I read the escaping element as a whole, so it failed `AllUsesAreTupleElementReads` and never became a nested node at all - nothing to retry, and the flat match worked. Reading it only through its elements is what makes it a node, and the first position is what puts every one of its leaves ahead of the assigned ones.

`EscapedTupleNodes()` now runs before the `deconstructionResults[0] == null` check, so an escaping node is demoted and retried; when nothing escaped, that check fails the pattern exactly as before. `LocalVariable_TupleInner_FirstElementUsedOutside` pins it as the first-position twin.

One note on the re-review: it was written against 00075660f, which included a commit replacing `ElementAtOrDefault` with an indexed helper. I have since dropped that commit - `Enumerable.ElementAtOrDefault` forwards to `TryGetElementAt`, which has an `IList<T>` fast path, and `InstructionCollection<T>` implements `IList<T>`, so there was no enumeration to avoid. Current head is f7eeb4518.

PR #3951, opened 2026-08-01
Add --dump-table metadata dump mode to ilspycmd

siegfriedpammer:
The GUI has the metadata-tables view; the CLI had nothing, so questions like "which MethodSemantics rows reference this Property row" required hand-written System.Reflection.Metadata scripts (item 2 of the internal ilspycmd agent-workflow wishlist).

`ilspycmd --dump-table <table> assembly.dll` prints every row of a metadata table - RID, token, resolved names, heap offsets, coded indexes - for the same 39 Cor tables the GUI shows. `<table>` is the ECMA-335 table name (case-insensitive) or the table number, decimal or `0x`-prefixed hex.

Output of `--dump-table TypeDef` against an F# sample assembly (aligned text table, default):

```
RID  Token       Attributes                  Name                Namespace                                           BaseType    FieldList  MethodList
---  ----------  --------------------------  ------------------  --------------------------------------------------  ----------  ---------  ----------
1    0x02000001  NotPublic                   <Module>                                                                0x01000001  1          1
2    0x02000002  Public, Abstract, Sealed    Repro3652                                                               0x01000001  1          1
3    0x02000003  NestedPublic, Serializable  Config                                                                  0x01000001  1          3
4    0x02000004  Abstract, Sealed            $Repro3652          <StartupCode$Repro3652>                             0x01000001  2          7
```

`--dump-table 0x18` (numeric form of MethodSemantics) - the motivating "which accessors belong to this property" query:

```
RID  Token       Semantics  Method      Association
---  ----------  ---------  ----------  -----------
1    0x18000001  Setter     0x06000005  0x17000001
2    0x18000002  Getter     0x06000004  0x17000001
```

`--dump-table Property --json`:

```json
{
  "assembly": "Repro3652.dll",
  "table": "Property",
  "rowCount": 1,
  "rows": [
    {
      "RID": 1,
      "Token": "0x17000001",
      "Attributes": "None",
      "Name": "Value'",
      "Signature": "0x000000CB"
    }
  ]
}
```

Design notes:
- Row enumeration for tables without public SRM row access (ClassLayout, EventMap, PropertyMap, NestedClass, FieldRVA, FieldMarshal, ImplMap, InterfaceImpl, the five *Ptr tables, and the SRM-hidden FieldList/MethodList/ParamList columns) lives in `MetadataExtensions` next to the existing `GetMethodSemantics` helper. The four GUI table nodes whose width rules diverged from SRM (TypeDef, MethodDef, ImplMap, FieldMarshal) already consume these shared readers; the remaining raw-reading nodes only use plain simple-index widths and can be folded in a follow-up.
- Every raw reader verifies that its computed column widths sum to the row size SRM derived from the tables-stream header and throws `BadImageFormatException` on mismatch, so a width bug (a large-heap flag over a small heap, EnC minimal-delta metadata) fails loudly instead of dumping plausible-looking garbage. String/Blob index widths are taken from SRM's computed row sizes of the single-column ModuleRef/TypeSpec tables, i.e. from the HeapSizes flags rather than the heap byte size.
- Every table's columns are spelled out explicitly in ECMA-335 declaration order; no reflection over SRM row structs, so the output is deterministic across runtime versions and trim/AOT-safe.
- JSON uses System.Text.Json from the shared framework - no new package reference, no lock-file changes. The `--json` flag is deliberately generic so it can later serve `-l` as well; for now it is gated to `--dump-table`.
- Unknown table names/numbers exit with EX_USAGE and list the supported tables with their numbers (`TypeDef (0x02), ...`); `-o` writes `.<table>.txt`/`.json`.

Also fixes a real bug found en route (own commit): `MetadataExtensions.GetMethodSemantics` read the Semantics and Association columns from metadata-root-relative offsets instead of per-row offsets, and used the plain-index width threshold for the coded Association column - the GUI's MethodSemantics table view was displaying garbage for both columns.

Tests: `DumpTableOptionTests` (red-first) in ICSharpCode.ILSpyCmd.Tests covering console and JSON output, the raw-read tiers (NestedClass, ClassLayout), the MethodSemantics accessor query, numeric addressing (decimal, hex, out-of-range), error paths, case-insensitivity and the `-o` multi-assembly output path; full CLI suite green (22/22), XPlat solution family builds warning-free.

### DIFF
--- a/ICSharpCode.Decompiler/Metadata/MetadataExtensions.cs
+++ b/ICSharpCode.Decompiler/Metadata/MetadataExtensions.cs
@@ -350,36 +350,54 @@ namespace ICSharpCode.Decompiler.Metadata
 
 		public static IEnumerable<(Handle Handle, MethodSemanticsAttributes Semantics, MethodDefinitionHandle Method, EntityHandle Association)> GetMethodSemantics(this MetadataReader metadata)
 		{
-			int offset = metadata.GetTableMetadataOffset(TableIndex.MethodSemantics);
-			int rowSize = metadata.GetTableRowSize(TableIndex.MethodSemantics);
 			int rowCount = metadata.GetTableRowCount(TableIndex.MethodSemantics);
+			var reader = metadata.AsBlobReader();
+			reader.Offset = metadata.GetTableMetadataOffset(TableIndex.MethodSemantics);
 
-			bool methodSmall = metadata.GetTableRowCount(TableIndex.MethodDef) <= ushort.MaxValue;
-			bool assocSmall = metadata.GetTableRowCount(TableIndex.Property) <= ushort.MaxValue && metadata.GetTableRowCount(TableIndex.Event) <= ushort.MaxValue;
-			int assocOffset = (methodSmall ? 2 : 4) + 2;
-			for (int row = 0; row < rowCount; row++)
+			int methodSize = SimpleIndexSize(metadata, TableIndex.MethodDef);
+			// HasSemantics coded index: 1 tag bit over Event (tag 0) and Property (tag 1).
+			int assocSize = CodedIndexSize(metadata, 1, TableIndex.Event, TableIndex.Property);
+			for (int rid = 1; rid <= rowCount; rid++)
 			{
-				yield return Read(row);
-			}
-
-			(Handle Handle, MethodSemanticsAttributes Semantics, MethodDefinitionHandle Method, EntityHandle Association) Read(int row)
-			{
-				var span = metadata.AsReadOnlySpan();
-				var methodDefSpan = span.Slice(offset + rowSize * row + 2);
-				int methodDef = methodSmall ? BinaryPrimitives.ReadUInt16LittleEndian(methodDefSpan) : (int)BinaryPrimitives.ReadUInt32LittleEndian(methodDefSpan);
-				var assocSpan = span.Slice(assocOffset);
-				int assocDef = assocSmall ? BinaryPrimitives.ReadUInt16LittleEndian(assocSpan) : (int)BinaryPrimitives.ReadUInt32LittleEndian(assocSpan);
+				var semantics = (MethodSemanticsAttributes)reader.ReadUInt16();
+				int methodRow = methodSize == 2 ? reader.ReadUInt16() : reader.ReadInt32();
+				int assocTag = assocSize == 2 ? reader.ReadUInt16() : reader.ReadInt32();
 				EntityHandle propOrEvent;
-				if ((assocDef & 0x1) == 1)
+				if ((assocTag & 0x1) == 1)
 				{
-					propOrEvent = MetadataTokens.PropertyDefinitionHandle(assocDef >> 1);
+					propOrEvent = MetadataTokens.PropertyDefinitionHandle(assocTag >> 1);
 				}
 				else
 				{
-					propOrEvent = MetadataTokens.EventDefinitionHandle(assocDef >> 1);
+					propOrEvent = MetadataTokens.EventDefinitionHandle(assocTag >> 1);
 				}
-				return (MetadataTokens.Handle(0x18000000 | (row + 1)), (MethodSemanticsAttributes)(BinaryPrimitives.ReadUInt16LittleEndian(span)), MetadataTokens.MethodDefinitionHandle(methodDef), propOrEvent);
+				yield return (MetadataTokens.Handle(((int)TableIndex.MethodSemantics << 24) | rid), semantics, MetadataTokens.MethodDefinitionHandle(methodRow), propOrEvent);
 			}
+		}
+
+		/// <summary>
+		/// Size in bytes of a column indexing <paramref name="table"/> directly
+		/// (ECMA-335 II.24.2.6: 2 bytes while the table has fewer than 2^16 rows).
+		/// </summary>
+		static int SimpleIndexSize(MetadataReader metadata, TableIndex table)
+		{
+			return metadata.GetTableRowCount(table) < (1 << 16) ? 2 : 4;
+		}
+
+		/// <summary>
+		/// Size in bytes of a coded-index column over <paramref name="tables"/> with
+		/// <paramref name="tagBits"/> tag bits (ECMA-335 II.24.2.6: 2 bytes while every
+		/// indexed table has fewer than 2^(16 - tagBits) rows).
+		/// </summary>
+		static int CodedIndexSize(MetadataReader metadata, int tagBits, params TableIndex[] tables)
+		{
+			int smallLimit = 1 << (16 - tagBits);
+			foreach (var table in tables)
+			{
+				if (metadata.GetTableRowCount(table) >= smallLimit)
+					return 4;
+			}
+			return 2;
 		}
 
 		public static IEnumerable<EntityHandle> GetFieldLayouts(this MetadataReader metadata)
--- a/ICSharpCode.Decompiler/Metadata/MetadataExtensions.cs
+++ b/ICSharpCode.Decompiler/Metadata/MetadataExtensions.cs
@@ -357,6 +357,7 @@ namespace ICSharpCode.Decompiler.Metadata
 			int methodSize = SimpleIndexSize(metadata, TableIndex.MethodDef);
 			// HasSemantics coded index: 1 tag bit over Event (tag 0) and Property (tag 1).
 			int assocSize = CodedIndexSize(metadata, 1, TableIndex.Event, TableIndex.Property);
+			CheckRowSize(metadata, TableIndex.MethodSemantics, 2 + methodSize + assocSize);
 			for (int rid = 1; rid <= rowCount; rid++)
 			{
 				var semantics = (MethodSemanticsAttributes)reader.ReadUInt16();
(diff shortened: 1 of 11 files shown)


christophwille:
## Review: `--dump-table` metadata dump mode

Reviewed at `b2189544`. Built `ILSpy.XPlat` family clean (0 warnings) and ran the CLI suite: **21/21 green**. Also smoke-tested the new mode against `ICSharpCode.Decompiler.dll` -- the `MethodSemantics` fix is real and verified (all `Association` values now land in `0x14`/`0x17`; before the fix both columns were read from the metadata root), and the `TypeDef` `FieldList`/`MethodList` columns come out monotonic with `MethodList` ending at `MethodDef rowcount + 1`, exactly as ECMA-335 requires.

### What it does

Adds `ilspycmd --dump-table <table> [--json]`, printing every row of any of the 39 Cor tables the GUI shows (RID, token, resolved names, heap offsets, coded indexes). Row readers for the tables SRM does not expose publicly (ClassLayout, EventMap, PropertyMap, NestedClass, FieldRVA, FieldMarshal, ImplMap, InterfaceImpl, the five `*Ptr` tables, and the hidden FieldList/MethodList/ParamList columns) land in `MetadataExtensions`. En route it fixes a genuine `GetMethodSemantics` bug that made the GUI's MethodSemantics view display garbage.

### What is good

- Red-first tests with real coverage: both raw-read tiers, the motivating accessor query, numeric/hex addressing, out-of-range, case-insensitivity, and both error paths.
- Columns spelled out explicitly in ECMA-335 order rather than reflected over SRM row structs -- deterministic, trim/AOT-safe.
- `System.Text.Json` from the shared framework, so no new `PackageReference` and no lock-file churn.
- `SimpleIndexSize`'s `< (1 << 16)` is the correct threshold (the GUI's existing copies use `<= ushort.MaxValue`, equivalent, and `< ushort.MaxValue` for heaps, which is off by one).
- The `GetMethodSemantics` fix is the right call to make here rather than deferring it.

### Findings

I have left inline comments; the ranking, highest first:

1. **`ListColumnSize` does not match the width rule SRM actually uses** for FieldList/MethodList/ParamList. Decompiling `System.Reflection.Metadata` 10.0.10 (`MetadataReader.InitializeTableReaders`) gives `fieldRefSize = GetReferenceSize(FieldPtr) > 2 ? 4 : GetReferenceSize(Field)` -- the `*Ptr` row count only decides whether the *pointer table itself* is large; otherwise the *member* table's row count governs. The PR instead switches the whole decision to the `*Ptr` table whenever it is non-empty, so a non-empty-but-small FieldPtr next to a >= 64K Field table yields 2 where the file stores 4, and every subsequent read is misaligned.

2. **Heap index widths are inferred from heap byte size, not the `HeapSizes` flags.** SRM reads only the flag bits (`(heapSizes & HeapSizes.BlobHeapLarge) != 0 ? 4 : 2`); the byte size is a producer convention, not a rule. There is an exact public-API derivation available -- see the inline comment.

3. **All computed widths ignore minimal-delta metadata**, where SRM forces 4 bytes for every table index and every coded index. `IsMinimalDelta` is `internal`, so the practical fix is the row-size derivation / self-check suggested inline.

   These three share one cheap mitigation: assert that the sum of the computed column widths equals `metadata.GetTableRowSize(table)` before reading, and fail loudly on mismatch. That converts every width bug in this family from silent garbage into a diagnosable error -- which is precisely the failure mode this PR is fixing in `GetMethodSemantics`.

4. **`-o` with more than one input assembly silently truncates every file but the last.** Pre-existing (`-l c` reproduces it identically on master), but the new branch adds another copy of the pattern. Repro and the local one-line fix are inline.

5. `PEFile` is not disposed in `DumpTable` (consistent with the other CLI helpers, but a `using` is free in new code).

6. **`ICSharpCode.ILSpyCmd/README.md` is not updated.** It mirrors the `--help` output option-by-option; `--dump-table` and `--json` are missing.

7. **Duplication.** `GetNestedClasses`, `GetTypeDefListColumns` and `GetMethodDefParamLists` are near-verbatim copies of `NestedClassTableTreeNode`, `TypeDefTableTreeNode` and `MethodTableTreeNode` (comments included), and the same holds for FieldMarshal/ImplMap/InterfaceImpl/PropertyMap/EventMap/FieldLayout/FieldRVA/Ptr. The PR defers the fold, which means findings 1-3 have to be fixed twice. Since the shared helpers now live in the library, folding at least the tables touched by those fixes in this PR would be cheaper than a follow-up. Separately, `DumpTableOptionTests.RunAsync` is a verbatim copy of `ILSpyCmdMemberOptionTests.RunAsync` -- worth a shared helper.

8. `FormatValue`'s per-handle-type cases are a latent trap -- flagged inline.

### Nits

- Header and separator lines keep trailing padding; only data rows are `TrimEnd()`ed.
- `WriteJson` round-trips through a `MemoryStream` and then materializes the whole document as a UTF-16 string. `new Utf8JsonWriter(...)` over the destination, or just writing to `Console.OpenStandardOutput()`, avoids holding a large table three times over.
- `--dump-table` sits last in the `else if` chain, so `-p`, `-il`, `-e`, `-d`, `--list-resources` and `--resource` all silently win over it. Consistent with the existing options, but the `--json` guard shows the intent to reject bad combinations -- a mutual-exclusion check would be more consistent than silent precedence.
- Fixture naming: the existing fixture is `ILSpyCmdMemberOptionTests`, the new one `DumpTableOptionTests`.

Nothing here is a blocker for the happy path -- the mode works, and findings 1-3 are all edge cases (unoptimized `#-` metadata, >= 64K member tables, EnC deltas). But they are exactly the inputs the raw readers exist to serve, and they fail silently, so the width self-check is worth adding before merge.

ICSharpCode.Decompiler/Metadata/MetadataExtensions.cs
```diff
@@ -350,35 +350,276 @@ public static IEnumerable<MethodSpecificationHandle> GetMethodSpecifications(thi
...
+			return 2;
+		}
+
+		/// <summary>
+		/// Size in bytes of the FieldList/MethodList/ParamList member-range columns: they index
+		/// the *Ptr table when the indirection is present, whose row count then governs the width.
+		/// </summary>
+		static int ListColumnSize(MetadataReader metadata, TableIndex ptrTable, TableIndex memberTable)
+		{
+			return SimpleIndexSize(metadata, metadata.GetTableRowCount(ptrTable) > 0 ? ptrTable : memberTable);
```
christophwille: This is not the rule SRM uses, and the doc comment above encodes the same wrong model.

Decompiling `System.Reflection.Metadata` 10.0.10 (`MetadataReader.InitializeTableReaders`):

```csharp
int fieldRefSize = ((GetReferenceSize(rowCounts, TableIndex.FieldPtr) > 2) ? 4 : GetReferenceSize(rowCounts, TableIndex.Field));
```

The `*Ptr` table's row count only decides whether the *pointer table itself* has crossed 2^16; if it has not, the width still comes from the **member** table. This helper instead hands the whole decision to the `*Ptr` table as soon as it is non-empty.

Divergence case: unoptimized metadata with `FieldPtr` = 10 rows and `Field` = 70,000 rows.
- SRM: `GetReferenceSize(FieldPtr)` = 2, not > 2, so `GetReferenceSize(Field)` = **4**.
- Here: `FieldPtr` row count 10 > 0, so `SimpleIndexSize(FieldPtr)` = **2**.

Two bytes short, and since `GetTypeDefListColumns` seeks relative to the row *end*, the FieldList read lands mid-column and MethodList picks up garbage. Note this is reachable exactly when the `*Ptr` tables exist -- i.e. the unoptimized (`#-`) metadata that `GetPtrRows` was added to serve.

Matching SRM is a one-liner:

```csharp
static int ListColumnSize(MetadataReader metadata, TableIndex ptrTable, TableIndex memberTable)
{
    return SimpleIndexSize(metadata, ptrTable) > 2 ? 4 : SimpleIndexSize(metadata, memberTable);
}
```

Same heuristic currently lives in `ILSpy/Metadata/CorTables/TypeDefTableTreeNode.cs` (`ListColumnWidth`) and `MethodTableTreeNode.cs`, so folding those onto this helper fixes all three copies at once.

ICSharpCode.Decompiler/Metadata/MetadataExtensions.cs
```diff
@@ -350,35 +350,276 @@ public static IEnumerable<MethodSpecificationHandle> GetMethodSpecifications(thi
...
+		}
+
+		public static IEnumerable<(EntityHandle Parent, BlobHandle NativeType)> GetFieldMarshals(this MetadataReader metadata)
+		{
+			int rowCount = metadata.GetTableRowCount(TableIndex.FieldMarshal);
+			var reader = metadata.AsBlobReader();
+			reader.Offset = metadata.GetTableMetadataOffset(TableIndex.FieldMarshal);
+			// HasFieldMarshal coded index: 1 tag bit over Field (tag 0) and Param (tag 1).
+			int parentSize = CodedIndexSize(metadata, 1, TableIndex.Field, TableIndex.Param);
+			int blobSize = metadata.GetHeapSize(HeapIndex.Blob) < (1 << 16) ? 2 : 4;
```
christophwille: Heap index width is a declared flag, not a derived value. SRM reads only the `HeapSizes` bits from the `#~` header:

```csharp
int stringHeapRefSize = (((heapSizes & HeapSizes.StringHeapLarge) == HeapSizes.StringHeapLarge) ? 4 : 2);
int blobHeapRefSize   = (((heapSizes & HeapSizes.BlobHeapLarge)   == HeapSizes.BlobHeapLarge)   ? 4 : 2);
```

Nothing in ECMA-335 forbids a producer from setting the large-heap bit over a heap smaller than 64 KB, and ilasm/obfuscators do. When that happens this reads 2-byte offsets out of 4-byte columns and prints plausible-looking garbage with no error.

Both widths can be derived exactly through public API, because SRM computes them for you and single-column tables expose them directly -- `ModuleRef` is `{ Name : String }` and `TypeSpec` is `{ Signature : Blob }`, and `GetTableRowSize` returns the computed row size whether or not the table has rows:

```csharp
int stringSize = metadata.GetTableRowSize(TableIndex.ModuleRef);
int blobSize   = metadata.GetTableRowSize(TableIndex.TypeSpec);
```

Same applies to `stringSize` in `GetImplMaps` (line 525). This also removes the last disagreement with the GUI, which uses `< ushort.MaxValue` here (`ImplMapTableTreeNode.cs:49`, `FieldMarshalTableTreeNode.cs:47`) and so differs from this file at a heap size of exactly 65535.

ICSharpCode.Decompiler/Metadata/MetadataExtensions.cs
```diff
@@ -350,35 +350,276 @@ public static IEnumerable<MethodSpecificationHandle> GetMethodSpecifications(thi
...
+			}
+		}
+
+		/// <summary>
+		/// Size in bytes of a column indexing <paramref name="table"/> directly
+		/// (ECMA-335 II.24.2.6: 2 bytes while the table has fewer than 2^16 rows).
+		/// </summary>
+		static int SimpleIndexSize(MetadataReader metadata, TableIndex table)
+		{
+			return metadata.GetTableRowCount(table) < (1 << 16) ? 2 : 4;
```
christophwille, line 385: `SimpleIndexSize` and `CodedIndexSize` both miss SRM's minimal-delta case. From the same decompile:

```csharp
private int GetReferenceSize(int[] rowCounts, TableIndex index)
{
    if ((long)rowCounts[(uint)index] >= 65536L || IsMinimalDelta) return 4;
    return 2;
}

private int ComputeCodedTokenSize(int largeRowSize, int[] rowCounts, TableMask tablesReferenced)
{
    if (IsMinimalDelta) return 4;
    ...
}
```

In EnC delta metadata every table index and coded index is 4 bytes regardless of row counts, so all of these helpers under-read. `MetadataReader.IsMinimalDelta` is `internal`, so there is no direct guard available.

Rather than chase each rule, I would add a self-check that covers this, the heap-flag case, and the `ListColumnSize` case in one go -- assert the widths you computed actually reconstruct the row:

```csharp
// A width bug reads plausible-looking garbage rather than failing, so verify the
// computed columns account for exactly the row size SRM derived from the header.
static void CheckRowSize(MetadataReader metadata, TableIndex table, int computed)
{
    int actual = metadata.GetTableRowSize(table);
    if (computed != actual)
        throw new BadImageFormatException($"Unexpected {table} row size: computed {computed}, actual {actual}.");
}
```

One call per reader. That turns every member of this bug family -- including the `GetMethodSemantics` bug this PR is fixing -- into a diagnosable error instead of silent nonsense, and it would have caught that one at authoring time.

ICSharpCode.ILSpyCmd/MetadataTableDumper.cs
```diff
@@ -0,0 +1,537 @@
...
+				table = (TableIndex)number;
+				return supportedTables.Contains(table);
+			}
+			return Enum.TryParse(name, ignoreCase: true, out table)
+				&& supportedTables.Contains(table);
+		}
+
+		public static int DumpTable(string assemblyFileName, TextWriter output, TableIndex table, bool asJson)
+		{
+			var module = new PEFile(assemblyFileName);
```
christophwille: `PEFile` is `IDisposable` (it owns the `PEReader` and the underlying `FileStream`) and this one is never released, so the handle lives until GC finalizes it. The other CLI helpers do the same thing, so this is not a regression -- but it is new code and the fix is a keyword:

```csharp
using var module = new PEFile(assemblyFileName);
```

It matters a little more here than elsewhere: `DumpTableOptionTests` calls `Main` in-process eleven times against the running test assembly, so each run leaks a live handle to it.

ICSharpCode.ILSpyCmd/MetadataTableDumper.cs
```diff
@@ -0,0 +1,537 @@
...
+					return "";
+				case string s:
+					return s;
+				case StringHandle sh:
+					return sh.IsNil ? "" : metadata.GetString(sh);
+				case BlobHandle bh:
+					return bh.IsNil ? "nil" : FormatHex(MetadataTokens.GetHeapOffset(bh));
+				case GuidHandle gh:
+					return gh.IsNil ? "nil" : metadata.GetGuid(gh).ToString();
+				case EntityHandle eh:
```
christophwille, line 450: Worth a comment in the file, because this block reads like a bug and is not one: `case EntityHandle eh` does **not** catch a boxed `MethodDefinitionHandle`. Pattern matching is a runtime type test and ignores SRM's user-defined conversion operators, so each concrete handle struct genuinely needs its own case -- which is why the six cases below are reachable and the build stays warning-free. Someone "simplifying" this by deleting them would silently change the output.

The flip side is the failure mode for handle types that are *not* listed (`TypeReferenceHandle`, `ParameterHandle`, `MemberReferenceHandle`, ...): they fall to `default:` and `Convert.ToString` renders them as `System.Reflection.Metadata.TypeReferenceHandle` instead of a token. Today every column happens to be covered -- the `(EntityHandle)constraint.Parameter` cast in the `GenericParamConstraint` case suggests you already hit this -- but the next column added is one cast away from printing type names into the dump.

Cheapest hardening: keep the conversion at the call sites (cast to `EntityHandle` when building the row, as `GenericParamConstraint` already does) and make `default:` reject anything from the `System.Reflection.Metadata` namespace, so an unhandled handle type fails loudly instead of formatting itself.

ICSharpCode.ILSpyCmd/IlspyCmdProgram.cs
```diff
@@ -363,6 +375,23 @@ int PerformPerFileAction(string fileName)
...
+					{
+						app.Error.WriteLine($"Unknown metadata table '{DumpTableName}'.");
+						app.Error.WriteLine($"Supported tables: {MetadataTableDumper.SupportedTableNames}");
+						return ProgramExitCodes.EX_USAGE;
+					}
+
+					if (outputDirectory != null)
+					{
+						string outputName = Path.GetFileNameWithoutExtension(fileName);
+						output = File.CreateText(Path.Combine(outputDirectory, outputName) + $".{table}.{(JsonOutputFlag ? "json" : "txt")}");
```
christophwille: This inherits a pre-existing `-o` bug that costs data. `output` is reassigned per input file, but only the last writer is closed (`finally { output.Close(); }`), so everything still sitting in the earlier `StreamWriter` buffers is dropped.

Reproduced on this branch:

```
$ ilspycmd ICSharpCode.Decompiler.dll ICSharpCode.ILSpyX.dll --dump-table Property -o out
$ wc -l out/ICSharpCode.Decompiler.Property.txt      # 3009
$ ilspycmd ICSharpCode.Decompiler.dll --dump-table Property | wc -l   # 3015
```

Six rows silently gone from the first file, with exit code 0. It is not introduced here -- `-l c` truncates identically on master (1266 vs 1289 lines) -- but the new branch adds another copy of the pattern, and it is contained locally:

```csharp
using var tableOutput = outputDirectory != null
    ? File.CreateText(Path.Combine(outputDirectory, Path.GetFileNameWithoutExtension(fileName)) + $".{table}.{(JsonOutputFlag ? "json" : "txt")}")
    : output;
return MetadataTableDumper.DumpTable(fileName, tableOutput, table, JsonOutputFlag);
```

(careful not to dispose `Console.Out` twice in the no-`-o` case). Fixing the shared `PerformPerFileAction` pattern for all options is a reasonable separate PR.

christophwille (commented):
Inline findings for PR #3951; see the summary comment for context, ranking, and the build/test results.

siegfriedpammer:
All findings addressed at `af620b50e` (fixes squashed into the commits that introduced the code, plus one new commit for the GUI fold):

1. `ListColumnSize` now follows SRM's rule: the `*Ptr` width only wins when the pointer table itself is large, otherwise the member table governs.
2. String/Blob index widths come from `GetTableRowSize(ModuleRef)` / `GetTableRowSize(TypeSpec)` as suggested, i.e. from the HeapSizes flags rather than the heap byte size.
3. Every raw reader (including the two row-end readers, whose full row width is reconstructible) verifies its computed widths against `GetTableRowSize` and throws `BadImageFormatException` on mismatch, which also turns the minimal-delta case into a diagnosable error.
4. `--dump-table -o` uses a per-file writer disposed per file. Regression test added red-first with two input assemblies; it reproduced the truncation (first file cut off mid-JSON).
5. `PEFile` is disposed.
6. README updated with `--dump-table` and `--json`.
7. The four GUI nodes whose width rules these fixes correct (TypeDef, MethodDef, ImplMap, FieldMarshal) are folded onto the shared readers in a new commit; the remaining raw-reading nodes only use plain simple-index widths and stay for the announced follow-up. `RunAsync` moved to a shared `CliTestRunner`.
8. `FormatValue` now carries a comment explaining why the concrete handle cases are reachable, and `default:` rejects unlisted `System.Reflection.Metadata` types instead of printing their type names.

Nits: the header line is `TrimEnd()`ed now. Left as-is: `WriteJson` buffering, option mutual exclusion (consistent with the surrounding options), fixture naming.

Verification: CLI suite 22/22, headless UI suite 1175/1178 (3 skipped, pre-existing), XPlat + ILSpy builds warning-free, and all 39 tables smoke-dumped against ICSharpCode.Decompiler.dll and the test assembly with the row-size self-check active.

*This comment was written by an AI agent (Claude Code) operating the account of @siegfriedpammer.*

PR #3956, opened 2026-08-05
Fix #3952: decompile VB.NET anonymous types and queries to valid C#

### LINKED ISSUE #3952: VB.NET anonymous types decompile to invalid C# (VB$AnonymousType_* use sites, hidden definitions)
Decompiling VB.NET-compiled assemblies that use anonymous types produces uncompilable C#: anonymous type *definitions* are hidden from the output, but *use sites* still reference them by their metadata name.

### Reproduction

- NuGet package `QuartzNetWebConsole` 1.0.2, assembly `QuartzNetWebConsole.Views.dll` (VB.NET-compiled), current master, default `DecompilerSettings`.
- Decompile e.g. `QuartzNetWebConsole.Views.Views` (or any member using VB anonymous types / LINQ `Select ... Into`).

### Observed

Use sites emit the raw generated names, which are not valid C#:

```csharp
select new VB$AnonymousType_2<TriggerWithState, ITrigger>(tr, tr.Trigger) into $VB$It
select new VB$AnonymousType_3<VB$AnonymousType_2<TriggerWithState, ITrigger>, bool>($VB$It, ...) into $VB$It
```

At the same time `DecompileTypeAsString(new FullTypeName("VB$AnonymousType_0`2"))` yields an empty string, i.e. the definition is suppressed, so the emitted code has no declaration to fall back on.

### Analysis

Two generated-name predicates disagree about VB names:

- `SRMExtensions.IsGeneratedName` treats names containing `$` as generated, and `SRMExtensions.IsAnonymousType` therefore matches `VB$AnonymousType_*`. This is what `CSharpDecompiler.MemberIsHidden` uses (CSharpDecompiler.cs:389), so the definitions are hidden when `settings.AnonymousTypes` is on.
- `NRExtensions.HasGeneratedName(IType)` only matches names starting with/containing `<`, so the type-system-level `NRExtensions.IsAnonymousType(IType)` returns false for `VB$AnonymousType_*`. As a result none of the anonymous-type translation paths in `CallBuilder`/`ExpressionBuilder` (e.g. CallBuilder.cs:1412/1838) fire for VB assemblies, and no `new { ... }` is synthesized.

Aligning `NRExtensions.IsAnonymousType`/`HasGeneratedName` with the SRM-level predicate should make VB anonymous types round-trip like the C# ones (`<>f__AnonymousType*`).

Related, likely a separate work item: VB also emits `VB$AnonymousDelegate_*` types (visible in the first snippet's generic arguments). C# has no equivalent construct, so those would additionally need mapping to `Func<...>`/`Action<...>` or a synthesized named delegate to make the output compilable.

*This issue was written by an AI agent (Claude) on Siegfried's behalf.*

siegfriedpammer:
Fixes #3952.

## The bug

Two predicates disagreed about what a compiler-generated name looks like:

- `SRMExtensions.IsGeneratedName` (metadata level) counts a `$` in the name, so
  `SRMExtensions.IsAnonymousType` matched `VB$AnonymousType_*` and `MemberIsHidden`
  dropped the definitions from the output.
- `NRExtensions.HasGeneratedName(IType)` (type system level) only looked for `<`, so
  `NRExtensions.IsAnonymousType` returned false and none of the anonymous-type
  translations in `CallBuilder`/`ExpressionBuilder` fired.

VB assemblies therefore lost the type definitions *and* kept the raw metadata names at
every use site — output that cannot be recompiled, since `$` is not a legal C# identifier
character.

## The fix

1. **One shared predicate** for both levels, so they cannot drift apart again. The
   metadata-level behaviour is kept exactly as it was: counting every name that merely
   *contains* `<` would newly capture explicit implementations of generic interface members.
2. **VB transparent identifiers.** The VB compiler carries query range variables in
   `$VB$It`, `$VB$It1`, `$VB$It2` and `$VB$ItAnonymous` (Roslyn's `GeneratedNameConstants`),
   the counterpart to C#'s `<>h__TransparentIdentifier`. Unrecognized, they survived into
   the output and left the queries uncompilable even once the anonymous types were fixed.

Without the second change the first one alone still yields invalid identifiers for every VB
query with more than one range variable, which is the shape the reported assembly is full of.

## Effect

A VB `From i In items Let square = i * i Where square > 4 Select i, square` went from

```csharp
IEnumerable<VB$AnonymousType_3<int, int>> enumerable = from i in items
    select new VB$AnonymousType_3<int, int>(i, checked(i * i)) into $VB$It
    where $VB$It.square > 4
    select new VB$AnonymousType_3<int, int>($VB$It.i, $VB$It.square);
```

to

```csharp
var enumerable = from i in items
                 let square = checked(i * i)
                 where square > 4
                 select new { i, square };
```

On the assembly from the issue (`QuartzNetWebConsole.Views` 1.0.2) every `VB$AnonymousType_*`
reference is gone.

## Tests

New `VBPretty/VBAnonymousTypes` fixture (20 configurations) covering mutable and `Key`
anonymous types, an anonymous type as an argument, and `Select`/`Let`+`Where`/`Join`/`Order By`
queries. Committed fixtures-first: the expected output fails on the first commit and passes on
the last. Full `ICSharpCode.Decompiler.Tests` suite green (3314 passed, 46 skipped, 0 failed).

The Roslyn 2.10 / .NET Core 2.2 configuration is branched with `#if`: there the query operator
calls are not restored to extension-method syntax, so no query expression is formed at all.
That is a pre-existing, config-specific gap unrelated to this change.

## Deliberately not in scope

- `VB$AnonymousDelegate_*` types (already called out in the issue as separate work). These
  are *not* a VB-only construct: a natural-typed C# lambda or method group that `Func`/`Action`
  cannot express makes Roslyn synthesize a delegate type too, and decompiling those is what
  the `natural-type-lambdas-methods` branch adds. Its `IsAnonymousDelegate` predicate matches
  on `Name.Contains("AnonymousDelegate")` through the same `HasGeneratedName()` this PR fixes,
  and VB's synthesized delegates satisfy its other conditions as well (empty namespace,
  `TypeKind.Delegate`, `[CompilerGenerated]`), so the two compose: this PR is what makes that
  machinery reachable for VB rather than something that has to be duplicated for it.
- Transparent identifiers that survive because the query transform bails, e.g. for lambdas
  with statement bodies (common in VB XML-literal code). This is **not** VB-specific: with
  `QueryExpressions` disabled, C# output emits 35 `<>h__TransparentIdentifier` references
  for the existing `QueryExpressions` fixture today. ILSpy applies `EscapeInvalidIdentifiers`
  only in the test harness, not in the default pipeline.
- VB closure display classes (`_Closure$__N-M`, `$VB$Local_*`) are still not recognized as
  such.

---

*This pull request was prepared by an AI agent (Claude) on Siegfried's behalf.*

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
+++ b/ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
@@ -250,6 +250,7 @@ namespace ICSharpCode.Decompiler.CSharp
 				new CombineQueryExpressions(),
 				new NormalizeBlockStatements(),
 				new FlattenSwitchBlocks(),
+				new RenameVisualBasicAnonymousTypes(), // must run before FixNameCollisions
 				new FixNameCollisions(),
 				new AddXmlDocumentationTransform(),
 			};
--- a/ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
+++ b/ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
@@ -645,8 +645,13 @@ namespace ICSharpCode.Decompiler.CSharp
 
 		internal static bool IsTransparentIdentifier(string identifier)
 		{
-			return identifier.StartsWith("<>", StringComparison.Ordinal)
-				&& (identifier.Contains("TransparentIdentifier") || identifier.Contains("TranspIdent"));
+			if (identifier.StartsWith("<>", StringComparison.Ordinal))
+			{
+				return identifier.Contains("TransparentIdentifier") || identifier.Contains("TranspIdent");
+			}
+			// The VB compiler names the carriers of its query range variables
+			// $VB$It, $VB$It1, $VB$It2 and $VB$ItAnonymous.
+			return identifier.StartsWith("$VB$It", StringComparison.Ordinal);
 		}
 		#endregion
 
--- a/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
+++ b/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
@@ -149,6 +149,7 @@
     <None Include="TestCases\VBPretty\Issue1906.vb" />
     <None Include="TestCases\VBPretty\Issue2192.vb" />
     <None Include="TestCases\VBPretty\Select.vb" />
+    <None Include="TestCases\VBPretty\VBAnonymousTypes.vb" />
     <None Include="TestCases\VBPretty\ParameterizedProperties.vb" />
     <None Include="TestCases\ILPretty\Unsafe.il" />
     <None Include="TestCases\ILPretty\Issue1389.il" />
--- a/ICSharpCode.Decompiler.Tests/TestCases/Ugly/NoLocalFunctions.Expected.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Ugly/NoLocalFunctions.Expected.cs
@@ -44,9 +44,9 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Ugly
 
 		private static void SimpleCapture()
 		{
-			_003C_003Ec__DisplayClass1_0 _003C_003Ec__DisplayClass1_1 = default(_003C_003Ec__DisplayClass1_0);
-			_003C_003Ec__DisplayClass1_1.x = 1;
-			_003CSimpleCapture_003Eg__F_007C1_0(ref _003C_003Ec__DisplayClass1_1);
+			_003C_003Ec__DisplayClass1_0 obj = default(_003C_003Ec__DisplayClass1_0);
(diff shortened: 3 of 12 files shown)


ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
```diff
@@ -628,8 +628,11 @@ static bool IsClosureType(SRM.TypeDefinition type, MetadataReader metadata)
 
 		internal static bool IsTransparentIdentifier(string identifier)
 		{
-			return identifier.StartsWith("<>", StringComparison.Ordinal)
-				&& (identifier.Contains("TransparentIdentifier") || identifier.Contains("TranspIdent"));
+			return (identifier.StartsWith("<>", StringComparison.Ordinal)
```
siegfriedpammer: Use if + return for readability

PR #3963, opened 2026-08-07
Substring optimizations: eliminate hot-path allocations (audit top 8)

christophwille:
Implements the first eight recommendations of a repository-wide `Substring` usage audit (105 production call sites classified by call frequency and by what the substring result feeds into, plus pattern micro-benchmarks). Each commit corresponds to one recommendation, ranked by (frequency x allocations saved) / effort:

1. **`AssignVariableNames.SplitName`** - the digit tail is already validated by the preceding loop, so the `Substring` + `int.TryParse` pair is replaced with an inline, overflow-guarded digit accumulation. Runs per variable and per reserved-name registration of every decompiled method.
2. **`AbstractSearchStrategy`** - operator-prefix stripping and (for `~` fuzzy terms) lowercasing now happen once per search run in the constructor instead of per candidate name per keystroke; the noncontiguous matcher compares spans with per-char `ToLowerInvariant`. This was the largest allocation source in interactive search (the `~` branch allocated two full `ToLower()` strings per entity). New `IsMatchTests` pin the `+`/`-`/`=`/`~` semantics.
3. **`ReflectionHelper` / `TopLevelTypeName`** - the arity after the backtick is parsed in place (netstandard2.0 has no span `int.TryParse`) and namespace/name are each cut exactly once, removing the throwaway intermediate per generic reflection name parsed. Micro-benchmark of this pattern: 74.2 -> 24.1 ns/op, 386 -> 0 B/op.
4. **`MethodBodyDisassembler.WriteOpCode`** - `ldarg.N`/`ldloc.N`/`stloc.N` shortcut forms render their digit and `param_`/`loc_` reference keys from static tables instead of two allocations per instruction.
5. **`TextWriterTokenWriter.EscapeIdentifier`** - returns the original string instance when nothing needs escaping (the overwhelmingly common case; previously it always built a `StringBuilder` plus a fresh result string per identifier token) and appends surrogate pairs as two chars instead of a 2-char substring.
6. **`CleanUpVariableName` chain** - all prefix/suffix cuts on the per-variable naming path are now slices; `ContainsNonPrintableIdentifierChar` and `IsValidName` gained span overloads, and only the final lowered name is materialized (single allocation). `IsKeyword` deliberately keeps its string parameter: it is called on that final string anyway, and the keyword `HashSet` has no span lookup on netstandard2.0.
7. **IL-with-C# view** - `ISmartTextOutput` gained `Write(ReadOnlySpan<char>)` as a default interface method (net10.0), overridden in `AvaloniaEditTextOutput` to append straight into its buffer; the highlighted-comment writer slices each source line instead of cutting up to four substrings per emitted line. The overload is on `ISmartTextOutput` rather than `ITextOutput` because the decompiler library targets netstandard2.0 (no DIMs), where a new interface member would break external implementers.
8. **`MetadataMethod`** - the explicit-interface-implementation `op_` check tests a span slice first and only allocates the short name for actual operators, instead of for every static non-generic method containing a dot.

Deliberate behavior changes (both called out in the respective commit messages):

- Fuzzy search matching (`~`) now lowercases with the invariant culture instead of the current culture.
- Reflection-name arity parsing accepts only plain ASCII digits; suffixes like ``Foo`+1`` that `int.TryParse` tolerated are now rejected (not legal reflection names).

Verification: full `ILSpy.sln` test suite (including the complete `PrettyTestRunner` golden suite, the disassembler golden tests, and the headless-Avalonia UI tests); new unit tests for the search matcher, reflection-name/arity parsing, and identifier escaping.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/Output/EscapeIdentifierTests.cs
+++ b/ICSharpCode.Decompiler.Tests/Output/EscapeIdentifierTests.cs
@@ -1,0 +1,69 @@
+// Copyright (c) 2026 Christoph Wille
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this
+// software and associated documentation files (the "Software"), to deal in the Software
+// without restriction, including without limitation the rights to use, copy, modify, merge,
+// publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons
+// to whom the Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all copies or
+// substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+// INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+// PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+// FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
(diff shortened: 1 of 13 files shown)


christophwille:
Original analysis document: [SubString.html](https://github.com/user-attachments/files/30830943/SubString.html)

PR #3964, opened 2026-08-07
Delay exceptions when copy-propagating an address into a StObj target

siegfriedpammer:
Decompiling with a language version below C# 7 can trip a debug assertion in `StObj.CheckTargetSlot`:

```
Debug.Assert(SemanticHelper.IsPure(this.Value.Flags))
  ICSharpCode.Decompiler/IL/Instructions/LdFlda.cs:93
```

Reproduces on the existing test assemblies, e.g. `CompoundAssignmentTest.ShortUnsignedRightShiftTest` at C# 6 and below, and `DynamicTests.StructMemberAccess` at C# 3 and below. Minimal case:

```csharp
public struct CustomStruct { public short ShortField; }
static void M(CustomStruct s) { s.ShortField >>>= 5; }   // decompiled at <= C# 6
```

`CanPerformCopyPropagation` permits copying `LdFlda`/`LdElema` when `UseRefLocalsForAccurateOrderOfEvaluation` is off, but `DoPropagate` puts the clone into a `StObj` target slot without consulting `StObj.SatisfiesSlotRestrictionForInlining`, leaving `stobj(ldflda F(ldloc C_0), <impure value>)`.

`ILInlining.FindLoadInNext` already handles the same conflict by setting `DelayExceptions` on the address instead of falling back to a ref local. Copy propagation now does the same, gated on the same setting that drives `InliningOptions.AllowChangingOrderOfEvaluationForExceptions`. The check sits in `DoPropagate` rather than `CanPerformCopyPropagation` because the public `Propagate()` entry point bypasses the latter -- `AsyncAwaitDecompiler` copies an `ldflda` of the builder field through it regardless of the setting.

Decompiled output for `DynamicTests` and `CompoundAssignmentTest` at C# 3 and C# 6 is byte-identical before and after; only the IL invariant changes.

New Ugly test `NoUnsignedRightShift` (decompiled at `LanguageVersion.CSharp6`) fails on master and passes with the fix.

---
_Prepared by an AI agent (Claude) on Siegfried's behalf._

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Ugly/NoUnsignedRightShift.Expected.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Ugly/NoUnsignedRightShift.Expected.cs
@@ -1,0 +1,30 @@
+namespace ICSharpCode.Decompiler.Tests.TestCases.Ugly
+{
+	internal class NoUnsignedRightShift
+	{
+		public struct CustomStruct
+		{
+			public short ShortField;
+		}
+
+		public class CustomClass
+		{
+			public short ShortField;
+		}
+
+		public static void ClassField(CustomClass c)
+		{
+			c.ShortField = (short)((uint)c.ShortField >> 5);
+		}
+
+		public static void StructField(CustomStruct s)
+		{
+			s.ShortField = (short)((uint)s.ShortField >> 5);
+		}
+
+		public static void ArrayElement(short[] a)
+		{
+			a[0] = (short)((uint)a[0] >> 5);
+		}
+	}
+}
--- a/ICSharpCode.Decompiler/IL/Transforms/CopyPropagation.cs
+++ b/ICSharpCode.Decompiler/IL/Transforms/CopyPropagation.cs
@@ -16,6 +16,7 @@
 // OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
@@ -145,6 +146,7 @@ namespace ICSharpCode.Decompiler.IL.Transforms
 		static void DoPropagate(ILVariable v, ILInstruction copiedExpr, Block block, ref int i, ILTransformContext context)
 		{
 			context.Step($"Copy propagate {v.Name}", copiedExpr);
+			int firstUninlinedArg = i;
 			// un-inline the arguments of the ldArg instruction
 			ILVariable[] uninlinedArgs = new ILVariable[copiedExpr.Children.Count];
 			for (int j = 0; j < uninlinedArgs.Length; j++)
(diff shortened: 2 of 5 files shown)


ICSharpCode.Decompiler/IL/Transforms/CopyPropagation.cs
```diff
@@ -168,6 +168,22 @@ static void DoPropagate(ILVariable v, ILInstruction copiedExpr, Block block, ref
 				}
 				// We are copying an expression from far away, reusing the ILRange would result in incorrect sequence points.
 				clone.SetILRange(new Interval());
+				if (!context.Settings.UseRefLocalsForAccurateOrderOfEvaluation
```
christophwille: Gating on `!UseRefLocalsForAccurateOrderOfEvaluation` protects the `RunOnBlock` path (where `CanPerformCopyPropagation` already refuses `LdFlda`/`LdElema` when the setting is on), but it leaves the public `Propagate()` entry point unprotected in the *other* direction.

`AsyncAwaitDecompiler.CleanUpBodyOfMoveNext` copy-propagates `ldflda <builderField>(ldloc this)` for VB state machines. That `LdFlda` comes from a real `ldflda` opcode, so `DelayExceptions == false`, and that call never passes through `CanPerformCopyPropagation`. With the default settings (any C# >= 7 profile) the new guard is skipped, so if one of those loads sits in a `StObj` target slot whose value isn't pure, we still build `stobj(ldflda F(...), <impure>)` and trip the very `StObj.CheckTargetSlot` assertion this PR fixes.

And there's no fallback available at this point: `DoPropagate` removes the defining `stloc` right after the loop, so every load *must* be replaced -- "honor the setting" here can only mean "emit IL that violates the invariant". Since the alternative (a ref local) isn't reachable from here anyway, I'd argue for making this unconditional and dropping the `context.Settings` check; the setting is already enforced where it can actually be acted on, in `CanPerformCopyPropagation`.

(Caveat: I could not construct a VB repro -- no VB toolset on this box -- and a C# corpus at every language version shows 0 assertions with the fix as written. So this is a soundness gap in the guard rather than an observed failure.)

ICSharpCode.Decompiler.Tests/TestCases/Ugly/NoUnsignedRightShift.Expected.cs
```diff
@@ -0,0 +1,33 @@
...
+		}
+
+		public static void ClassField(CustomClass c)
+		{
+			c.ShortField = (short)((uint)c.ShortField >> 5);
+		}
+
+		public static void StructField(CustomStruct s)
+		{
+			ref CustomStruct reference = ref s;
```
christophwille: Worth flagging: this expected output is decompiled at `LanguageVersion.CSharp6`, but `ref CustomStruct reference = ref s;` is C# 7 syntax -- the expectation records output that can't be compiled at the language version that was requested. I confirmed this is the real output (and that it is unchanged by the fix -- master produces the same, just with the assertion on top).

It comes from `DoPropagate` un-inlining the address's own argument (`stloc C_0(ldloca s)`) into a ref-typed stack slot with two loads, so it can never be inlined back. That makes the new comment in `CopyPropagation` -- "in order to avoid introducing a ref local" -- read oddly next to the very fixture being added, which introduces one anyway.

No other `TestCases/Ugly/*.Expected.cs` currently declares a ref local, so this is the first one. Fine to land as-is if the intent is to document existing ugliness, but a one-line note in the fixture saying the ref local is a known artifact of the un-inlining (and not something C# 6 could compile) would stop the next reader from treating it as correct-by-construction.

christophwille (commented):
## Review summary

I reproduced the whole thing locally (worktree on `fix-copypropagation-stobj-target`, Debug build with a trace listener that reports `Debug.Assert` failures) and the fix holds up:

* **Regression test is genuinely red/green.** `UglyTestRunner.NoUnsignedRightShift` fails on the parent commit for all 8 `roslynOnlyOptions` cases and passes with the fix (the 4 `TargetNet40` cases only failed here because the .NET 4.0 reference assemblies aren't available on this machine). The assertion that fires on master is `StObj.CheckTargetSlot` on `NoUnsignedRightShift.StructField`, i.e. the `LdFlda` case.
* **Both arms of the new code are live.** Instrumenting the `if` shows the `LdFlda` arm and the `LdElema` arm are both hit, and the `else` never is -- consistent with `StObj.SatisfiesSlotRestrictionForInlining` only rejecting those two opcodes.
* **No output change.** Decompiling `ICSharpCode.Decompiler.dll` (183k lines) at `LanguageVersion.CSharp6` gives a **byte-identical** result before and after, while the `CheckTargetSlot` assertion count drops 10 -> 0. A hand-written probe assembly covering `+=` / `>>>=` on struct fields, nested struct fields, class fields, arrays, `ref` parameters and array-of-struct elements is also byte-identical, 80 -> 0 assertions.
* **No leftover assertions at any language version.** Same assembly at CSharp1..CSharp8/Latest: 0 assertions with the fix.
* One thing I specifically went looking for and did **not** find: setting `DelayExceptions` only on the target-slot clone leaves it structurally unequal (`LdFlda.PerformMatch` compares `DelayExceptions`) to the sibling clone in the value slot, which could have broken `TransformAssignment.IsMatchingCompoundLoad`. Empirically it doesn't -- compound assignments still round-trip in every probe case.

Two comments below; neither blocks the fix.

dgrunwald (changes requested):
This fixes the assertion when !UseRefLocalsForAccurateOrderOfEvaluation, but don't we still have the same problem when the option is enabled?
ILInlining has 3 code paths:
 * slot restriction satisfied
 * semantic-changing reordering allowed -> set DelayExceptions
 * neither: don't inline
Copy propagation should do the same, but this PR only adds the second case and still lacks the third.

Edit: ah the last case is currently only reachable on the VB state machine code path. Maybe instead use an assertion that the option is disabled? Or add the necessary validation to the public entry point?

dgrunwald (approved):

PR #3967, opened 2026-08-08
Keep ref-returning switches as statements

### LINKED ISSUE #3966: Ref-returning switch is emitted as an invalid switch expression (CS1525)
### Input code

```csharp
using System;

public sealed class C
{
	private int _first;
	private int _second;

	public ref int Get(int index)
	{
		switch (index)
		{
			case 0:
				return ref _first;
			case 1:
				return ref _second;
			default:
				throw new ArgumentOutOfRangeException(nameof(index));
		}
	}
}
```

### Erroneous output

```csharp
public ref int Get(int index) => index switch
{
	0 => ref _first,
	1 => ref _second,
	_ => throw new ArgumentOutOfRangeException("index")
};
```

This does not compile:

```text
error CS1525: Invalid expression term 'ref'
```

A switch expression cannot return by reference. The decompiler should keep the valid switch statement in this case.

### Details

Product: ICSharpCode.Decompiler / ilspycmd.

Version: master `fdc4c7c`.

The input was compiled in Release for `net10.0`.

I can submit a small fix and regression test.

sailro:
tl;dr: C# has no ref-returning switch expression, so keep these switches as statements.

Fixes #3966

### Problem

`HandleSwitchExpression` also converts switches whose result is `StackType.Ref`. This emits arms such as `0 => ref field`, which C# rejects with CS1525.

### Solution

Skip switch-expression conversion for `StackType.Ref`. The existing statement path then emits a valid switch with `return ref` branches.

The regression case covers a ref-returning switch with multiple branches and a throwing default.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/RefLocalsAndReturns.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/RefLocalsAndReturns.cs
@@ -282,6 +282,19 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 			return ref DefaultInt;
 		}
 
+		public static ref int SwitchRefReturn(int index)
+		{
+			switch (index)
+			{
+				case 0:
+					return ref numbers[0];
+				case 1:
+					return ref numbers[1];
+				default:
+					throw new InvalidOperationException();
+			}
+		}
+
 		public static ref int LastOrDefault()
 		{
 			if (numbers.Length != 0)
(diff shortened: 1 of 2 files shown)


sailro:
@siegfriedpammer, I've set all my other PRs to “draft” status, since they'll require further discussion (they're all tagged with `ILSpy Dev Meeting 2026`). We can also delete them if you think it's too much work. I'm completely fine with that.

Going forward, I’ll only submit minor edit requests like this one. There will only be one “active” request at a time, so that shouldn’t be too overwhelming. (I currently have 3 similar PR candidates set aside, with some interesting fixes).

Thanks for your work !
Seb

siegfriedpammer:

PR #3969, opened 2026-08-08
Keep null-conditional LINQ chains in method syntax

### LINKED ISSUE #3968: Null-conditional LINQ chain over a value type is emitted as invalid query syntax (CS1936)
### Input code

```csharp
using System.Collections.Immutable;
using System.Linq;

public sealed class Box
{
	public ImmutableArray<int> Values { get; init; }
}

public static class C
{
	public static string? FirstPositive(Box? box) =>
		box?.Values
			.Where(value => value > 0)
			.Select(value => value.ToString())
			.FirstOrDefault();
}
```

### Erroneous output

```csharp
public static string? FirstPositive(Box? box)
{
	return (from value in box?.Values
		where value > 0
		select value.ToString()).FirstOrDefault();
}
```

This does not compile:

```text
error CS1936: Could not find an implementation of the query pattern for source type 'ImmutableArray<int>?'
```

The null-conditional access lifts `Values` to `ImmutableArray<int>?`. The method chain must remain inside the conditional access.

### Details

Product: ICSharpCode.Decompiler / ilspycmd.

Version: master `9ea8d7c`.

The input was compiled in Release for `net10.0`.

I can submit a small fix and regression test.

sailro:
tl;dr: Keep LINQ method syntax when its receiver chain contains a null-conditional access.

Fixes #3968

### Problem

`IntroduceQueryExpressions` only checks the immediate LINQ receiver for `?.`. A chained `Where(...).Select(...)` can therefore be converted to query syntax even when its source contains a null-conditional access over a value type.

The query source is then nullable and does not implement the query pattern.

### Solution

Make `IsNullConditional` inspect the member and invocation receiver chain before introducing query syntax.

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/Transforms/IntroduceQueryExpressions.cs
+++ b/ICSharpCode.Decompiler/CSharp/Transforms/IntroduceQueryExpressions.cs
@@ -380,7 +380,8 @@ namespace ICSharpCode.Decompiler.CSharp.Transforms
 		bool IsNullConditional(Expression target) => target switch {
 			UnaryOperatorExpression { Operator: UnaryOperatorType.NullConditional } => true,
 			MemberReferenceExpression member => IsNullConditional(member.Target),
-			InvocationExpression { Target: { } invocationTarget } => IsNullConditional(invocationTarget),
+			InvocationExpression invocation => IsNullConditional(invocation.Target),
+			IndexerExpression { Target: { } indexerTarget } => IsNullConditional(indexerTarget),
 			_ => false
 		};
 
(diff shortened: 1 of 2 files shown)


ICSharpCode.Decompiler/CSharp/Transforms/IntroduceQueryExpressions.cs
```diff
@@ -377,10 +377,12 @@ public override AstNode VisitIdentifier(Identifier identifier)
 			}
 		}
 
-		bool IsNullConditional(Expression target)
-		{
-			return target is UnaryOperatorExpression uoe && uoe.Operator == UnaryOperatorType.NullConditional;
-		}
+		bool IsNullConditional(Expression target) => target switch {
```
siegfriedpammer, line 380: I think it would be easy to avoid recursion here. But since this is not an ILAst transform it probably doesn't matter as much.

sailro: It's up to you. Personally, I think it's more in line with the tree-like representation I have of the model :)

siegfriedpammer:

PR #3970, opened 2026-08-09
Generate DecompilerSettings boilerplate from [DecompilerSetting] attributes

siegfriedpammer:
Every version-gated decompiler setting was bookkept in four places that had to stay in sync by hand: the property boilerplate, `SetLanguageVersion`, `GetMinimumRequiredVersion`, and the `[Category]` display string. A survey of the file showed that sync had already drifted in four settings.

This PR adds a second source generator to `ICSharpCode.Decompiler.Generators` that derives all four from a single `[DecompilerSetting]` attribute on a partial property: the backing field, the accessors with change notification, the version-derived `[Category]`, and both version methods. `DecompilerSettings.cs` shrinks from ~2450 to ~990 lines, and adding a new setting is now ~5 lines in one place:

```csharp
/// <summary>
/// Use C# 9 <c>nint</c>/<c>nuint</c> types.
/// </summary>
[Description("DecompilerSettings.NativeIntegers")]
[DecompilerSetting(CSharp.LanguageVersion.CSharp9_0)]
public partial bool NativeIntegers { get; set; }
```

`[Description]` stays handwritten: 51 of the 121 resource keys do not follow the `DecompilerSettings.{PropertyName}` convention, and the keys are what you grep for from the resx. Settings without a version argument (VB/F#/UI options) keep their handwritten `[Category]`. New diagnostics keep the scheme honest: DSTG002 (attribute target must be a partial instance bool property), DSTG003 (version-gated settings must not also declare a handwritten `[Category]`, which would make `GetCustomAttribute<CategoryAttribute>()` throw), DSTG004 (a new `LanguageVersion` needs a display-category mapping in the generator), DSTG005 (the containing class must be a non-nested partial class).

The first commit is a strict 1:1 translation - the pre-existing inconsistencies are reproduced exactly. The three follow-up commits then remove them, each independently droppable:

1. `ExtensionMethods`, `UseLambdaSyntax` and `UseEnhancedUsing` were disabled by `SetLanguageVersion` for old targets but never raised `GetMinimumRequiredVersion` while enabled, unlike every comparable syntax-preference setting.
2. `SwitchOnReadOnlySpanChar` carried the C# 11.0 category but was missing from both version methods, so decompiling for an older target could still produce switches over `ReadOnlySpan<char>` that the requested compiler cannot compile.
3. Four settings used the bare category string `"Other"` instead of the `DecompilerSettings.Other` resource key; both currently resolve to the same English text, but the keys would split into two options groups the moment their translations diverge.

Verification: a throwaway characterization harness recorded, on the old code, the full property vector after `SetLanguageVersion` for every `LanguageVersion` and the `GetMinimumRequiredVersion` result for each setting toggled alone, and was re-run green against the generated code (red/green around each behavior-changing commit); a reflection diff of the per-property `Category`/`Description`/`Browsable` surface between the old and new build came back identical for all 122 bool properties (also proving the merged partial-property attributes stay unambiguous for the options UI and settings XML round-trip). Full decompiler suite (3365 tests) and ILSpy UI suite (1178 tests) pass.

Later commits from the review rounds re-validate an explicitly set `LanguageVersion` when project export starts, document the two roles the language version plays in the settings API, and derive the C# 1.0 `[Category]` strings from the generator's version map as well (observably identical: `CSharp1` is the enum floor, so the new cascade bucket cannot fire and the new minimum-version arm returns what the fallback already did).

Identified but out of scope: with `UseLambdaSyntax` enabled, statement-bodied anonymous functions still decompile to `delegate` syntax; extending the setting to emit statement lambdas is a candidate follow-up.

---

### DIFF
--- a/ICSharpCode.Decompiler.Generators/DecompilerSettingsGenerator.cs
+++ b/ICSharpCode.Decompiler.Generators/DecompilerSettingsGenerator.cs
@@ -1,0 +1,385 @@
+// Copyright (c) 2026 Siegfried Pammer
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this
+// software and associated documentation files (the "Software"), to deal in the Software
+// without restriction, including without limitation the rights to use, copy, modify, merge,
+// publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons
+// to whom the Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all copies or
+// substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+// INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+// PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+// FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+using System.Collections.Immutable;
+using System.Text;
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Text;
+
+namespace ICSharpCode.Decompiler.Generators;
+
+/// <summary>
+/// Generates the boilerplate behind [DecompilerSetting] partial properties: the backing field,
+/// the accessors with change notification, and - from the per-setting language version - the
+/// [Category] attribute plus the SetLanguageVersion and GetMinimumRequiredVersion methods,
+/// so that a setting's version is declared in exactly one place.
+/// </summary>
+[Generator]
+internal class DecompilerSettingsGenerator : IIncrementalGenerator
+{
+	static readonly DiagnosticDescriptor InvalidSettingProperty = new(
+		id: "DSTG002",
+		title: "[DecompilerSetting] target must be a partial instance bool property",
+		messageFormat: "Setting property '{0}' must be a partial instance bool property with get and set accessors and an uppercase-start name (the generated backing field uses the camelCase form)",
+		category: "DecompilerSettingsGenerator",
+		defaultSeverity: DiagnosticSeverity.Error,
+		isEnabledByDefault: true);
+
+	// The generated [Category] would join a handwritten one on the merged partial property, and
+	// GetCustomAttribute<CategoryAttribute>() (used by the settings UI) throws on duplicates.
+	static readonly DiagnosticDescriptor CategoryOnVersionedSetting = new(
+		id: "DSTG003",
+		title: "Version-gated setting must not declare [Category]",
+		messageFormat: "Setting '{0}' derives its [Category] from the language version; remove the handwritten [Category] attribute",
+		category: "DecompilerSettingsGenerator",
+		defaultSeverity: DiagnosticSeverity.Error,
+		isEnabledByDefault: true);
+
+	static readonly DiagnosticDescriptor UnsupportedLanguageVersion = new(
+		id: "DSTG004",
+		title: "Language version has no display category",
+		messageFormat: "Language version '{0}' has no display category; gate settings on a released C# version, or add the new version to DecompilerSettingsGenerator.CategoryByVersion",
+		category: "DecompilerSettingsGenerator",
+		defaultSeverity: DiagnosticSeverity.Error,
+		isEnabledByDefault: true);
+
+	// The generated implementation is emitted as a top-level partial class; a nested or
+	// non-partial containing type would make it merge nowhere (or into a stray new type).
+	static readonly DiagnosticDescriptor InvalidContainingType = new(
+		id: "DSTG005",
+		title: "Setting must be declared in a non-nested partial class",
+		messageFormat: "Setting '{0}' must be declared in a partial, non-nested class so the generated implementation merges into it",
+		category: "DecompilerSettingsGenerator",
+		defaultSeverity: DiagnosticSeverity.Error,
+		isEnabledByDefault: true);
+
+	static readonly Dictionary<string, DiagnosticDescriptor> DescriptorsById =
+		new DiagnosticDescriptor[] { InvalidSettingProperty, CategoryOnVersionedSetting, UnsupportedLanguageVersion, InvalidContainingType }
+			.ToDictionary(d => d.Id);
+
+	// Display category per released C# version; the settings UI groups options by these strings.
+	static readonly Dictionary<string, string> CategoryByVersion = new() {
+		["CSharp1"] = "C# 1.0 / VS .NET",
+		["CSharp2"] = "C# 2.0 / VS 2005",
+		["CSharp3"] = "C# 3.0 / VS 2008",
+		["CSharp4"] = "C# 4.0 / VS 2010",
+		["CSharp5"] = "C# 5.0 / VS 2012",
+		["CSharp6"] = "C# 6.0 / VS 2015",
+		["CSharp7"] = "C# 7.0 / VS 2017",
+		["CSharp7_1"] = "C# 7.1 / VS 2017.3",
+		["CSharp7_2"] = "C# 7.2 / VS 2017.4",
+		["CSharp7_3"] = "C# 7.3 / VS 2017.7",
+		["CSharp8_0"] = "C# 8.0 / VS 2019",
+		["CSharp9_0"] = "C# 9.0 / VS 2019.8",
+		["CSharp10_0"] = "C# 10.0 / VS 2022",
+		["CSharp11_0"] = "C# 11.0 / VS 2022.4",
+		["CSharp12_0"] = "C# 12.0 / VS 2022.8",
+		["CSharp13_0"] = "C# 13.0 / VS 2022.12",
+		["CSharp14_0"] = "C# 14.0 / VS 2026",
+	};
+
+	readonly record struct SettingInfo(
+		string Namespace, string ClassName, string Accessibility, string PropertyName, string FieldName,
+		bool DefaultValue, int VersionValue, string? VersionName, string? Category, bool AffectsMinimumRequiredVersion,
+		string FilePath, int SpanStart);
+
+	// A diagnostic captured during the transform; kept as plain values so the pipeline stays cacheable.
+	readonly record struct DiagInfo(string Id, string MessageArg, string FilePath, int SpanStart, int SpanLength,
+		int StartLine, int StartChar, int EndLine, int EndChar);
+
+	readonly record struct SettingResult(SettingInfo? Setting, EquatableArray<DiagInfo>? Diagnostics);
+
+	public void Initialize(IncrementalGeneratorInitializationContext context)
+	{
+		context.RegisterPostInitializationOutput(i => i.AddSource("DecompilerSettingsGeneratorAttributes.g.cs", RoslynHelpers.EmbeddedAttributeSource + @"
+namespace ICSharpCode.Decompiler
+{
+	[global::Microsoft.CodeAnalysis.EmbeddedAttribute]
+	[global::System.AttributeUsage(global::System.AttributeTargets.Property)]
+	sealed class DecompilerSettingAttribute : global::System.Attribute
+	{
+		public DecompilerSettingAttribute() { }
+
+		public DecompilerSettingAttribute(global::ICSharpCode.Decompiler.CSharp.LanguageVersion introducedIn) { }
+
+		/// <summary>Initial value of the setting. Defaults to true.</summary>
+		public bool DefaultValue { get; set; } = true;
+
+		/// <summary>
+		/// Whether enabling the setting raises GetMinimumRequiredVersion() to the version the
+		/// setting was introduced in. Defaults to true; only meaningful on version-gated settings.
+		/// </summary>
+		public bool AffectsMinimumRequiredVersion { get; set; } = true;
+	}
+}
+
+"));
+
+		var settings = context.SyntaxProvider.ForAttributeWithMetadataName(
+			"ICSharpCode.Decompiler.DecompilerSettingAttribute",
+			(n, ct) => n is PropertyDeclarationSyntax,
+			GetSetting);
+
+		context.RegisterSourceOutput(settings.Collect(), WriteSettingsClasses);
+	}
+
+	static SettingResult GetSetting(GeneratorAttributeSyntaxContext context, CancellationToken cancellationToken)
+	{
+		var property = (IPropertySymbol)context.TargetSymbol;
+		var node = (PropertyDeclarationSyntax)context.TargetNode;
+		var diagnostics = new List<DiagInfo>();
+
+		if (property.Type.SpecialType != SpecialType.System_Boolean || property.IsStatic
+			|| property.GetMethod == null || property.SetMethod == null || property.SetMethod.IsInitOnly
+			|| !char.IsUpper(property.Name[0])
+			|| !node.Modifiers.Any(m => m.IsKind(SyntaxKind.PartialKeyword)))
+		{
+			diagnostics.Add(MakeDiagInfo(InvalidSettingProperty.Id, property.Name, node));
+			return new SettingResult(null, diagnostics.ToEquatableArray());
+		}
+
+		if (property.ContainingType.ContainingType != null
+			|| node.Parent is not ClassDeclarationSyntax containingClass
+			|| !containingClass.Modifiers.Any(m => m.IsKind(SyntaxKind.PartialKeyword)))
+		{
+			diagnostics.Add(MakeDiagInfo(InvalidContainingType.Id, property.Name, node));
+			return new SettingResult(null, diagnostics.ToEquatableArray());
+		}
+
+		var attribute = context.Attributes[0];
+		int versionValue = 0;
+		string? versionName = null;
+		string? category = null;
+		if (attribute.ConstructorArguments.Length == 1)
+		{
+			var versionArgument = attribute.ConstructorArguments[0];
+			if (versionArgument.Kind == TypedConstantKind.Error || versionArgument.Value is not int boundVersion || versionArgument.Type is null)
+			{
+				// The argument did not bind (e.g. a typo'd enum member); the compiler already
+				// reports that error at the argument, so just skip the setting instead of
+				// crashing the whole generator.
+				return new SettingResult(null, null);
+			}
+			versionValue = boundVersion;
+			versionName = VersionNameFromSyntax(attribute)
+				?? versionArgument.Type.GetMembers()
+					.OfType<IFieldSymbol>()
+					.FirstOrDefault(f => f.HasConstantValue && Equals(f.ConstantValue, versionValue))?.Name
+				?? versionValue.ToString();
+			if (!CategoryByVersion.TryGetValue(versionName, out category))
+			{
+				diagnostics.Add(MakeDiagInfo(UnsupportedLanguageVersion.Id, versionName, node));
+				return new SettingResult(null, diagnostics.ToEquatableArray());
+			}
+			if (property.GetAttributes().Any(a => a.AttributeClass?.ToDisplayString() == "System.ComponentModel.CategoryAttribute"))
+			{
+				diagnostics.Add(MakeDiagInfo(CategoryOnVersionedSetting.Id, property.Name, node));
+				// The compiler would otherwise also flag the generated [Category] as a duplicate;
+				// suppress it so the mistake surfaces as the single DSTG003.
+				category = null;
+			}
+		}
+
+		bool defaultValue = true;
+		bool affectsMinimumRequiredVersion = true;
+		foreach (var named in attribute.NamedArguments)
+		{
+			// A named argument that failed to bind is already a compiler error; ignore it here.
+			if (named.Value.Value is not bool namedValue)
+				continue;
+			if (named.Key == "DefaultValue")
+				defaultValue = namedValue;
+			else if (named.Key == "AffectsMinimumRequiredVersion")
+				affectsMinimumRequiredVersion = namedValue;
+		}
+
+		string fieldName = char.ToLowerInvariant(property.Name[0]) + property.Name.Substring(1);
+		if (SyntaxFacts.GetKeywordKind(fieldName) != SyntaxKind.None)
+			fieldName = "@" + fieldName;
+
+		var setting = new SettingInfo(
+			property.ContainingNamespace.IsGlobalNamespace ? "" : property.ContainingNamespace.ToDisplayString(),
+			property.ContainingType.Name,
+			SyntaxFacts.GetText(property.DeclaredAccessibility),
+			property.Name,
+			fieldName,
+			defaultValue,
+			versionValue,
+			versionName,
+			category,
+			affectsMinimumRequiredVersion,
+			node.SyntaxTree.FilePath,
+			node.SpanStart);
+		return new SettingResult(setting, diagnostics.Count == 0 ? null : diagnostics.ToEquatableArray());
+	}
+
+	// Prefer the enum member name as spelled at the use site: constant values are not unique in
+	// LanguageVersion (CSharp15_0 and Preview share a value), so a value-based reverse lookup can
+	// name an alias the user never wrote.
+	static string? VersionNameFromSyntax(AttributeData attribute)
+	{
+		if (attribute.ApplicationSyntaxReference?.GetSyntax() is not AttributeSyntax { ArgumentList.Arguments: { Count: >= 1 } arguments })
+			return null;
+		if (arguments[0].NameEquals != null)
+			return null;
+		return arguments[0].Expression switch {
+			MemberAccessExpressionSyntax memberAccess => memberAccess.Name.Identifier.Text,
+			IdentifierNameSyntax identifier => identifier.Identifier.Text,
+			_ => null,
+		};
+	}
+
+	static DiagInfo MakeDiagInfo(string id, string messageArg, SyntaxNode node)
+	{
+		var lineSpan = node.GetLocation().GetLineSpan();
+		return new DiagInfo(id, messageArg, node.SyntaxTree.FilePath, node.Span.Start, node.Span.Length,
+			lineSpan.StartLinePosition.Line, lineSpan.StartLinePosition.Character,
+			lineSpan.EndLinePosition.Line, lineSpan.EndLinePosition.Character);
+	}
+
+	static void WriteSettingsClasses(SourceProductionContext context, ImmutableArray<SettingResult> results)
+	{
+		foreach (var result in results)
+		{
+			if (result.Diagnostics is not { } resultDiagnostics)
+				continue;
+			foreach (var diag in resultDiagnostics)
+			{
+				// Indexer lookup so a diagnostic id missing from the map fails loudly instead of
+				// being reported under an unrelated descriptor.
+				var descriptor = DescriptorsById[diag.Id];
+				var location = Location.Create(diag.FilePath, new TextSpan(diag.SpanStart, diag.SpanLength),
+					new LinePositionSpan(new LinePosition(diag.StartLine, diag.StartChar), new LinePosition(diag.EndLine, diag.EndChar)));
+				context.ReportDiagnostic(Diagnostic.Create(descriptor, location, diag.MessageArg));
+			}
+		}
+
+		var settings = results
+			.Where(r => r.Setting != null)
+			.Select(r => r.Setting!.Value)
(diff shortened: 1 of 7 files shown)


ICSharpCode.Decompiler.Generators/DecompilerSettingsGenerator.cs
```diff
@@ -0,0 +1,333 @@
...
+			return new SettingResult(null, diagnostics.ToEquatableArray());
+		}
+
+		var attribute = context.Attributes[0];
+		int versionValue = 0;
+		string? versionName = null;
+		string? category = null;
+		if (attribute.ConstructorArguments.Length == 1)
+		{
+			versionValue = (int)attribute.ConstructorArguments[0].Value!;
```
christophwille: **[correctness, CONFIRMED]** Null-forgiving casts crash the generator on error-typed attribute arguments.

`(int)attribute.ConstructorArguments[0].Value!` here, `.Type!` on the next line, and `(bool)named.Value.Value!` at lines 180/182 dereference values that are null whenever the attribute argument fails to bind. A contributor who typos a version (e.g. `[DecompilerSetting(CSharp.LanguageVersion.CSharp15)]` when that member does not exist) or writes a non-constant/unresolved `DefaultValue` gets an error `TypedConstant` with null `Value`/`Type`; the cast throws `NullReferenceException` inside `GetSetting`, Roslyn reports AD0001 and suppresses the generator's entire output, and the build/IDE floods with ~90 cascading CS9248 "partial property has no implementation" errors that mask the actual typo. (Reproduced empirically against the pinned Microsoft.CodeAnalysis.CSharp under `CSharpGeneratorDriver`.)

Guard for `TypedConstant.Kind == TypedConstantKind.Error` / null values and either skip the setting or report a targeted diagnostic instead.

ICSharpCode.Decompiler.Generators/DecompilerSettingsGenerator.cs
```diff
@@ -0,0 +1,333 @@
...
+		if (versionBuckets.Length > 0)
+		{
+			WriteSetLanguageVersion(builder, versionBuckets);
+			builder.AppendLine();
+			WriteGetMinimumRequiredVersion(builder, versionBuckets);
+		}
+
+		builder.AppendLine("\t}");
+		builder.AppendLine("}");
+		context.AddSource($"{className}.Settings.g.cs", SourceText.From(builder.ToString().Replace("\r\n", "\n"), Encoding.UTF8));
```
christophwille: **[correctness, CONFIRMED]** The hint name uses only the class name (`{className}.Settings.g.cs`) while settings are grouped by `(Namespace, ClassName)` at line 238.

If a second class with the same name in a different namespace (e.g. in a test or plugin namespace inside the same compilation) declares a `[DecompilerSetting]` property, `WriteSettingsClass` calls `context.AddSource` twice with the identical hint name; `AddSource` throws `ArgumentException`, the generator is reported failed via AD0001, all generated settings code disappears, and the build fails with hundreds of cascading partial-property errors rather than pointing at the conflicting class.

Include the namespace in the hint name, e.g. `{ns}.{className}.Settings.g.cs`.

ICSharpCode.Decompiler/DecompilerSettings.cs
```diff
@@ -1198,1097 +455,435 @@ public bool ExtensionMethodsInCollectionInitializers {
...
-				{
-					switchOnReadOnlySpanChar = value;
-					OnPropertyChanged();
-				}
-			}
-		}
-
-		bool unsignedRightShift = true;
+		[DecompilerSetting(CSharp.LanguageVersion.CSharp11_0)]
+		public partial bool SwitchOnReadOnlySpanChar { get; set; }
```
christophwille, line 496: **[correctness, CONFIRMED]** `SwitchOnReadOnlySpanChar` was previously in neither the `SetLanguageVersion` cascade nor `GetMinimumRequiredVersion`; gating it on `CSharp11_0` (with the default `AffectsMinimumRequiredVersion = true`) changes both directions inside an otherwise mechanical refactor.

Consequences: a language-version profile of C# 10.0 or lower now renders `ReadOnlySpan<char>` string-switch methods as if/else comparison chains instead of the switch statements every prior release produced for that profile; and a profile where this is the only enabled C# 11 feature now makes `WholeProjectDecompiler.LanguageVersion = CSharp10_0` throw `InvalidOperationException` instead of exporting.

The follow-up commit declares this intentional -- if so, it deserves a test capturing the new gating (and ideally a separate commit/changelog mention); if not, `[DecompilerSetting]` without a version restores the old behavior.

ICSharpCode.Decompiler/DecompilerSettings.cs
```diff
@@ -430,766 +135,318 @@ public bool LifetimeAnnotations {
...
-				{
-					useEnhancedUsing = value;
-					OnPropertyChanged();
-				}
-			}
-		}
-
-		bool alwaysUseBraces = true;
+		[DecompilerSetting(CSharp.LanguageVersion.CSharp8_0)]
+		public partial bool UseEnhancedUsing { get; set; }
```
christophwille, line 311: **[correctness, CONFIRMED]** The deleted handwritten `GetMinimumRequiredVersion` deliberately excluded `useLambdaSyntax`, `extensionMethods`, and `useEnhancedUsing` from the version ladder (they were disabled by `SetLanguageVersion` but never raised the minimum). The generated replacement includes all three, because no `[DecompilerSetting(AffectsMinimumRequiredVersion = false)]` was applied -- the knob built for exactly this exclusion is currently used by zero settings.

Concrete regression: a settings profile that disables every C# 8.0 feature but leaves `UseEnhancedUsing` enabled (previously harmless) now makes `WholeProjectDecompiler.LanguageVersion = CSharp7_3` throw `InvalidOperationException` ("The chosen settings require at least CSharp8_0") where the same call succeeded before; project export likewise pins a higher LangVersion into the generated .csproj than the code requires.

Apply `AffectsMinimumRequiredVersion = false` to these three settings to preserve the old ladder.

ICSharpCode.Decompiler.Generators/DecompilerSettingsGenerator.cs
```diff
@@ -0,0 +1,333 @@
...
+		isEnabledByDefault: true);
+
+	// The generated [Category] would join a handwritten one on the merged partial property, and
+	// GetCustomAttribute<CategoryAttribute>() (used by the settings UI) throws on duplicates.
+	static readonly DiagnosticDescriptor CategoryOnVersionedSetting = new(
+		id: "DSTG003",
+		title: "Version-gated setting must not declare [Category]",
+		messageFormat: "Setting '{0}' derives its [Category] from the language version; remove the handwritten [Category] attribute",
+		category: "DecompilerSettingsGenerator",
+		defaultSeverity: DiagnosticSeverity.Error,
```
christophwille, line 53: **[test-coverage, CONFIRMED]** The deleted handwritten `SetLanguageVersion`/`GetMinimumRequiredVersion` ladders were directly exercised by the decompiler test suite reading their output, but the generator that now produces them -- including bucket ordering by raw enum value, the reversed minimum-version scan, keyword field-name escaping, and diagnostics DSTG002-004 -- ships with zero tests.

A future edit (e.g. breaking the `OrderBy(g => g.Key)`/`Reverse()` pairing or the `CategoryByVersion` map) would regenerate wrong version logic with no failing test; it would only surface as wrong decompiled output or wrong LangVersion in exported projects. A small `CSharpGeneratorDriver`-based test project verifying the generated source for a representative settings class (plus one test per diagnostic) would lock this down.

ICSharpCode.Decompiler.Generators/DecompilerSettingsGenerator.cs
```diff
@@ -0,0 +1,333 @@
...
+				.FirstOrDefault(f => f.HasConstantValue && Equals(f.ConstantValue, versionValue))?.Name
+				?? versionValue.ToString();
+			if (!CategoryByVersion.TryGetValue(versionName, out category))
+			{
+				diagnostics.Add(MakeDiagInfo(UnsupportedLanguageVersion.Id, versionName, node));
+				return new SettingResult(null, diagnostics.ToEquatableArray());
+			}
+			if (property.GetAttributes().Any(a => a.AttributeClass?.Name == "CategoryAttribute"))
+			{
+				diagnostics.Add(MakeDiagInfo(CategoryOnVersionedSetting.Id, property.Name, node));
```
christophwille, line 188: **[cleanup, CONFIRMED]** On DSTG003 the setting is still emitted with the generated `[Category]`, adding a redundant duplicate-attribute compile error.

This branch only adds the diagnostic without clearing `category` (unlike the DSTG004 path above, which returns early), so `WriteSettingsClass` emits `[global::System.ComponentModel.Category(...)]` on the generated declaration anyway and the developer sees both DSTG003 and a CS0579 "Duplicate Category attribute" for the same mistake. Emitting the setting with `Category = null` here leaves only the single intended diagnostic.

ICSharpCode.Decompiler.Generators/DecompilerSettingsGenerator.cs
```diff
@@ -0,0 +1,333 @@
...
+
+	public void Initialize(IncrementalGeneratorInitializationContext context)
+	{
+		context.RegisterPostInitializationOutput(i => i.AddSource("DecompilerSettingsGeneratorAttributes.g.cs", @"
+
+using System;
+
+namespace Microsoft.CodeAnalysis
+{
+    internal sealed partial class EmbeddedAttribute : global::System.Attribute
```
christophwille: **[cleanup, CONFIRMED]** This `Microsoft.CodeAnalysis.EmbeddedAttribute` post-init declaration duplicates the textually identical copy `DecompilerSyntaxTreeGenerator` already emits in its own `RegisterPostInitializationOutput`.

The two copies compile only because both stay `partial` and textually compatible; if one is later edited (made non-partial, moved namespace, given a conflicting member) the merged partial declarations stop compiling with confusing CS0260/CS0101 errors far from the edit. A shared post-init source or small shared helper would keep a single definition.

ICSharpCode.Decompiler.Generators/DecompilerSettingsGenerator.cs
```diff
@@ -0,0 +1,333 @@
...
+		}
+
+		var attribute = context.Attributes[0];
+		int versionValue = 0;
+		string? versionName = null;
+		string? category = null;
+		if (attribute.ConstructorArguments.Length == 1)
+		{
+			versionValue = (int)attribute.ConstructorArguments[0].Value!;
+			versionName = attribute.ConstructorArguments[0].Type!.GetMembers()
```
christophwille: **[cleanup, CONFIRMED]** The enum-name reverse lookup via `FirstOrDefault(f => Equals(f.ConstantValue, versionValue))` picks the first alias for duplicate constant values.

`LanguageVersion` declares `CSharp15_0 = 1500` and `Preview = 1500`, so a contributor who writes `[DecompilerSetting(LanguageVersion.Preview)]` gets a DSTG004 message naming `CSharp15_0` instead of the identifier they actually wrote. Reading the name from the attribute argument syntax (or preferring the syntax-spelled name when available) would report what the user typed.

ICSharpCode.Decompiler.Generators/DecompilerSettingsGenerator.cs
```diff
@@ -0,0 +1,333 @@
...
+
+	static void WriteSettingsClasses(SourceProductionContext context, ImmutableArray<SettingResult> results)
+	{
+		foreach (var result in results)
+		{
+			if (result.Diagnostics is not { } resultDiagnostics)
+				continue;
+			foreach (var diag in resultDiagnostics)
+			{
+				var descriptor = diag.Id switch {
```
christophwille: **[cleanup, PLAUSIBLE]** The diagnostic-id round-trip decodes with a switch whose default arm maps every unknown id to `UnsupportedLanguageVersion`.

If a DSTG005 is ever added to `GetSetting` but this switch is forgotten, the diagnostic is silently reported with DSTG004's id and messageFormat -- a wrong error message for an unrelated problem. A static `Dictionary<string, DiagnosticDescriptor>` keyed by `descriptor.Id` with no fallback (or carrying the descriptor itself in `DiagInfo`) keeps the mapping single-sourced and fails loudly on a missing entry.

christophwille (commented):
## Code review (high effort, multi-agent)

The review of the DecompilerSettings source-generator migration surfaced two real generator robustness bugs: unguarded null-forgiving casts on error-typed attribute arguments crash the entire generator (flooding the build with cascading partial-property errors instead of one targeted diagnostic), and the `AddSource` hint name drops the namespace half of the `(Namespace, ClassName)` grouping key, so same-named settings classes in different namespaces would collide deterministically.

The PR also folds two disclosed but untested behavior changes into the refactor: `SwitchOnReadOnlySpanChar` is newly gated on C# 11, and `ExtensionMethods`/`UseLambdaSyntax`/`UseEnhancedUsing` now raise `GetMinimumRequiredVersion` -- both of which can change decompiled output or make `WholeProjectDecompiler.LanguageVersion` assignment throw where it previously succeeded. The 333-line generator with three diagnostics also ships with zero tests.

Remaining findings are cleanups: DSTG003 double-reports via a redundant CS0579, a duplicated `EmbeddedAttribute` post-init declaration, alias-ambiguous enum-name recovery in diagnostics, and a catch-all diagnostic-id switch that would misreport future ids.

9 findings posted as inline comments (4 correctness, 1 test coverage, 4 cleanup). 16 candidates were independently verified; 15 survived, merged into these 9.

ICSharpCode.Decompiler/DecompilerSettings.cs
siegfriedpammer: Intentional, kept: with the setting enabled, the decompiler emits switch-on-`ReadOnlySpan<char>` code that requires C# 11, so a lower version profile could previously export a project whose stamped LangVersion cannot compile it. The gating brings the setting in line with the other version-gated syntax options; the throw on an explicit lower LanguageVersion is the deliberate safety net and is re-validated when export starts (8602ef750). Behavior was pinned red/green around this commit with the throwaway characterization harness described in the PR text; for committed tests, see the test-coverage thread.

*This reply was written by the Claude Code agent session working on this PR.*

ICSharpCode.Decompiler/DecompilerSettings.cs
siegfriedpammer: Intentional: the old exclusion was the drift being fixed. These are syntax-preference settings that emit that version's syntax when enabled (an enhanced using needs C# 8 to compile), so the old ladder could stamp a LangVersion under which the exported code does not compile; the export-time re-validation (8602ef750) covers the mutate-after-assign case. The suggested `AffectsMinimumRequiredVersion = false` escape hatch is now removed entirely in this same commit (5436eb19a) - it existed only to reproduce the old inconsistency and no setting uses it. Reverting this one commit restores both the old ladder and the knob.

*This reply was written by the Claude Code agent session working on this PR.*

ICSharpCode.Decompiler.Generators/DecompilerSettingsGenerator.cs
siegfriedpammer: Decided to keep committed generator tests out for now: the generated SetLanguageVersion cascade is exercised by every version-profiled fixture in the decompiler suite, DSTG002-004 misuse surfaces as build errors, and the migration itself was verified with the characterization harness and attribute-surface reflection diff described in the PR text. If the generator grows beyond DecompilerSettings, a CSharpGeneratorDriver-based test project is the natural next step.

*This reply was written by the Claude Code agent session working on this PR.*

ICSharpCode.Decompiler/DecompilerSettings.cs
```diff
@@ -430,766 +154,318 @@ public bool LifetimeAnnotations {
...
-				{
-					useLambdaSyntax = value;
-					OnPropertyChanged();
-				}
-			}
-		}
-
-		bool expressionTrees = true;
+		[DecompilerSetting(CSharp.LanguageVersion.CSharp3)]
+		public partial bool UseLambdaSyntax { get; set; }
```
christophwille, line 197: **[correctness, CONFIRMED -- carried over]** `ExtensionMethods`, `UseLambdaSyntax` (here), and `UseEnhancedUsing` now participate in `GetMinimumRequiredVersion`, which the deleted handwritten ladder deliberately excluded (they were disabled by `SetLanguageVersion` but never raised the minimum). Commit 5436eb19a makes the change intentional and explains it well -- but it ships with no regression test.

Observable change: a caller with `UseLambdaSyntax` or `ExtensionMethods` enabled but all other C# 3+ features disabled (or `UseEnhancedUsing` with the other C# 8 flags off) now gets a higher LangVersion pinned into exported project files, and a `WholeProjectDecompiler.LanguageVersion` assignment that was accepted before this PR (e.g. `CSharp2` with `UseLambdaSyntax` on) now throws `InvalidOperationException` -- in the setter and again at `DecompileProject` start.

A small test pinning `GetMinimumRequiredVersion()` for these profiles would document the new contract and catch accidental regressions.

ICSharpCode.Decompiler/DecompilerSettings.cs
```diff
@@ -1198,1097 +474,435 @@ public bool ExtensionMethodsInCollectionInitializers {
...
-				{
-					switchOnReadOnlySpanChar = value;
-					OnPropertyChanged();
-				}
-			}
-		}
-
-		bool unsignedRightShift = true;
+		[DecompilerSetting(CSharp.LanguageVersion.CSharp11_0)]
+		public partial bool SwitchOnReadOnlySpanChar { get; set; }
```
christophwille, line 496: **[correctness, CONFIRMED -- carried over]** `SwitchOnReadOnlySpanChar` was previously in neither the `SetLanguageVersion` cascade nor `GetMinimumRequiredVersion`; it is now gated on `CSharp11_0`. Commit 3baa9133c makes this intentional -- but it also ships without a regression test.

Observable change: a C# 10-or-lower profile now renders `ReadOnlySpan<char>` string-switches as comparison chains (changed decompiled text for the same assembly), and a stored per-flag profile with `SwitchOnReadOnlySpanChar=true` but all other C# 11 flags off sees `GetMinimumRequiredVersion()` jump to `CSharp11_0`, so an exported .csproj carries LangVersion 11.0 and a previously valid `WholeProjectDecompiler.LanguageVersion = CSharp10_0` assignment now throws.

Same ask as above: one test pinning the new gating.

ICSharpCode.Decompiler.Generators/DecompilerSettingsGenerator.cs
```diff
@@ -0,0 +1,343 @@
...
+		}
+	}
+
+	static void WriteSettingsClass(SourceProductionContext context, string ns, string className, SettingInfo[] settings)
+	{
+		var builder = new StringBuilder();
+		builder.AppendLine("// <auto-generated/>");
+		builder.AppendLine("#nullable enable");
+		builder.AppendLine();
+		builder.AppendLine($"namespace {ns}");
```
christophwille: **[correctness, CONFIRMED]** `WriteSettingsClass` emits an invalid empty namespace declaration (`namespace ` followed by `{`) when the settings class is in the global namespace -- even though the hint-name logic at line 305 explicitly supports the `ns.Length == 0` case.

A `[DecompilerSetting]` property on a class in the global namespace produces a generated file that fails to parse (CS1001/CS1514 inside an unopenable generated file) plus a cascading CS9248, instead of working output or a targeted diagnostic. Either wrap the class in the namespace only when `ns.Length > 0`, or reject the global-namespace case with a DSTG diagnostic to match the hint-name handling.

christophwille (commented):
## Re-review (high effort, multi-agent)

Follow-up to [the previous review](https://github.com/icsharpcode/ILSpy/pull/3970#pullrequestreview-4890751443), against head f7e6a8a.

**Six of the nine previous findings are fixed** at this head: error-typed attribute arguments are now guarded (skip instead of crash), the hint name carries the namespace, DSTG003 suppresses the generated `[Category]`, the `EmbeddedAttribute` source is shared via `RoslynHelpers`, the enum name is recovered from syntax (alias-safe), and the diagnostic-id lookup is a fail-loud dictionary. Nice turnaround.

**Three findings carry over**: the two behavior changes (`ExtensionMethods`/`UseLambdaSyntax`/`UseEnhancedUsing` raising the minimum version; `SwitchOnReadOnlySpanChar` gated on C# 11) are now explicit, well-described commits -- but both still ship without a regression test pinning the new semantics, and the generator itself remains untested.

**Five new findings**: two latent generator input-validation gaps (invalid output for global-namespace classes and for camelCase-named settings), a missing containing-type check, duplicated `LanguageVersion` validation in `WholeProjectDecompiler`, and 14 handwritten C# 1.0 `[Category]` literals the generator's category map could single-source.

8 findings posted as inline comments (4 correctness, 1 test coverage, 3 cleanup). 18 candidates were independently verified; the one candidate targeting the since-removed `AffectsMinimumRequiredVersion` knob was dropped as obsolete (the branch moved mid-review).

ICSharpCode.Decompiler.Generators/DecompilerSettingsGenerator.cs
```diff
@@ -0,0 +1,343 @@
...
+		foreach (var named in attribute.NamedArguments)
+		{
+			// A named argument that failed to bind is already a compiler error; ignore it here.
+			if (named.Value.Value is not bool namedValue)
+				continue;
+			if (named.Key == "DefaultValue")
+				defaultValue = namedValue;
+		}
+
+		string fieldName = char.ToLowerInvariant(property.Name[0]) + property.Name.Substring(1);
```
christophwille, line 205: **[correctness, CONFIRMED]** The backing-field name derivation only lower-cases the first character and escapes reserved keywords, so a property whose name already starts lowercase produces a field with the identical name as the property -- and the DSTG002 shape check does not reject that case.

`public partial bool switchExpressions { get; set; }` with `[DecompilerSetting]` passes DSTG002 (bool, instance, partial, get+set), and the generated file then declares both `bool switchExpressions;` and `partial bool switchExpressions {...}` in the same class, failing the build with CS0102 duplicate-member errors pointing into generated code. Extending DSTG002 to require an uppercase-start property name (repo settings are all PascalCase) would surface one targeted diagnostic instead.

ICSharpCode.Decompiler.Generators/DecompilerSettingsGenerator.cs
```diff
@@ -0,0 +1,343 @@
...
+namespace ICSharpCode.Decompiler.Generators;
+
+/// <summary>
+/// Generates the boilerplate behind [DecompilerSetting] partial properties: the backing field,
+/// the accessors with change notification, and - from the per-setting language version - the
+/// [Category] attribute plus the SetLanguageVersion and GetMinimumRequiredVersion methods,
+/// so that a setting's version is declared in exactly one place.
+/// </summary>
+[Generator]
+internal class DecompilerSettingsGenerator : IIncrementalGenerator
```
christophwille, line 36: **[test-coverage, CONFIRMED -- carried over]** The generator now produces `SetLanguageVersion`, `GetMinimumRequiredVersion`, all property boilerplate, and diagnostics DSTG002-004, but no test project or fixture in the PR exercises it -- not the bucket ordering, the reversed minimum-version scan, keyword escaping, or any diagnostic path.

The previously handwritten ladders were at least indirectly pinned by decompiler tests reading their output; a future edit to the generator now compiles green while silently changing decompiler behavior for every consumer. A `CSharpGeneratorDriver`-based test verifying the generated source for a representative settings class (plus one test per diagnostic) would lock this down -- and would also cover the global-namespace and camelCase edge cases flagged in this review.

ICSharpCode.Decompiler.Generators/DecompilerSettingsGenerator.cs
```diff
@@ -0,0 +1,343 @@
...
+		context.RegisterSourceOutput(settings.Collect(), WriteSettingsClasses);
+	}
+
+	static SettingResult GetSetting(GeneratorAttributeSyntaxContext context, CancellationToken cancellationToken)
+	{
+		var property = (IPropertySymbol)context.TargetSymbol;
+		var node = (PropertyDeclarationSyntax)context.TargetNode;
+		var diagnostics = new List<DiagInfo>();
+
+		if (property.Type.SpecialType != SpecialType.System_Boolean || property.IsStatic
```
christophwille, line 144: **[cleanup, CONFIRMED]** DSTG002 validates only the property (bool, instance, get/set, partial) but never the containing type, while `WriteSettingsClass` unconditionally emits `partial class {ContainingType.Name}` directly under the namespace.

A `[DecompilerSetting]` on a partial property of a *nested* class makes the generator silently create a brand-new top-level class of the same simple name containing the implementation; the real property fails with CS9248 plus a stray generated type, with no DSTG diagnostic pointing at the actual mistake. A non-partial containing class similarly surfaces as CS0260 on the handwritten file. Extending DSTG002 (or a sibling diagnostic) to require a non-nested partial containing type would fail with one targeted error, matching the diligence already applied to the property itself.

ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
```diff
@@ -154,6 +161,15 @@ protected WholeProjectDecompiler(
 			{
 				throw new InvalidOperationException("Must set TargetDirectory");
 			}
+			// The LanguageVersion setter already rejects a version below what the settings require,
+			// but Settings is mutable and shared, so re-validate against the settings actually in
+			// effect now - otherwise the exported project would carry a LangVersion under which the
+			// emitted code cannot compile.
+			if (languageVersion is { } explicitVersion && explicitVersion < Settings.GetMinimumRequiredVersion())
```
christophwille: **[cleanup, CONFIRMED]** This re-validation duplicates the `LanguageVersion` setter's check and its exception message verbatim (lines 73-76). When one copy is later edited, the other silently drifts, giving users two different errors for the same condition.

Minor extras in the same spot: the throw path calls `Settings.GetMinimumRequiredVersion()` a second time just for the message instead of reusing the value from the condition, and the second string segment (`$" Please change the DecompilerSettings accordingly."`) is an interpolated string with no interpolation holes. A shared `ValidateLanguageVersion(...)` helper (or shared message builder) would remove the copy-paste.

ICSharpCode.Decompiler/DecompilerSettings.cs
```diff
@@ -430,766 +154,318 @@ public bool LifetimeAnnotations {
...
-		}
-
-		bool decimalConstants = true;
+		[DecompilerSetting(CSharp.LanguageVersion.CSharp8_0)]
+		public partial bool AsyncEnumerator { get; set; }
 
 		/// <summary>
 		/// Decompile [DecimalConstant(...)] as simple literal values.
 		/// </summary>
 		[Category("C# 1.0 / VS .NET")]
```
christophwille: **[cleanup, CONFIRMED]** 14 handwritten `[Category("C# 1.0 / VS .NET")]` literals remain (DecimalConstants, FixedBuffers, StringConcat, AutomaticEvents, UsingStatement, ForEachStatement, LockStatement, SwitchStatementOnString, SparseIntegerSwitch, UsingDeclarations, ExpandParamsArguments, ArrayInitializers, ForStatement, DoWhileStatement), duplicating exactly the per-version category knowledge the generator was built to single-source.

A typo in any one copy silently creates a new settings-UI group, and renaming the category means 14 coordinated edits instead of one map entry. Since `LanguageVersion` has no value below `CSharp1 = 1`, adding `CSharp1` to `CategoryByVersion` and writing `[DecompilerSetting(CSharp.LanguageVersion.CSharp1)]` derives the category with observably identical `SetLanguageVersion`/`GetMinimumRequiredVersion` semantics (the `< CSharp1` bucket can never fire, and the ladder entry returns the same `CSharp1` the fallback already returns).

ICSharpCode.Decompiler/DecompilerSettings.cs
siegfriedpammer: Now pinned by DecompilerSettingsTests.GetMinimumRequiredVersionReturnsTheHighestEnabledFeatureVersion (7a7cb44a2): enabling UseEnhancedUsing alone raises the minimum to C# 8.0. Rationale as in the round-1 thread: an enabled syntax-preference setting emits that version's syntax, so the old lower stamp could produce an exported project that does not compile.

*This reply was written by the Claude Code agent session working on this PR.*

ICSharpCode.Decompiler/DecompilerSettings.cs
siegfriedpammer: The C# 11 gating is likewise pinned by the new settings test (7a7cb44a2). The output change for pre-C# 11 profiles is the intent: with the setting enabled the decompiler emits C# 11-only syntax, so it now behaves like every other version-gated syntax option.

*This reply was written by the Claude Code agent session working on this PR.*

ICSharpCode.Decompiler.Generators/DecompilerSettingsGenerator.cs
siegfriedpammer: Deliberate scope decision, with one addition. Structural generator mistakes are compile-time by design: DSTG002-005 plus partial-member matching turn every malformed input from the two review rounds into a targeted diagnostic. Emission regressions (defaults, change notification, the SetLanguageVersion cascade) light up the fixture suite, which decompiles with version-profiled settings - and bucket order in SetLanguageVersion is semantically irrelevant, since the buckets are independent if statements. The one failure mode that compiles green and stays suite-silent - dropping the reversed scan in GetMinimumRequiredVersion - is now pinned by DecompilerSettingsTests (7a7cb44a2). Beyond that, keeping a CSharpGeneratorDriver harness out of the repo is intentional; if the generator gains more consumers than DecompilerSettings, that is the point to revisit.

*This reply was written by the Claude Code agent session working on this PR.*

PR #3971, opened 2026-08-09
Fix #3510: keep exporting a project when a member cannot be decompiled

### LINKED ISSUE #3510: Save project - single fail lead to total crash
### Steps to reproduce
1. get any binary
2. save as project
3. it should fail & stop

### Error message shown

regular ILSpy error

### Expectations
Is it possible to skip a single fail on a single function and continue project saving until the end?
right  now it fully stops and there is no any chanve to move forward even having a single error in millions of routines...

### Details
* ILSpy version 10.0.0.8039-preview1+fa4108e450e0045d1632f5f0207cc8a840f73196
* .NET version 10.0.0-preview.5.25277.114+ddf39a1b4690fbe23aea79c78da67004a5c31094

siegfriedpammer:
### Problem

Save Code / `ilspycmd -p` aborted the entire export on the first member it could not decompile. One unsupported method in a large assembly left the user with nothing: no sources, no `.csproj`, and no way to work around it (#3510).

### What changed

- `CSharpDecompiler.DecompileBody` records the failure instead of throwing. The member keeps its signature and the exception takes the place of its body - a comment with the full trace and where to report it, in the same spot warnings about the code go. The remaining members decompile normally, and the failures are exposed as `CSharpDecompiler.Errors` (reset per decompilation).
- `WholeProjectDecompiler` applies the same rule per file, per resource and around the assembly-info file, so an export always runs to completion. What it recovered from is on `WholeProjectDecompiler.Errors`.
- Those failures are surfaced where the user looks when the export finishes: the export report in ILSpy (one headline per failure, each with its full exception in a collapsed "Exception details" fold) and stderr in `ilspycmd`. The `ITextOutput` handed to the language is discarded by both export callers, so the errors travel on `DecompilationOptions.DecompilationErrors` and are rendered by the caller that owns the report.
- `ErrorTolerantOutputVisitor` keeps a file well-formed when writing a member throws: the error takes that member's place, the braces it opened are closed, and the following members are still written.
- The consumers that relied on the exception keep their failure signal: the PowerShell cmdlets raise one error record per failure, and `RoundtripAssembly` asserts the export reported none - otherwise a decompiler crash on a method the round-trip's own tests never call would ship green.
- **`ilspycmd` keeps its failure signal**: it lists the failures on stderr and exits non-zero, as it did when the decompiler threw. The new `--ignore-decompilation-errors` flag exits with success instead, for automation that wants the partial output to count as a success; the failures are still listed either way.

Recovering silently would trade one bad outcome for a worse one, so the inline comment names `https://github.com/icsharpcode/ILSpy/issues/new` and every summary repeats it.

This also improves the text view: a single bad method used to replace the whole type with an error page; the type is now shown with the error in place of that one member.

### Release notes

> **Project export no longer stops at the first member it cannot decompile.** The affected member is written out with the error text in place of its code and the export runs to completion; ILSpy lists the failures in the export report and `ilspycmd` prints them to stderr. `ilspycmd` exits with a non-zero status when this happens - pass `--ignore-decompilation-errors` to treat such a run as successful.

### Drive-by fixes

- `ilspycmd` wrote the `.csproj` through `File.OpenWrite`, which does not truncate: re-exporting into the same directory could leave a shorter project file with trailing bytes from the previous run.
- `SmartTextOutputExtensions.WriteExceptionDetails` split the exception text without trimming, so for exceptions that render a trailing newline (`DecompilerException` does) the fold reached one line past the last frame and a collapsed fold swallowed the line behind it. That affected the PDB generator and the assembly tree node too.
- `DecompilerTabPageModel`'s decompilation-failure path dumped the raw stack trace into the text view instead of using that folded helper.

### Tests

- `DecompilationErrorRecoveryTests` - a failing method body keeps the rest of the type, is recorded as an error, and does not carry over into the next decompilation.
- `WholeProjectDecompilerTests.FailuresDoNotAbortTheExport` - a failing decompile, two failing file creations (a type file and the assembly-info file) and a failing resource enumeration are all reported, and every other file is still written.
- `WholeProjectDecompilerTests.OneFailingResourceDoesNotDropTheOthers` - a resource that cannot be written costs that resource alone.
- `ProjectExportTests.Export_Report_Names_The_Failures_And_Where_To_Report_Them` - the export report names the failure and the report URL, and the export still counts as successful.
- `DecompilationErrorRecoveryTests.FailingOutputKeepsTheFileWellFormed` - a member whose output throws is replaced in place, the file closes every brace it opens, and the members after it are still written.
- `SmartTextOutputExtensionsTests.Exception_Fold_Stops_At_The_Last_Frame` - the fold does not extend past the last frame.

Failures are injected with a throwing `IILTransform` / `IAstTransform` and an overridden `CreateFile`, so the tests do not depend on a decompiler bug that may get fixed.

Decompiler suite green (3368 tests); `ilspycmd` tests and the affected UI tests green.

---

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
+++ b/ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
@@ -291,6 +291,59 @@ namespace ICSharpCode.Decompiler.CSharp
 		}
 
 		/// <summary>
+		/// Method bodies that could not be decompiled. Instead of aborting the surrounding type,
+		/// such a member is emitted with the error text in place of its body (see
+		/// <see cref="GetErrorCommentLines"/>) and the exception is collected here, so callers
+		/// decompiling many members - the project exporter above all - can tell the user how many
+		/// members are affected.
+		/// </summary>
+		public IReadOnlyList<DecompilerException> Errors => errors;
+
+		readonly List<DecompilerException> errors = new List<DecompilerException>();
+
+		/// <summary>
+		/// Where users are asked to report decompilation failures; part of the error text emitted
+		/// into the output, because a failure nobody reports is a failure nobody fixes.
+		/// </summary>
+		public const string DecompilationErrorReportUrl = "https://github.com/icsharpcode/ILSpy/issues/new";
+
+		/// <summary>
+		/// The headline a front end puts above the list of failures it recovered from. Shared so the
+		/// UI, the command line and any other consumer say the same thing and point at the same URL.
+		/// </summary>
+		public static IEnumerable<string> GetErrorSummaryLines(int errorCount)
+		{
+			yield return $"{errorCount} error(s) occurred; the affected code was replaced by the error text in the output.";
+			yield return $"Please report them at {DecompilationErrorReportUrl}:";
+		}
+
+		/// <summary>
+		/// The one-line description of a single failure, so the UI and the command line name it the
+		/// same way.
+		/// </summary>
+		public static string GetErrorHeadline(DecompilerException error)
+		{
+			if (error == null)
+				throw new ArgumentNullException(nameof(error));
+			return error.InnerException == null ? error.Message : $"{error.Message}: {error.InnerException.Message}";
+		}
+
+		/// <summary>
+		/// Renders <paramref name="error"/> as the lines of a comment block: an explanation, the
+		/// request to report it, and the full exception including its stack trace, which is what
+		/// makes such a report actionable.
+		/// </summary>
+		internal static IEnumerable<string> GetErrorCommentLines(Exception error)
+		{
+			yield return "ILSpy could not decompile this. Please report the exception below,";
+			yield return "along with the assembly it came from, at " + DecompilationErrorReportUrl;
+			foreach (string line in error.ToString().Split('\n'))
+			{
+				yield return line.TrimEnd('\r');
+			}
+		}
+
+		/// <summary>
 		/// Creates a new <see cref="CSharpDecompiler"/> instance from the given <paramref name="fileName"/> using the given <paramref name="settings"/>.
 		/// </summary>
 		public CSharpDecompiler(string fileName, DecompilerSettings settings)
@@ -752,6 +805,10 @@ namespace ICSharpCode.Decompiler.CSharp
 
 		DecompileRun CreateDecompileRun(HashSet<string> namespaces)
 		{
+			// Every public Decompile* entry point starts here, so this is where the failures of the
+			// previous one stop counting - otherwise a reused instance reports them again against
+			// members that decompiled cleanly.
+			errors.Clear();
 			List<INamespace> resolvedNamespaces = new List<INamespace>();
 			foreach (var ns in namespaces)
 			{
@@ -2283,9 +2340,34 @@ namespace ICSharpCode.Decompiler.CSharp
 
 				CleanUpMethodDeclaration(entityDecl, body, function, localSettings.DecompileMemberBodies);
 			}
-			catch (Exception innerException) when (!(innerException is OperationCanceledException || innerException is DecompilerException))
+			catch (Exception innerException) when (!(innerException is OperationCanceledException))
 			{
-				throw new DecompilerException(module, method, innerException);
+				// One method the decompiler cannot handle must not cost the user the type or, when
+				// exporting a project, the assembly around it: keep the signature, put the error in
+				// front of it, and let the remaining members decompile.
+				errors.Add(innerException as DecompilerException ?? new DecompilerException(module, method, innerException));
+				entityDecl.GetChild(Slots.Body)?.Remove();
+				if (settings.DecompileMemberBodies)
+				{
+					// The error goes where the code would have been, the same way a warning about the
+					// code does - and the body keeps the member's shape intact.
+					var errorBody = new BlockStatement();
+					var errorStatement = new EmptyStatement();
+					foreach (string line in GetErrorCommentLines(innerException))
+					{
+						errorStatement.AddTrailingTrivia(new Comment(" " + line));
+					}
+					errorBody.Statements.Add(errorStatement);
+					entityDecl.AddChild(errorBody, Slots.Body);
+				}
+				else
+				{
+					// Definitions-only output has no body to put the error in.
+					foreach (string line in GetErrorCommentLines(innerException))
+					{
+						entityDecl.AddLeadingTrivia(new Comment(" " + line));
+					}
+				}
 			}
 		}
 
--- a/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
+++ b/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
@@ -151,7 +151,81 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 
 		// per-run members
 		HashSet<string> directories = new HashSet<string>(Platform.FileNameComparer);
+		readonly List<DecompilerException> errors = new List<DecompilerException>();
 		readonly IProjectFileWriter projectWriter;
+
+		/// <summary>
+		/// Everything that went wrong during the last <see cref="DecompileProject(MetadataFile, string, CancellationToken)"/>.
+		/// An export never aborts on a member, file or resource it cannot handle; it writes the
+		/// error text where the content would have gone and continues, so a single unsupported
+		/// method still yields a complete project. Callers should show this list to the user -
+		/// otherwise the failures ship silently and never get reported.
+		/// </summary>
+		public IReadOnlyList<DecompilerException> Errors => errors;
+
+		void RecordError(DecompilerException error)
+		{
+			lock (errors)
+			{
+				errors.Add(error);
+			}
+		}
+
+		/// <summary>
+		/// Yields the items of <paramref name="items"/> until one of them throws; the failure is
+		/// recorded instead of aborting the export.
+		/// </summary>
+		IEnumerable<T> RecordingErrors<T>(IEnumerable<T> items, MetadataFile file, string what)
+		{
+			using var enumerator = items.GetEnumerator();
+			bool lastMoveFailed = false;
+			while (true)
+			{
+				T item;
+				try
+				{
+					if (!enumerator.MoveNext())
+						yield break;
+					item = enumerator.Current;
+				}
+				catch (Exception ex) when (!(ex is OperationCanceledException))
+				{
+					RecordError(ex as DecompilerException ?? new DecompilerException(file, $"Error writing {what}", ex));
+					// Skip the item that failed and try the next one, but give up once two attempts
+					// in a row fail: an enumerator that throws without advancing - which nothing
+					// stops an override from being - would otherwise loop forever.
+					if (lastMoveFailed)
+						yield break;
+					lastMoveFailed = true;
+					continue;
+				}
+				lastMoveFailed = false;
+				yield return item;
+			}
+		}
+
+		/// <summary>
+		/// Puts the error text where the file's contents would have gone. The writer itself may be
+		/// what failed - a full disk, a stream already closed - so a second failure while reporting
+		/// the first is dropped rather than allowed to take the export down.
+		/// </summary>
+		static void WriteErrorComment(TextWriter? writer, Exception error)
+		{
+			if (writer == null)
+				return;
+			try
+			{
+				// The failure may have interrupted the output visitor mid-line.
+				writer.WriteLine();
+				foreach (string line in CSharpDecompiler.GetErrorCommentLines(error))
+				{
+					writer.WriteLine("// " + line);
+				}
+			}
+			catch (Exception ex) when (!(ex is OperationCanceledException))
+			{
+			}
+		}
 
 		public void DecompileProject(MetadataFile file, string targetDirectory, CancellationToken cancellationToken = default(CancellationToken))
 		{
@@ -182,7 +256,8 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 			{
 				TargetDirectory = targetDirectory;
 				directories.Clear();
-				var resources = WriteResourceFilesInProject(file).ToList();
+				errors.Clear();
+				var resources = RecordingErrors(WriteResourceFilesInProject(file), file, "resource files").ToList();
 				resourceFileCount = resources.Count;
 				var files = WriteCodeFilesInProject(file, resources.SelectMany(r => r.PartialTypes ?? Enumerable.Empty<PartialTypeInfo>()).ToList(), cancellationToken).ToList();
 				codeFileCount = files.Count;
@@ -190,7 +265,7 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 				var module = file as PEFile;
 				if (module != null)
 				{
-					files.AddRange(WriteMiscellaneousFilesInProject(module));
+					files.AddRange(RecordingErrors(WriteMiscellaneousFilesInProject(module), file, "miscellaneous files"));
 				}
 				if (StrongNameKeyFile != null)
 				{
@@ -277,6 +352,7 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 			var progressReporter = ProgressIndicator;
 			var progress = new DecompilationProgress { TotalUnits = files.Count, Title = "Exporting project..." };
 			DecompilerTypeSystem ts = new DecompilerTypeSystem(module, AssemblyResolver, Settings);
+			var missingFiles = new ConcurrentBag<string>();
 			var workList = new HashSet<TypeDefinitionHandle>();
 			var processedTypes = new HashSet<TypeDefinitionHandle>();
 			ProcessFiles(files);
@@ -290,7 +366,21 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 				progress.TotalUnits = files.Count;
 			}
 
-			return files.Select(f => new ProjectItemInfo("Compile", f.Key)).Concat(WriteAssemblyInfo(ts, cancellationToken));
+			// The assembly-level attributes are a single file like any other: failing to decompile
+			// them costs that file, not the export.
+			IEnumerable<ProjectItemInfo> assemblyInfo;
+			try
+			{
+				assemblyInfo = WriteAssemblyInfo(ts, cancellationToken);
+			}
+			catch (Exception ex) when (!(ex is OperationCanceledException))
+			{
+				RecordError(ex as DecompilerException ?? new DecompilerException(module, "Error decompiling the module and assembly attributes", ex));
+				assemblyInfo = Enumerable.Empty<ProjectItemInfo>();
+			}
+
+			return files.Select(f => f.Key).Except(missingFiles, Platform.FileNameComparer)
+				.Select(f => new ProjectItemInfo("Compile", f)).Concat(assemblyInfo);
 
 			string GetFileFileNameForHandle(TypeDefinitionHandle h)
 			{
@@ -325,10 +415,14 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 					delegate (IGrouping<string, TypeDefinitionHandle> file) {
 						var declaredTypes = file.ToArray();
 						DecompilerEventSource.Log.ProjectFileStart(file.Key, declaredTypes.Length);
+						// Everything that can fail for this one file - creating it included, which is
+						// where a path too long for the file system surfaces - belongs inside the try.
+						TextWriter? w = null;
+						CSharpDecompiler? decompiler = null;
 						try
 						{
-							using var w = CreateFile(Path.Combine(TargetDirectory, file.Key));
-							CSharpDecompiler decompiler = CreateDecompiler(ts);
+							w = CreateFile(Path.Combine(TargetDirectory, file.Key));
+							decompiler = CreateDecompiler(ts);
 
 							foreach (var partialType in partialTypes)
 							{
@@ -356,14 +450,44 @@ namespace ICSharpCode.Decompiler.CSharp.ProjectDecompiler
 								}
 							}
 
-							syntaxTree.AcceptVisitor(new CSharpOutputVisitor(w, Settings.CSharpFormattingOptions));
+							// A member the output visitor cannot write is replaced by the error text
+							// rather than truncating the file where it failed.
+							var outputVisitor = new ErrorTolerantOutputVisitor(w, Settings.CSharpFormattingOptions);
+							syntaxTree.AcceptVisitor(outputVisitor);
+							foreach (var outputError in outputVisitor.Errors)
+							{
+								RecordError(new DecompilerException(module, $"Error writing '{file.Key}'", outputError));
+							}
 						}
-						catch (Exception innerException) when (!(innerException is OperationCanceledException || innerException is DecompilerException))
+						catch (Exception innerException) when (!(innerException is OperationCanceledException))
 						{
-							throw new DecompilerException(module, $"Error decompiling for '{file.Key}'", innerException);
+							// Whatever the decompiler could not cope with here, the remaining files
+							// are unaffected and the user still gets a complete project; the error
+							// takes the place of the file's contents.
+							RecordError(innerException as DecompilerException ?? new DecompilerException(module, $"Error decompiling for '{file.Key}'", innerException));
+							if (w == null)
+							{
+								// Nothing was written, so nothing can carry the error text - and the
+								// project must not claim a file that is not there.
+								missingFiles.Add(file.Key);
+							}
+							WriteErrorComment(w, innerException);
 						}
 						finally
 						{
+							foreach (var error in decompiler?.Errors ?? (IReadOnlyList<DecompilerException>)Array.Empty<DecompilerException>())
+							{
+								RecordError(error);
+							}
+							try
+							{
+								w?.Dispose();
+							}
+							catch (Exception ex) when (!(ex is OperationCanceledException))
+							{
+								// Dispose flushes: on a full disk this is where the write actually fails.
+								RecordError(new DecompilerException(module, $"Error writing '{file.Key}'", ex));
+							}
 							DecompilerEventSource.Log.ProjectFileStop(file.Key);
 						}
 						progress.Status = file.Key;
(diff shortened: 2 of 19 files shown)


christophwille (commented):
## Code review (high effort, multi-agent, Claude Code)

The PR's error-recovery machinery has structural gaps that defeat its own goal: `RecordingErrors` wraps an eager method so code-file failures still abort the whole export, `CreateFile` was hoisted out of the per-file `try` so one I/O failure kills the entire run, the catch handler writes to the possibly-failed writer, and the yield-break wrapper silently drops all resources after the first failure.

Additionally, the removed rethrow silently changes exit-code behavior for ilspycmd and PowerShell consumers, `CSharpDecompiler.Errors` accumulates across calls without a reset, and the UI exporter drops recovered errors when the export later throws. A pre-existing `File.OpenWrite` truncation bug and two minor cleanups (duplicated summary phrasing, a history-referencing comment) round out the findings.

**10 findings** (8 correctness, 2 cleanup) posted as inline comments; all were independently verified against the PR head.

---
*Automated review run via Claude Code (claude-fable-5); findings verified by independent verifier agents.*

ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
```diff
@@ -160,15 +203,16 @@ protected WholeProjectDecompiler(
 			{
 				TargetDirectory = targetDirectory;
 				directories.Clear();
-				var resources = WriteResourceFilesInProject(file).ToList();
+				errors.Clear();
+				var resources = RecordingErrors(WriteResourceFilesInProject(file), file, "resource files").ToList();
 				resourceFileCount = resources.Count;
-				var files = WriteCodeFilesInProject(file, resources.SelectMany(r => r.PartialTypes ?? Enumerable.Empty<PartialTypeInfo>()).ToList(), cancellationToken).ToList();
+				var files = RecordingErrors(WriteCodeFilesInProject(file, resources.SelectMany(r => r.PartialTypes ?? Enumerable.Empty<PartialTypeInfo>()).ToList(), cancellationToken), file, "code files").ToList();
```
christophwille: **[correctness] `RecordingErrors` wraps an eager method here, so it records nothing.**

`WriteCodeFilesInProject` is a regular method, not an iterator: the `DecompilerTypeSystem` construction, `ProcessFiles` (the whole `Parallel.ForEach`), and `WriteAssemblyInfo` all execute during the call itself, *before* `RecordingErrors` ever wraps the returned enumerable. Only the pure `files.Select` projection (which cannot throw) is deferred and thus guarded.

Failure scenario: an assembly whose module/assembly attributes fail to decompile (`WriteAssemblyInfo` -> `DecompileModuleAndAssemblyAttributes` throws, e.g. obfuscated attribute blobs), or where creating `Properties/AssemblyInfo.cs` fails, throws straight out of `DecompileProject`. The entire export aborts with an unhandled exception and nothing lands in `Errors` - contradicting the `Errors` doc-comment contract ("An export never aborts on a member, file or resource it cannot handle"). In ilspycmd the command dies with a stack trace instead of printing the error summary, which is exactly the #3510 symptom this PR sets out to fix.

ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
```diff
@@ -303,11 +347,10 @@ void ProcessFiles(List<IGrouping<string, TypeDefinitionHandle>> files)
 					delegate (IGrouping<string, TypeDefinitionHandle> file) {
 						var declaredTypes = file.ToArray();
 						DecompilerEventSource.Log.ProjectFileStart(file.Key, declaredTypes.Length);
+						using var w = CreateFile(Path.Combine(TargetDirectory, file.Key));
```
christophwille: **[correctness] Moving `CreateFile` out of the per-file `try` lets one file-creation failure abort the whole export.**

`CreateFile` (and `CreateDecompiler`) now run before the `try`, so a per-file file-creation failure escapes the per-file catch, cancels the whole `Parallel.ForEach` with an `AggregateException`, and - since the `RecordingErrors` wrapper at the call site is ineffective (see comment there) - aborts the entire export unrecorded.

Failure scenario: on Windows, a type/namespace whose cleaned-up file name pushes the combined path past MAX_PATH (or an invalid/locked target file) makes `CreateFile` throw inside the `Parallel.ForEach` delegate; all remaining files are lost and no error is recorded or reported. PathTooLong during export is a known real-world case - `SolutionWriter` has a dedicated `PathTooLongException` catch for it - and the PR's whole point is that one bad file must not cost the user the project.

ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
```diff
@@ -336,12 +379,25 @@ void ProcessFiles(List<IGrouping<string, TypeDefinitionHandle>> files)
...
-						catch (Exception innerException) when (!(innerException is OperationCanceledException || innerException is DecompilerException))
+						catch (Exception innerException) when (!(innerException is OperationCanceledException))
 						{
-							throw new DecompilerException(module, $"Error decompiling for '{file.Key}'", innerException);
+							// Whatever the decompiler could not cope with here, the remaining files
+							// are unaffected and the user still gets a complete project; the error
+							// takes the place of the file's contents.
+							RecordError(innerException as DecompilerException ?? new DecompilerException(module, $"Error decompiling for '{file.Key}'", innerException));
+							// The failure may have interrupted the output visitor mid-line.
+							w.WriteLine();
```
christophwille: **[correctness] The catch handler writes the error comment into the same writer that may have just failed.**

If the exception came from `w` itself (e.g. `IOException` from disk-full or a closed stream during `AcceptVisitor`), the handler's `w.WriteLine()` throws a second `IOException` outside any handler, escapes the delegate, and aborts the whole export.

Failure scenario: the disk fills up mid-export; the output visitor throws `IOException`, the catch records it and then writes to the same failed writer; `Parallel.ForEach` aborts with an `AggregateException` and the export terminates with an unhandled exception instead of the recorded-error report the PR promises.

ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
```diff
@@ -137,8 +137,51 @@ protected WholeProjectDecompiler(
...
+				try
+				{
+					if (!enumerator.MoveNext())
+						yield break;
+					item = enumerator.Current;
+				}
+				catch (Exception ex) when (!(ex is OperationCanceledException))
+				{
+					RecordError(ex as DecompilerException ?? new DecompilerException(file, $"Error writing {what}", ex));
+					yield break;
```
christophwille: **[correctness] `yield break` on first failure silently drops every subsequent item.**

`RecordingErrors` stops enumerating at the first item that throws, so a single bad resource or misc item drops every subsequent item from the export while recording only one error.

Failure scenario: an assembly with many embedded resources where one `.resources` blob fails to decode - every resource after the failing one is silently omitted from the exported project (no file on disk, no csproj item), yet the report shows a single "Error writing resource files" entry. The user believes one resource failed when most of the project's resources are actually missing. Since the wrapped iterators yield per item, per-item recovery (catch inside a loop over the enumerator, continue on failure) is achievable.

ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
```diff
@@ -2283,9 +2315,23 @@ void DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun de
 
 				CleanUpMethodDeclaration(entityDecl, body, function, localSettings.DecompileMemberBodies);
 			}
-			catch (Exception innerException) when (!(innerException is OperationCanceledException || innerException is DecompilerException))
+			catch (Exception innerException) when (!(innerException is OperationCanceledException))
```
christophwille, line 2343: **[correctness] The removed rethrow drops the failure signal for every consumer except project export.**

Only `WholeProjectDecompiler` / ilspycmd project mode / the UI exporter read the new `Errors` list. Plain ilspycmd decompilation (`IlspyCmdProgram.cs:653` path) and the PowerShell cmdlets previously surfaced the `DecompilerException` as a hard error.

Failure scenario: `ilspycmd broken.dll -o out` on an assembly with a method body the decompiler cannot handle used to exit with EX_SOFTWARE and an error on stderr (OnExecute's catch); it now exits 0 with silent stderr, so CI scripts gating on the exit code accept output containing empty method bodies (invalid C# for non-void methods). `Get-DecompiledSource` in the PowerShell module likewise returns broken source without raising an error record.

ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
```diff
@@ -2283,9 +2315,23 @@ void DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun de
...
 			{
-				throw new DecompilerException(module, method, innerException);
+				// One method the decompiler cannot handle must not cost the user the type or, when
+				// exporting a project, the assembly around it: keep the signature, put the error in
+				// front of it, and let the remaining members decompile.
+				errors.Add(innerException as DecompilerException ?? new DecompilerException(module, method, innerException));
```
christophwille, line 2348: **[correctness] The `errors` list is never cleared, so a reused `CSharpDecompiler` accumulates errors across calls.**

This is the only mutation of the `errors` field in the file, and there is no reset anywhere - so a reused instance attributes earlier failures to later, successful decompilations.

Failure scenario: a caller reusing one `CSharpDecompiler` to decompile types one at a time and checking `decompiler.Errors` after each call (the exact consumption pattern `WholeProjectDecompiler` models per file) sees the first type's `DecompilerException` reported again after every subsequent type, and writes the same error comment/report against types that decompiled cleanly. `WholeProjectDecompiler.Errors` got an explicit `errors.Clear()` per run; `CSharpDecompiler.Errors` has no equivalent.

ILSpy/Commands/ProjectExporter.cs
```diff
@@ -150,6 +175,7 @@ static SolutionExportResult ExportProject(LoadedAssembly assembly, ProjectExport
 				};
 				language.DecompileAssembly(assembly, new PlainTextOutput(new StringWriter()), decompileOptions);
 				report.AppendLine("Project written to " + options.OutputDirectory);
+				errors = decompileOptions.DecompilationErrors;
```
christophwille: **[correctness] Recovered per-member errors are dropped from the report when `DecompileAssembly` later throws.**

`errors = decompileOptions.DecompilationErrors` executes only on the success path; the catch sets `success = false` and leaves `errors` empty.

Failure scenario: an export recovers from several member failures (error text already written into the .cs files on disk) but then fails at the end (e.g. `IOException` writing the .csproj). `SolutionExportResult.Errors` is `[]` and the result report never mentions the "ILSpy could not decompile this" blocks sitting in the partially written sources - the failures ship silently, exactly what the `Errors` documentation warns callers not to allow.

ICSharpCode.ILSpyCmd/IlspyCmdProgram.cs
```diff
@@ -645,8 +645,21 @@ ProjectId DecompileAsProject(string assemblyFileName, string projectFileName)
 			{
 				decompiler = new WholeProjectDecompiler(settings, resolver, null, resolver, debugInfo);
 			}
+			ProjectId projectId;
 			using (var projectFileWriter = new StreamWriter(File.OpenWrite(projectFileName)))
```
christophwille: **[correctness, pre-existing] `File.OpenWrite` does not truncate, so a shorter .csproj leaves trailing bytes from the previous run.**

`File.OpenWrite` opens with `FileMode.OpenOrCreate`/`FileAccess.Write` and never truncates. This line is unchanged, but it sits inside the function this PR modifies, and the PR's recovery path makes re-running exports over existing output more likely.

Failure scenario: a user re-runs `ilspycmd -p` into the same output directory after the assembly shrank (or settings changed) and the new project file content is shorter than the old - the resulting .csproj contains the new XML followed by leftover garbage from the previous run, and MSBuild/Visual Studio fails to load the corrupt project. `File.Create` (or `new StreamWriter(path, append: false)`) fixes it.

ICSharpCode.ILSpyCmd/IlspyCmdProgram.cs
```diff
@@ -645,8 +645,21 @@ ProjectId DecompileAsProject(string assemblyFileName, string projectFileName)
...
-				return decompiler.DecompileProject(module, Path.GetDirectoryName(projectFileName), projectFileWriter);
+				projectId = decompiler.DecompileProject(module, Path.GetDirectoryName(projectFileName), projectFileWriter);
+			// Members that could not be decompiled carry the error text in the generated sources
+			// instead of aborting the export; say so here, or nobody notices and nobody reports it.
+			if (decompiler.Errors.Count > 0)
+			{
+				Console.Error.WriteLine($"{decompiler.Errors.Count} error(s) occurred while decompiling {assemblyFileName}.");
```
christophwille: **[cleanup] Error-summary phrasing is duplicated from `ProjectExporter.WriteDecompilationErrors`.**

The "N error(s) occurred" / "the affected members were exported with the error text in place" / "Please report them at <url>" headline now exists twice - once here for stderr and once in `ILSpy/Commands/ProjectExporter.cs` - and the two copies already differ slightly ("members" vs "code"). Future wording or URL changes must be made in both places or the front ends drift apart. A shared summary-line helper next to `GetErrorCommentLines`/`DecompilationErrorReportUrl` in ICSharpCode.Decompiler (which both projects reference) would keep one copy.

ILSpy.Tests/SmartTextOutputExtensionsTests.cs
```diff
@@ -0,0 +1,65 @@
...
+		text[fold.StartOffset..fold.EndOffset].Should().NotEndWith("\n",
+			"the fold must end with the last frame, not with the newline behind it");
+		text.Should().Contain("this line must stay visible");
+	}
+
+	static Exception ExceptionWithTrace() => new TrailingNewlineException();
+
+	/// <summary>
+	/// Stands in for the exceptions that actually reach this helper: <see cref="DecompilerException"/>
+	/// renders a trailing newline, and that is what used to push the fold one line too far.
```
christophwille: **[cleanup] Comment references pre-fix behavior ("what used to push the fold one line too far").**

CLAUDE.md's comment rule: comments must stand on their own, with no memory of how the code was written - never reference the previous version. The doc comment on `TrailingNewlineException` only makes sense relative to the pre-PR version of `WriteExceptionDetails`; a cold reader has no "used to" to compare against. Stating the invariant instead ("a trailing newline would push the fold one line too far") keeps the same information without the history reference.

ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
```diff
@@ -268,7 +338,20 @@ IEnumerable<ProjectItemInfo> WriteCodeFilesInProject(MetadataFile module, IList<
...
+			{
+				assemblyInfo = WriteAssemblyInfo(ts, cancellationToken);
+			}
+			catch (Exception ex) when (!(ex is OperationCanceledException))
+			{
+				RecordError(ex as DecompilerException ?? new DecompilerException(module, "Error decompiling the module and assembly attributes", ex));
+				assemblyInfo = Enumerable.Empty<ProjectItemInfo>();
+			}
+
+			return files.Select(f => new ProjectItemInfo("Compile", f.Key)).Concat(assemblyInfo);
```
christophwille: **[correctness] The .csproj still references .cs files whose creation failed.**

`files.Select` emits a `Compile` `ProjectItemInfo` for every file group unconditionally, including groups whose `CreateFile` or decompilation failed and were skipped by the recovery.

Failure scenario: export with the classic (non-SDK) project format where `CreateFile` throws for one file (e.g. `PathTooLongException` for a deeply nested namespace, or the `IOException` path the new `FailuresDoNotAbortTheExport` test itself simulates). `ProcessFiles` records the error and skips the file, but `ProjectFileWriterDefault` writes an explicit `<Compile Include="..."/>` item for it - the exported project fails to load/build in Visual Studio/MSBuild with a missing-file error, while `GetErrorSummaryLines` told the user "the affected code was replaced by the error text in the output" even though no such file exists. Before this PR the export aborted instead of producing a broken .csproj.

ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
```diff
@@ -2283,9 +2329,34 @@ void DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun de
 
 				CleanUpMethodDeclaration(entityDecl, body, function, localSettings.DecompileMemberBodies);
 			}
-			catch (Exception innerException) when (!(innerException is OperationCanceledException || innerException is DecompilerException))
+			catch (Exception innerException) when (!(innerException is OperationCanceledException))
```
christophwille, line 2343: **[correctness] Consumers not updated to read `Errors` silently lose the failure signal - the PowerShell cmdlets in this repo among them.**

`DecompileBody` no longer throws, so any library consumer that relied on the exception and was not taught to inspect the new `Errors` list loses failures entirely. In this repo, the PowerShell cmdlets' catch/`WriteError(DecompilationFailed)` path is now dead code.

Failure scenario: `Get-DecompiledSource` (or `Get-DecompiledProject`) on an assembly containing a method the decompiler cannot handle. Before, the cmdlet raised a PowerShell error record (`DecompilationFailed`) that scripts could trap; now it completes successfully, returns source with error comments buried inside, and never inspects `Decompiler.Errors`/`WholeProjectDecompiler.Errors` - automated pipelines treat known-broken output as clean.

ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
```diff
@@ -160,15 +229,16 @@ protected WholeProjectDecompiler(
...
+				var resources = RecordingErrors(WriteResourceFilesInProject(file), file, "resource files").ToList();
 				resourceFileCount = resources.Count;
 				var files = WriteCodeFilesInProject(file, resources.SelectMany(r => r.PartialTypes ?? Enumerable.Empty<PartialTypeInfo>()).ToList(), cancellationToken).ToList();
 				codeFileCount = files.Count;
 				files.AddRange(resources);
 				var module = file as PEFile;
 				if (module != null)
 				{
-					files.AddRange(WriteMiscellaneousFilesInProject(module));
+					files.AddRange(RecordingErrors(WriteMiscellaneousFilesInProject(module), file, "miscellaneous files"));
```
christophwille, line 268: **[correctness] A failure writing one misc file silently drops the rest (app.manifest, app.config).**

`WriteMiscellaneousFilesInProject` got only the `RecordingErrors` enumeration wrapper, not per-item recovery. A throw while writing app.ico leaves the compiler-generated iterator finished, so `RecordingErrors` records one icon error and stops - app.manifest and app.config, yielded after the icon in the same iterator, are never written and never mentioned. This is the exact truncation the PR's own resource test pins against for resources.

Failure scenario: `File.WriteAllBytes` for app.ico throws (disk full, read-only target, or corrupt icon-group data in `CreateApplicationIcon`). The user reads a report naming only the icon failure and ships an exported project that silently lost its application manifest and config file; the removed abort at least made the incompleteness loud.

ILSpy/Languages/CSharpLanguage.cs
```diff
@@ -558,6 +558,12 @@ public override void DecompileType(ITypeDefinition type, ITextOutput output, Dec
 			using (var writer = new System.IO.StreamWriter(projectFileName))
 				id = decompiler.DecompileProject(module, targetDirectory, writer, options.CancellationToken);
 			output.WriteLine("// Project written to " + targetDirectory);
+			// The export does not abort on what it cannot decompile; hand the failures to the caller,
+			// whose result report - not this ITextOutput - is what the user sees when it finishes.
+			foreach (var error in decompiler.Errors)
```
christophwille: **[correctness] Recovered errors are lost from the UI report when `DecompileProject` itself throws.**

`decompiler.Errors` is copied into `options.DecompilationErrors` only after `DecompileProject` returns; when it throws (e.g. `projectWriter.Write` or the strong-name-key `File.Copy` fails), this copy loop never runs.

Failure scenario: during a UI project export, several methods fail decompilation (recovered, error stubs written into the .cs files on disk), then writing the .csproj throws an `IOException`. `ProjectExporter.ExportProject`'s catch path builds the report from `decompileOptions.DecompilationErrors`, which is empty - the user sees only the fatal IO error and never learns that the source files already written to their output directory contain silently stubbed-out members. Copying errors in a finally (or reading them in the catch path) would preserve them.

ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
```diff
@@ -336,12 +423,21 @@ void ProcessFiles(List<IGrouping<string, TypeDefinitionHandle>> files)
...
 						}
-						catch (Exception innerException) when (!(innerException is OperationCanceledException || innerException is DecompilerException))
+						catch (Exception innerException) when (!(innerException is OperationCanceledException))
 						{
-							throw new DecompilerException(module, $"Error decompiling for '{file.Key}'", innerException);
+							// Whatever the decompiler could not cope with here, the remaining files
+							// are unaffected and the user still gets a complete project; the error
+							// takes the place of the file's contents.
+							RecordError(innerException as DecompilerException ?? new DecompilerException(module, $"Error decompiling for '{file.Key}'", innerException));
+							WriteErrorComment(w, innerException);
```
christophwille, line 474: **[correctness] The error comment is appended after partial content, but the message claims the code was "replaced".**

When the output visitor throws midway through writing a type (reachable after part of the file was already emitted - e.g. an exception from a resolve-annotation lookup during output, or an `IOException` mid-write), `WriteErrorComment` appends the comment after whatever `CSharpOutputVisitor` already wrote, rather than replacing the file's contents.

Failure scenario: the exported .cs file contains half a class with unclosed braces plus the trailing error comment; the exported project fails to compile with cascading syntax errors in that file, while both the error summary shown to the user ("the affected code was replaced by the error text in the output") and the code comment claim the error text took the place of the contents. Truncating the stream (`w.Flush(); stream.SetLength(0)` or reopening the file) before writing the comment would match the message.

christophwille (commented):
## Re-review (high effort, multi-agent, Claude Code)

Re-reviewed at head 9874542 after the updates. The earlier round's structural issues (eager `RecordingErrors` wrap, `CreateFile` outside the per-file try, catch writing to the failed writer, yield-break dropping items) are addressed. This round surfaced **10 findings** (8 correctness, 2 cleanup), all independently verified, posted as inline comments.

The most severe are correctness gaps in the recovery itself:

- the non-SDK .csproj still references source files whose creation failed (a broken deliverable the old abort never produced),
- the PowerShell cmdlets were never taught to read the new `Errors` lists, so failures become fully silent,
- one mid-iterator failure in `WriteMiscellaneousFilesInProject` silently drops app.manifest/app.config,
- recovered errors are lost from the UI report when the export later throws, and
- an unguarded writer `Dispose` in the finally block can re-introduce the whole-export abort on disk-full.

Remaining findings cover truncated files contradicting the "replaced by error text" message, `RecordingErrors`' unbounded retry against non-compiler-iterator overrides, a stderr/exit-code documentation contradiction in ilspycmd, lost regression coverage in the round-trip suite, and a duplicated per-error headline format.

---
*Automated review run via Claude Code (claude-fable-5); findings verified by independent verifier agents.*

ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
```diff
@@ -137,8 +137,77 @@ protected WholeProjectDecompiler(
...
+		/// </summary>
+		IEnumerable<T> RecordingErrors<T>(IEnumerable<T> items, MetadataFile file, string what)
+		{
+			using var enumerator = items.GetEnumerator();
+			while (true)
+			{
+				T item;
+				try
+				{
+					if (!enumerator.MoveNext())
```
christophwille, line 187: **[correctness] `RecordingErrors` can loop forever on non-compiler-generated iterators.**

The "an iterator that throws is left in its finished state" assumption only holds for compiler-generated iterators. But this wrapper is applied to `WriteResourceFilesInProject`/`WriteMiscellaneousFilesInProject`, which are protected-virtual extension points of a published NuGet library - overrides can return any `IEnumerable`.

Failure scenario: a `WholeProjectDecompiler` subclass overrides `WriteResourceFilesInProject` with a hand-written enumerator that throws the same exception on every `MoveNext` call (e.g. one wrapping a broken stream that throws `IOException` without advancing, or one that validates state and throws `ObjectDisposedException` each call). `RecordingErrors` catches, records, and calls `MoveNext` again in the `while (true)` loop: the export hangs forever and the errors list grows unboundedly until `OutOfMemoryException`, instead of reporting one error and finishing. A bounded retry count (or bailing out after the first catch per enumerator) would keep the guarantee.

ICSharpCode.ILSpyCmd/IlspyCmdProgram.cs
```diff
@@ -622,6 +627,34 @@ int ShowIL(string assemblyFileName, TextWriter output)
...
+		/// Lists what the decompiler could not handle. The output was still written - with the error
+		/// text in place of the affected code - so this is the only sign anything went wrong, and
+		/// <see cref="ExitCodeForDecompilationErrors"/> keeps it from passing silently in a script.
+		/// </summary>
+		void ReportDecompilationErrors(string assemblyFileName, IReadOnlyList<DecompilerException> errors)
+		{
+			if (errors.Count == 0)
+				return;
+			decompilationErrors.AddRange(errors);
+			Console.Error.WriteLine($"While decompiling {assemblyFileName}:");
```
christophwille, line 642: **[correctness] `--ignore-decompilation-errors` still writes the failure list to stderr, contradicting its documentation.**

`ReportDecompilationErrors` writes to stderr unconditionally; the help text and README say "without this option the failures are listed on stderr and the exit status is non-zero", implying the flag suppresses both.

Failure scenario: a CI script runs `ilspycmd -p --ignore-decompilation-errors -o out asm.dll` and, per the documented contract, treats any stderr output as a real failure (a common scripting pattern, e.g. `2>err.txt && [ -s err.txt ]`). The decompiler recovers from one method-body failure: exit code is 0 as promised, but stderr still carries the full error listing, so the script reports the export as failed even though the user explicitly opted into ignoring decompilation errors. Either suppress the listing under the flag or reword the docs to say only the exit status is affected.

ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
```diff
@@ -137,8 +137,77 @@ protected WholeProjectDecompiler(
...
 		readonly IProjectFileWriter projectWriter;
 
+		/// <summary>
+		/// Everything that went wrong during the last <see cref="DecompileProject(MetadataFile, string, CancellationToken)"/>.
+		/// An export never aborts on a member, file or resource it cannot handle; it writes the
+		/// error text where the content would have gone and continues, so a single unsupported
+		/// method still yields a complete project. Callers should show this list to the user -
+		/// otherwise the failures ship silently and never get reported.
+		/// </summary>
+		public IReadOnlyList<DecompilerException> Errors => errors;
```
christophwille, line 164: **[correctness] The round-trip suite loses its regression signal: nothing asserts `Errors` is empty.**

`ICSharpCode.Decompiler.Tests/RoundtripAssembly.cs` calls `DecompileProject` without asserting the new `Errors` list is empty, so decompiler regressions that previously failed the suite via a thrown `DecompilerException` can now pass green.

Failure scenario: a regression makes `CSharpDecompiler` crash on a void method in e.g. FSharp.Core that the round-trip run's own NUnit tests never execute. Previously `DecompileProject` threw and the `RoundtripAssembly` test failed immediately; now the method body is replaced by `{ ; }` plus comments, the project recompiles, the round-trip tests still pass, and the regression ships undetected by CI. An `Assert.That(decompiler.Errors, Is.Empty)` after the export restores the signal.

ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs
```diff
@@ -336,12 +423,21 @@ void ProcessFiles(List<IGrouping<string, TypeDefinitionHandle>> files)
...
+							RecordError(innerException as DecompilerException ?? new DecompilerException(module, $"Error decompiling for '{file.Key}'", innerException));
+							WriteErrorComment(w, innerException);
 						}
 						finally
 						{
+							foreach (var error in decompiler?.Errors ?? (IReadOnlyList<DecompilerException>)Array.Empty<DecompilerException>())
+							{
+								RecordError(error);
+							}
+							w?.Dispose();
```
christophwille: **[cleanup] Unguarded `w?.Dispose()` in the finally block can re-introduce the whole-export abort.**

`StreamWriter.Dispose` flushes buffered output; on a full disk or dropped network share it throws `IOException` from the finally block, outside the catch that is supposed to confine failures to one file. The exception propagates out of the `Parallel.ForEach` body, stops the remaining files, and `DecompileProject` throws - the whole-project abort that #3510 set out to eliminate, triggered by exactly the disk-full scenario `WriteErrorComment`'s own guard anticipates two hunks earlier. Wrapping the dispose in the same try/catch-and-`RecordError` treatment closes the gap.

ILSpy/Commands/ProjectExporter.cs
```diff
@@ -125,6 +125,32 @@ static string SkipReport(IEnumerable<(LoadedAssembly Assembly, MetadataFile? Fil
...
+				return;
+			output.WriteLine();
+			foreach (string line in CSharpDecompiler.GetErrorSummaryLines(errors.Count))
+			{
+				output.WriteLine(line);
+			}
+			foreach (var error in errors)
+			{
+				output.WriteLine();
+				output.WriteLine($"{error.Message}: {error.InnerException?.Message}");
```
christophwille: **[cleanup] Per-error headline format is duplicated between UI and CLI.**

The `$"{error.Message}: {error.InnerException?.Message}"` format appears both here in `WriteDecompilationErrors` and in `IlspyCmdProgram.ReportDecompilationErrors`, despite the PR introducing `GetErrorSummaryLines` precisely so all front ends render failures identically. The two front ends' per-error lines only agree by copy-paste; the next tweak (adding the member's declaring type, trimming the inner message) lands in one and not the other, and the CLI and UI start describing the same failure differently. A companion helper next to `GetErrorSummaryLines` (e.g. `CSharpDecompiler.GetErrorHeadline(DecompilerException)`) removes the duplication.

PR #3972, opened 2026-08-09
Add support for C# 14 user-defined compound assignment operators

siegfriedpammer:
Implements decompiler support for C# 14 [user-defined compound assignment operators](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-14.0/user-defined-compound-assignment.md) (#829): instance operator declarations decompile to `public void operator +=(T rhs)` / `operator checked +=` / `operator ++()`, and call sites fold back to `x += y;` / `x++;`. The binding rules are taken from the proposal and validated against Roslyn throughout.

The branch is five commits, one per architecture layer.

### Classify (type system)

C# 14 emits an instance compound assignment operator as a void-returning `op_*Assignment` method. `MetadataMethod` classifies such a method as an operator only when it has the shape C# requires — an instance, void-returning method with the right arity (one parameter, none for `++`/`--`) and no `ref`/`params` parameter. F# mangles `static member (+=)` to a static, value-returning `op_AdditionAssignment` and C++/CLI emits value-returning instance operators; both stay plain methods. Explicit interface implementations, whose metadata carries only the dotted name and no `specialname`, are recognized too, so `void ICompound<int>.operator +=(int rhs)` round-trips.

Recognition is gated by a `UserDefinedCompoundAssignmentOperators` decompiler setting and a matching `TypeSystemOptions` flag (the way the extension-method classification already is). Below C# 14 the methods stay plain `op_*Assignment` methods and keep their `specialname` flag as a `[SpecialName]` attribute; `[IsReadOnly]` is now surfaced for operators so a `readonly` struct operator prints with the modifier.

### Keep the receiver a variable (IL)

Once classified, an operator has no callable spelling (CS0571), so every call site must end as `x op= y` / `x++` with `x` an assignable variable whose static type binds the operator the call names. The reader materializes a reference-type receiver into a stack slot typed with the operator's declaring type, and `CallInstruction` vetoes unfit receiver replacements through `SatisfiesSlotRestrictionForInlining`: not writable (`this` in a class, foreach/using/fixed variables, `in` parameters, readonly fields outside their constructor), not a storage location, or of a type that declares a hiding operator of the same name *or its checked/unchecked sibling* — which of the pair `x += y` binds depends on the checked context the statement ends up in, so both count. That last test is a declaration-existence walk on the type system, deliberately not a binding question; the exact, overload-resolution-based form of the same question lives on `CSharpResolver` for the C# layer. Copy propagation asks the same check per receiver load, and `foreach`/`using` hand the operator a writable copy where the loop or using variable would otherwise be read-only.

### Bind (builders and resolver)

`CSharpResolver` gains the two-phase rules next to `GetUserDefinedOperatorCandidates`: `x op= y` and a *prefix* increment consider the instance operators reachable from the static type of `x` first — whether or not the result is used — and the static operators only when none is applicable. The one form that always binds a static operator is a *postfix* increment whose result is used. `CallBuilder` models the first phase when checking that recompiling an instance operator call binds the same method.

The same rule governs the folds built from static operator calls: where a type declares both `static C operator +(C, int)` and `void operator +=(int)`, the fold is written in a static-binding form. Binary operators become `x = x + y` (any position); an increment becomes a postfix increment whose result goes to a discard, `_ = x++;`. That form only exists as a statement, so folds that would embed a shadowed increment in an expression hold back and `FixRemainingIncrements` gives the increment a statement of its own. This closes a silent rebinding: `UserDefinedCompoundAssignmentInheritance` runs each shape before and after decompilation and previously printed `1` before and `100` after.

### Fold calls (C# transforms)

A call to an instance operator becomes `x op= y` / `x++` in `ReplaceMethodCallsWithOperators` only when C# accepts the receiver in that position and binds the same operator there; the operator form cannot pick an overload by parameter modifier, so a call to an `in` overload whose by-value sibling is applicable stays a call, and a non-public operator (CS9308) stays a call too. In the other direction, `PrettifyAssignments` keeps `x = x + y` spelled out wherever folding it would hand the statement to an applicable instance operator.

### Render (UI)

`TypeSystemAstBuilder` writes the `operator +=` declaration form behind a support flag like the other version-gated operator syntax; non-public operators fall back to plain methods (except explicit interface implementations), instance operators can carry `new`, and the `[CompilerFeatureRequired]` marker is removed. Tree and tooltips show `operator +=(int) : void` via a new `ConversionFlags` bit — widening `ConversionFlags.All` also turns on the existing checked-operator and `>>>` flags, which the tooltip ambience now sets explicitly. ilspycmd's `-lv` help lists the C# 14/15 language versions. The checked and `>>>=` names follow `CheckedOperators` / `UnsignedRightShift` like every other operator name, at call sites and declarations alike.

### Not covered

Result-used forms like `d = (c += 5)` decompile as two equivalent statements. Hand-written IL with no C# spelling keeps an explicit call even though it will not recompile — an operator called on an `object`-typed receiver, or a shadowed increment in a method that declares a local named `_` (the discard is the only static-binding spelling): visibly broken output over silently binding the wrong operator.

### Tests

The `UserDefinedCompoundAssignment` pretty fixture covers all 19 operators, the call-site shapes above, inheritance, explicit interface implementation, ref locals and ref returns, a checked operator hidden by an unchecked sibling, and types declaring both the static and the instance operator. `UserDefinedCompoundAssignmentInheritance` executes every combination of where a static `operator +` and an instance `operator +=` are declared across two levels, before and after decompilation — the only test kind that catches a silent rebinding, since the decompiled text looks fine. `CompoundAssignmentOperatorEdgeCases` pins hand-written-IL cases: the F#/C++ method shapes, `base` calls, non-public operators, mismatched receiver types, unconstrained generic receivers, covariant increment returns, and the statement-level shapes that print `_ = x++;`. `NoUserDefinedCompoundAssignmentOperators` pins the output with the setting off, byte-identical to master. Each shadowing guard has a fixture that fails when that guard is disabled. Full ICSharpCode.Decompiler.Tests sweep: 3520 total, 0 failed, 45 skipped.


ICSharpCode.Decompiler/CSharp/Syntax/TypeMembers/OperatorDeclaration.cs
```diff
@@ -136,6 +157,25 @@ static OperatorDeclaration()
 			names[(int)OperatorType.Implicit] = new string[] { "implicit", "op_Implicit" };
 			names[(int)OperatorType.Explicit] = new string[] { "explicit", "op_Explicit" };
 			names[(int)OperatorType.CheckedExplicit] = new string[] { "explicit", "op_CheckedExplicit" };
+			names[(int)OperatorType.AdditionAssignment] = new string[] { "+=", "op_AdditionAssignment" };
```
christophwille, line 163: **[correctness]** The new operator-name table entries feed MetadataMethod's SpecialName branch, but the non-SpecialName fallback branch for explicit interface implementations (MetadataMethod.cs:99) still requires `MethodAttributes.Static`, so an explicit interface implementation of an instance compound-assignment operator is classified as a plain method.

Failure scenario: decompiling a C# 14 class containing `void ICompound<int>.operator +=(int rhs)` (Roslyn emits explicit operator implementations without SpecialName, per the existing comment in MetadataMethod) yields `void ICompound<int>.op_AdditionAssignment(int rhs)` - an explicit implementation of a nonexistent interface method - so the decompiled output fails to compile (CS0539).

Same root cause also affects the name table at line 124.

ICSharpCode.Decompiler/CSharp/CallBuilder.cs
```diff
@@ -1586,7 +1586,7 @@ OverloadResolutionErrors IsUnambiguousCall(ExpectedTargetDetails expectedTargetD
 					}
 				}
 			}
-			else if (method.IsOperator)
+			else if (method.IsOperator && method.IsStatic)
```
christophwille, line 1661: **[correctness]** Instance operator calls always fail `IsUnambiguousCall` (MemberLookup.Lookup's filter at MemberLookup.cs:377 excludes `SymbolKind.Operator`), so the disambiguation loop casts the receiver to the declaring type whenever the receiver's static type differs from it, and ReplaceMethodCallsWithOperators then uses that cast as the assignment target.

Failure scenario: `class C` declares `public virtual void operator +=(int)`; `class D : C`; code `D d; d += n;` compiles to `callvirt C::op_AdditionAssignment` on a D-typed receiver. Decompiling: overload resolution always returns AmbiguousMatch, the repair loop rewrites the target to `((C)d).op_AdditionAssignment(n)`, and the new transform emits `(C)d += n` - a cast expression as assignment target, so the decompiled output does not compile (CS0131).

ICSharpCode.Decompiler/CSharp/Transforms/ReplaceMethodCallsWithOperators.cs
```diff
@@ -192,6 +192,61 @@ void ProcessInvocationExpression(InvocationExpression invocationExpression)
...
+					context.Step("Replace instance operator method with compound assignment", invocationExpression);
+					if (isChecked)
+					{
+						invocationExpression.AddAnnotation(AddCheckedBlocks.CheckedAnnotation);
+					}
+					else if (HasCheckedEquivalent(method))
+					{
+						invocationExpression.AddAnnotation(AddCheckedBlocks.UncheckedAnnotation);
+					}
+					var assignment = new AssignmentExpression(
```
christophwille, line 247: **[correctness]** `assignmentTarget` is never validated as an assignable expression (variable/field/property/indexer), so non-lvalue receivers are rewritten into an invalid assignment LHS.

Failure scenario: C# 14 source `CompoundClass c = GetC(); c += n;` where `c` is single-use: ILInlining inlines the receiver, giving the AST `GetC().op_AdditionAssignment(n)`, which this transform rewrites to `GetC() += n;` - a CS0131 compile error on recompilation (LHS must be a variable, property or indexer). Likewise `new C().op_...(n)` becomes `new C() += n;` and an IL base-call `base.op_AdditionAssignment(y)` becomes the invalid `base += y;`.

ICSharpCode.Decompiler/CSharp/Transforms/ReplaceMethodCallsWithOperators.cs
```diff
@@ -192,6 +192,61 @@ void ProcessInvocationExpression(InvocationExpression invocationExpression)
 				context.EndStep(binaryOperator);
 				return;
 			}
+			if (context.Settings.UserDefinedCompoundAssignmentOperators && method is { IsOperator: true, IsStatic: false }
+				&& invocationExpression.Target is MemberReferenceExpression targetMemberRef)
+			{
+				Expression assignmentTarget = targetMemberRef.Target;
+				if (assignmentTarget is CastExpression castExpr && castExpr.Expression.GetResolveResult().Type is ITypeParameter)
```
christophwille: **[correctness]** The interface cast on a generic receiver is stripped whenever the operand is a type parameter, without verifying the cast type is among the type parameter's constraints (or is the operator's declaring type).

Failure scenario: IL ``box !!T; castclass ICompound`1<int32>; callvirt op_AdditionAssignment`` inside `void M<T>(T x) where T : IOther` decompiles to `((ICompound<int>)x).op_AdditionAssignment(n)`; the transform strips the cast, emitting `x += n;` which does not compile because T has no ICompound constraint. If T happens to be constrained to a different interface that also declares `operator +=`, the recompiled code silently binds to that other operator instead of the one behind the runtime cast - wrong method called at runtime.

ICSharpCode.Decompiler/IL/Transforms/TransformAssignment.cs
```diff
@@ -386,7 +386,7 @@ internal static bool HandleCompoundAssign(ILInstruction compoundStore, Statement
 					binary, target, targetKind, binary.Right,
 					targetType, CompoundEvalMode.EvaluatesToNewValue);
 			}
-			else if (setterValue is Call operatorCall && operatorCall.Method.IsOperator)
+			else if (setterValue is Call operatorCall && operatorCall.Method.IsOperator && operatorCall.Method.IsStatic)
```
christophwille, line 390: **[correctness]** `HandleCompoundAssign` still prints `x op= y` for static operator calls without checking whether the target type also declares a C# 14 instance compound-assignment operator, which C# 14 overload resolution prefers for `x op= y`.

Failure scenario: a type declares both `public static C operator +(C c, int n)` and `public void operator +=(int n)`. IL for `c = c + n` (an explicit static-operator call plus store) is decompiled to `c += n;`. Recompiling that output under C# 14 binds to the instance `operator +=` instead of the static operator-plus-assignment the original assembly executed, so the recompiled program mutates the existing object in place (visible through every alias of `c`) instead of storing a new instance - silently different runtime behavior with no compile error.

Same root cause also affects the increment/decrement path around line 891.

ICSharpCode.Decompiler/CSharp/Syntax/TypeSystemAstBuilder.cs
```diff
@@ -2428,6 +2433,8 @@ EntityDeclaration ConvertOperator(IMethod op)
 				return ConvertMethod(op);
 			if (opType == OperatorType.UnsignedRightShift && !SupportUnsignedRightShift)
 				return ConvertMethod(op);
+			if (!SupportUserDefinedCompoundAssignmentOperators && OperatorDeclaration.IsCompoundAssignment(opType.Value))
```
christophwille, line 2566: **[correctness]** `ConvertOperator` gains the compound-assignment gate but no shape check: because the extended name table makes MetadataMethod classify any SpecialName method named `op_AdditionAssignment` etc. as `SymbolKind.Operator`, a static or value-returning `op_*Assignment` method (emitted by F# `static member (+=)` and C++/CLI `operator+=`) is now rendered as an operator declaration.

Failure scenario: decompiling an existing F# or C++/CLI assembly that defines a static `op_AdditionAssignment` produces `public static T operator +=(T a, int b)` - invalid C# (compound assignment operators must be instance and void-returning), so the output no longer compiles; before this PR the same method decompiled as a normal callable method `public static T op_AdditionAssignment(...)`.

ICSharpCode.Decompiler/CSharp/Transforms/ReplaceMethodCallsWithOperators.cs
```diff
@@ -192,6 +192,61 @@ void ProcessInvocationExpression(InvocationExpression invocationExpression)
 				context.EndStep(binaryOperator);
 				return;
 			}
+			if (context.Settings.UserDefinedCompoundAssignmentOperators && method is { IsOperator: true, IsStatic: false }
```
christophwille: **[correctness]** The compound-assignment rewrite never checks that the instance operator method returns void, though C# 14 only recognizes void-returning instance operators as compound-assignment operators.

Failure scenario: decompiling a C++/CLI or IL-authored assembly with an instance `op_AdditionAssignment(X)` that returns a value: `X r = a.op_AdditionAssignment(b);` is rewritten to `X r = a += b;` (and even a bare statement call becomes `a += b;`). On recompilation the non-void instance operator is not an eligible C# 14 compound operator, so the code either fails to compile or binds to a different (static) operator with different behavior.

ICSharpCode.Decompiler/CSharp/Transforms/ReplaceMethodCallsWithOperators.cs
```diff
@@ -517,6 +572,50 @@ static bool ToStringIsKnownEffectFree(IType type)
...
 		}
 
+		static AssignmentOperatorType? GetCompoundAssignmentOperatorTypeFromMetadataName(string name, out bool isChecked)
+		{
+			isChecked = false;
+			switch (name)
+			{
+				case "op_AdditionAssignment":
+					return AssignmentOperatorType.Add;
+				case "op_CheckedAdditionAssignment":
```
christophwille: **[settings gating]** Checked compound-assignment names are not gated on `settings.CheckedOperators`, unlike the adjacent `GetBinaryOperatorTypeFromMetadataName`/`GetUnaryOperatorTypeFromMetadataName` which use `when settings.CheckedOperators` on every `op_Checked*` case.

Failure scenario: with CheckedOperators=false and UserDefinedCompoundAssignmentOperators=true (independent toggles in the UI), `TypeSystemAstBuilder.ConvertOperator` demotes `op_CheckedAdditionAssignment` to a plain method declaration (the `!SupportOperatorChecked && IsChecked` branch), but `GetCompoundAssignmentOperatorTypeFromMetadataName` still rewrites its call sites into `checked { x += y; }`. The recompiled `x += y` can only bind to the unchecked operator, so the decompiled program silently calls a different method than the original IL (or fails to compile when the IL declares only the checked variant). The same gap exists in the `op_CheckedIncrementAssignment`/`op_CheckedDecrementAssignment` switch at line 227.

ICSharpCode.Decompiler/CSharp/Transforms/ReplaceMethodCallsWithOperators.cs
```diff
@@ -192,6 +192,61 @@ void ProcessInvocationExpression(InvocationExpression invocationExpression)
...
+				}
+				AssignmentOperatorType? aop = GetCompoundAssignmentOperatorTypeFromMetadataName(method.Name, out isChecked);
+				if (aop != null && arguments.Length == 1)
+				{
+					context.Step("Replace instance operator method with compound assignment", invocationExpression);
+					if (isChecked)
+					{
+						invocationExpression.AddAnnotation(AddCheckedBlocks.CheckedAnnotation);
+					}
+					else if (HasCheckedEquivalent(method))
```
christophwille, line 243: **[correctness]** `HasCheckedEquivalent` is name-only (signature-blind), so an operator overload with no checked counterpart for its own signature still gets an UncheckedAnnotation because a checked variant with a different signature exists on the type.

Failure scenario: for a type where `op_CheckedAdditionAssignment(int)` exists but only `op_AdditionAssignment(long)` for long, original source `checked { c += 2L; ... }` decompiles with an unchecked annotation on `c += 2L`, so the output pulls the statement into an `unchecked {}` wrapper / out of the checked block - noisy, misleading decompiled output that misrepresents the original checked context of the statement.

ICSharpCode.Decompiler/CSharp/Transforms/ReplaceMethodCallsWithOperators.cs
```diff
@@ -517,6 +572,50 @@ static bool ToStringIsKnownEffectFree(IType type)
...
+					return AssignmentOperatorType.BitwiseAnd;
+				case "op_BitwiseOrAssignment":
+					return AssignmentOperatorType.BitwiseOr;
+				case "op_ExclusiveOrAssignment":
+					return AssignmentOperatorType.ExclusiveOr;
+				case "op_LeftShiftAssignment":
+					return AssignmentOperatorType.ShiftLeft;
+				case "op_RightShiftAssignment":
+					return AssignmentOperatorType.ShiftRight;
+				case "op_UnsignedRightShiftAssignment":
```
christophwille: **[settings gating]** `op_UnsignedRightShiftAssignment` is not gated on `settings.UnsignedRightShift`, unlike `op_UnsignedRightShift` in `GetBinaryOperatorTypeFromMetadataName` (`when settings.UnsignedRightShift`), and `TypeSystemAstBuilder.ConvertOperator` only gates `OperatorType.UnsignedRightShift`, not `UnsignedRightShiftAssignment`.

Failure scenario: a user who disables the C# 11 'unsigned right shift' setting but leaves compound-assignment operators enabled still gets `x >>>= y` expressions and `public void operator >>>=` declarations in the output, i.e. syntax the setting explicitly turned off; every other settings-gated operator name in these helpers respects its feature toggle.

christophwille (commented):
Automated high-effort review (multi-agent, each finding independently verified against the PR head). Overall: the PR handles the straightforward Roslyn lvalue cases well, but the new pattern matchers are loose at several boundaries. The most severe defects break recompilation of plain C# 14 code (explicit interface operator implementations, virtual operators called through a derived-typed receiver), and the static-operator compound-assign path can now silently change runtime semantics when a type declares both static and instance operators. Secondary issues are settings-gating gaps (`CheckedOperators`, `UnsignedRightShift`) and the signature-blind checked-equivalent probe inherited by the new paths.

10 findings posted as inline comments, ordered by severity there. Summary:

1. **Explicit interface implementations of instance compound operators decompile as plain methods** (OperatorDeclaration.cs) - output fails to compile (CS0539).
2. **Derived-typed receivers get a cast that becomes the assignment target** (CallBuilder.cs) - emits `(C)d += n` (CS0131).
3. **Non-lvalue receivers become an invalid assignment LHS** (ReplaceMethodCallsWithOperators.cs) - `GetC() += n;` (CS0131).
4. **Type-parameter cast stripped without checking constraints** (ReplaceMethodCallsWithOperators.cs) - non-compiling or wrong-binding output.
5. **`x op= y` printed for static operator calls even when an instance compound operator exists** (TransformAssignment.cs) - recompiled code binds to the instance operator, silently different runtime behavior.
6. **Static / value-returning `op_*Assignment` (F#, C++/CLI) now rendered as operator declarations** (TypeSystemAstBuilder.cs) - invalid C#.
7. **No void-return check on the instance-operator rewrite** (ReplaceMethodCallsWithOperators.cs).
8. **Checked compound-assignment names not gated on `settings.CheckedOperators`** (ReplaceMethodCallsWithOperators.cs).
9. **`HasCheckedEquivalent` is signature-blind** (ReplaceMethodCallsWithOperators.cs) - spurious `unchecked` wrappers.
10. **`op_UnsignedRightShiftAssignment` not gated on `settings.UnsignedRightShift`** (ReplaceMethodCallsWithOperators.cs).

ICSharpCode.Decompiler.Tests/TestCases/ILPretty/CompoundAssignmentOperatorEdgeCases.cs
```diff
@@ -0,0 +1,66 @@
...
+public class CompoundTarget
+{
+	public int Value;
+
+	public void operator +=(int rhs)
+	{
+		Value += rhs;
+	}
+
+	[SpecialName]
```
dgrunwald, line 49: This could use a comment explaining why we expect this to not turn into `operator +=` (because C# requires that syntax to use void return type)

ICSharpCode.Decompiler.Tests/TestCases/Pretty/UserDefinedCompoundAssignment.cs
```diff
@@ -0,0 +1,293 @@
...
+			x++;
+		}
+
+		public static void UseInheritedOperator(DerivedCompoundClass d, int n)
+		{
+			d += n;
+			d++;
+		}
+
+		public static BothOperators UseStaticOperator(BothOperators b, int n)
```
dgrunwald, line 436: I'd like to see another test like UseStaticOperator with an array element as target instead of the local variable (since local variable compound assignments are somewhat of a special case in the decompiler)

ICSharpCode.Decompiler.Tests/TestCases/ILPretty/CompoundAssignmentOperatorEdgeCases.cs
siegfriedpammer: Added that comment above the method in 597d50598 (comment-only lines are ignored by the expected-output comparison, so it lives in the .cs next to the declaration).

_Reply written by an AI agent (Claude) on Siegfried's behalf._

ICSharpCode.Decompiler.Tests/TestCases/Pretty/UserDefinedCompoundAssignment.cs
siegfriedpammer: Added `UseStaticOperatorOnArrayElement` (and a field variant) in 597d50598: `arr[0] = arr[0] + n` stays spelled out, so the array path is covered next to the local one.

_Reply written by an AI agent (Claude) on Siegfried's behalf._

ICSharpCode.Decompiler/CSharp/Transforms/ReplaceMethodCallsWithOperators.cs
siegfriedpammer: Left as is on purpose: \`HasCheckedEquivalent\` has always been name-only, and the same signature-blindness applies to every static operator on master - the compound assignment names just reuse it. Fixing it means matching signatures for all operators, which is a separate change.

_Reply written by an AI agent (Claude) on Siegfried's behalf._

ICSharpCode.Decompiler/CSharp/CallBuilder.cs
```diff
@@ -1622,6 +1622,15 @@ OverloadResolutionErrors IsUnambiguousCall(ExpectedTargetDetails expectedTargetD
 					or.AddCandidate(m);
 				}
 			}
+			else if (method.IsOperator && !method.IsStatic && target != null)
```
dgrunwald: Using separate candidate sets for static vs. non-static operators seems questionable -- the set of candidates must be all potential candidates that the C# compiler will consider when recompiling the code we're generating; so it shouldn't depend on details of the method called in IL that don't impact the syntactic form we'll be using.

ICSharpCode.Decompiler/CSharp/Resolver/CSharpResolver.cs
```diff
@@ -1289,7 +1289,8 @@ public IEnumerable<IParameterizedMember> GetUserDefinedOperatorCandidates(IType
 				return EmptyList<IMethod>.Instance;
 			}
 			// C# spec (draft-v11): §12.4.6 Candidate user-defined operators
-			var operators = type.GetMethods(m => m.IsOperator && m.Name == operatorName).ToList();
+			// C# 14 instance compound-assignment operators are not candidates for the static operator forms.
+			var operators = type.GetMethods(m => m.IsOperator && m.IsStatic && m.Name == operatorName).ToList();
```
dgrunwald: This function is used in the ambiguity check, e.g. for increment operator. I don't think it's correct to just skip non-static operators there?

ICSharpCode.Decompiler/CSharp/Transforms/ReplaceMethodCallsWithOperators.cs
```diff
@@ -192,6 +192,68 @@ void ProcessInvocationExpression(InvocationExpression invocationExpression)
...
+			{
+				Expression assignmentTarget = targetMemberRef.Target;
+				if (assignmentTarget is CastExpression castExpr
+					&& castExpr.Expression.GetResolveResult().Type is ITypeParameter typeParameter
+					&& typeParameter.DirectBaseTypes.Any(t => t.Equals(castExpr.Type.GetResolveResult().Type)))
+				{
+					// Calls through a generic type parameter constrained to an interface declaring the
+					// operator are built as "((I)x).op_...(y)"; the cast is not a valid assignment
+					// target, and "x op= y" binds to the same operator via the constraint.
+					// Without that constraint the cast selects the operator and cannot be dropped.
```
dgrunwald: This seems like a weird special case -- I think we should instead investigate why CallBuilder is emitting an unnecessary cast.

ICSharpCode.Decompiler/IL/Transforms/ILInlining.cs
```diff
@@ -264,6 +264,13 @@ static bool DoInline(ILVariable v, ILInstruction inlinedExpression, ILInstructio
 				{
 					return false;
 				}
+				if (context.Settings.UserDefinedCompoundAssignmentOperators && IsCompoundAssignmentOperatorReceiver(loadInst))
+				{
+					// The receiver of a C# 14 instance compound assignment operator call becomes the
+					// target of "x op= y", which must remain a variable: inlining the expression that
+					// initializes it would produce "GetX() += y".
```
dgrunwald: We could still inline if `inlinedExpression` is classified as a variable (use ClassifyExpression)

ICSharpCode.Decompiler.Tests/TestCases/ILPretty/CompoundAssignmentOperatorEdgeCases.cs
```diff
@@ -0,0 +1,68 @@
...
+public class EdgeCases
+{
+	public static CompoundTarget GetTarget()
+	{
+		return new CompoundTarget();
+	}
+
+	public static void NonVariableReceiver(int n)
+	{
+		GetTarget().op_AdditionAssignment(n);
```
dgrunwald: It would be better if we had avoided inlining.
There's already a change in inlining that should do so -> investigate why that doesn't work.

ICSharpCode.Decompiler.Tests/TestCases/Pretty/UserDefinedCompoundAssignment.cs
```diff
@@ -0,0 +1,308 @@
...
+			// "b += n" and "b++" would bind to the instance operators instead.
+			b = b + n;
+			b = b + 1;
+			return b;
+		}
+
+		public static void UseStaticOperatorOnArrayElement(BothOperators[] arr, int n)
+		{
+			// Same shape as UseStaticOperator, but the target is an array element rather than a
+			// local: compound assignments to locals take a separate path through the decompiler.
```
dgrunwald: This doesn't look like a compound assignment to the decompiler -- for a real compound assignment on a class without operator+=, csc would emit only a single ldelema.

ICSharpCode.Decompiler/CSharp/Syntax/TypeMembers/OperatorDeclaration.cs
```diff
@@ -201,10 +241,25 @@ public static bool IsChecked(OperatorType type)
...
 			};
 		}
 
+		/// <summary>
+		/// Gets whether the operator type is a C# 14 user-defined compound assignment operator
+		/// (a void-returning instance operator, including the increment/decrement forms).
+		/// </summary>
+		public static bool IsCompoundAssignment(OperatorType type)
+		{
+			return type >= OperatorType.AdditionAssignment;
```
dgrunwald, line 263: If we have code sensitive to the order of enumerators like this, that deserves a comment in the enum definition.

dgrunwald (changes requested):

ICSharpCode.Decompiler/CSharp/Syntax/TypeMembers/OperatorDeclaration.cs
siegfriedpammer: Added to the enum itself in 9cbc574ad: the compound assignment members now carry a note that they
have to stay last as a block, since `IsCompoundAssignment` tells them apart by comparing against
the first of them.

_Reply written by an AI agent (Claude) on Siegfried's behalf._

ICSharpCode.Decompiler.Tests/TestCases/Pretty/UserDefinedCompoundAssignment.cs
siegfriedpammer: Right - written out in full it is a plain store and never reaches the compound assignment path.
9cbc574ad adds `UseCompoundAssignmentOnArrayElement` and `UseCompoundAssignmentOnStructArrayElement`
on element types that declare no instance operator, so `arr[0] += n` compiles to the shapes the
decompiler has to recognize: the array spilled to a local for a reference type, a single `ldelema`
for a value type. The original case is kept next to them, where the element type does declare an
instance operator and the static call has to stay spelled out.

_Reply written by an AI agent (Claude) on Siegfried's behalf._

ICSharpCode.Decompiler/CSharp/Transforms/ReplaceMethodCallsWithOperators.cs
siegfriedpammer: Investigated - the cast came from the receiver having lost the type the call resolved against: an
optimized build has no local at all, and in a debug build copy propagation replaced the local with
a value of a more derived type. CallBuilder then inserted the cast to repair the call, and a cast
is not a valid assignment target.

9cbc574ad materializes the receiver into a variable typed as the type the call resolved against and
keeps copy propagation and the C# layer's stack-slot type refinement from taking that type away
again. No cast appears any more, so the special case is gone.

_Reply written by an AI agent (Claude) on Siegfried's behalf._

ICSharpCode.Decompiler/IL/Transforms/ILInlining.cs
siegfriedpammer: Done in 9cbc574ad - the receiver is inlined whenever the expression taking its place denotes a
storage location (`ldloc`/`ldobj`/`ldflda`/`ldsflda`) and binds the same operator. A temporary is
introduced only where a shadowing operator would otherwise be picked up.

_Reply written by an AI agent (Claude) on Siegfried's behalf._

ICSharpCode.Decompiler.Tests/TestCases/ILPretty/CompoundAssignmentOperatorEdgeCases.cs
siegfriedpammer: Found it: inlining refused to touch the receiver at all, because the assignment target has to stay a
variable and no cheaper test was available. 9cbc574ad replaces that with a check that the
replacement denotes a storage location and binds the same operator, so this case is no longer
inlined away - it decompiles to `CompoundTarget target = GetTarget(); target += n;`.

_Reply written by an AI agent (Claude) on Siegfried's behalf._

ICSharpCode.Decompiler/CSharp/CallBuilder.cs
siegfriedpammer: Checked this against Roslyn rather than the proposal text. `x op= y` is resolved in two phases: the
instance compound assignment operators reachable from the static type of x first, the static
operators only if none of them is applicable. Measured, one method per case, reading back which one
the IL calls:

* instance `+=(long)` beats static `+(T, int)` for an `int` argument - so it is not one merged set
  ranked by conversion quality;
* instance `+=(int)` with a `long` argument is not applicable and the static operator binds - phase
  one is decided by applicability, not by the name being present;
* an instance operator inherited from a base beats a static operator declared on the derived class;
* for a property or indexer target phase one does not happen at all, since an instance operator
  needs a variable to mutate.

So the candidate set does follow the syntactic form we emit - but you are right that it must not be
picked from what the IL happened to call. a68dd1cc7 makes both directions consult the same set:
the fold that turns `x = x op y` into `x op= y` now runs overload resolution over the instance
candidates of the *target's* type, instead of matching names against the static operator's
declaring type.

That was hiding a miscompile. With a static operator on the base returning the derived type and the
instance operator on the derived class, `arr[0] = arr[0] + 1` folded to `arr[0] += 1` and silently
rebound to the instance operator - running the program before and after decompilation printed `1`
and `100`. `UserDefinedCompoundAssignmentInheritance` covers it now; the same change also stops the
fold being refused for arguments the instance operator cannot take, and for property and indexer
targets.

_Reply written by an AI agent (Claude) on Siegfried's behalf._

ICSharpCode.Decompiler/CSharp/Resolver/CSharpResolver.cs
siegfriedpammer: Went through the callers. `GetUserDefinedOperatorCandidates` feeds the static-operator branch of
`IsUnambiguousCall` and `ResolveUnaryOperator`/`ResolveBinaryOperator`, which model `x op y`, `-x`
and casts - forms C# resolves against static operators only. Verified: `f + 1` binds the static
operator even where the type declares an instance `operator +=`. `ResolveUnaryOperator` is only
ever reached with `BitNot` in the decompiler, never with `Increment`.

The increment case is real, but it sits at the sites that emit the form rather than here: `x++` in
statement position does bind an instance `operator ++()`. a68dd1cc7 checks that in the folds in
`TransformAssignment` and `FixRemainingIncrements`, and `ReplaceMethodCallsWithOperators` already
refuses to turn a leftover `op_Increment(a)` call into `++a`. A shadowed static `op_Increment` call
therefore stays an explicit call rather than binding the wrong operator - `ShadowedIncrement` in
`CompoundAssignmentOperatorEdgeCases` pins that.

Left the function as it is on that basis; happy to change it if you would rather the candidate set
took the emitted form into account here too.

_Reply written by an AI agent (Claude) on Siegfried's behalf._

ICSharpCode.Decompiler/IL/Transforms/TransformAssignment.cs
```diff
@@ -386,30 +387,40 @@ internal static bool HandleCompoundAssign(ILInstruction compoundStore, Statement
...
 				if (!IsMatchingCompoundLoad(operatorCall.Arguments[0], compoundStore, out var target, out var targetKind, out var finalizeMatch, forbiddenVariable: storeInSetter?.Variable))
 					return false;
 				ILInstruction rhs;
+				IType valueType;
 				if (operatorCall.Arguments.Count == 2)
 				{
 					if (CSharp.ExpressionBuilder.GetAssignmentOperatorTypeFromMetadataName(operatorCall.Method.Name, context.Settings) == null)
 						return false;
 					rhs = operatorCall.Arguments[1];
+					valueType = operatorCall.GetParameter(1).Type;
```
christophwille, line 403: `operatorCall.GetParameter(1).Type` is a `ByReferenceType` when the static operator takes `in Foo` / `ref readonly Foo`. `IsShadowedByInstanceOperator` wraps that in a plain `ResolveResult`, and `OverloadResolution.CheckApplicability` (OverloadResolution.cs:702-728) only strips the *parameter's* by-ref and then asks for `ImplicitConversion(ByReferenceType(Foo) -> Foo)`, which is `None` - so no candidate is ever applicable and the shadow check silently returns false.

Repro: a type with `static Foo operator +(Foo a, in Foo b)` and `public void operator +=(in Foo b)` (or `+=(Foo)`); source `x = x + y` compiles to `stloc x(call op_Addition(ldloc x, ldloca y))`, this transform emits `x += y`, and C# 14 binds that to the instance operator (instance phase first) - recompiled code calls a different method.

`PrettifyAssignments` uses `binary.Right.GetResolveResult().Type` (by-value) and is fine. Fix here: `GetParameter(1).Type.UnwrapByRef()` (TypeSystemExtensions.cs:459, as CallBuilder.cs:1752 already does). None of the new fixtures has an `in`-parameter operator; worth adding one next to `BothOperators`.

ICSharpCode.Decompiler/CSharp/Resolver/CSharpResolver.cs
```diff
@@ -1245,6 +1245,191 @@ public IEnumerable<IParameterizedMember> GetUserDefinedOperatorCandidates(IType
...
+		///
+		/// A type that carries no operator of this name at all is not a rebinding: several ILAst
+		/// nodes are typed by their stack type rather than by the C# type they will be written as,
+		/// and the C# layer resolves those itself.
+		/// </remarks>
+		public static bool WouldRebindOperator(IMethod called, IType receiverType, ICompilation compilation)
+		{
+			ResolveResult[] arguments = called.Parameters.Count == 0
+				? []
+				: [new ResolveResult(called.Parameters[0].Type)];
```
christophwille, line 1369: Same by-ref issue as in `HandleCompoundAssign`: for `operator +=(in T)` `called.Parameters[0].Type` is a `ByReferenceType`, so no candidate is applicable, `PruneCandidatesHiddenByDerivedApplicable` prunes nothing, and rebinding through a `new` operator on a derived receiver type is never detected from `ILInlining.CanReplaceCompoundAssignmentReceiver` / `CopyPropagation`.

`class Base { public void operator +=(in Foo f) }`, `class Derived : Base { public new void operator +=(in Foo f) }`, source `Base b = derived; b += f;` -> inlining substitutes `ldloc derived` for the receiver slot, `ReplaceMethodCallsWithOperators` then re-checks with the real argument resolve result (231-237), sees the `new` operator and keeps the call -> `derived.op_AdditionAssignment(in f)` (not valid C#) instead of the foldable `b += f`.

`called.Parameters[0].Type.UnwrapByRef()` (a by-value `ResolveResult` is the right model; a `ByReferenceResolveResult(In)` would wrongly skip by-value siblings, see OverloadResolution.cs:686-689). Only the `ReplaceMethodCallsWithOperators` overload is exercised by the `CallInOverload` ILPretty case.

ICSharpCode.Decompiler/IL/ILReader.cs
```diff
@@ -1854,6 +1863,12 @@ ILInstruction[] PrepareArguments(bool firstArgumentIsStObjTarget)
 						{
 							firstArgumentInstruction = new LdObjIfRef(firstArgumentInstruction, typeOfThis);
 						}
+						else if (materializeReceiver)
```
christophwille, line 1866: Evaluation-order bug for the zero-parameter operators (`op_IncrementAssignment` / `op_DecrementAssignment` and checked forms): `AllocateStackSlot` appends `stloc S(receiver)` to the current block, but the `FlushExpressionStack()` at 1844-1847 is inside the per-parameter loop, which has zero iterations here - so the receiver read is hoisted above pending side effects on the expression stack.

Reproduced with this branch's `ilspycmd` on hand-assembled IL `ldarg.0; call object Test::A(); ldarg.0; ldfld Counter Test::x; callvirt void Counter::op_IncrementAssignment(); ldarg.0; ldfld x; call Foo(object, Counter)` where `A()` reassigns `this.x`: output is `Counter counter = x; object o = A(); counter++; Foo(o, x);` - increments the *old* `x`, the original increments the new one. The same IL with a 1-arg `op_AdditionAssignment` decompiles correctly (`object o = A(); x += 1; Foo(o, x);`).

Roslyn emits `dup` for `Foo(A(), ++x)` so C#-compiled input happens to be unaffected, but any other compiler/weaver/hand IL is not. Fix: `if (materializeReceiver) FlushExpressionStack();` before the loop, independent of `Parameters.Count`.

ICSharpCode.Decompiler/IL/ILReader.cs
```diff
@@ -1831,9 +1836,13 @@ ILInstruction[] PrepareArguments(bool firstArgumentIsStObjTarget)
 					&& !firstArgumentIsStObjTarget
 					&& UseRefLocalsForAccurateOrderOfEvaluation
 					&& expectedStackType == StackType.Ref && typeOfThis.IsReferenceType != false;
+				// Only an object reference is worth materializing: a value-type receiver is passed by
+				// address, so it already denotes a variable, and copying it into another one would
+				// make the operator mutate the copy.
+				bool materializeReceiver = IsNonStaticOperatorCall() && expectedStackType == StackType.O;
```
christophwille, line 1842: `materializeReceiver` keys on `IsNonStaticOperatorCall()` (`IsOperator && !IsStatic`) and is not gated by any setting; `UserDefinedCompoundAssign.IsCompoundAssignmentReceiverUse` (CompoundAssignmentInstruction.cs:357) uses the same predicate. Neither checks the compound-assignment name or the `void` return, so every value-returning instance operator - C++/CLI `R^ operator+(R^)`, still `SymbolKind.Operator` via MetadataMethod.cs:84-90 exactly as before this PR - gets its receiver forced into a stack slot that `ILInlining.CanReplaceCompoundAssignmentReceiver` (351-361) then refuses to inline unless it is `LdLoc`/`LdObj`/`LdFlda`/`LdsFlda`.

Net effect on a C++/CLI assembly: `GetR().op_Addition(x)` / `r.Prop.op_Addition(x)` / `new R().op_Addition(x)` regress to `R r = GetR(); r.op_Addition(x);`, and foreach/using receivers get forced local copies (StatementBuilder 1112-1118, PatternStatementTransform 326-332/674-676, UsingTransform 193-205, CopyPropagation 163-176) - although `ReplaceMethodCallsWithOperators` will never emit an `op=` form for a non-`*Assignment` name. The `InstanceOperatorCall` ILPretty fixture only has `ldarg.0` receivers, so it cannot catch this.

Suggest restricting both predicates to the C# 14 shape (`OperatorDeclaration.IsCompoundAssignment(GetOperatorType(name))` + void return) and gating the reader on the setting like every other site.

ICSharpCode.Decompiler/IL/Transforms/FixRemainingIncrements.cs
```diff
@@ -20,12 +20,25 @@
...
 	public class FixRemainingIncrements : IILTransform
 	{
+		/// <summary>
+		/// Gets the type "++x" binds its operator against once the call is rewritten: the first
+		/// rewrite below increments the variable the result is stored in, whose type can be more
+		/// derived than the operator's parameter when the operator has a covariant return type.
+		/// </summary>
+		static IType GetIncrementTargetType(Call call)
+		{
+			if (call.SlotInfo == StLoc.ValueSlot && call.Parent!.SlotInfo == Block.InstructionSlot)
```
christophwille, line 37: `GetIncrementTargetType` returns `((StLoc)call.Parent).Variable.Type`; for an ILReader `StackSlot` that is still the `System.Object` placeholder, so `IsShadowedByInstanceOperator` finds no candidates, the call is rewritten to `stloc S(ldloc x); ++S` (86-89), and `ExpressionBuilder.VisitStLoc` (872-880, `HasDefaultStackSlotType` true for object) later retypes `S` to `x`'s real type.

Scenario: `Both` declares `static Both operator ++(Both)` and `void operator ++()`. `Foo(++x)` reads as `stloc S(call op_Increment(ldloc x)); stloc x(ldloc S); call Foo(ldloc S)` (S has two loads, not inlined). With `MakeAssignmentExpressions=false` (TransformAssignment.cs:46-54 skips both inline-assignment transforms) or when `TransformInlineAssignmentStObjOrCall` bails (impure/used-within target, parameterized setter, 139-199), the object-typed block-level stloc reaches this transform, passes the shadow check against `object`, and the output `Both s = x; ++s; x = s; Foo(s);` binds the instance `operator ++()` under C# 14 instead of the static `op_Increment` the IL called.

For a `StackSlot` store variable use `call.Arguments[0].InferType(...)` / chase the alias like `ILInlining.GetReceiverType` (ILInlining.cs:375-385), or fall back to `call.GetParameter(0).Type`.

christophwille (commented):
## Review: C# 14 user-defined compound assignment operators

Overall the feature is solid: the two-phase binding model, the instance/static shadowing guard and the receiver-lvalue protection are well thought out and well covered by fixtures. The problems below all sit at the seams of that guard. I went through the three commits (recognize / decompile / render, 40 files) with several independent passes; one candidate (the `HasDefaultStackSlotType` change) was refuted empirically (no output diff master vs PR) and is not listed.

### Correctness (inline comments carry the details)

1. **`in`-parameter operators defeat the shadow/rebind check** - `HandleCompoundAssign` (TransformAssignment.cs:403) and `WouldRebindOperator(IMethod, IType, ICompilation)` (CSharpResolver.cs:1369) feed a `ByReferenceType` into overload resolution, which is applicable to nothing, so the check is always "not shadowed". `x = x + y` with `static operator +(Foo, in Foo)` next to `operator +=(in Foo)` folds to `x += y`, which C# 14 binds to the *instance* operator. `UnwrapByRef()` (as CallBuilder.cs:1752 already does) fixes both. No `in`-parameter operator exists in the fixtures.
2. **Statement-level `x++` / `++s` / foreach-local receiver escape the guard** on three paths: the dead-store branch of `TransformPostIncDecOperator` (and `...WithInlineStore`), `FixRemainingIncrements` when the store variable is still an object-typed stack slot, and `CanBeDeconstructedInForeach` (the deconstruction branch runs before the stloc branch that has the guard).
3. **Receiver materialization for 0-parameter operators** (`op_IncrementAssignment` etc.) stores the receiver slot *before* `FlushExpressionStack()` runs, because the flush sits inside the per-parameter loop. Reproduced with hand-assembled IL: `Foo(A(), ++x)` where `A()` reassigns `x` decompiles to increment the *old* `x`. Roslyn happens to emit `dup` for this shape, so C#-compiled input is unaffected; other compilers/weavers are not.
4. **Receiver machinery keys on `IsOperator && !IsStatic`, not on the compound-assignment shape** (ILReader.cs:1842, `UserDefinedCompoundAssign.IsCompoundAssignmentReceiverUse`), so C++/CLI-style value-returning instance operators get forced into stack slots that inlining then refuses to fold (`R r = GetR(); r.op_Addition(x);`), and with the setting *off* such calls no longer reach any operator branch in `CallBuilder` (-> `AmbiguousMatch` -> casts).

### Cleanup

- `IsShadowedByInstanceOperator` is not gated by the setting at its four callers and walks the type hierarchy twice per call (plain + checked name) plus an O(n^2) dedup; every `x = x + y` / `++x` on `decimal`/`DateTime`/`BigInteger`/... pays for it even when the feature is off.
- `CopyPropagation.CannotReplaceCompoundAssignmentReceiver` re-inlines the match that `UserDefinedCompoundAssign.IsCompoundAssignmentReceiverUse` (added by this PR, used by ILInlining/UsingTransform) already expresses.
- `MetadataMethod.IsUserDefinedCompoundAssignmentOperator` carries two stacked `<summary>` blocks; the first belongs on `IsCompoundAssignmentOperatorSignature`, which has none.
- `OperatorDeclaration.IsCompoundAssignment` relies on enum order (`type >= AdditionAssignment`).
- `CSharpResolver.PruneCandidatesHiddenByDerivedApplicable` / `IsApplicable` duplicate `OverloadResolution.AddMethodLists`; `ReplaceMethodCallsWithOperators.IsValidAssignmentTarget` / `IsAssignableTarget` overlap each other and restate `ILInlining.IsReadonlyCompoundAssignmentTarget`.
- `ConversionFlags.All = 0xffffff` now also switches on `UsePrivateProtectedAccessibility` / `SupportExtensionDeclarations` for tooltips and compare - probably desirable, but worth calling out in the PR description.
- `CSharpAmbience` prints `operator +=` for a non-public compound operator while `TypeSystemAstBuilder` writes it as a method.
- `CorrectnessTestRunner.roslyn5OrNewerOptions` omits `executesCompiledOutput: true`.
- CallBuilder.cs:1703: "modelled" -> "modeled" (en-US rule in CLAUDE.md).

ICSharpCode.Decompiler/IL/Transforms/TransformAssignment.cs
```diff
@@ -1088,6 +1108,9 @@ bool TransformPostIncDecOperator(Block block, int i)
 			{
 				if (!operatorCall.Arguments[0].MatchLdLoc(tmpVar))
 					return false;
+				// No shadowing check here: "x++" whose result is used hands back the value from
+				// before the increment, which an instance operator mutating in place cannot
+				// produce, so C# does not consider one for that shape.
```
christophwille, line 1113: The reasoning in this comment holds while the result is used, but the dead-store branch just below (`if (inst.Variable.IsSingleDefinition && inst.Variable.LoadCount == 0) inst.ReplaceWith(inst.Value)`) leaves a block-level `UserDefinedCompoundAssign(EvaluatesToOldValue)` that `ExpressionBuilder` prints as the statement `x++;` - and per the C# 14 proposal a statement-level `x++` on a variable considers instance increment operators *first*. Same asymmetry in `TransformPostIncDecOperatorWithInlineStore` (~995).

Repro with the fixture's `BothOperators`: consumer IL `stloc tmp(ldsfld X); stsfld X(call op_Increment(ldloc tmp))` with `tmp` otherwise unused - exactly the legacy-csc / "Roslyn for most post-increments" shape the method's own doc comment names - becomes the statement `X++;`, which recompiles to the instance operator (mutate in place) instead of the static one the IL called. No downstream pass re-checks (`FixRemainingIncrements` only sees remaining `Call`s, `PrettifyAssignments` only `AssignmentExpression`s).

Apply `IsShadowedByInstanceOperator` in the dead-store branch (and after `WithInlineStore` when the stloc variable is dead) and keep the call / temp when shadowed.

ICSharpCode.Decompiler/CSharp/StatementBuilder.cs
```diff
@@ -1085,6 +1109,13 @@ RequiredGetCurrentTransformation DetectGetCurrentTransformation(BlockContainer u
 				if (stloc.Parent == loopBody && VariableIsOnlyUsedInBlock(stloc, usingContainer, loopContainer))
 				{
 					foreachVariable = stloc.Variable;
+					if (settings.UserDefinedCompoundAssignmentOperators
```
christophwille, line 1112: This guard only covers the stloc branch, but `DetectGetCurrentTransformation` checks the `Deconstruction` branch (1095-1099) *before* it, and `CanBeDeconstructedInForeach` (1132-1167) accepts a Local/StackSlot with a single store plus a `ldloc`/`ldloca` receiver use; `TranslateDeconstructionDesignation` (935) then makes it a read-only `ForeachLocal`.

`foreach (var item in list) { var (a, b) = item; a += 1; }` over `List<(Counter, int)>` where `Counter` has an instance `operator +=`: the reader's receiver slot is inlined back to `ldloc a` while `a` is still `VariableKind.Local` (ILInlining.cs:268-270 accepts it), `DeconstructionTransform` matches the tuple-field-copy shape (DeconstructionTests.cs:983 `foreach (var (text, num) in tuples)` shows this path fires), `a` becomes a `ForeachLocal`, and `ReplaceMethodCallsWithOperators.IsAssignableTarget` (680-702) rejects it -> `foreach (var (a, b) in list) { a.op_AdditionAssignment(1); }`, CS0571 on recompile (and `a += 1` would be CS1656).

Mirror this guard inside `CanBeDeconstructedInForeach`'s assignment loop.

ICSharpCode.Decompiler/CSharp/CallBuilder.cs
```diff
@@ -1694,6 +1694,23 @@ OverloadResolutionErrors IsUnambiguousCall(ExpectedTargetDetails expectedTargetD
 					or.AddCandidate(m);
 				}
 			}
+			else if (settings.UserDefinedCompoundAssignmentOperators && method.IsOperator && !method.IsStatic && target != null)
```
christophwille, line 1697: With `UserDefinedCompoundAssignmentOperators` *off*, a call to an instance method that is still classified `Operator` (C++/CLI instance `op_Addition`) no longer reaches any operator branch - 1661 now requires `IsStatic`, this branch requires the setting - and falls through to `lookup.Lookup` (1722-1727), which `MemberLookup.cs:376` filters to non-operators -> `UnknownMemberResolveResult` -> `AmbiguousMatch` -> the retry loop (1302-1392) adds casts.

Before this PR `else if (method.IsOperator)` ran `GetUserDefinedOperatorCandidates(argType, name)` (CSharpResolver.cs:1231-1245; its `m.IsOperator && m.Name == operatorName` filter does not exclude instance methods), so for `R op_Addition(R)` the candidate was applicable, `foundMember.Equals(method)` held and the call rendered as `r.op_Addition(x)`. At language version <= C# 13 (or the `NoUserDefinedCompoundAssignmentOperators` Ugly configuration) the same call now renders as `((R)r).op_Addition((R)x)` or gives up.

Either drop the settings gate here (the type system already decides classification) or keep instance operators in the 1661 branch when the setting is off.

ICSharpCode.Decompiler/CSharp/CallBuilder.cs
```diff
@@ -1694,6 +1694,23 @@ OverloadResolutionErrors IsUnambiguousCall(ExpectedTargetDetails expectedTargetD
...
+			{
+				// An instance operator has no callable spelling of its own: it is only ever emitted
+				// as "x op= y" or "x++". C# resolves that form in two phases, the instance operators
+				// on the static type of x first and the static operators only if none of them is
+				// applicable, so these candidates are the whole of what recompilation considers here.
+				// The fallback phase is not modelled: a call this set cannot account for gives up the
```
christophwille, line 1703: nit: "modelled" -> "modeled" (en-US per CLAUDE.md).

ICSharpCode.Decompiler/IL/Transforms/TransformAssignment.cs
```diff
@@ -386,30 +387,40 @@ internal static bool HandleCompoundAssign(ILInstruction compoundStore, Statement
...
 					rhs = new LdcI4(1);
+					valueType = null;
 				}
 				else
 				{
 					return false;
 				}
+				// IsCompoundStore accepts a store to a local (StLoc) or to a field, array element,
+				// ref or pointer (StObj), which are variables, and a setter call, which is not.
+				if (CSharpResolver.IsShadowedByInstanceOperator(operatorCall.Method, targetType, valueType,
```
christophwille, line 419: Cost/gating: `IsShadowedByInstanceOperator` is called from four sites (here, 906, FixRemainingIncrements.cs:51, PrettifyAssignments.cs:112) without a `Settings.UserDefinedCompoundAssignmentOperators` gate, and each call walks the full type hierarchy twice (plain name + checked name via two `GetInstanceOperatorCandidates` calls, CSharpResolver.cs:1304-1306) plus an O(n^2) `GetBaseMembers` dedup (1266-1270).

So every `x = x + y` / `++x` on any user-defined type (`decimal`, `DateTime`, `TimeSpan`, `BigInteger`, ...) reaching these transforms now pays two `BaseTypeCollector` walks over every method of every non-interface base type - even when the setting is off, in which case `DecompilerTypeSystem.GetOptions` never classifies `op_*Assignment` as `Operator` and a candidate can never be found. ILInlining / CopyPropagation / StatementBuilder / UsingTransform all gate on the setting; these four do not.

Cheaper: gate on the setting (parameter or call site), collect both names in one `GetMethods` pass (`m.Name == name || m.Name == checkedName`), and `if (candidates.Count <= 1) return candidates;` before the dedup.

ICSharpCode.Decompiler/IL/Transforms/CopyPropagation.cs
```diff
@@ -143,6 +147,34 @@ static bool CanPerformCopyPropagation(ILVariable target, ILInstruction value, Ha
...
+		/// ("this" in a class, a foreach or using variable, an "in" parameter) cannot replace a
+		/// receiver at all: it is not a valid target of "x op= y".
+		/// </remarks>
+		static bool CannotReplaceCompoundAssignmentReceiver(ILVariable target, ILInstruction value, ILTransformContext context)
+		{
+			if (value is not LdLoc)
+				return false;
+			foreach (var load in target.LoadInstructions)
+			{
+				if (load.Parent is not CallInstruction { Method: { IsOperator: true, IsStatic: false } } call
```
christophwille, line 169: This re-inlines the `Parent is CallInstruction { Method: { IsOperator: true, IsStatic: false } } && Arguments[0] == load` match instead of calling `UserDefinedCompoundAssign.IsCompoundAssignmentReceiverUse`, which this PR adds and which ILInlining.cs:269 and UsingTransform.cs:195/197 already use. Tightening the shared helper later (e.g. to the compound-assignment/void shape, or adding the `Accessibility == Public` requirement ReplaceMethodCallsWithOperators.cs:199 has) would leave CopyPropagation on the old rule. `if (!UserDefinedCompoundAssign.IsCompoundAssignmentReceiverUse(load)) continue; var call = (CallInstruction)load.Parent!;`

ICSharpCode.Decompiler/TypeSystem/Implementation/MetadataMethod.cs
```diff
@@ -118,6 +122,53 @@ internal MetadataMethod(MetadataModule module, MethodDefinitionHandle handle)
...
 
+		/// <summary>
+		/// Gets whether this method has the shape C# requires of a user-defined compound assignment
+		/// operator: an instance method returning void, taking one parameter (none for the
+		/// increment and decrement operators). Other languages give unrelated methods the same
+		/// metadata names: F# mangles
+		/// "static member (+=)" to a static, value-returning op_AdditionAssignment, and C++/CLI emits
+		/// value-returning instance operators. Those are plain methods as far as C# is concerned.
+		/// </summary>
+		/// <summary>
```
christophwille, line 133: Two stacked `<summary>` blocks: the first one (shape: instance, void, one/zero parameters, F# / C++/CLI note) describes `IsCompoundAssignmentOperatorSignature`, which currently has no doc comment; move it up there.

ICSharpCode.Decompiler/CSharp/Syntax/TypeMembers/OperatorDeclaration.cs
```diff
@@ -201,10 +244,25 @@ public static bool IsChecked(OperatorType type)
...
+				OperatorType.CheckedDecrementAssignment => true,
 				_ => false,
 			};
 		}
 
+		/// <summary>
+		/// Gets whether the operator type is a C# 14 user-defined compound assignment operator
+		/// (a void-returning instance operator, including the increment/decrement forms).
+		/// </summary>
+		public static bool IsCompoundAssignment(OperatorType type)
```
christophwille, line 261: `type >= AdditionAssignment` depends on the enum order staying as it is; an explicit switch (like `IsChecked` next to it) or deriving from the names table would not break silently when someone appends a non-assignment member to `OperatorType`.

ICSharpCode.Decompiler.Tests/TestCases/Pretty/UserDefinedCompoundAssignment.cs
dgrunwald: I think there's still no test for the ref-local `operator+` call that could get mis-decompiled into an `operator+=` call.
I'm asking for a test that would fail if we didn't have IsShadowedByInstanceOperator.

ICSharpCode.Decompiler/IL/ILReader.cs
dgrunwald: Caution: this PR is already overly complicated. If complexity is exploded further for minor gains (like this AI comment seems to want), the whole feature will be rejected.

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -853,10 +853,23 @@ protected internal override TranslatedExpression VisitLdLoca(LdLoca inst, Transl
 				.WithRR(new ByReferenceResolveResult(expr.ResolveResult, ReferenceKind.Ref));
 		}
 
+		/// <summary>
+		/// Gets whether the variable still carries the type the ILReader hands a stack slot, which is
+		/// derived from the stack type alone and therefore says nothing about the value stored in it.
+		/// A slot that was given a type on purpose has to keep it.
+		/// </summary>
+		bool HasDefaultStackSlotType(ILVariable variable)
```
dgrunwald, line 861: Maybe no longer needed after my ILAst typing improvements?

PR #3973, opened 2026-08-09
Fix the remaining deconstruction issues

### LINKED ISSUE #2322: Incorrect deconstruction of foreach item
### Input code
```cs
static class DeconstructForeachItemTest
{
    static void Test(Dictionary<string, object> data)
    {
        var dictionary = new Dictionary<string, object>();

        foreach (var (key, value) in data)
        {
            dictionary[key] = value;
        }
    }

    static void Deconstruct<K, V>(this KeyValuePair<K, V> pair, out K k, out V v)
    {
        k = pair.Key;
        v = pair.Value;
    }
}
```

### Erroneous output
```cs
internal static class DeconstructForeachItemTest
{
    private static void Test(Dictionary<string, object> data)
    {
        Dictionary<string, object> dictionary = new Dictionary<string, object>();
        foreach (KeyValuePair<string, object> datum in data)
        {
            string key;
            (key, dictionary[key]) = datum;
        }
    }

    private static void Deconstruct<K, V>(this KeyValuePair<K, V> pair, out K k, out V v)
    {
        k = pair.Key;
        v = pair.Value;
    }
}
```
Error CS0165: Use of unassigned local variable 'key'

### Details
* ILSpy version 7.0.0.6372-preview3

### LINKED ISSUE #3037: System.InvalidOperationException: inst is not an assignment! in DeconstructInstruction.CheckInvariant
### Steps to reproduce
1. Compile the below code and view `Class1.<Test1>d__1.MoveNext()` in `C#` or `IL with C#` mode.
Seems 
```cs
public class Class1
{
	public (int, int) Test2()
	{
		return default((int, int));
	}

	public async Task<int> Test1(string file)
	{
		int a = 0;
		int b = 0;
		await Task.Run(delegate
		{
			(a, b) = Test2();
		});
		return a + b;
	}
}
```

### Error message shown

```
Error decompiling @0600000A Class1.<Test1>d__1.MoveNext
in assembly "F:\z1\bin\Release\net8.0\z1.dll"
 ---> System.InvalidOperationException: inst is not an assignment!
   at ICSharpCode.Decompiler.IL.DeconstructInstruction.CheckInvariant(ILPhase phase) in DeconstructInstruction.cs:line 298
   at ICSharpCode.Decompiler.IL.ILInstruction.CheckInvariant(ILPhase phase) in ILInstruction.cs:line 95
   at ICSharpCode.Decompiler.IL.Block.CheckInvariant(ILPhase phase) in Block.cs:line 108
   at ICSharpCode.Decompiler.IL.ILInstruction.CheckInvariant(ILPhase phase) in ILInstruction.cs:line 95
   at ICSharpCode.Decompiler.IL.BlockContainer.CheckInvariant(ILPhase phase) in BlockContainer.cs:line 196
   at ICSharpCode.Decompiler.IL.ILInstruction.CheckInvariant(ILPhase phase) in ILInstruction.cs:line 95
   at ICSharpCode.Decompiler.IL.ILFunction.CheckInvariant(ILPhase phase) in ILFunction.cs:line 260
   at ICSharpCode.Decompiler.IL.ILInstruction.CheckInvariant(ILPhase phase) in ILInstruction.cs:line 95
   at ICSharpCode.Decompiler.IL.CallInstruction.CheckInvariant(ILPhase phase) in CallInstruction.cs:line 123
   at ICSharpCode.Decompiler.IL.ILInstruction.CheckInvariant(ILPhase phase) in ILInstruction.cs:line 95
   at ICSharpCode.Decompiler.IL.CallInstruction.CheckInvariant(ILPhase phase) in CallInstruction.cs:line 123
   at ICSharpCode.Decompiler.IL.ILInstruction.CheckInvariant(ILPhase phase) in ILInstruction.cs:line 95
   at ICSharpCode.Decompiler.IL.StLoc.CheckInvariant(ILPhase phase) in StLoc.cs:line 41
   at ICSharpCode.Decompiler.IL.ILInstruction.CheckInvariant(ILPhase phase) in ILInstruction.cs:line 95
   at ICSharpCode.Decompiler.IL.Block.CheckInvariant(ILPhase phase) in Block.cs:line 108
   at ICSharpCode.Decompiler.IL.ILInstruction.CheckInvariant(ILPhase phase) in ILInstruction.cs:line 95
   at ICSharpCode.Decompiler.IL.IfInstruction.CheckInvariant(ILPhase phase) in IfInstruction.cs:line 58
   at ICSharpCode.Decompiler.IL.ILInstruction.CheckInvariant(ILPhase phase) in ILInstruction.cs:line 95
   at ICSharpCode.Decompiler.IL.Block.CheckInvariant(ILPhase phase) in Block.cs:line 108
   at ICSharpCode.Decompiler.IL.ILInstruction.CheckInvariant(ILPhase phase) in ILInstruction.cs:line 95
   at ICSharpCode.Decompiler.IL.BlockContainer.CheckInvariant(ILPhase phase) in BlockContainer.cs:line 196
   at ICSharpCode.Decompiler.IL.ILInstruction.CheckInvariant(ILPhase phase) in ILInstruction.cs:line 95
   at ICSharpCode.Decompiler.IL.ILInstruction.CheckInvariant(ILPhase phase) in ILInstruction.cs:line 95
   at ICSharpCode.Decompiler.IL.Block.CheckInvariant(ILPhase phase) in Block.cs:line 108
   at ICSharpCode.Decompiler.IL.ILInstruction.CheckInvariant(ILPhase phase) in ILInstruction.cs:line 95
   at ICSharpCode.Decompiler.IL.BlockContainer.CheckInvariant(ILPhase phase) in BlockContainer.cs:line 196
   at ICSharpCode.Decompiler.IL.ILInstruction.CheckInvariant(ILPhase phase) in ILInstruction.cs:line 95
   at ICSharpCode.Decompiler.IL.ILFunction.CheckInvariant(ILPhase phase) in ILFunction.cs:line 260
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun decompileRun, ITypeResolveContext decompilationContext) in CSharpDecompiler.cs:line 1721
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun decompileRun, ITypeResolveContext decompilationContext) in CSharpDecompiler.cs:line 1756
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DoDecompile(IMethod method, DecompileRun decompileRun, ITypeResolveContext decompilationContext) in CSharpDecompiler.cs:line 1605
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.<>c__DisplayClass74_0.<DoDecompile>g__DoDecompileMember|0(IEntity entity, RecordDecompiler recordDecompiler, PartialTypeInfo partialType) in CSharpDecompiler.cs:line 1480
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DoDecompile(ITypeDefinition typeDef, DecompileRun decompileRun, ITypeResolveContext decompilationContext) in CSharpDecompiler.cs:line 1369
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.Decompile(IEnumerable`1 definitions) in CSharpDecompiler.cs:line 1034
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.Decompile(EntityHandle[] definitions) in CSharpDecompiler.cs:line 1006
   at ICSharpCode.ILSpy.CSharpLanguage.DecompileType(ITypeDefinition type, ITextOutput output, DecompilationOptions options) in CSharpLanguage.cs:line 334
   at ICSharpCode.ILSpy.TreeNodes.TypeTreeNode.Decompile(Language language, ITextOutput output, DecompilationOptions options) in TypeTreeNode.cs:line 132
   at ICSharpCode.ILSpy.TextView.DecompilerTextView.DecompileNodes(DecompilationContext context, ITextOutput textOutput) in DecompilerTextView.cs:line 914
   at ICSharpCode.ILSpy.TextView.DecompilerTextView.<>c__DisplayClass52_0.<DecompileAsync>b__0() in DecompilerTextView.cs:line 884

```

### Details
* Product in use: ILSpy
* Version in use: 8.0.0.7444(last CI build)

### LINKED ISSUE #3208: Wrong decompilation
### Details
ILSpy version 9.0.0.7618-preview1

### LINKED ISSUE #3275: ILSpy crashes when it parses deconstruction constructs
### Steps to reproduce
1. Create and compile the code below using .NET 8.
2. Try to decompile it with ILSpy:
```
ilspycmd.exe -p -o .\decompiled2 ConsoleApp1.dll
```
3. A crash occurs

### Code:
ConsoleApp1:
```
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net8.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>
</Project>
```

Program.cs:
```
class Test
{
    private WorkerProfileBounds SqlDbBounds => new(1024, 4096);
    public readonly record struct WorkerProfileInfo(WorkerProfile WorkerProfile, WorkerProfileBounds Bounds);
    public readonly record struct WorkerProfileBounds(ulong SmallProfileUpperBound, ulong LargeProfileLowerBound);
    public enum WorkerProfile
    {
        Small,
        Medium,
        Large,
        Archive
    }
    public WorkerProfileInfo GetDeploymentWorkerProfileForSql(ulong? protectedResourceSize)
        => GetWorkerProfileInfoForSize(protectedResourceSize, SqlDbBounds);
    private static WorkerProfileInfo GetWorkerProfileInfoForSize(ulong? protectedResourceSize, WorkerProfileBounds bounds)
        => new(WorkerProfile.Small, bounds);
}

class Program { 
    static void Main() {
        Test test = new Test();
        var (workerProfile, (smallBound, largeBound)) = test.GetDeploymentWorkerProfileForSql(777);

        Console.WriteLine(workerProfile.ToString());
        Console.WriteLine(smallBound);
        Console.WriteLine(largeBound);
    }

}
```
### Error message shown

```
System.AggregateException: One or more errors occurred. (Error decompiling @06000005 Program.Main)
 ---> Error decompiling @06000005 Program.Main
in assembly "ConsoleApp1.dll"
 ---> System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
   at System.ThrowHelper.ThrowArgumentOutOfRange_IndexException() in offset 0
   at ICSharpCode.Decompiler.CSharp.ExpressionBuilder.<VisitDeconstructInstruction>g__ConstructTuple|159_0(MatchInstruction matchInstruction, <>c__DisplayClass159_0&) in ExpressionBuilder.cs:line 4496
   at ICSharpCode.Decompiler.CSharp.ExpressionBuilder.VisitDeconstructInstruction(DeconstructInstruction inst, TranslationContext context) in ExpressionBuilder.cs:line 4481
   at ICSharpCode.Decompiler.IL.DeconstructInstruction.AcceptVisitor[C,T](ILVisitor`2 visitor, C context) in Instructions.cs:line 6729
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.Default(ILInstruction inst) in StatementBuilder.cs:line 89
   at ICSharpCode.Decompiler.IL.ILVisitor`1.VisitDeconstructInstruction(DeconstructInstruction inst) in Instructions.cs:line 7614
   at ICSharpCode.Decompiler.IL.DeconstructInstruction.AcceptVisitor[T](ILVisitor`1 visitor) in Instructions.cs:line 6725
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.Convert(ILInstruction inst) in StatementBuilder.cs:line 78
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertBlockContainer(BlockStatement blockStatement, BlockContainer container, IEnumerable`1 blocks, Boolean isLoop) in StatementBuilder.cs:line 1429
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertBlockContainer(BlockContainer container, Boolean isLoop) in StatementBuilder.cs:line 1367
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.VisitBlockContainer(BlockContainer container) in StatementBuilder.cs:line 1253
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.Convert(ILInstruction inst) in StatementBuilder.cs:line 78
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertAsBlock(ILInstruction inst) in StatementBuilder.cs:line 83
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun decompileRun, ITypeResolveContext decompilationContext) in CSharpDecompiler.cs:line 1750
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun decompileRun, ITypeResolveContext decompilationContext) in CSharpDecompiler.cs:line 1773
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DoDecompile(IMethod method, DecompileRun decompileRun, ITypeResolveContext decompilationContext) in CSharpDecompiler.cs:line 1622
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.<>c__DisplayClass74_0.<DoDecompile>g__DoDecompileMember|0(IEntity entity, RecordDecompiler recordDecompiler, PartialTypeInfo partialType) in CSharpDecompiler.cs:line 1497
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DoDecompile(ITypeDefinition typeDef, DecompileRun decompileRun, ITypeResolveContext decompilationContext) in CSharpDecompiler.cs:line 1364
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DoDecompileTypes(IEnumerable`1 types, DecompileRun decompileRun, ITypeResolveContext decompilationContext, SyntaxTree syntaxTree) in CSharpDecompiler.cs:line 651
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DecompileTypes(IEnumerable`1 types) in CSharpDecompiler.cs:line 950
   at ICSharpCode.Decompiler.CSharp.ProjectDecompiler.WholeProjectDecompiler.<>c__DisplayClass46_0.<WriteCodeFilesInProject>b__5(IGrouping`2 file) in WholeProjectDecompiler.cs:line 274
   at System.Threading.Tasks.Parallel.<>c__DisplayClass44_0`2.<PartitionerForEachWorker>b__1(IEnumerator& partitionState, Int32 timeout, Boolean& replicationDelegateYieldedBeforeCompletion) in offset 259
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw() in offset 17
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw(Exception source) in offset 0
   at System.Threading.Tasks.Parallel.<>c__DisplayClass44_0`2.<PartitionerForEachWorker>b__1(IEnumerator& partitionState, Int32 timeout, Boolean& replicationDelegateYieldedBeforeCompletion) in offset 609
   at System.Threading.Tasks.TaskReplicator.Replica`1.ExecuteAction(Boolean& yieldedBeforeCompletion) in offset 0
   at System.Threading.Tasks.TaskReplicator.Replica.Execute() in offset 0
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.TaskReplicator.Run[TState](ReplicatableUserAction`1 action, ParallelOptions options, Boolean stopOnFirstFailure)
   at System.Threading.Tasks.Parallel.PartitionerForEachWorker[TSource,TLocal](Partitioner`1 source, ParallelOptions parallelOptions, Action`1 simpleBody, Action`2 bodyWithState, Action`3 bodyWithStateAndIndex, Func`4 bodyWithStateAndLocal, Func`5 bodyWithEverything, Func`1 localInit, Action`1 localFinally)
--- End of stack trace from previous location ---
   at System.Threading.Tasks.Parallel.ThrowSingleCancellationExceptionOrOtherException(ICollection exceptions, CancellationToken cancelToken, Exception otherException)
   at System.Threading.Tasks.Parallel.PartitionerForEachWorker[TSource,TLocal](Partitioner`1 source, ParallelOptions parallelOptions, Action`1 simpleBody, Action`2 bodyWithState, Action`3 bodyWithStateAndIndex, Func`4 bodyWithStateAndLocal, Func`5 bodyWithEverything, Func`1 localInit, Action`1 localFinally)
   at System.Threading.Tasks.Parallel.ForEach[TSource](Partitioner`1 source, ParallelOptions parallelOptions, Action`1 body)
   at ICSharpCode.Decompiler.CSharp.ProjectDecompiler.WholeProjectDecompiler.<>c__DisplayClass46_0.<WriteCodeFilesInProject>g__ProcessFiles|3(List`1 files) in /_/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs:line 255
   at ICSharpCode.Decompiler.CSharp.ProjectDecompiler.WholeProjectDecompiler.WriteCodeFilesInProject(PEFile module, IList`1 partialTypes, CancellationToken cancellationToken) in /_/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs:line 221
   at ICSharpCode.Decompiler.CSharp.ProjectDecompiler.WholeProjectDecompiler.DecompileProject(PEFile moduleDefinition, String targetDirectory, TextWriter projectFileWriter, CancellationToken cancellationToken) in /_/ICSharpCode.Decompiler/CSharp/ProjectDecompiler/WholeProjectDecompiler.cs:line 154
   at ICSharpCode.ILSpyCmd.ILSpyCmdProgram.DecompileAsProject(String assemblyFileName, String projectFileName) in D:\a\ILSpy\ILSpy\ICSharpCode.ILSpyCmd\IlspyCmdProgram.cs:line 318
   at ICSharpCode.ILSpyCmd.ILSpyCmdProgram.OnExecuteAsync(CommandLineApplication app) in D:\a\ILSpy\ILSpy\ICSharpCode.ILSpyCmd\IlspyCmdProgram.cs:line 147
```

### Details
* ilspycmd: 8.2.0.7535
* Microsoft.NETCore.App 8.0.6

### LINKED ISSUE #3453: Wrong decompilation of deconstruction in foreach
### Input code
```cs
using System.Collections.Immutable;

class EnumeratorDeconstructionTest
{
    struct Data
    {
        public int A;
        public int B;

        public void Deconstruct(out int a, out int b)
        {
            a = A;
            b = B;
        }
    }

    void Test(ImmutableArray<Data> array)
    {
        foreach (var item in array)
        {
            var (a, b) = item;

            Dummy(a);
        }
    }

    void Dummy(int v) { }
}
```

### Erroneous output
```cs
using System.Collections.Immutable;

internal class EnumeratorDeconstructionTest
{
	private struct Data
	{
		public int A;

		public int B;

		public void Deconstruct(out int a, out int b)
		{
			a = A;
			b = B;
		}
	}

	private void Test(ImmutableArray<Data> array)
	{
		ImmutableArray<Data>.Enumerator enumerator = array.GetEnumerator();

		while (enumerator.MoveNext())
		{
			var (a, _) = (Data)(ref enumerator.Current);
			Dummy(a);
		}
	}

	private void Dummy(int v)
	{
	}
}
```

### Details
Tested with ILSpy at commit fd6070d2292edcf817763b191ea4d491368970ec

### LINKED ISSUE #3962: Nested tuple designation is not reconstructed when the inner tuple has 8 or more elements
Follow-up to #3950, which reconstructs nested tuple designations. An inner tuple of eight or more elements is not reconstructed, because its last element is reached through the `Rest` field rather than an `Item_N` field of the temporary itself.

### Repro

```csharp
static (T1, T2) GetTuple<T1, T2>() => default;

public static void RestChainedInner()
{
    var (x, (a, b, c, d, e, f, g, h)) = GetTuple<int, (int, int, int, int, int, int, int, int)>();
    Console.WriteLine(x);
    Console.WriteLine(a);
    // ... all eight inner elements read
    Console.WriteLine(h);
}
```

### Actual

```csharp
(int, (int, int, int, int, int, int, int, int)) tuple = GetTuple<int, (int, int, int, int, int, int, int, int)>();
(int, int, int, int, int, int, int, int) item = tuple.Item2;
int item2 = tuple.Item1;
int item3 = item.Item1;
// ...
int item10 = item.Rest.Item1;
```

### Expected

```csharp
var (x, (a, b, c, d, e, f, g, h)) = GetTuple<int, (int, int, int, int, int, int, int, int)>();
```

Seven or fewer inner elements reconstruct correctly, so this is specific to the `Rest` chaining of long tuples.

### Notes for whoever picks this up

`AllUsesAreTupleElementReads` in `DeconstructionTransform` is the guard that rejects it, and it is the obvious suspect since it requires every use of the temporary to be `ldobj(ldflda Item_N(...))` while a `Rest`-chained read is `ldobj(ldflda Item1(ldflda Rest(...)))`. Extending it to step over `Rest` accesses is **not** sufficient on its own: with that change the match still bails at the same guard, so the use shape differs from what that reading suggests and the cause needs to be established before fixing.

The index arithmetic is already in place: `TupleTransform.MatchTupleFieldAccess` walks the `Rest` chain and folds it back into a flat position (`position += TupleType.RestPosition - 1`), and `TupleType` counts cardinality flattened.

Only sugar is lost; the output above is correct and recompiles.

---

siegfriedpammer:
Fixes #3208, fixes #3453, fixes #3962. Closes #2322, closes #3037, closes #3275: those three no longer reproduce on master (#3275 was fixed by #3949), and this PR adds the missing regression fixtures for them, so they can be closed with it.

Triage first: each issue's repro was compiled and decompiled against current master. The three crashes/wrong-code reports (#2322, #3037, #3275) are already fixed - #2322 is covered by the existing `DeconstructDictionaryForEach` fixture, and this PR adds fixtures for #3037 (deconstruction assignment to locals captured by a lambda in an async method) and #3275 (nested deconstruction of readonly record structs). What remained collapses into two defects:

**Value-semantics copy left behind (#3453, #3208).** Deconstruction assignment copies the right-hand side into a temporary before calling `Deconstruct`. When the RHS is a call, inlining folds the temporary; for a local or parameter it survived into the output as a separate statement (`CalibrationPoint calibrationPoint = point; var (num, num2) = calibrationPoint;`). `DeconstructionTransform` now consumes the copy into the pattern - rendering the copied value as the RHS recompiles to the identical temporary. Because blocks are matched back to front, the call-position match defers to the attempt starting at the copy, mirroring the existing nested-deconstruction defer guard.

**`Rest`-chained inner tuple designation (#3962).** Element 8+ of a long tuple is read through the `Rest` field, which Roslyn loads *by value*; ILSpy rendered that as an `addressof(ldobj(...))` sandwich that hid the tuple field chain from every downstream matcher (including the existing `Rest` flattening in `MatchTupleFieldAccess`). Two-part fix at the failure origin: `EarlyExpressionTransforms` elides the copy when the enclosing expression only reads through it, and the deconstruction transform's use-shape guard walks the resulting `ldflda` chain instead of requiring exactly one field access. `var (x, (a, b, c, d, e, f, g, h)) = ...` now reconstructs fully; as a side effect, plain reads render as `item.Item8` instead of `item.Rest.Item1`.

Each fix was developed fixture-first (red on master, green after); the full decompiler suite passes on every commit.

---

### DIFF
--- a/ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
+++ b/ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
@@ -129,6 +129,17 @@ namespace ICSharpCode.Decompiler.IL.Transforms
 			// exists (see the guard for the precision guarantees).
 			if (IsConsumableByEnclosingDeconstruction(block, pos))
 				return false;
+			// Same idea for the value-semantics copy before a root Deconstruct call: blocks are
+			// processed back to front, so defer the call-only match to the attempt starting at
+			// the copy, which can consume both (see MatchDeconstruction). Defer only when that
+			// attempt actually reaches its match: an attempt that is itself deferred to an
+			// enclosing pattern bails without consuming this position, and the back-to-front
+			// walk never comes back, which would lose the deconstruction at both positions.
+			if (pos > 0 && IsRootDeconstructionCopy(block, pos - 1, out _, out _)
+				&& !IsConsumableByEnclosingDeconstruction(block, pos - 1))
+			{
+				return false;
+			}
 			if (!MatchDeconstructionSequence(block, startPos, out pos, out var rootCall,
 				out var rootTestedOperand, out var conversionStLocs, out var delayedActions))
 			{
@@ -431,7 +442,19 @@ namespace ICSharpCode.Decompiler.IL.Transforms
 		void MatchDeconstruction(Block block, ref int pos, out DeconstructionCall? rootCall,
 			out ILInstruction? testedOperand)
 		{
-			rootCall = MatchDeconstructionCall(block.Instructions[pos], out testedOperand);
+			// Deconstruction assignment has value semantics, so Roslyn copies the deconstructed
+			// value into a temporary and calls Deconstruct on that. When the value is a call
+			// result, inlining already folds the temporary away; when it is a local or parameter,
+			// the copy survives to here. Consume it into the pattern: rendering the copied value
+			// as the deconstruction target recompiles to the identical temporary.
+			rootCall = null;
+			testedOperand = null;
+			if (IsRootDeconstructionCopy(block, pos, out var copiedValue, out rootCall))
+			{
+				testedOperand = copiedValue;
+				pos++;
+			}
+			rootCall ??= MatchDeconstructionCall(block.Instructions[pos], out testedOperand);
 			if (rootCall == null)
 				return;
 			rootedInDeconstructCall = true;
@@ -458,6 +481,46 @@ namespace ICSharpCode.Decompiler.IL.Transforms
 						leaves.Add(call.Results[i]);
 				}
 			}
+		}
+
+		/// <summary>
+		/// stloc copy(value)                          at pos
+		/// call Deconstruct(ldloc(a) copy, ...)       a root Deconstruct call on the copy
+		/// where the copy has no other use: the value-semantics temporary Roslyn emits for a
+		/// deconstruction whose right-hand side is not already a temporary. On success,
+		/// <paramref name="copiedValue"/> is the deconstructed value and <paramref name="call"/>
+		/// the matched call, so callers need not re-match either.
+		/// </summary>
+		bool IsRootDeconstructionCopy(Block block, int pos, out ILInstruction? copiedValue,
+			out DeconstructionCall? call)
+		{
+			copiedValue = null;
+			call = null;
+			if (pos + 1 >= block.Instructions.Count)
+				return false;
+			if (!block.Instructions[pos].MatchStLoc(out var copy, out var value))
+				return false;
+			if (copy.Kind is not (VariableKind.Local or VariableKind.StackSlot))
+				return false;
+			// A byref temporary is not a copy: Deconstruct called through it acts on the original,
+			// which is not what a value-semantics deconstruction of the referenced expression does.
+			if (copy.StackType == StackType.Ref)
+				return false;
+			if (!(copy.StoreCount == 1 && copy.LoadCount + copy.AddressCount == 1))
+				return false;
+			// The defensive copy of a struct element of an enclosing Deconstruct call has this
+			// exact shape. It belongs to the enclosing call's nested designation, so consuming it
+			// here would commit the inner call on its own and break the designation for good.
+			if (TryFindEnclosingDeconstructionCall(block, pos + 1, out _))
+				return false;
+			var matchedCall = MatchDeconstructionCall(block.Instructions[pos + 1], out var testedOperand);
+			if (matchedCall == null)
+				return false;
+			if (!MatchLdLocOrLdLoca(testedOperand!, out var receiver) || receiver != copy)
+				return false;
+			copiedValue = value;
+			call = matchedCall;
+			return true;
 		}
 
 		/// <summary>
--- a/ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
+++ b/ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
@@ -731,7 +731,16 @@ namespace ICSharpCode.Decompiler.IL.Transforms
 			{
 				foreach (var use in temp.AddressInstructions.Concat<ILInstruction>(temp.LoadInstructions))
 				{
-					if (!(use.Parent is LdFlda elementAccess && elementAccess.Parent is LdObj))
+					// Walk the ldflda chain up to the reading ldobj: element 8+ of a long tuple
+					// is accessed through the Rest field, i.e. through more than one ldflda.
+					// Whether each read is really a tuple element access (and consumed by the
+					// pattern) is verified by MatchTupleElementRead and the escape check.
+					ILInstruction? node = use.Parent;
+					if (node is not LdFlda)
+						return false;
+					while (node is LdFlda ldflda)
+						node = ldflda.Parent;
+					if (node is not LdObj)
 						return false;
 				}
 				return true;
--- a/ICSharpCode.Decompiler/IL/Transforms/EarlyExpressionTransforms.cs
+++ b/ICSharpCode.Decompiler/IL/Transforms/EarlyExpressionTransforms.cs
@@ -163,17 +163,37 @@ namespace ICSharpCode.Decompiler.IL.Transforms
 				temp = ldfldaTarget;
 				range = range.Concat(temp.ILRanges);
 			}
-			if (temp.MatchAddressOf(out var addressOfTarget, out _) && addressOfTarget.MatchLdLoc(out var v))
+			if (!temp.MatchAddressOf(out var addressOfTarget, out var addressOfType))
+				return;
+			ILInstruction replacement;
+			switch (addressOfTarget)
 			{
-				context.Step($"ldobj(...(addressof(ldloca {v.Name}))) => ldobj(...(ldloca {v.Name}))", inst);
-				var replacement = new LdLoca(v).WithILRange(addressOfTarget);
-				foreach (var r in range)
-				{
-					replacement = replacement.WithILRange(r);
-				}
-				temp.ReplaceWith(replacement);
-				context.EndStep(replacement);
+				case LdLoc { Variable: var v }:
+					context.Step($"ldobj(...(addressof(ldloca {v.Name}))) => ldobj(...(ldloca {v.Name}))", inst);
+					replacement = new LdLoca(v).WithILRange(addressOfTarget);
+					break;
+				// The enclosing ldobj only reads through the temporary the addressof creates, so
+				// the copy can be elided and the read go directly through the inner address.
+				// This shape arises when a struct field chain is read by value (ldfld, then a
+				// field of the loaded value), e.g. reading element 8+ of a long tuple via Rest.
+				// The type check keeps the elision on that shape: re-rooting a read of an
+				// incompatible type onto the inner address would read memory behind it that the
+				// copy never covered.
+				case LdObj { UnalignedPrefix: 0, IsVolatile: false } innerLdObj
+					when TypeUtils.IsCompatibleTypeForMemoryAccess(innerLdObj.Type, addressOfType):
+					context.Step("ldobj(...(addressof(ldobj(addr)))) => ldobj(...(addr))", inst);
+					replacement = innerLdObj.Target;
+					range = range.Concat(addressOfTarget.ILRanges);
+					break;
+				default:
+					return;
 			}
+			foreach (var r in range)
+			{
+				replacement = replacement.WithILRange(r);
+			}
+			temp.ReplaceWith(replacement);
+			context.EndStep(replacement);
 		}
 
 		protected internal override void VisitCall(Call inst)
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/DeconstructionTests.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/DeconstructionTests.cs
@@ -996,6 +996,20 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 			return a + b;
 		}
 
+		public void NestedDesignation_RestChainedInnerTuple()
+		{
+			var (value, (value2, value3, value4, value5, value6, value7, value8, value9)) = GetTuple<int, (int, int, int, int, int, int, int, int)>();
+			Console.WriteLine(value);
+			Console.WriteLine(value2);
+			Console.WriteLine(value3);
+			Console.WriteLine(value4);
+			Console.WriteLine(value5);
+			Console.WriteLine(value6);
+			Console.WriteLine(value7);
+			Console.WriteLine(value8);
+			Console.WriteLine(value9);
+		}
+
 		public bool DeconstructStructParameter(StructDeconstructionSource<int, string> point)
 		{
 			var (num2, value) = point;
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/DeconstructionTests.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/DeconstructionTests.cs
@@ -19,6 +19,7 @@
 using System;
 using System.Collections.Generic;
 using System.Runtime.InteropServices;
+using System.Threading.Tasks;
 
 namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 {
@@ -417,6 +418,26 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 			Console.WriteLine(value);
 			Console.WriteLine(myInt4);
 			Console.WriteLine(value2);
+		}
+
+		public void LocalVariable_Nested_StructInnerFirstElement()
+		{
+			var ((value, value2), value3) = GetSource<StructDeconstructionSource<int, string>, int>();
+			Console.WriteLine(value);
+			Console.WriteLine(value2);
+			Console.WriteLine(value3);
+		}
+
+		public void LocalVariable_ElementOfElementRead_ThenDeconstruct()
+		{
+			((StructDeconstructionSource<int, string>, int), int) tuple = GetTuple<(StructDeconstructionSource<int, string>, int), int>();
+			(StructDeconstructionSource<int, string>, int) item = tuple.Item1;
+			StructDeconstructionSource<int, string> item2 = item.Item1;
+			var (value, value2) = item2;
+			Console.WriteLine(value);
+			Console.WriteLine(value2);
+			Console.WriteLine(item.Item2);
+			Console.WriteLine(tuple.Item2);
 		}
 
 		public void LocalVariable_Nested_Depth3()
@@ -963,6 +984,30 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 			{
 				Console.WriteLine(text + ": " + num);
 			}
+		}
+
+		public async Task<int> DeconstructionAssignmentToCapturedLocals(string file)
+		{
+			int a = 0;
+			int b = 0;
+			await Task.Run(delegate {
+				(a, b) = GetTuple<int, int>();
+			});
+			return a + b;
+		}
+
+		public bool DeconstructStructParameter(StructDeconstructionSource<int, string> point)
+		{
+			var (num2, value) = point;
+			Console.WriteLine(value);
+			return num2 >= 0;
+		}
+
+		public void DeconstructStructLocal()
+		{
+			StructDeconstructionSource<int, string> structSource = GetStructSource<int, string>();
+			var (num2, text2) = structSource;
+			Console.WriteLine(num2 + text2 + structSource.Dummy);
 		}
 	}
 }
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/Records.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/Records.cs
@@ -285,6 +285,10 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 
 		public record struct PairWithPrimaryCtor<A, B>(A First, B Second);
 
+		public readonly record struct BoundsInfo(int Profile, Bounds Bounds);
+
+		public readonly record struct Bounds(ulong Small, ulong Large);
+
 		public record struct PrimaryCtor(int A, string B);
 
 		public record struct MultipleCtorsNoPrimaryCtor
@@ -446,6 +450,19 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 			{
 				C = 1.41;
 			}
+		}
+
+		private static BoundsInfo GetInfo()
+		{
+			return new BoundsInfo(1, new Bounds(2uL, 3uL));
+		}
+
+		public static void NestedDeconstruction()
+		{
+			var (value, (value2, value3)) = GetInfo();
+			Console.WriteLine(value);
+			Console.WriteLine(value2);
+			Console.WriteLine(value3);
 		}
 	}
 


ICSharpCode.Decompiler/IL/Transforms/EarlyExpressionTransforms.cs
```diff
@@ -174,6 +174,22 @@ internal static void AddressOfLdLocToLdLoca(LdObj inst, ILTransformContext conte
 				temp.ReplaceWith(replacement);
 				context.EndStep(replacement);
 			}
+			else if (temp.MatchAddressOf(out addressOfTarget, out _)
+				&& addressOfTarget is LdObj { UnalignedPrefix: 0, IsVolatile: false } innerLdObj)
```
christophwille: **Correctness (verified):** this new `ldobj(...(addressof(ldobj(addr)))) => ldobj(...(addr))` elision only checks the inner `LdObj`'s volatile/unaligned prefixes, but never checks type compatibility between the type the `addressof` temporary holds and the enclosing read. The sibling transforms (`LdObjToLdLoc`, `StObjToStLoc`) guard exactly this with `TypeUtils.IsCompatibleTypeForMemoryAccess`.

For type-punning IL (hand-written or obfuscated, not Roslyn-emitted) - e.g. `ldobj T2(addressof(ldobj T1(addr)))` with `sizeof(T2) > sizeof(T1)`, or an `ldflda` chain whose declaring type does not match the inner `ldobj`'s type - the rewrite re-roots the read directly onto `addr`: the original program read bytes out of a temporary copy of `T1`, while the transformed expression reads adjacent live memory behind `addr` that the original never touched. A type-compatibility check between the inner `ldobj` type and the `addressof` type (mirroring the sibling transforms) would restrict the elision to the Roslyn shape it targets.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
```diff
@@ -129,6 +129,11 @@ bool TransformDeconstruction(Block block, int pos)
 			// exists (see the guard for the precision guarantees).
 			if (IsConsumableByEnclosingDeconstruction(block, pos))
 				return false;
+			// Same idea for the value-semantics copy before a root Deconstruct call: blocks are
+			// processed back to front, so defer the call-only match to the attempt starting at
+			// the copy, which can consume both (see MatchDeconstruction).
+			if (pos > 0 && IsRootDeconstructionCopy(block.Instructions[pos - 1], block.Instructions[pos]))
```
christophwille: **Correctness (verified, reproduced end-to-end):** this guard unconditionally defers the call-only match at `pos` whenever the copy+call shape holds at `pos - 1`, but the deferred-to attempt at `pos - 1` can itself be aborted before `MatchDeconstruction` runs: when the copy's value is a tuple-element read of an earlier designation temporary, `IsConsumableByEnclosingDeconstruction` (line 130) returns true for `pos - 1` via `HasEnclosingTupleDesignation`, and that attempt bails. The back-to-front walk never returns to `pos`, so the deconstruction is lost at *both* positions.

Repro shape: `var inner = big.Item1; var (x, y) = inner.Item1;` (element type has a `Deconstruct` method). On master the call position matched and emitted `var (x, y) = ...;`; with this guard the output regresses to an explicit temporary plus a raw `.Deconstruct(out ...)` call. Deferring only when the `pos - 1` attempt will actually consume the pair (i.e. not itself deferred/aborted) would fix it.

**Minor, related:** the guard here and the consuming check inside `MatchDeconstruction` encode the same adjacency test at shifted offsets (`(pos-1, pos)` vs `(pos, pos+1)`), coupled only by comments. If the consuming side is ever tightened without mirroring this guard, positions defer to an attempt that no longer consumes them and the sugar is silently lost. Keeping both sides on the identical shared predicate is what holds this together - worth preserving deliberately.

christophwille (commented):
Automated high-effort review (Claude Code, multi-agent find + independent verify pass). Three correctness findings and four cleanups survived verification; three further candidates were investigated and refuted. Details are in the inline comments; the short version:

**Correctness**
1. `EarlyExpressionTransforms.AddressOfLdLocToLdLoca` - the new `addressof(ldobj)` elision has no type-compatibility guard between the inner `ldobj` and the enclosing read, unlike its siblings `LdObjToLdLoc`/`StObjToStLoc`. Type-punning IL gets silently re-rooted onto memory the original never read.
2. `DeconstructionTransform` defer guard - when the deferred-to attempt at `pos - 1` is itself aborted by `IsConsumableByEnclosingDeconstruction`, the deconstruction is lost at *both* positions (reproduced end-to-end); master sugared this shape.
3. `IsRootDeconstructionCopy` also matches the defensive copy of a *first-position* nested struct element of an enclosing Deconstruct call, so the inner call gets committed as a standalone root deconstruction and the enclosing nested designation is permanently broken. Existing fixtures only cover struct elements in non-first positions, so the suite stays green.

**Cleanups**
- `MatchDeconstruction` re-matches the copy+call pair `IsRootDeconstructionCopy` just matched (up to three times per position); outing the copied value / matched call from the helper removes the redundancy and the drift hazard.
- `IsRootDeconstructionCopy` discards the copied value (`out _`), so a byref-typed temp also matches - an in-place Deconstruct through a reference renders as a value-semantics deconstruction.
- The two `AddressOfLdLocToLdLoca` branches duplicate `MatchAddressOf` plus the whole replacement/ILRange/ReplaceWith tail.

Findings 2 and 3 each come with a concrete source shape that regresses versus master - both would make good fixture additions.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
```diff
@@ -431,7 +436,22 @@ sealed class DeconstructionCall
...
-			rootCall = MatchDeconstructionCall(block.Instructions[pos], out testedOperand);
+			// Deconstruction assignment has value semantics, so Roslyn copies the deconstructed
+			// value into a temporary and calls Deconstruct on that. When the value is a call
+			// result, inlining already folds the temporary away; when it is a local or parameter,
+			// the copy survives to here. Consume it into the pattern: rendering the copied value
+			// as the deconstruction target recompiles to the identical temporary.
+			rootCall = null;
+			testedOperand = null;
+			if (pos + 1 < block.Instructions.Count
+				&& IsRootDeconstructionCopy(block.Instructions[pos], block.Instructions[pos + 1]))
```
christophwille: **Correctness (verified):** `IsRootDeconstructionCopy` also matches the *defensive copy of a nested struct element* of an enclosing Deconstruct call, so the attempt starting at that copy commits the inner call as a standalone root deconstruction and permanently breaks the enclosing nested designation. The back-to-front walk visits the copy before the enclosing call, and nothing defers at the copy position: `IsConsumableByEnclosingDeconstruction` needs a `CallInstruction` at `pos`, and the `pos - 1` guard sees the enclosing call, not a `stloc`.

Repro shape: `var ((a, b), c) = GetSource<StructDeconstructionSource<int, string>, int>();` - a struct element with its own `Deconstruct` in **first** position, conversion-free leaves. At the copy position, `MatchDeconstruction` consumes copy+inner call and commits `(a, b) = s;`; the enclosing call then has no assignments left and stays explicit. Output regresses from master's `var ((a, b), c) = src;` to `src.Deconstruct(out var s, out var num); (a, b) = s; int c = num;`.

The existing fixtures only place struct elements in non-first positions (`BothElementsNested` has the class element first), so the suite stays green while this shape regresses - a first-position fixture would catch it.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
```diff
@@ -431,7 +436,22 @@ sealed class DeconstructionCall
...
+			{
+				block.Instructions[pos].MatchStLoc(out _, out var copiedValue);
```
christophwille: **Cleanup:** this branch re-matches instructions `IsRootDeconstructionCopy` just matched - a `MatchStLoc` whose bool result is discarded, plus a second `MatchDeconstructionCall` on the same call instruction (a third counting the `TransformDeconstruction` guard). Having `IsRootDeconstructionCopy` return the copied value and the matched `DeconstructionCall` via out parameters removes the redundant work and the lockstep hazard: today, if the helper is ever loosened (say, to accept a `stobj`-shaped copy), the ignored-result `MatchStLoc(out _, out var copiedValue)` silently leaves `copiedValue` wrong instead of failing the match.

The `|| testedOperand == null` check below is also dead: `MatchDeconstructionCall` sets `testedOperand` from `call.Arguments[0]` after checking `Arguments.Count >= 3`.

ICSharpCode.Decompiler/IL/Transforms/DeconstructionTransform.cs
```diff
@@ -460,6 +480,24 @@ static void CollectLeaves(DeconstructionCall call, List<ILVariable> leaves)
...
 		}
 
+		/// <summary>
+		/// stloc copy(value) directly followed by a root Deconstruct call on the copy, where the
+		/// copy has no other use: the value-semantics temporary Roslyn emits for a deconstruction
+		/// whose right-hand side is not already a temporary.
+		/// </summary>
+		bool IsRootDeconstructionCopy(ILInstruction copyInst, ILInstruction callInst)
+		{
+			if (!copyInst.MatchStLoc(out var copy, out _))
```
christophwille: **Strictness:** the copied value is discarded (`out _`), so the helper never checks that this is actually a by-value copy. A byref-typed temp - `stloc r(ldflda f(x)); call Deconstruct(ldloc r, ...)` - satisfies every condition here (`MatchLdLocOrLdLoca` accepts the plain `ldloc r` receiver), and an in-place `Deconstruct` call through a reference gets rendered as a value-semantics deconstruction of `x.f`. If `Deconstruct` mutates its receiver struct, the decompiled `(a, b) = x.f;` recompiles to different behavior.

The XML doc describes the target shape as the value-semantics temporary for a local or parameter; constraining the copied value (non-byref copy variable type, or value matching `ldloc`/`ldobj`) keeps the matcher at that guaranteed shape.

ICSharpCode.Decompiler/IL/Transforms/EarlyExpressionTransforms.cs
```diff
@@ -174,6 +174,22 @@ internal static void AddressOfLdLocToLdLoca(LdObj inst, ILTransformContext conte
 				temp.ReplaceWith(replacement);
 				context.EndStep(replacement);
 			}
+			else if (temp.MatchAddressOf(out addressOfTarget, out _)
```
christophwille: **Cleanup:** the two branches are near-duplicates - `temp.MatchAddressOf` is evaluated twice (every visited `LdObj` pays the second match when the first branch fails), and the replacement/ILRange-transfer/`ReplaceWith`/`EndStep` boilerplate is duplicated. A single `if (temp.MatchAddressOf(out var addressOfTarget, out var type))` that dispatches on `addressOfTarget` (`MatchLdLoc` vs `LdObj`) to compute the replacement, sharing one range-transfer loop and one `ReplaceWith`/`EndStep` tail, does the same job with half the surface - and gives the missing type check (see the comment on the `LdObj` branch) a single place to live.

siegfriedpammer:
All six findings are addressed in the force-pushed branch (rebased on master, review fixes squashed into the commits they belong to). Full decompiler suite green: 3369 tests, 0 failed.

**1. `AddressOfLdLocToLdLoca` type compatibility** — the `addressof(ldobj)` elision now requires `TypeUtils.IsCompatibleTypeForMemoryAccess(innerLdObj.Type, addressOfType)`, mirroring `LdObjToLdLoc`/`StObjToStLoc`. No fixture: the shape needs hand-written type-punning IL, and the guard only narrows an existing rewrite.

**2. Defer guard losing the deconstruction at both positions** — confirmed and fixed: the call position now defers to `pos - 1` only when that attempt is not itself deferred (`&& !IsConsumableByEnclosingDeconstruction(block, pos - 1)`). New fixture `LocalVariable_ElementOfElementRead_ThenDeconstruct` reproduced it exactly as described (`item2.Deconstruct(out var a, out var b); int value = a; ...`) and now pins the sugared form.

**3. `IsRootDeconstructionCopy` matching a nested first-position struct element** — confirmed and fixed: the helper rejects the pair when `TryFindEnclosingDeconstructionCall` finds the following call belongs to an enclosing deconstruction. New fixture `LocalVariable_Nested_StructInnerFirstElement` reproduced the regression (`GetSource<...>().Deconstruct(out var a, out var b); var (value, value2) = a; int value3 = b;`) and now pins `var ((value, value2), value3) = ...`.

**4. Redundant re-matching** — `IsRootDeconstructionCopy` now takes `(Block, int pos)` and returns the copied value and the matched `DeconstructionCall` via out parameters; both call sites use them. The dead `|| testedOperand == null` check is gone.

**5. Byref temp accepted as a value copy** — rejected explicitly via `copy.StackType == StackType.Ref`.

**6. Duplicated `AddressOfLdLocToLdLoca` branches** — merged into one `MatchAddressOf` with a `switch` on the target, sharing the range-transfer and `ReplaceWith`/`EndStep` tail; the new type check lives in that single place.

PR #3974, opened 2026-08-09
Extend UseLambdaSyntax to emit statement lambdas

siegfriedpammer:
With `UseLambdaSyntax` enabled, anonymous functions were emitted as lambdas only when an expression body was possible; every statement-bodied anonymous function still decompiled to C# 2 `delegate` syntax. This PR extends the setting to emit statement lambdas: any anonymous function whose parameter shape a lambda can express now uses lambda syntax. `delegate` syntax remains for `ref`/`out`/`in` and `params` parameters, and for pre-C# 3 language profiles (`SetLanguageVersion` still disables the setting there).

Two latent issues surfaced by the wider lambda coverage and are fixed here:

1. `DeclareVariables.ResolveCollisions` stopped its insertion-point walk at any node whose parent is a `LambdaExpression`, assuming an expression body that declaration insertion then converts to a block. A statement-bodied lambda's own block now reaches that path and tripped a `Debug.Assert`; the stop now applies only to actual expression bodies.
2. Anonymous methods declared without a parameter list carry compiler-generated parameter names (`<p0>`, `<sender>`) that are not valid C# identifiers. The old output hid them by omitting the parameter list; a lambda cannot, so `AssignVariableNames` now treats an invalid metadata name like a missing one and regenerates it from the parameter type: `(object obj, EventArgs e) => {...}`.

A visible side effect captured in the fixtures: an explicit parameter list can make a delegate-creation cast redundant that bare `delegate` needed for overload resolution - `new Thread((ThreadStart)delegate { })` becomes `new Thread(() => { })`.

Verification: full decompiler suite green (3317 passed). Differential check with decompdiff (master vs. this branch, fresh Release builds) over 12 real-world assemblies / 2341 types - Newtonsoft.Json 9, EntityFramework 6.2, Autofac, Castle.Core, Moq, NLog, log4net, Dapper, RestSharp, SharpZipLib, Mono.Cecil, NUnit 3.5: 68 types changed, every changed line is the delegate-to-lambda rewrite (parameter names from metadata are preserved), 0 new errors, and no deltas in lines, gotos, `//IL_` warnings, or compiler-generated name leaks.

---

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
@@ -2550,8 +2550,9 @@ namespace ICSharpCode.Decompiler.CSharp
 			}
 			else if (settings.UseLambdaSyntax && ame.Parameters.All(p => p.ParameterModifier == ReferenceKind.None && !p.IsParams))
 			{
-				// otherwise use lambda only if an expression lambda is possible
-				isLambda = (body.Statements.Count == 1 && body.Statements.Single() is ReturnStatement);
+				// Lambdas cover statement bodies too; anonymous method syntax remains only for
+				// parameter shapes a lambda cannot express (ref/out/in and params modifiers).
+				isLambda = true;
 			}
 			// Remove the parameter list from an AnonymousMethodExpression if the parameters are not used in the method body
 			var parameterReferencingIdentifiers =
--- a/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
@@ -2537,6 +2537,12 @@ namespace ICSharpCode.Decompiler.CSharp
 				attributeSections.Add(new AttributeSection(astBuilder.ConvertAttribute(attr)) { AttributeTarget = "return" });
 			}
 
+			bool parametersAreUsed = (
+				from ident in body.Descendants.OfType<IdentifierExpression>()
+				let v = ident.GetILVariable()
+				where v != null && v.Function == function && v.Kind == VariableKind.Parameter
+				select ident).Any();
+
 			bool isLambda = false;
 			if (ame.Parameters.Any(p => p.Type is null))
 			{
@@ -2548,19 +2554,19 @@ namespace ICSharpCode.Decompiler.CSharp
 				// C# 10 lambdas can have attributes, but anonymous methods cannot
 				isLambda = true;
 			}
-			else if (settings.UseLambdaSyntax && ame.Parameters.All(p => p.ParameterModifier == ReferenceKind.None && !p.IsParams))
+			else if (settings.UseLambdaSyntax && ame.Parameters.All(p => p.ParameterModifier == ReferenceKind.None && !p.IsParams)
+				&& (parametersAreUsed || (ParameterTypesAreAccessible(function) && ParametersAreNameable(function))))
 			{
-				// Lambdas cover statement bodies too; anonymous method syntax remains only for
-				// parameter shapes a lambda cannot express (ref/out/in and params modifiers).
+				// Lambdas cover statement bodies too. Anonymous method syntax remains where
+				// dropping the parameter list is the better rendering: for ref/out/in and
+				// params parameters (expressible in an explicitly typed lambda list, but
+				// conservatively left alone), and for unused parameters that a list would have
+				// to name or type from nothing to keep. The parameter-list-less "delegate {}"
+				// form is compatible with any delegate signature, so it is always legal there.
 				isLambda = true;
 			}
 			// Remove the parameter list from an AnonymousMethodExpression if the parameters are not used in the method body
-			var parameterReferencingIdentifiers =
-				from ident in body.Descendants.OfType<IdentifierExpression>()
-				let v = ident.GetILVariable()
-				where v != null && v.Function == function && v.Kind == VariableKind.Parameter
-				select ident;
-			if (!isLambda && !parameterReferencingIdentifiers.Any())
+			if (!isLambda && !parametersAreUsed)
 			{
 				ame.Parameters.Clear();
 			}
@@ -2606,6 +2612,47 @@ namespace ICSharpCode.Decompiler.CSharp
 			TranslatedExpression translatedLambda = replacement.WithILInstruction(function).WithRR(rr);
 			return new CastExpression(ConvertType(delegateType), translatedLambda)
 				.WithRR(new ConversionResolveResult(delegateType, rr, LambdaConversion.Instance));
+		}
+
+		/// <summary>
+		/// True when every parameter carries a name that can be written out as-is. Unused
+		/// parameters whose metadata names are missing or not identifiers (ilasm's synthetic
+		/// A_0/A_1 for unnamed Param rows, "&lt;p0&gt;" from an anonymous method declared without a
+		/// parameter list, obfuscated names) would have to be invented for a lambda's mandatory
+		/// parameter list; "delegate {}" drops the list instead of presenting a made-up name as
+		/// if it came from the source.
+		/// </summary>
+		static bool ParametersAreNameable(ILFunction function)
+		{
+			return function.Parameters.All(
+				p => !string.IsNullOrWhiteSpace(p.Name) && AssignVariableNames.IsValidName(p.Name));
+		}
+
+		bool ParameterTypesAreAccessible(ILFunction function)
+		{
+			var currentTypeDefinition = resolver.CurrentTypeDefinition;
+			if (currentTypeDefinition == null)
+				return true;
+			var lookup = new MemberLookup(currentTypeDefinition, currentTypeDefinition.ParentModule);
+			return function.Parameters.All(p => IsAccessible(p.Type));
+
+			bool IsAccessible(IType type)
+			{
+				switch (type)
+				{
+					case ParameterizedType pt:
+						return IsAccessible(pt.GenericType) && pt.TypeArguments.All(IsAccessible);
+					case TypeWithElementType t:
+						return IsAccessible(t.ElementType);
+					default:
+						for (var td = type.GetDefinition(); td != null; td = td.DeclaringTypeDefinition)
+						{
+							if (!lookup.IsAccessible(td, allowProtectedAccess: true))
+								return false;
+						}
+						return true;
+				}
+			}
 		}
 
 		protected internal override TranslatedExpression VisitILFunction(ILFunction function, TranslationContext context)
--- a/ICSharpCode.Decompiler/CSharp/Transforms/DeclareVariables.cs
+++ b/ICSharpCode.Decompiler/CSharp/Transforms/DeclareVariables.cs
@@ -472,8 +472,11 @@ namespace ICSharpCode.Decompiler.CSharp.Transforms
 			{
 				// We can only insert variable declarations in blocks, but FindInsertionPoints() didn't
 				// guarantee that it finds only blocks.
-				// Fix that up now.
-				while (!(v.InsertionPoint.nextNode.Parent is BlockStatement or LambdaExpression))
+				// Fix that up now. A lambda is a valid stop only for its expression body (insertion
+				// will convert that body to a block); a point at a statement-bodied lambda's block
+				// itself must keep moving up into the enclosing scope.
+				while (!(v.InsertionPoint.nextNode.Parent is BlockStatement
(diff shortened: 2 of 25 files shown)


ICSharpCode.Decompiler/CSharp/Transforms/DeclareVariables.cs
```diff
@@ -472,8 +472,11 @@ void ResolveCollisions()
 			{
 				// We can only insert variable declarations in blocks, but FindInsertionPoints() didn't
 				// guarantee that it finds only blocks.
-				// Fix that up now.
-				while (!(v.InsertionPoint.nextNode.Parent is BlockStatement or LambdaExpression))
+				// Fix that up now. A lambda is a valid stop only for its expression body (insertion
+				// will convert that body to a block); a point at a statement-bodied lambda's block
+				// itself must keep moving up into the enclosing scope.
+				while (!(v.InsertionPoint.nextNode.Parent is BlockStatement
+					|| (v.InsertionPoint.nextNode.Parent is LambdaExpression && v.InsertionPoint.nextNode is Expression)))
```
christophwille, line 479: **Correctness (confirmed):** when the expression-lambda body preserved by this stop condition is itself the matching assignment, the `CombineDeclarationAndInitializer` path replaces the lambda's body `Expression` with a `VariableDeclarationStatement` without first converting the body to a block, emitting invalid C# such as `x => int num = x;`.

Repro shape: IL `stloc num(ldarg x); ldloc num; ret` inside a delegate (a single `return num = x;`, with `num` a non-captured local of the lambda) decompiles to the expression lambda `x => num = x`, whose body node is the `AssignmentExpression`. `IsMatchingAssignment` then matches the body node itself and the replacement puts a statement in expression position.

This was already reachable on master for single-return expression lambdas, but this PR touches exactly this stop condition and the new comment documents the body-to-block conversion assumption, so it seems worth guarding here (e.g. do not stop at a lambda when the insertion point is the matching assignment itself, or convert the body to a block before combining).

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -2548,18 +2554,18 @@ internal ExpressionWithResolveResult TranslateFunction(IType delegateType, ILFun
...
+				&& (parametersAreUsed || ParameterTypesAreAccessible(function)))
 			{
-				// otherwise use lambda only if an expression lambda is possible
-				isLambda = (body.Statements.Count == 1 && body.Statements.Single() is ReturnStatement);
+				// Lambdas cover statement bodies too; anonymous method syntax remains for
+				// parameter shapes a lambda cannot express (ref/out/in and params modifiers)
+				// and for unused parameters whose types the current context cannot name:
+				// only the parameter-list-less "delegate {}" form, which is compatible with
+				// any delegate signature, can legally occur in such code.
+				isLambda = true;
```
christophwille, line 2566: **Output regression (confirmed):** making `isLambda` unconditional here removes the parameterless `delegate { }` form for anonymous methods whose parameters are unused but whose types are accessible, so machine-generated-but-valid parameter names now leak into output. The `IsValidName` guard in `AssignVariableNames` only catches syntactically invalid names, and the accessibility guard on this branch only restores `delegate {}` when a parameter type cannot be named.

This PR's own Issue1038 expectation shows the effect: `delegate { }` becomes `(TK A_0, TR A_1) => {}`, surfacing ilasm's synthetic `A_0`/`A_1` names (from unnamed Param rows) as if they were source-level names. Obfuscated assemblies get the same treatment for every unused-parameter anonymous method.

Consider also keeping `delegate {}` (or regenerating names) when all parameters are unused and their metadata names are absent or generated.

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -2548,18 +2554,18 @@ internal ExpressionWithResolveResult TranslateFunction(IType delegateType, ILFun
...
 				isLambda = true;
 			}
-			else if (settings.UseLambdaSyntax && ame.Parameters.All(p => p.ParameterModifier == ReferenceKind.None && !p.IsParams))
+			else if (settings.UseLambdaSyntax && ame.Parameters.All(p => p.ParameterModifier == ReferenceKind.None && !p.IsParams)
+				&& (parametersAreUsed || ParameterTypesAreAccessible(function)))
 			{
-				// otherwise use lambda only if an expression lambda is possible
-				isLambda = (body.Statements.Count == 1 && body.Statements.Single() is ReturnStatement);
+				// Lambdas cover statement bodies too; anonymous method syntax remains for
+				// parameter shapes a lambda cannot express (ref/out/in and params modifiers)
```
christophwille: **Comment accuracy:** explicitly-typed lambda parameters have supported `ref`/`out` modifiers since C# 3 (`(ref int x) => { ... }` is valid) and `in` since C# 7.2; only the omitted parameter list and `params` (before C# 13) are genuinely delegate-only. Keeping delegate syntax for these modifiers is a fine conservative choice, but the "shapes a lambda cannot express" justification is factually wrong for `ref`/`out`/`in` and could mislead a future change.

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -2548,18 +2554,18 @@ internal ExpressionWithResolveResult TranslateFunction(IType delegateType, ILFun
 				// C# 10 lambdas can have attributes, but anonymous methods cannot
 				isLambda = true;
 			}
-			else if (settings.UseLambdaSyntax && ame.Parameters.All(p => p.ParameterModifier == ReferenceKind.None && !p.IsParams))
+			else if (settings.UseLambdaSyntax && ame.Parameters.All(p => p.ParameterModifier == ReferenceKind.None && !p.IsParams)
```
christophwille, line 2557: **Test coverage:** after this PR no fixture exercises the retained delegate-syntax path for `ref`/`out`/`in`/`params` parameters - every `delegate` occurrence in TestCases was converted to lambda syntax, and no fixture contains an anonymous method with a modifier-carrying parameter list. (The new `InaccessibleParameterTypes` fixture covers the `delegate {}` / `Parameters.Clear()` fallback, but the modifier condition of this branch is untested.) A small Pretty fixture with e.g. a delegate type taking a `ref` parameter would pin the behavior this branch claims to preserve.

ICSharpCode.Decompiler/IL/Transforms/AssignVariableNames.cs
```diff
@@ -234,6 +234,15 @@ static bool IsSetOrEventAccessor(IMethod method)
 							if (variables.TryGetValue(i, out var v))
 								variableMapping[v] = name;
 						}
+						else if (!IsValidName(name))
```
christophwille, line 237: **Test coverage:** this new branch is exercised by no fixture. The only fixture that produced `<p0>`-style names, `Bug971_DelegateWithoutParameterList`, was rewritten in this same PR from `delegate { }` to `(object obj) => { }`, so no test now compiles an anonymous method declared without a parameter list against a parameterful delegate. The new `InaccessibleParameterTypes.il` does not cover it either: its anonymous-method parameters have empty names, which take the `variables[i].Name` fallback above instead of reaching this `IsValidName` check. An ILPretty fixture with a literally `'<p0>'`-named parameter would keep this branch pinned.

ICSharpCode.Decompiler/IL/Transforms/AssignVariableNames.cs
```diff
@@ -234,6 +234,15 @@ static bool IsSetOrEventAccessor(IMethod method)
...
+						{
+							// Compiler-generated parameter names (e.g. "<p0>" on an anonymous method
+							// declared without a parameter list) are not valid C# identifiers; generate
+							// a fresh name from the parameter type instead of escaping the raw name.
+							if (variables.TryGetValue(i, out var invalidlyNamed))
+								invalidlyNamed.HasGeneratedName = true;
```
christophwille: **Simplification:** setting `HasGeneratedName = true` here is redundant - `AssignName` already regenerates any name that fails `IsValidName` (it checks `v.HasGeneratedName || !IsValidName(newName)`), so the load-bearing part of this branch is the `continue` that skips name reservation/mapping. Dropping the assignment and keeping only the skip would express the fix exactly; as written, a reader may assume the flag is what triggers regeneration and simplify away the `continue`.

christophwille (commented):
**Automated high-effort review** (Claude Code, multi-agent find + adversarial-verify) of head `0116f24`.

The direction looks good, and the head commit ("Keep 'delegate {}' when a parameter type is inaccessible") already resolves one issue the review had flagged against the earlier state: lambda expansion forcing unnameable delegate parameter types into an emitted parameter list. Six findings remain, posted inline:

1. **Correctness (confirmed):** `DeclareVariables.ResolveCollisions` can emit invalid C# like `x => int num = x;` when the expression-lambda body preserved by the new stop condition is itself the matching assignment. Pre-existing on master, but this PR touches exactly that stop condition.
2. **Output regression (confirmed):** unused-parameter anonymous methods no longer collapse to `delegate {}` when the parameter types are accessible, so synthetic-but-valid parameter names (ilasm `A_0`/`A_1`, obfuscators) now leak into output - visible in this PR's own Issue1038 expectation change.
3. **Comment accuracy:** `ref`/`out`/`in` are expressible in explicitly-typed lambda parameter lists; the comment's "shapes a lambda cannot express" rationale is wrong for those modifiers.
4. **Test coverage:** the new invalid-name (`<p0>`) branch in `AssignVariableNames` is exercised by no fixture after the Bug971 rewrite (the new `InaccessibleParameterTypes.il` uses empty parameter names, which take the `variables[i].Name` fallback instead).
5. **Test coverage:** the retained delegate-syntax fallback for `ref`/`out`/`in`/`params` parameters has no fixture.
6. **Simplification:** `HasGeneratedName = true` in the invalid-name branch is redundant; the `continue` is the load-bearing part.

siegfriedpammer:
All six findings are addressed in the force-pushed branch (rebased on master, fixes squashed into the commits they belong to). Full decompiler suite green locally (3371 tests) and CI is green on all platforms.

**1. `x => int num = x;`** — fixed in `CombineDeclarationAndInitializer`: it now declines when the insertion point's parent is a `LambdaExpression`, so the separate-declaration path handles it, and that path already converts the expression body to a block. Output becomes `(int x) => { int num; return num = x; }`. New ILPretty fixture `AnonymousMethodEdgeCases.AssignmentIsTheLambdaBody` pins it - verified red before the fix with exactly the reported shape (`return (int x) => int num = x;`).

**2. Synthetic parameter names leaking** — `ParametersAreNameable` now gates the lambda form: when no parameter is used and any metadata name is missing or not a valid identifier, `delegate {}` is kept. `Bug971_DelegateWithoutParameterList` is restored to `delegate { }` in the fixture, which also restores the parameter-list-less input the suite had lost.

One honest limitation: `Issue1038`'s `A_0`/`A_1` are real names in the metadata (ilasm writes them for the empty parameter names in the .il), not absent ones, so nothing distinguishes them from source-level names without guessing at name shapes. That case still renders as `(TK A_0, TR A_1) => {}`. The absent/invalid half of the finding is fixed; the "generated-looking but valid" half is left alone deliberately.

**3. Comment accuracy** — rewritten. It no longer claims `ref`/`out`/`in` are inexpressible in a lambda; it now says the delegate form is kept where dropping the parameter list is the better rendering, with the modifiers called out as a deliberately conservative choice.

**4. `<p0>` branch untested** — new ILPretty fixture `AnonymousMethodEdgeCases.UsedParameterWithInvalidName`: a literally `'<p0>'`-named parameter that IS used, so the name has to be regenerated (`(object value) => ...`). A C#-authored fixture cannot reach this branch, since a parameter-list-less anonymous method can never use its parameters.

**5. ref/out/in/params fallback untested** — new Pretty fixture `AnonymousMethodWithByRefParameters` with `RefAction`/`OutAction` delegate types, pinning the retained `delegate (ref int reference) { ... }` form. `params` is not covered: C# rejects it in an anonymous-method parameter list (CS1670), so only hand-written IL could exercise that half.

**6. Redundant `HasGeneratedName = true`** — dropped; only the `continue` remains, with a comment naming `AssignName` as what regenerates the name.

Worth recording from the CI round-trip: restoring `delegate { }` in Bug971 broke the **legacy csc** configurations, which are Windows-only. Legacy csc names the parameter of a parameter-list-less anonymous method `param0` - a perfectly valid identifier that therefore becomes a lambda - while Roslyn emits `<p0>`. The fixture member is now guarded with `#if ROSLYN`.

PR #3975, opened 2026-08-09
Fix #3624: decompile semi-auto properties with the C# 14 field keyword

### LINKED ISSUE #3624: `field` keyword in auto properties
**Feature request**

https://github.com/dotnet/csharplang/issues/8635

**Additional context**

* I think this depends on https://github.com/icsharpcode/ILSpy/pull/3588.
* I'm willing to work on it.

siegfriedpammer:
Fixes #3624 (also part of #829).

Property accessors that reference their compiler-generated backing field now decompile using the C# 14 `field` keyword, and the backing-field declaration disappears. Compiler-generated trivial accessors collapse per accessor, so mixed shapes like `{ get; set { if (field != value) { field = value; ... } } }` come out naturally; `[field:]` attribute targets, property initializers (constructor stores lift to `} = value;`, and setter-less properties keep `P = x;` assignments in constructors), statics, init accessors, readonly struct accessors, generics, and lambdas/local functions reading `field` are all covered by the new `FieldKeyword` pretty fixture (compiled with Roslyn 5.9 at `-langversion:latest`).

Design notes:
- Emission happens at IL-to-AST translation time in `ExpressionBuilder.ConvertField`: inside the owning accessor the backing field is printed as bare `field` (never qualified — `this.field` would mean a real member). Recognition deliberately does **not** mirror the ILAst analysis used for automatic events: event recognition must prove compiler-generated bodies before discarding them, while the field keyword discards nothing, so name association plus accessor context suffices. An `AutoPropertyDecompiler` consolidation (records dedup, exact UI-tree hiding) remains possible as a follow-up without redesign.
- New `FieldKeyword` setting (C# 14.0, default on, via the `[DecompilerSetting]` generator). With it off — or below C# 14 — the field declaration survives **under its metadata name**, so the UI keeps showing the truth; `EscapeInvalidIdentifiers` (the transform the compilable-output flows — project export, VS, tests — already add) now maps `<P>k__BackingField` to the readable `P__BackingField` instead of the generic character escape. Exercised by a new `NoFieldKeyword` ugly test.
- A genuine field literally named `field` is qualified as `this.field` inside accessors, and locals are no longer named `field` there, because C# 14 rebinds the bare identifier (this fixed a real self-decompilation bug in ILSpy's own `TransformDisplayClassUsage`).
- Two independent fixes surfaced by the feature: referenced members are enqueued by definition in `DoDecompile(ITypeDefinition)`'s work list (in generic types the re-added declaration was silently dropped), and bodiless accessors mixed into multi-line properties now get their own line instead of `get; set {`.

Full ICSharpCode.Decompiler.Tests sweep on the branch: 3367 tests, 0 failed, 46 skipped (Windows-only/submodule-gated).

This PR was authored by an AI agent (Claude Code) operated by @siegfriedpammer.

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
+++ b/ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
@@ -458,8 +458,14 @@ namespace ICSharpCode.Decompiler.CSharp
 							return true;
 						if (settings.UsePrimaryConstructorSyntaxForNonRecordTypes && IsPrimaryConstructorParameterBackingField(field, metadata))
 							return true;
-						if (settings.AutomaticProperties && module.PropertyAndEventBackingFieldLookup.IsPropertyBackingField(fieldHandle, out var propertyHandle))
+						if ((settings.AutomaticProperties || settings.FieldKeyword)
+							&& module.PropertyAndEventBackingFieldLookup.IsPropertyBackingField(fieldHandle, out var propertyHandle))
 						{
+							// GetterOnlyAutomaticProperties exists so output stays compilable on
+							// toolchains that predate C# 6 getter-only auto-properties. Switching it off
+							// is a stronger statement than leaving FieldKeyword at its default, and it
+							// wins: accessors needing the C# 14 field keyword would not compile on such
+							// a toolchain either.
 							if (!settings.GetterOnlyAutomaticProperties)
 							{
 								PropertyAccessors accessors = metadata.GetPropertyDefinition(propertyHandle).GetAccessors();
--- a/ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
+++ b/ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
@@ -2030,7 +2030,11 @@ namespace ICSharpCode.Decompiler.CSharp
 							&& mrr.Member.DeclaringTypeDefinition == typeDef
 							&& !(mrr.Member is IMethod { IsLocalFunction: true }))
 						{
-							workList.Enqueue(mrr.Member);
+							// In generic types the reference is to a member specialized by the type's
+							// own type parameters, but entityMap and the dequeue dedupe are keyed by
+							// the definition; enqueueing the specialized member would decompile the
+							// member under a key the output pass never looks up.
+							workList.Enqueue(mrr.Member.MemberDefinition);
 						}
 						else if (rr is TypeResolveResult trr
 							&& trr.Type.GetDefinition()?.DeclaringTypeDefinition == typeDef)
--- a/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
@@ -299,6 +299,21 @@ namespace ICSharpCode.Decompiler.CSharp
 			return !(target.Expression is ThisReferenceExpression || target.Expression is BaseReferenceExpression);
 		}
 
+		/// <summary>
+		/// True when the field access has no target at all (already-collapsed `this` access) or
+		/// loads it from `this`, directly or through the address of `this` that a struct
+		/// accessor uses.
+		/// </summary>
+		static bool TargetIsThis(ILInstruction? targetInstruction)
+		{
+			return targetInstruction switch {
+				null => true,
+				var inst when inst.MatchLdThis() => true,
+				LdLoca { Variable.Kind: VariableKind.Parameter, Variable.Index: < 0 } => true,
+				_ => false,
+			};
+		}
+
 		ExpressionWithResolveResult ConvertField(IField field, ILInstruction? targetInstruction = null)
 		{
 			if (settings.AutomaticEvents && IsBackingFieldOfAutomaticEvent(field, out var ev))
@@ -317,6 +332,29 @@ namespace ICSharpCode.Decompiler.CSharp
 				return eventReference.WithRR(eventResolveResult);
 			}
 
+			if (settings.FieldKeyword
+				&& decompilationContext.CurrentMember is IProperty accessedProperty
+				&& accessedProperty.Parameters.Count == 0
+				// Ask exactly the question PatternStatementTransform asks when it decides whether
+				// the field declaration can go away. A looser test here prints `field` inside a
+				// property whose declaration then keeps explicit accessors and its field: on
+				// recompile the keyword binds to a freshly synthesized backing field while the
+				// original one stays declared and unwritten - silently different storage.
+				&& PatternStatementTransform.TryGetBackingField(accessedProperty, out var backingField)
+				&& field.MemberDefinition.Equals(backingField.MemberDefinition)
+				// Only THIS instance's field is the `field` keyword. IL can load another
+				// instance's backing field inside an accessor (weavers, obfuscators, hand-written
+				// IL); rendering that as `field` would redirect the access, and drop whatever
+				// side effect producing the target had.
+				&& (field.IsStatic || TargetIsThis(targetInstruction)))
+			{
+				// Inside its own property's get/set/init accessor (including nested lambdas and
+				// local functions), the backing field is the C# 14 "field" keyword. It must stay
+				// unqualified: "this.field" would refer to a real member named "field".
+				return new IdentifierExpression("field")
+					.WithRR(new MemberResolveResult(null, field));
+			}
+
 			var target = TranslateTarget(targetInstruction,
 				nonVirtualInvocation: true,
 				memberStatic: field.IsStatic,
@@ -332,6 +370,13 @@ namespace ICSharpCode.Decompiler.CSharp
 				&& (property.CanSet || settings.GetterOnlyAutomaticProperties))
 			{
 				requireTarget = RequiresQualifier(property, target);
+			}
+			else if (settings.FieldKeyword && field.Name == "field"
+				&& decompilationContext.CurrentMember is IProperty { Parameters.Count: 0 })
+			{
+				// In a C# 14 property accessor a bare "field" identifier binds to the backing
+				// field keyword, so a genuine field of that name needs a qualifier.
+				requireTarget = true;
 			}
 			else
 			{
--- a/ICSharpCode.Decompiler/CSharp/Transforms/EscapeInvalidIdentifiers.cs
+++ b/ICSharpCode.Decompiler/CSharp/Transforms/EscapeInvalidIdentifiers.cs
@@ -18,6 +18,7 @@
 
 #nullable enable
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 
@@ -44,8 +45,24 @@ namespace ICSharpCode.Decompiler.CSharp.Transforms
 			return false;
 		}
 
-		string ReplaceInvalid(string s)
+		string ReplaceInvalid(string s, ISet<string>? existingNames = null)
 		{
+			// A property backing field that stayed declared (the property is not an
+			// auto-property and the "field" keyword is unavailable or insufficient) gets a
+			// readable name instead of the generic escape of "<P>k__BackingField". The mapping
+			// is name-deterministic like the escape itself, so declarations and references
+			// stay consistent without symbol tracking.
+			if (s.StartsWith("<", System.StringComparison.Ordinal)
+				&& PatternStatementTransform.NameCouldBeBackingFieldOfAutomaticProperty(s, out string? propertyName))
+			{
+				var readable = ReplaceInvalid(propertyName) + "__BackingField";
+				// "P__BackingField" is a plausible identifier a human could have written, unlike
+				// the generic escape it replaces. If the tree already contains that name, the
+				// rename would produce two members with one name; fall back to the escape, which
+				// keeps the output compilable at the cost of readability.
+				if (existingNames?.Contains(readable) != true)
+					return readable;
+			}
 			string name = string.Concat(s.Select(ch => IsValid(ch) ? ch.ToString() : string.Format("_{0:X4}", (int)ch)));
 			if (name.Length >= 1 && !(char.IsLetter(name[0]) || name[0] == '_'))
 				name = "_" + name;
@@ -54,9 +71,13 @@ namespace ICSharpCode.Decompiler.CSharp.Transforms
 
 		public void Run(AstNode rootNode, TransformContext context)
 		{
+			// Every name already in the tree, so the readable backing-field rename can detect
+			// that its target name is taken.
+			var existingNames = new HashSet<string>(
+				rootNode.DescendantsAndSelf.OfType<Identifier>().Select(i => i.Name), StringComparer.Ordinal);
 			foreach (var ident in rootNode.DescendantsAndSelf.OfType<Identifier>())
 			{
-				string newName = ReplaceInvalid(ident.Name);
+				string newName = ReplaceInvalid(ident.Name, existingNames);
 				if (newName != ident.Name)
 				{
 					context.Step($"Escape identifier '{ident.Name}'", ident);
--- a/ICSharpCode.Decompiler/CSharp/Transforms/PatternStatementTransform.cs
+++ b/ICSharpCode.Decompiler/CSharp/Transforms/PatternStatementTransform.cs
@@ -863,14 +863,19 @@ namespace ICSharpCode.Decompiler.CSharp.Transforms
 			var setter = propertyDeclaration.Setter;
 			if (context.Settings.AutomaticProperties)
 			{
-				CollapseTrivialAccessor(getter, trivialFieldGetterBody, field);
+				// VB auto-properties do not mark their accessors [CompilerGenerated]; the
+				// pre-C# 14 transform recognizes them by their "_<PropertyName>" backing field
+				// instead. Keep that rule, or a VB auto-property grows explicit "field"
+				// accessors where every other compiler's collapses to "{ get; set; }".
+				bool accessorsMustBeCompilerGenerated = field.Name != "_" + property.Name;
+				CollapseTrivialAccessor(getter, trivialFieldGetterBody, field, accessorsMustBeCompilerGenerated);
 				// A readonly setter cannot become an auto-accessor (same rule as the pre-C# 14
 				// transform); readonly getters collapse fine because auto-getters are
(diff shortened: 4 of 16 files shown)


ICSharpCode.Decompiler/CSharp/Transforms/PatternStatementTransform.cs
```diff
@@ -814,6 +819,168 @@ bool CanTransformToAutomaticProperty(IProperty property, bool accessorsMustBeCom
...
+			return false;
+		}
+
+		/// <summary>
+		/// The "field" keyword cannot express backing-field accesses outside the owning
+		/// property's accessors. The only shapes C# can express are stores in a constructor of
+		/// the declaring type (property initializers, or assignments to setter-less
+		/// properties); anything else means the field declaration has to be kept.
+		/// </summary>
+		bool OutsideReferencesAreExpressible(PropertyDeclaration propertyDeclaration, IProperty property, IField field)
```
christophwille: **Removed backing field can stay referenced in the ctor.** `OutsideReferencesAreExpressible` approves any plain ctor store, but `TransformFieldAndConstructorInitializers` (which runs later) can decline to lift that store, leaving a reference to the already-removed backing field.

Example: a class with an explicit static ctor (type not beforefieldinit) and a static semi-auto property with an initializer, e.g. `public static int P { get => field; set => field = value + 1; } = 5; static C() { ... }`. This transform sees the cctor store `<P>k__BackingField = 5` as expressible and deletes the field declaration, but TFCI's `Analyze` sets `onlyMoveConstants` (`AlwaysMoveInitializer` defaults false, `!IsBeforeFieldInit`, static) and skips the non-const store, so it stays in the cctor as `P__BackingField = 5;` referencing a field that no longer exists -- non-compilable output. The same hole exists for instance ctors whose leading assignment sequences don't match across multiple non-chained ctors, and for parameter-dependent initializers when the primary-constructor transform doesn't fire.

ICSharpCode.Decompiler/CSharp/Transforms/TransformFieldAndConstructorInitializers.cs
```diff
@@ -187,6 +187,17 @@ class InitializerSequence
 				return sequence;
 			}
 
+			internal static bool IsDefaultValueInitializer(Expression initializer)
```
christophwille, line 190: **`IsDefaultValueInitializer` drops observably non-default initializers.** The constant patterns (`0m`, `0.0`, `0f`) match with `==` semantics, so `0.00m` matches `0m` and `-0.0` matches `0.0`.

Example: a struct with `public decimal P { get => field; set => field = value; } = 0.00m;` (or `= -0.0` for double) -- the ctor store is classified as an implicit struct default initialization and dropped instead of becoming the property initializer. Recompiling yields `default(decimal)` = `0m`, so `P.ToString()` returns `"0"` instead of `"0.00"` (and `1.0/P` flips sign for `-0.0`) -- a silent behavioral change in round-tripped code.

christophwille (commented):
Automated high-effort review (Claude Code) of the C# 14 `field`-keyword support. Summary of findings; details are in the inline comments.

**Correctness (8 findings).** The core weakness: the property transform removes backing-field declarations based on optimistic assumptions that downstream transforms do not uphold.

- `OutsideReferencesAreExpressible` approves any plain ctor store, but `TransformFieldAndConstructorInitializers` can later decline to lift that store (explicit static ctors, non-matching multi-ctor assignment sequences), leaving the ctor referencing a removed field -- non-compiling output even for plain Roslyn-compiled C#.
- `IsDefaultValueInitializer` treats `0.00m` / `-0.0` struct property initializers as redundant defaults and silently drops them.
- `ReplaceBackingFieldUsage` rewrites ctor stores into property assignments even when the property transform bailed and kept the field declared (CS0200 on recompile).
- `ExpressionBuilder.ConvertField` prints the bare `field` keyword under a looser predicate than the transform uses (`IsBackingFieldOfAutomaticProperty` vs `TryGetBackingField`), stranding `field` keywords when the transform bails; it also never checks the target is `this`, and the field-type-vs-property-type check the old auto-property path guaranteed structurally is re-established nowhere.
- The `<P>k__BackingField` -> `P__BackingField` rename in `EscapeInvalidIdentifiers` is collision-blind against existing user members.
- `FieldKeyword` (default-on) silently bypasses an explicitly disabled `GetterOnlyAutomaticProperties` setting in `MemberIsHidden`.

**Performance / cleanup (2 findings).** `OutsideReferencesAreExpressible` walks the entire syntax tree once per property (quadratic; a real regression for whole-module output of property-heavy assemblies), and the constructor-store predicate is duplicated (inline + `IsConstructorStoreTarget`), which is the direct source of the transform-disagreement hazards above.

ICSharpCode.Decompiler/CSharp/Transforms/PatternStatementTransform.cs
```diff
@@ -894,11 +1061,27 @@ static bool NameCouldBeBackingFieldOfAutomaticProperty(string name, [NotNullWhen
...
-					&& CanTransformToAutomaticProperty(property, !(field.IsCompilerGenerated() && field.Name == "_" + property.Name))
 					&& currentMethod?.AccessorOwner != property)
 				{
-					if (!property.CanSet && !context.Settings.GetterOnlyAutomaticProperties)
+					if (CanTransformToAutomaticProperty(property, !(field.IsCompilerGenerated() && field.Name == "_" + property.Name)))
+					{
+						if (!property.CanSet && !context.Settings.GetterOnlyAutomaticProperties && !context.Settings.FieldKeyword)
+							return null;
+					}
+					else if (context.Settings.FieldKeyword && !property.CanSet && IsConstructorStoreTarget(parent, field))
```
christophwille: **Ctor store rewritten to a property assignment even when the property transform bailed.** `ReplaceBackingFieldUsage` rewrites constructor stores to setter-less field-backed properties without the old `CanTransformToAutomaticProperty` gate and without checking that `TransformFieldBackedProperty` actually succeeded (`OutsideReferencesAreExpressible`).

Example: a getter-only property with a non-trivial accessor body (uses `field`) whose backing field is also read from another method. The transform bails and keeps the field declared with explicit accessors, but the ctor's `<P>k__BackingField = x` store is still replaced with `P = x`. The emitted ctor assigns a read-only property that is not auto/field-backed in the output -- CS0200 on recompile.

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -317,6 +317,19 @@ ExpressionWithResolveResult ConvertField(IField field, ILInstruction? targetInst
 				return eventReference.WithRR(eventResolveResult);
 			}
 
+			if (settings.FieldKeyword
```
christophwille, line 335: Two related issues with the bare-`field` substitution here:

**1. Predicate mismatch with the cleanup transform.** The condition for printing the bare `field` keyword (`IsBackingFieldOfAutomaticProperty`: name match + CompilerGenerated, no static check) is looser than what `TransformFieldBackedProperty` relies on (`TryGetBackingField`: additionally requires `candidate.IsStatic == property.IsStatic`, and picks the first name-matching field). When `TryGetBackingField` fails or picks a different field, the transform returns null via a path that never renames the already-emitted `field` identifiers back. For IL where a property's accessors reference a static CG `<P>k__BackingField` (staticness mismatch), or where both a CG `_P` and a `<P>k__BackingField` exist and the accessors use `_P`: the accessor bodies print bare `field` but the property keeps explicit accessors, so on recompile `field` binds to a freshly synthesized backing field while the original field stays declared and unwritten -- silently different storage.

**2. Target expression discarded without checking it is `this`.** The branch returns before `TranslateTarget(targetInstruction, ...)` runs, so IL inside property P's own accessor that loads/stores *another instance's* backing field (IL weavers/obfuscators, hand-written IL) decompiles to bare `field`, which recompiles as access to this instance's backing field; if the target was a call (`GetOther().<P>k__BackingField`), its side effect is silently dropped.

ICSharpCode.Decompiler/CSharp/Transforms/PatternStatementTransform.cs
```diff
@@ -814,6 +819,168 @@ bool CanTransformToAutomaticProperty(IProperty property, bool accessorsMustBeCom
...
+				return;
+			}
+			RemoveCompilerGeneratedAttribute(accessor.Attributes);
+			// Auto-accessors are implicitly readonly.
+			accessor.Modifiers &= ~Modifiers.Readonly;
+			// Clearing the accessor body turns it into an auto-property accessor.
+			accessor.Body = null;
+		}
+
+		internal static bool TryGetBackingField(IProperty property, [NotNullWhen(true)] out IField? field)
```
christophwille, line 929: **No field-type-vs-property-type check.** The old auto-property path only removed a backing field whose trivial get/set bodies structurally guaranteed the field's type matched the property's. `TryGetBackingField` matches any compiler-generated field named `<P>k__BackingField`/`_P` with matching staticness but never checks that the field type equals the property type, and the arbitrary-body field-keyword path re-establishes that check nowhere.

Example: an assembly (VB-style `_Name` naming or obfuscated IL) with a string property `Name` backed by a CG int field `_Name` whose getter returns `_Name.ToString()`: the int field declaration is deleted and the accessor prints as `return field.ToString();`, where `field` now has type string. The output either recompiles with silently different behavior (fresh string backing field, original int value lost) or fails to compile where conversions do not line up.

ICSharpCode.Decompiler/CSharp/Transforms/EscapeInvalidIdentifiers.cs
```diff
@@ -46,6 +46,16 @@ bool IsValid(char ch)
 
 		string ReplaceInvalid(string s)
 		{
+			// A property backing field that stayed declared (the property is not an
+			// auto-property and the "field" keyword is unavailable or insufficient) gets a
+			// readable name instead of the generic escape of "<P>k__BackingField". The mapping
+			// is name-deterministic like the escape itself, so declarations and references
+			// stay consistent without symbol tracking.
+			if (s.StartsWith("<", System.StringComparison.Ordinal)
+				&& PatternStatementTransform.NameCouldBeBackingFieldOfAutomaticProperty(s, out string? propertyName))
```
christophwille, line 56: **Collision-blind rename.** The name-deterministic mapping `<P>k__BackingField` -> `P__BackingField` can collide with an existing user member. A type that declares both a kept backing field `<P>k__BackingField` (property not transformable) and a genuine field or property already named `P__BackingField` (a legal, human-plausible identifier, unlike the old `_003CP_003Ek__BackingField` escape) ends up with two members of the same name after renaming -- non-compilable output or references silently rebinding to the wrong member.

ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
```diff
@@ -458,9 +458,12 @@ public static bool MemberIsHidden(MetadataFile? module, EntityHandle member, Dec
...
 						if (settings.UsePrimaryConstructorSyntaxForNonRecordTypes && IsPrimaryConstructorParameterBackingField(field, metadata))
 							return true;
-						if (settings.AutomaticProperties && module.PropertyAndEventBackingFieldLookup.IsPropertyBackingField(fieldHandle, out var propertyHandle))
+						if ((settings.AutomaticProperties || settings.FieldKeyword)
+							&& module.PropertyAndEventBackingFieldLookup.IsPropertyBackingField(fieldHandle, out var propertyHandle))
 						{
-							if (!settings.GetterOnlyAutomaticProperties)
+							// With the field keyword, even a getter-only property can express
+							// arbitrary accessor bodies, so the backing field is always hidden.
+							if (!settings.FieldKeyword && !settings.GetterOnlyAutomaticProperties)
```
christophwille: **`FieldKeyword` bypasses an explicitly disabled `GetterOnlyAutomaticProperties`.** The carve-out that kept getter-only property backing fields visible is guarded with `!settings.FieldKeyword && !settings.GetterOnlyAutomaticProperties`, so with `FieldKeyword` at its default (on), unchecking GetterOnlyAutomaticProperties no longer has any effect: the backing field is still hidden and the property still collapses to `get;`. The decompiled output contradicts the chosen setting and no longer compiles on the older toolchain that setting exists to support.

ICSharpCode.Decompiler/CSharp/Transforms/PatternStatementTransform.cs
```diff
@@ -814,6 +819,168 @@ bool CanTransformToAutomaticProperty(IProperty property, bool accessorsMustBeCom
...
+		/// <summary>
+		/// The "field" keyword cannot express backing-field accesses outside the owning
+		/// property's accessors. The only shapes C# can express are stores in a constructor of
+		/// the declaring type (property initializers, or assignments to setter-less
+		/// properties); anything else means the field declaration has to be kept.
+		/// </summary>
+		bool OutsideReferencesAreExpressible(PropertyDeclaration propertyDeclaration, IProperty property, IField field)
+		{
+			var root = propertyDeclaration.Ancestors.LastOrDefault() ?? propertyDeclaration;
+			foreach (var node in root.Descendants)
```
christophwille, line 1019: **Quadratic full-tree walk, default-on.** `OutsideReferencesAreExpressible` scans the entire syntax tree (`root.Descendants`, resolving symbols per node) once per property. For whole-module decompilation (ilspycmd single-file output, 'save whole assembly') of a property-heavy assembly -- thousands of DTO/EF auto-properties in one tree -- every property triggers a full AST walk, turning a seconds-long decompile into minutes. This is a significant performance regression versus the pre-PR `AutomaticProperties` path, which never scanned outside the property.

ICSharpCode.Decompiler/CSharp/Transforms/PatternStatementTransform.cs
```diff
@@ -814,6 +819,168 @@ bool CanTransformToAutomaticProperty(IProperty property, bool accessorsMustBeCom
...
+				if (node is not (IdentifierExpression or MemberReferenceExpression))
+					continue;
+				if (node.GetSymbol() is not IField referencedField
+					|| !field.Equals(referencedField.MemberDefinition))
+				{
+					continue;
+				}
+				if (node.Ancestors.Contains(propertyDeclaration))
+					continue;
+				if (node.Parent is not AssignmentExpression { Operator: AssignmentOperatorType.Assign } assignment
```
christophwille: **Constructor-store predicate encoded twice.** This inline check (assignment-LHS + `IsConstructor` + `IsStatic` match + `DeclaringTypeDefinition` match) restates the same rule as `IsConstructorStoreTarget` (~line 1094), differing only in how the enclosing method is obtained (ancestor walk vs `currentMethod`). If one copy is later refined and the other is not, `OutsideReferencesAreExpressible` and `ReplaceBackingFieldUsage` disagree: the field declaration is removed while the store stays a field reference (or kept while the store is rewritten to a property assignment) -- output that does not compile. Consider sharing one helper taking the resolved enclosing method.

siegfriedpammer:
All nine findings are addressed in the force-pushed branch (rebased on master, fixes squashed into the implementation commit). Full decompiler suite green: 3375 tests, 0 failed.

**1. Removed backing field still referenced in the constructor** — `OutsideReferencesAreExpressible` now only approves a constructor store that something will actually consume: a setter-less property's store (rewritten to a property assignment by this same transform, no later transform involved), or a store `TransformFieldAndConstructorInitializers` will lift. The reported hole is closed for the explicit-static-constructor case, which is checked directly (`AlwaysMoveInitializer` off plus a non-`beforefieldinit` type, mirroring TFCI's `onlyMoveConstants`).

Not closed, and worth stating plainly: TFCI can still decline for reasons this transform does not predict - instance constructors whose leading assignment sequences do not match across several non-chained constructors, and parameter-dependent initializers when the primary-constructor transform does not fire. Predicting the rest would duplicate TFCI's analysis, and the transform order (PatternStatementTransform runs first, and by then the accessors are already collapsed) rules out deciding this after the fact. If you would rather have it airtight than pretty, the conservative option is to refuse removal whenever the declaring type has more than one non-chained instance constructor.

**2. `IsDefaultValueInitializer` dropping observable initializers** — now compares representations, not values: `BitConverter` bits for float/double and `decimal.GetBits` for the scale/sign word, so `-0.0` and `0.00m` are no longer mistaken for defaults. New fixture case `FieldKeyword.PreciseDefaults`, verified red before the fix.

**3. Constructor store rewritten although the property transform bailed** — the rewrite is now gated on `BackingFieldWillBeRemoved`, which asks the same two questions `TransformFieldBackedProperty` asks (`TryGetBackingField` plus the outside-reference verdict), so a property that keeps explicit setter-less accessors keeps its field store.

**4. `ConvertField` printing bare `field` too eagerly** — the branch now calls `TryGetBackingField` (the transform's own predicate) instead of the looser `IsBackingFieldOfAutomaticProperty`, and requires the access target to be `this` (or the field to be static), so another instance's backing field is no longer rendered as `field` and no target side effect is dropped.

**5. Field type vs property type** — `TryGetBackingField` now requires `candidate.Type.Equals(propertyType)`, comparing against the property DEFINITION's return type so a specialized property in a generic type still matches its own field.

**6. Collision-blind rename** — `EscapeInvalidIdentifiers` collects the names already present in the tree and falls back to the generic escape when `P__BackingField` is taken, so the readable rename can no longer produce two members with one name.

**7. `FieldKeyword` bypassing a disabled `GetterOnlyAutomaticProperties`** — the carve-out is now guarded on `GetterOnlyAutomaticProperties` alone. Switching it off is a statement about the target toolchain, and accessors needing the C# 14 field keyword would not compile there either, so the explicit setting wins.

**8. Quadratic full-tree walk** — replaced by one indexed pass per syntax tree (`BuildOutsideReferenceIndex`), consulted per property. Later transforms only ever remove references to the field of the property they rewrite, so a verdict computed up front stays valid for every other property.

**9. Duplicated constructor-store predicate** — both sides now call one `IsConstructorStore(node, field, enclosingMethod)`, which takes the enclosing method as a parameter since that was the only real difference.

On test coverage: findings 3, 4, 5 and 6 need shapes C# cannot express - a member outside the accessors referencing the backing field, a staticness or type mismatch between field and property, a name collision with `P__BackingField` - so they are not covered by a Pretty fixture. Say the word if you want ILPretty cases for them and I will add them.

Issue #3986, opened 2026-08-12
XML files are almost unreadable in dark mode


SylwesterZarebski:
### Steps to reproduce
1. Open ILSpy.
2. Set dark mode.
3. Open any library with XML files in embedded resources.
4. View XML file.

### Error message shown

<img width="1224" height="267" alt="Image" src="https://github.com/user-attachments/assets/3180d1a8-65c4-4b2b-b73b-90fcca206a1d" />

### Details
* Product in use: ILSpy
* Version in use: 11.0.0.9375

christophwille:
Is that Windows 10 by any chance?

Edit: For comparison, Windows 11 with resource from ILSpy itself: 

<img width="1108" height="588" alt="Image" src="https://github.com/user-attachments/assets/2190afd3-7739-4ed5-a48e-7b5e4bca6198" />

SylwesterZarebski:
Unfortunately no, Windows 11 here. Only changed options is mode and font to *Source Code Pro*.
Also other things, like highlighting some buttons or rows are the same as light theme.  
Here is highlighted left Decompiler tab button:
<img width="816" height="327" alt="Image" src="https://github.com/user-attachments/assets/d1fb3112-865b-4eea-a58d-e178a6ccce54" />

I've just reset Display options and set mode plus font to no avail.

christophwille:
That "highlighting box" (where the Decompiler options are), that one I see as well (text unreadable just like in your screenshot).

Interestingly when I changed the font size to 16, the decompiled text view did change the color closer to your screenshot - and from then on, it stays that way. Even when I delete ilspy.xml. Most curious.

<img width="740" height="327" alt="Image" src="https://github.com/user-attachments/assets/ea663055-de7f-46da-a3e2-7d1e6301783b" />

Edit: Switching to Light and then Dark restores the colors I sent in the first screenshot.

christophwille:
I know this is slightly strange - could you please try builds for PR #3987? Those would be at https://github.com/icsharpcode/ILSpy/actions/workflows/build-ilspy.yml?query=branch%3Awin-a11y-textsize I was able to repro the "decolorization/flattening of colors" with master but no longer with builds from that branch and I'd like to know if that is coincidence on my machine.

christophwille:
[...]

PR #3989, opened 2026-08-12
Prevent double dark conversion of syntax colors

christophwille:
## What

Fixes and hardens the `ThemeManager` / `ThemeAwareHighlightingColorizer` contract against double dark conversion of syntax colors — including a live, user-visible bug: **with dark theme preselected, the first document of a session rendered washed out** (resources across every token; C# on the tokens outside the hand-authored dark palette).

The two components split responsibility for dark mode: `ThemeManager` darkens a *registered* definition's named `HighlightingColor` instances in place (snapshot-based restore), while the colorizer per-paint-remaps colors of *unregistered* definitions. Any path that runs both mechanisms — or the in-place remap twice — on the same colors washes the palette out.

## Commit 1: check theme-awareness per paint, not at colorizer creation

`ThemeAwareHighlightingColorizer` captured `IsThemeAware(definition)` once, in its constructor, into a readonly field. A colorizer constructed *before* its definition's first `RegisterThemableDefinition` call that survives a *later* registration would keep remapping colors that `ThemeManager` had meanwhile darkened in place. Reachable today only by violating a calling convention (every current path registers before the colorizer exists), but nothing enforced that. Now the flag is read per paint, alongside the existing per-paint `IsDarkTheme` read.

The originally suspected companion hazard — the colorizer's `darkColors` cache serving conversions computed from superseded color values — turned out to be a non-issue: `HighlightingColor` overrides `Equals`/`GetHashCode` with content-based equality, so an in-place recolor changes the key's hash, misses the cache, and reconverts. A characterization test pins that, so an equality-semantics change in AvaloniaEdit surfaces as a red test rather than stale colors.

## Commit 2: fix washed-out colors when dark theme is active at startup

The startup bug had a different mechanism, on the `ThemeManager` side: **the same definition was registered under two identities.** AvaloniaEdit's `HighlightingManager` hands out a `DelayLoadedHighlightingDefinition` *wrapper* whose members all forward to the inner definition; ILSpy's `HighlightingService.Load` registers the *inner* definition during materialization, and `GetByExtension` then registers the *wrapper* it got from the manager. The wrapper registration snapshots the shared live colors **after** the inner registration has already themed them.

- Theme switched in-session: first touch happens in Light, both snapshots are pristine → no visible harm (just double bookkeeping). This is why the bug never reproduced by toggling.
- Dark active at first touch (dark preselected at startup): the wrapper's snapshot captures already-dark values as "light originals", and the rewrite dark-converts them a second time → washed out from the first paint.

The initial fix guarded `RegisterThemableDefinition` on the theme-aware marker; commit 3 (review feedback) replaces the load-bearing part of that guard with something structural.

## Commit 3 (review follow-up): snapshot original colors per color *instance*

Per review: the marker-based guard conflated "XSHD opts out" / "already themed" / "already registered", and leaned on two non-contractual AvaloniaEdit details (the wrapper's `Properties` forwarding, materialize-on-touch). The original-color snapshots are now keyed by `HighlightingColor` instance in a `ConditionalWeakTable` (reference identity — immune to the content-based `Equals`; weak keys die with the definition) instead of by definition. Theming is idempotent no matter how many definition objects expose the same colors or in which order they register, so correctness no longer depends on the marker or registration order. The guard remains only to honor the XSHD `ILSpy.IsThemeAware` opt-out and to skip redundant list entries, and `ApplyHighlightingColors` is now private so nothing can set the marker outside a registration.

## Tests

`ILSpy.Tests/Themes/DoubleDarkConversionTests.cs`, driving the paint decision through an internal seam (`GetEffectiveColor`) and `ThemeManager` through its public surface:

- `LazyLoadedDefinitionIsNotDarkenedTwiceAtDarkStartup` — replicates the delay-loaded-wrapper materialization flow with dark active before first touch; red before the startup fix, green after.
- `ThemeSwitchRestoresPristineColorsAfterDarkStartup` — same dark-startup flow, then Light→Dark round-trip; pins that every switch is computed from the pristine snapshot (Light restores the .xshd values exactly, Dark yields the single conversion).
- `RegistrationAfterConstructionDisablesPerPaintRemap` — red before commit 1 (constructor-cached flag ignored late registration), green after.
- `DarkConversionCacheMissesAfterInPlaceRecolor` — characterization guard for the content-keyed conversion cache: after an in-place recolor, the served conversion must equal the conversion of the *new* values (exists to catch an equality-semantics change in AvaloniaEdit).

Full headless `ILSpy.Tests` suite green (1167 passed, 4 pre-existing environment skips). Verified in the running app: with `<Theme>Dark</Theme>` preselected in the config, the first decompiled document of the session renders with the same vibrant dark palette as an in-session switch.

### DIFF
--- a/ILSpy.Tests/ThemeAwareHighlightingColorizerTests.cs
+++ b/ILSpy.Tests/ThemeAwareHighlightingColorizerTests.cs
@@ -64,6 +64,39 @@ public class ThemeAwareHighlightingColorizerTests
 		public HighlightingColor GetNamedColor(string name) => colors.Find(c => c.Name == name)!;
 	}
 
+	// Mimics AvaloniaEdit's DelayLoadedHighlightingDefinition: every member forwards through
+	// materialization, and materializing runs a load callback that (like
+	// HighlightingService.Load) registers the INNER definition with the theme manager. The
+	// wrapper and the inner definition are distinct objects sharing the same live colours
+	// and Properties dictionary.
+	sealed class LazyLoadedDefinition : IHighlightingDefinition
+	{
+		readonly StubHighlightingDefinition inner;
+		bool materialized;
+
+		public LazyLoadedDefinition(StubHighlightingDefinition inner)
+		{
+			this.inner = inner;
+		}
+
+		StubHighlightingDefinition GetDefinition()
+		{
+			if (!materialized)
+			{
+				materialized = true;
+				ThemeManager.Current.RegisterThemableDefinition(inner);
+			}
+			return inner;
+		}
+
+		public string Name => GetDefinition().Name;
+		public HighlightingRuleSet MainRuleSet => GetDefinition().MainRuleSet;
+		public IEnumerable<HighlightingColor> NamedHighlightingColors => GetDefinition().NamedHighlightingColors;
+		public IDictionary<string, string> Properties => GetDefinition().Properties;
+		public HighlightingRuleSet GetNamedRuleSet(string name) => GetDefinition().GetNamedRuleSet(name);
+		public HighlightingColor GetNamedColor(string name) => GetDefinition().GetNamedColor(name);
+	}
+
 	static HighlightingColor MakeColor(string name, Color foreground)
 		=> new() { Name = name, Foreground = new SimpleHighlightingBrush(foreground) };
 
@@ -74,6 +107,37 @@ public class ThemeAwareHighlightingColorizerTests
 		var settings = new SessionSettings();
 		ThemeManager.Current.Attach(settings);
 		return settings;
+	}
+
+	[AvaloniaTest]
+	public void LazyLoadedDefinitionIsNotDarkenedTwiceAtDarkStartup()
+	{
+		var settings = AttachThemeSettings();
+		try
+		{
+			// Dark is active BEFORE the definition is first touched -- the "dark theme
+			// preselected, session restores a document" startup ordering.
+			settings.Theme = "Dark";
+
+			var color = MakeColor("String", Colors.Red);
+			var pristine = (HighlightingColor)color.Clone();
+			var wrapper = new LazyLoadedDefinition(new StubHighlightingDefinition(color));
+
+			// HighlightingService.GetByExtension registers what the HighlightingManager
+			// returns: the wrapper. Its first member access materializes the inner
+			// definition, whose own registration has then ALREADY darkened the shared
+			// colours in place -- so the wrapper registration must not treat those dark
+			// values as light originals and convert them a second time.
+			ThemeManager.Current.RegisterThemableDefinition(wrapper);
+
+			var singleConversion = ThemeManager.GetColorForDarkTheme(pristine);
+			color.Should().Be(singleConversion,
+				"registering the lazy wrapper after its inner definition must not compound the dark conversion");
+		}
+		finally
+		{
+			settings.Theme = "Light";
+		}
 	}
 
 	[AvaloniaTest]
--- a/ILSpy.Tests/ThemeAwareHighlightingColorizerTests.cs
+++ b/ILSpy.Tests/ThemeAwareHighlightingColorizerTests.cs
@@ -1,0 +1,140 @@
+// Copyright (c) 2026 Christoph Wille
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this
+// software and associated documentation files (the "Software"), to deal in the Software
+// without restriction, including without limitation the rights to use, copy, modify, merge,
+// publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons
+// to whom the Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all copies or
+// substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+// INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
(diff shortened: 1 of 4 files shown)


siegfriedpammer:
Reviewed both commits. The diagnosis matches what I found independently while investigating #3986: the delay-loading wrapper and the inner definition are two identities over one set of `HighlightingColor` instances, and the wrapper's snapshot captures already-darkened values. I had a parallel fix in progress and am dropping it in favour of this PR. Notes:

**1. The theme-aware marker now carries three meanings.** `if (IsThemeAware(definition)) return;` conflates "XSHD opts out of theming", "already themed" and "already registered". `ApplyHighlightingColors` is public and sets the marker itself, so a caller reaching it before `RegisterThemableDefinition` makes the definition unregisterable: the early return precedes the `themableDefinitions.Add`, so it never enters the list and is never re-themed on a theme switch. No current path does that, so this is latent, not live. The guard also relies on two AvaloniaEdit implementation details — that the wrapper's `Properties` forwards to the inner dictionary, and that touching it forces materialization. Both hold in 12.0.0 (checked against the decompiled `DelayLoadedHighlightingDefinition`), neither is contractual.

An alternative that removes the ordering dependency: key the original-colour snapshot by the colour instance instead of by the definition — `ConditionalWeakTable<HighlightingColor, HighlightingColor>` with `GetValue(color, static c => c.Clone())` in place of `Dictionary<IHighlightingDefinition, Dictionary<string, HighlightingColor>>`. Double conversion then cannot happen regardless of how many definition objects expose the colours, and no marker is involved. I verified this variant against the real `HighlightingService` flow (red before, green after, full headless suite 1165 passed / 0 failed), so it works — take it or leave it, this PR fixes the reported bug either way.

**2. No coverage of the switch path.** Both `ThemeManager` tests stop at first touch. The observable half of the bug was that Light→Dark recovered the palette while a later re-registration (e.g. a font-size change) re-broke it, so a test that registers, switches Light→Dark→Light and asserts the colour returns to its pristine value would pin the part that regressed twice.

**3. Duplicate test class name.** `ILSpy.Tests/ThemeAwareHighlightingColorizerTests.cs` (namespace `ICSharpCode.ILSpy.Tests`) has the same class name as the existing `ILSpy.Tests/Themes/ThemeAwareHighlightingColorizerTests.cs` (namespace `ICSharpCode.ILSpy.Tests.Themes`), which already covers this class's editor wiring. It compiles, but test output and navigation become ambiguous.

**4. `DarkCacheDoesNotSurviveThemeSwitch` does not exercise a theme switch.** The stub definition is never registered, so `UpdateTheme` iterates a list that does not contain it and the two `settings.Theme` assignments are inert; the assertion is driven solely by the manual `color.Foreground = Lime`. `NotBeSameAs` also passes when the served conversion is wrong — asserting the content equals `GetColorForDarkTheme(lime)` would pin "no stale conversion". The underlying claim is correct: `SimpleHighlightingBrush` overrides `Equals`/`GetHashCode` by colour value, so the content-keyed cache does miss after an in-place recolour.

**5. Nits.** A mutated key stays in its old hash bucket, so an in-place recolour strands the old `darkColors` entry instead of replacing it (bounded by colours x switches; the class doc mentions the miss but not the leftover entry). `using ICSharpCode.ILSpy;` is redundant inside `namespace ICSharpCode.ILSpy.Tests`.

**6. Adjacent, not this PR.** `CSharp-Mode.xshd` paints with colours owned by another definition (`<Begin color="XmlDoc/DocComment">`, `<Import ruleSet="XmlDoc/DocCommentSet"/>`). Those instances are not in the C# definition's `NamedHighlightingColors`, and nothing registers the built-in `XmlDoc` definition with `ThemeManager`, while the colorizer skips the per-paint remap because the C# definition is theme-aware. I have not checked what that looks like on screen.

christophwille:
Thanks for the thorough review — all points addressed in 31e4a2526.

**1. Adopted the `ConditionalWeakTable` variant.** `originalColors` is now keyed by `HighlightingColor` instance (`GetValue(color, static c => c.Clone())`), so theming is idempotent across definition identities and registration order — correctness no longer rides on the marker or on the wrapper's `Properties` forwarding / materialize-on-touch. The `IsThemeAware` early return stays, but demoted to a non-load-bearing filter: it honors the XSHD opt-out and skips a redundant list entry; re-theming would now be harmless either way. The latent unregisterable path is gone structurally: `ApplyHighlightingColors` is private, so the marker can no longer be set outside a registration.

**2. Switch path covered.** New `ThemeSwitchRestoresPristineColorsAfterDarkStartup`: dark-startup wrapper flow, then Light→Dark round-trip, asserting Light restores the pristine .xshd values exactly and Dark yields the single conversion. (Red under the pre-fix code on both legs: the wrapper's dark snapshot re-broke the palette on every switch.)

**3. Renamed.** The fixture is now `ILSpy.Tests/Themes/DoubleDarkConversionTests.cs` — unique class name, and it sits next to the existing editor-wiring tests.

**4. Cache test reworked** into `DarkConversionCacheMissesAfterInPlaceRecolor`: the inert theme switches are gone (the assertion is driven by the in-place recolor alone, which is the actual mechanism under test), and it now asserts the served conversion *content-equals* `GetColorForDarkTheme` of the new values instead of `NotBeSameAs`.

**5. Nits:** the colorizer's class doc now mentions the stranded old-content cache entry (bounded, dies with the colorizer); the redundant `using` didn't survive the move.

**6. XmlDoc cross-definition colors:** agreed it's out of scope here; nothing registers the built-in `XmlDoc` definition and the C# definition's theme-aware marker suppresses the per-paint fallback for the imported colors, so doc comments likely render with light-theme colors in dark mode. Worth a look on screen and an issue if it reproduces.

Full headless suite: 1167 passed / 0 failed / 4 pre-existing skips.

_Posted by an AI agent (Claude) on Christoph's behalf._

PR #3997, opened 2026-08-14
Fix #3993, #3994: main-menu shortcuts and dark-mode hover tooltips

### LINKED ISSUE #3993: Some hotkeys not working in 11.0.0.9375
Just downloaded the new version, and it seems the hotkeys on the File menu (Ctrl+O, Ctrl+S, F5) are not working. That is, pressing any of these has no effect with either the Treeview or the Disassembly focused. Selecting the related menu items works as expected. Also Alt+LeftArrow and Alt+RightArrow seem to working correctly.

### LINKED ISSUE #3994: ILSpy version 11.0.0.9375: ToolTip rendering not readable.
In the latest release (Avalonia), tooltips are not readable, and there seems to be no way to configure their display attributes.

Image

christophwille:
Fixes #3993
Fixes #3994

Two regressions of the same shape from the Avalonia port: UI affordances that WPF wired up implicitly need explicit wiring in Avalonia, and both fell through the gap.

## Fix 1: Main-menu keyboard shortcuts do nothing on Windows/Linux (#3993)

**Symptom:** Ctrl+O, Ctrl+S, and F5 show next to their File-menu items but pressing them has no effect. Alt+Left/Right and Ctrl+E keep working.

**Root cause:** The main menu is built as a `NativeMenu`, and `NativeMenuItem.Gesture` is display-only outside macOS. Decompiling Avalonia 12.1.1's `NativeMenuBarPresenter` (the managed fallback that renders the menu inline on Windows/Linux) confirms it: the presenter binds `NativeMenuItem.Gesture` to `MenuItem.InputGesture`, which only paints the shortcut text -- it never registers a key handler. Only macOS gets working shortcuts, because there the gestures become real NSMenu key equivalents. The Avalonia docs state this explicitly: *"`InputGesture` only displays the text. You must also register the actual key binding separately for the shortcut to function."* The shortcuts that kept working were exactly the ones hardcoded as `KeyBindings` in `MainWindow.axaml`; everything coming from `[ExportMainMenuCommand(InputGestureText = ...)]` metadata was display-only.

**Fix:** After the menu is built, `MainMenu.RegisterGestureKeyBindings` walks it recursively and adds a window-level `KeyBinding` for every item that has both a gesture and a command. Details:

- It runs after `TranslateGesturesForMacOS`, so the bound gesture always matches the one the menu displays.
- On macOS the system menu bar consumes its key equivalents before they reach the window, so the extra bindings stay dormant there -- no double dispatch.
- Alt+Left/Right exist both in `MainWindow.axaml` and as menu gestures; that duplication is harmless because `KeyboardDevice` stops dispatching key bindings once one marks the event handled (verified in the decompiled Avalonia source), so only the first match executes.
- MEF metadata (`InputGestureText`) stays the single source of truth -- no parallel shortcut list to maintain.

**Test:** `Menu_Gestures_Are_Registered_As_Window_KeyBindings` walks the built menu and asserts every displayed gesture has a matching window `KeyBinding` wired to the same command instance. Red before the fix (failed on Ctrl+O for File > Open), green after.

## Fix 2: Documentation hover tooltip unreadable in dark mode (#3994)

**Symptom:** Hovering a symbol in the decompiled view with the dark theme active shows light-gray/blue text on a near-white box (see issue screenshot). Light mode looks fine.

**Root cause:** The hover popup is not an Avalonia `ToolTip` (those are already themed via the `ToolTip` style in `App.axaml`) -- it is a custom rich `Popup` whose content chrome was hardcoded in `DocumentationRenderer.CreateView`: background `#FCFCFC`, border `#AAAAAA`. The signature text inside it is coloured by the active *highlighting* theme, so in dark mode the light-on-dark syntax colours land on a light box. The hyperlink colour in rendered XML docs was hardcoded too (`#0066CC`), which would be equally unreadable on a dark popup.

**Fix:**

- New theme-variant brushes in `App.axaml`: `ILSpy.DocTooltipBackground` / `ILSpy.DocTooltipBorder` / `ILSpy.DocLinkForeground`. The Light dictionary keeps the exact previous values (`#FCFCFC` / `#AAAAAA` / `#0066CC`), so light mode is pixel-identical. The Dark dictionary uses `#252526` / `#3F3F46` (matching the themed tooltip surface, slightly lighter than the `#1E1E1E` editor canvas so the popup reads as raised) and a brightened link blue `#6FA8DC`.
- These are deliberately separate from the existing `ILSpy.Tooltip*` brushes: the plain tooltip uses the classic yellow `#FFFFE1` fill, which is wrong for a popup showing syntax-coloured code.
- `DocumentationRenderer.CreateView` now binds the popup chrome via `GetResourceObservable` (the same pattern `DecompilerTextEditor` uses for the editor background), and the link colour moved into a `TextBlock.doc-link` style -- both follow live theme switches.

**Tests:** `DocumentationRendererThemeTests.Popup_Chrome_Follows_The_Active_Theme_Variant` pins the dark chrome under `ThemeVariant.Dark` and pins light mode to the unchanged near-white values; the existing `SeeCref` link test gained an assertion that doc links use the themed foreground. Both red before the fix, green after.

Full `ILSpy.Tests` suite green on Windows with both fixes in place: 1176 tests, 0 failed, 4 skipped.

### DIFF
--- a/ILSpy.Tests/Editor/DocumentationRendererThemeTests.cs
+++ b/ILSpy.Tests/Editor/DocumentationRendererThemeTests.cs
@@ -1,0 +1,88 @@
+// Copyright (c) 2026 Christoph Wille
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this
+// software and associated documentation files (the "Software"), to deal in the Software
+// without restriction, including without limitation the rights to use, copy, modify, merge,
+// publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons
+// to whom the Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all copies or
+// substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+// INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+// PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+// FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+using Avalonia;
+using Avalonia.Controls;
+using Avalonia.Headless.NUnit;
+using Avalonia.Media;
+using Avalonia.Styling;
+using Avalonia.Threading;
+
+using AvaloniaEdit.Highlighting;
+
+using AwesomeAssertions;
+
+using ICSharpCode.Decompiler.CSharp.OutputVisitor;
+
+using ICSharpCode.ILSpy.TextView;
+
+using NUnit.Framework;
+
+namespace ICSharpCode.ILSpy.Tests.TextView;
+
+/// <summary>
+/// The hover popup's signature text is coloured by the active highlighting theme, so its
+/// chrome must follow the same theme variant: dark-theme (light-on-dark) text on a
+/// hardcoded near-white background is unreadable (issue #3994). Light mode keeps the
+/// established near-white look.
+/// </summary>
+[TestFixture]
+public class DocumentationRendererThemeTests
+{
+	[AvaloniaTest]
+	public void Popup_Chrome_Follows_The_Active_Theme_Variant()
+	{
+		var renderer = new DocumentationRenderer(
+			new CSharpAmbience(),
+			new FontFamily("Consolas, Menlo, Monospace"),
+			12);
+		renderer.AddSignatureBlock(new RichText("(parameter) string matchText"));
+
+		var view = (Border)renderer.CreateView();
+		var window = new Window { Content = view };
+		var app = Application.Current!;
+		var originalVariant = app.RequestedThemeVariant;
+		try
+		{
+			window.Show();
+
+			app.RequestedThemeVariant = ThemeVariant.Dark;
+			Dispatcher.UIThread.RunJobs();
+
+			window.TryFindResource("ILSpy.DocTooltipBackground", ThemeVariant.Dark, out var darkBackground)
+				.Should().BeTrue("the popup chrome must route through themed brushes");
+			view.Background.Should().Be(darkBackground,
+				"dark-theme signature colours are unreadable on a light popup background");
+			window.TryFindResource("ILSpy.DocTooltipBorder", ThemeVariant.Dark, out var darkBorder)
+				.Should().BeTrue();
+			view.BorderBrush.Should().Be(darkBorder);
+
+			app.RequestedThemeVariant = ThemeVariant.Light;
+			Dispatcher.UIThread.RunJobs();
+
+			((ISolidColorBrush)view.Background!).Color.Should().Be(Color.FromRgb(0xFC, 0xFC, 0xFC),
+				"light mode keeps the established near-white popup chrome");
+			((ISolidColorBrush)view.BorderBrush!).Color.Should().Be(Color.FromRgb(0xAA, 0xAA, 0xAA));
+		}
+		finally
+		{
+			window.Close();
+			app.RequestedThemeVariant = originalVariant;
+		}
+	}
+}
--- a/ILSpy.Tests/MainWindow/MainMenuTests.cs
+++ b/ILSpy.Tests/MainWindow/MainMenuTests.cs
@@ -95,6 +95,48 @@ public class MainMenuTests
 		openItem.Gesture!.Should().Be(expected);
 	}
 
+	// NativeMenuItem.Gesture is display-only when NativeMenuBar renders the menu inline
+	// (the managed fallback binds it to MenuItem.InputGesture, which never handles input),
+	// so every menu gesture must also be registered as a window-level KeyBinding or the
+	// shortcut silently does nothing on Windows / Linux (issue #3993: Ctrl+O, Ctrl+S, F5).
+	[AvaloniaTest]
+	public void Menu_Gestures_Are_Registered_As_Window_KeyBindings()
+	{
+		var window = AppComposition.Current.GetExport<MainWindow>();
+		window.Show();
+
+		var nativeMenu = NativeMenu.GetMenu(window)
+			?? throw new InvalidOperationException("MainMenu.Attach should have set NativeMenu on the window");
+
+		var gestureItems = new System.Collections.Generic.List<(string Path, NativeMenuItem Item)>();
+		CollectItemsWithGesture(nativeMenu, parentPath: "", gestureItems);
+
+		gestureItems.Should().NotBeEmpty("File > Open (Ctrl+O), Reload (F5) and Save (Ctrl+S) declare InputGestureText");
+
+		foreach (var (path, item) in gestureItems)
+		{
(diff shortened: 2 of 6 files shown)


ILSpy/Views/MainMenu.axaml.cs
```diff
@@ -76,9 +76,38 @@ public static void Attach(Window window)
...
+		foreach (var element in menu.Items)
+		{
+			if (element is NativeMenuItemSeparator || element is not NativeMenuItem item)
+				continue;
+			if (item.Gesture is { } gesture && item.Command is { } command)
+			{
+				window.KeyBindings.Add(new KeyBinding {
+					Gesture = gesture,
+					Command = command,
+					CommandParameter = item.CommandParameter!,
```
christophwille: **low / correctness (latent)** -- this snapshots the value, but `AppendRegistryCommands` sets that same property as a *live binding* for `IProvideParameterBinding` commands (`menuItem.Bind(NativeMenuItem.CommandParameterProperty, parameterBinding.ParameterBinding)`).

A menu click reads the live property (`NativeMenuItem.RaiseClicked` -> `Command.Execute(CommandParameter)`), while the shortcut would execute with whatever the property held at menu-build time -- which, for a binding that resolves from the visual tree, is `null` (the menu is not in a visual tree yet). Latent today: nothing currently implements `IProvideParameterBinding`.

One-line fix:

```csharp
keyBinding.Bind(KeyBinding.CommandParameterProperty, item.GetObservable(NativeMenuItem.CommandParameterProperty));
```

or a comment recording the constraint.

ILSpy.Tests/Editor/DocumentationRendererThemeTests.cs
```diff
@@ -0,0 +1,88 @@
+// Copyright (c) 2026 AlphaSierraPapa for the SharpDevelop Team
```
christophwille: **low / convention** -- CLAUDE.md requires line 1 to name the human contributor and "never AlphaSierraPapa". The sibling file in the same folder (`DocumentationLinkTests.cs`) uses `// Copyright (c) 2026 Siegfried Pammer`.

ILSpy.Tests/MainWindow/MainMenuTests.cs
```diff
@@ -95,6 +95,46 @@ public void File_Open_Carries_The_Ctrl_O_Gesture()
...
+	}
+
+	static void CollectItemsWithGesture(NativeMenu menu, string parentPath, System.Collections.Generic.List<(string, NativeMenuItem)> result)
+	{
+		foreach (var element in menu.Items)
+		{
+			if (element is NativeMenuItemSeparator || element is not NativeMenuItem item)
+				continue;
+			var path = string.IsNullOrEmpty(parentPath) ? (item.Header ?? "<unnamed>") : $"{parentPath} > {item.Header}";
+			if (item.Gesture != null)
```
christophwille: **low / test brittleness** -- this collects every item with a gesture, but `RegisterGestureKeyBindings` deliberately registers only items that have *both* a gesture and a command. A future display-only gesture (a submenu header, or a command whose export throws and is skipped via the `CompositionErrors.Report` path) would fail this test even though production behaves as designed.

Mirror the production condition: `if (item.Gesture != null && item.Command != null)`.

christophwille (commented):
## Code review

Read every hunk and verified the claims by decompiling Avalonia 12.1.1 (`KeyBinding`, `KeyboardDevice`, `NativeMenuItem`, `Popup`, `Avalonia.Native`'s menu-item proxy), plus empirical runs on macOS (arm64).

**Both fixes are correct, and both work on macOS.** Full `ILSpy.Tests` on this branch: 1173 passed, 0 failed, 3 skipped.

Probes for what the PR's own tests do not cover:

- Rendering the doc popup inside a real `Popup` (the new test only uses a `Window`): dark -> border `#252526`, chrome `#3f3f46`, body `#dedede`, link `#6fa8dc`; light -> `#fcfcfc` / black / `#0066cc`. Foreground inheritance does cross the `PopupRoot` boundary, so the body text is readable in both variants -- the fix is complete, not just the chrome.
- Window-level `KeyBinding` dispatch while a `TextBox` has focus: fires. `KeyboardDevice.ProcessRawEvent` walks focused-element -> visual ancestors and runs `KeyBindings` *before* raising `KeyDown`, so the fix is not shadowed by focused controls.

### macOS specifics

- `__MicroComIAvnMenuItemProxy.UpdateGesture` -> `SetGesture(key, modifiers)` confirms `NativeMenuItem.Gesture` becomes a real NSMenu key equivalent, so the menu consumes it and the extra window bindings are inert there. Even in the inverted case (Avalonia's `RawKeyEvent` returns a handled-bool to native, so `AvnView` could run managed input first), exactly one path executes: `KeyBinding.TryHandle` gates on `Command.CanExecute`, and `NativeMenuItem.IsEnabled` is itself `SetCurrentValue(..., Command?.CanExecute(...))`, so the two gates always agree. No double dispatch either way.
- `RegisterGestureKeyBindings` is correctly placed *after* `TranslateGesturesForMacOS`, so the registered gesture matches the displayed one (Cmd on macOS).
- The Alt+Left/Alt+Right duplication with `MainWindow.axaml` is harmless: `KeyBinding.TryHandle` only sets `Handled` when `CanExecute` is true and `KeyboardDevice` stops the loop on `Handled`; both route to the same inner `NavigateBackCommand`.

Nothing in the diff is a correctness bug. Three low-severity items inline.

### Note outside the diff (macOS-relevant)

`ILSpy/Views/MainWindow.axaml:25-32` hardcodes `Ctrl+E`, `Ctrl+W`, `Ctrl+Shift+F` as `KeyBinding`s. Unlike the menu gestures these do **not** go through `TranslateGesturesForMacOS`, so on macOS they still require Ctrl rather than Cmd. Pre-existing, not introduced here, but it is the remaining half of the same "shortcuts on macOS" story and the new comment block in `MainMenu.axaml.cs` reasons about exactly these bindings -- worth a follow-up issue.

ILSpy/Views/MainMenu.axaml.cs
christophwille: Fixed in the squashed #3993 commit: the KeyBinding's CommandParameter is now a live binding to the item's CommandParameter property (via GetObservable), so an IProvideParameterBinding command resolves its parameter at shortcut time exactly like a menu click would. A comment in RegisterGestureKeyBindings records the constraint.

ILSpy.Tests/Editor/DocumentationRendererThemeTests.cs
christophwille: Fixed in the squashed #3994 commit: the header now credits Christoph Wille per the repo convention.

ILSpy.Tests/MainWindow/MainMenuTests.cs
christophwille: Fixed in the squashed #3993 commit: the test's collector now mirrors the production condition (gesture AND command), with a comment noting that display-only gestures are deliberately excluded.

christophwille:
All three inline items are addressed and squashed into their respective commits (branch force-pushed, now `36b870d3e` / `0d244f4d7`):

- **CommandParameter snapshot** -- the shortcut's `KeyBinding` now binds `CommandParameterProperty` to the item's property via `GetObservable`, so `IProvideParameterBinding` commands resolve their parameter at shortcut time the same way a menu click does.
- **File header** -- the new test file now credits Christoph Wille.
- **Test brittleness** -- the gesture collector mirrors the production condition (`Gesture != null && Command != null`), so display-only gestures can't fail the assertion.

`MainMenuTests` + `DocumentationRendererThemeTests` re-run green after the changes.

On the out-of-diff note: agreed that the hardcoded `Ctrl+E` / `Ctrl+W` / `Ctrl+Shift+F` bindings in `MainWindow.axaml` are the remaining half of the macOS shortcut story, since they bypass `TranslateGesturesForMacOS`. Left out of this PR to keep it scoped to the two linked issues -- worth its own follow-up issue.

siegfriedpammer (approved):

PR #3998, opened 2026-08-14
Fix #3996: edit the display font size in points, not pixels

### LINKED ISSUE #3996: Options dialog edits the font size in pixels, not points
The font size in Options > Display is edited as device-independent pixels, while the WPF host presented it in points. The stored value is unchanged, so only the number the user sees and types is wrong.

`DisplaySettings.SelectedFontSize` holds device-independent pixels and is consumed as such (`DecompilerTextEditor`, `EditorZoom.EffectiveFontSize`, and the preview `TextBlock`), with the default written as `10.0 * 4 / 3` — 10 pt expressed in pixels:

https://github.com/icsharpcode/ILSpy/blob/3ec7e5f82/ILSpy/Options/DisplaySettings.cs#L40

The options panel binds that value straight to a `NumericUpDown`:

https://github.com/icsharpcode/ILSpy/blob/3ec7e5f82/ILSpy/Options/DisplaySettingsPanel.axaml#L29-L31

The WPF panel put a converter in between, `Math.Round(d / 4 * 3)` for display and `d * 4 / 3` on the way back (`FontSizeConverter` in the pre-Avalonia `DisplaySettingsPanel.xaml.cs`), so the box showed points. That converter has no counterpart in the Avalonia port.

Consequences:

* On a fresh profile the size box reads `13.333…` rather than `10`.
* `Increment="1"` steps by one pixel, i.e. 0.75 pt.
* `Minimum="6"` / `Maximum="72"` are pixels, so the reachable range is 4.5 pt to 54 pt — the usual 6-72 pt range is not.
* A settings file written by ILSpy 9.x (WPF) round-trips correctly, since the persisted attribute is pixels in both hosts; it is only the dialog that disagrees about the unit.

Either convert for display as WPF did, or keep pixels and drop the pt-derived default and range so the dialog is at least self-consistent.

_Reported by an AI agent (Claude) on Siegfried's behalf._

christophwille:
Fixes #3996.

The options dialog bound `DisplaySettings.SelectedFontSize` (device-independent pixels) straight into a `NumericUpDown`, so a fresh profile showed `13.333...` instead of `10`, the increment stepped by 0.75 pt, and the 6-72 bounds were pixels (4.5-54 pt). The WPF host presented the value in points via `FontSizeConverter`; this restores that behavior on Avalonia, matching the Windows font dialogs (Notepad et al.).

### What changed

- **`ILSpy/Options/DisplaySettingsViewModel.cs`** - new `SelectedFontSizePoints` string proxy that shows `Math.Round(px * 3 / 4)` and stores `pt * 4 / 3`, mirroring the WPF `FontSizeConverter`. Non-numeric input is ignored (usually a transient typing state); numeric input is clamped to 6-72 pt. External changes to `SelectedFontSize` (reset-to-defaults, settings load) refresh the text via `PropertyChanged`, while writes originating from the size box suppress the echo so typing isn't clobbered mid-keystroke. Also adds `FontSizes` (6-24), the same dropdown list the WPF host offered.
- **`ILSpy/Options/DisplaySettingsPanel.axaml`** - replaces the pixel-unit `NumericUpDown` with a "Size" label (existing `Resources.Size` string) plus an editable `ComboBox` (`IsEditable="True"`, `Text` bound to the points proxy, items 6-24), i.e. the WPF panel's shape on Avalonia 12. The preview `TextBlock` still binds the raw pixel value, so it renders the true size.
- **`ILSpy.Tests/Options/DisplayFontSizeTests.cs`** - new tests (written red-first): default 13.33 px displays as "10", typed "12" stores 16 px, external pixel change raises `PropertyChanged` and shows the right points, dialog-originated writes don't echo, garbage input is ignored, clamping at both ends, WPF-parity size list, and a headless UI test verifying the panel's size box is the editable ComboBox showing "10".

`DisplaySettings.SelectedFontSize` and its persistence are untouched, so settings files written by ILSpy 9.x still round-trip; `EditorZoom` and the live editor wire-up are unaffected.

Full `ILSpy.Tests` suite: 1178 passed, 0 failed, 4 skipped.

### DIFF
--- a/ILSpy.Tests/Options/DisplayFontSizeTests.cs
+++ b/ILSpy.Tests/Options/DisplayFontSizeTests.cs
@@ -1,0 +1,216 @@
+// Copyright (c) 2026 Christoph Wille
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this
+// software and associated documentation files (the "Software"), to deal in the Software
+// without restriction, including without limitation the rights to use, copy, modify, merge,
+// publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons
+// to whom the Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all copies or
+// substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+// INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+// PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+// FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+using Avalonia.Controls;
+using Avalonia.Headless.NUnit;
+using Avalonia.Threading;
+
+using AwesomeAssertions;
+
+using ICSharpCode.ILSpy.AppEnv;
+using ICSharpCode.ILSpy.Commands;
+using ICSharpCode.ILSpy.Docking;
+using ICSharpCode.ILSpy.Options;
+using ICSharpCode.ILSpy.Options.Panels;
+using ICSharpCode.ILSpy.Properties;
+using ICSharpCode.ILSpy.ViewModels;
+using ICSharpCode.ILSpy.Views;
+
+using NUnit.Framework;
+
+namespace ICSharpCode.ILSpy.Tests;
+
+/// <summary>
+/// The options dialog edits the font size in points (like the Windows font dialogs and the
+/// WPF host's FontSizeConverter), while <see cref="DisplaySettings.SelectedFontSize"/> keeps
+/// storing device-independent pixels so persisted settings round-trip with ILSpy 9.x.
+/// The pt/px conversion lives in <see cref="DisplaySettingsViewModel.SelectedFontSizePoints"/>.
+/// </summary>
+[TestFixture]
+public class DisplayFontSizeTests
+{
+	static (DisplaySettingsViewModel page, DisplaySettings settings) CreatePage()
+	{
+		var service = AppComposition.Current.GetExport<SettingsService>();
+		var page = new DisplaySettingsViewModel();
+		page.Load(service);
+		return (page, service.DisplaySettings);
+	}
+
+	[AvaloniaTest]
+	public void Default_Pixel_Size_Displays_As_10_Points()
+	{
+		var (page, settings) = CreatePage();
+		var original = settings.SelectedFontSize;
+		try
+		{
+			settings.SelectedFontSize = 10.0 * 4 / 3;
+			page.SelectedFontSizePoints.Should().Be("10",
+				"the default 13.33 px must be presented as the 10 pt it actually is");
+		}
+		finally
+		{
+			settings.SelectedFontSize = original;
+		}
+	}
+
+	[AvaloniaTest]
+	public void Typed_Point_Size_Is_Stored_As_Pixels()
+	{
+		var (page, settings) = CreatePage();
+		var original = settings.SelectedFontSize;
+		try
+		{
+			page.SelectedFontSizePoints = "12";
+			settings.SelectedFontSize.Should().BeApproximately(12.0 * 4 / 3, 1e-9,
+				"the stored value stays device-independent pixels for 9.x round-tripping");
+			page.SelectedFontSizePoints.Should().Be("12");
+		}
+		finally
+		{
+			settings.SelectedFontSize = original;
+		}
+	}
+
+	[AvaloniaTest]
+	public void External_Pixel_Change_Refreshes_The_Points_Text()
+	{
+		// Reset-to-defaults and LoadFromXml write SelectedFontSize directly; the dialog text
+		// must follow via PropertyChanged on SelectedFontSizePoints.
+		var (page, settings) = CreatePage();
+		var original = settings.SelectedFontSize;
+		try
+		{
+			var notified = new List<string?>();
+			page.PropertyChanged += (_, e) => notified.Add(e.PropertyName);
+
+			settings.SelectedFontSize = 20;
+
+			notified.Should().Contain(nameof(DisplaySettingsViewModel.SelectedFontSizePoints));
+			page.SelectedFontSizePoints.Should().Be("15");
+		}
+		finally
+		{
+			settings.SelectedFontSize = original;
+		}
+	}
+
+	[AvaloniaTest]
+	public void Setting_Points_Through_The_Dialog_Does_Not_Echo_A_Text_Notification()
+	{
+		// While the user is typing in the size box, the setter must not raise PropertyChanged
+		// for SelectedFontSizePoints - the binding would rewrite the box mid-keystroke
+		// (typing "10." would snap back to "10" before the fraction can be completed).
+		var (page, settings) = CreatePage();
+		var original = settings.SelectedFontSize;
+		try
+		{
+			var notified = new List<string?>();
+			page.PropertyChanged += (_, e) => notified.Add(e.PropertyName);
+
+			page.SelectedFontSizePoints = "14";
+
+			notified.Should().NotContain(nameof(DisplaySettingsViewModel.SelectedFontSizePoints));
+		}
+		finally
+		{
+			settings.SelectedFontSize = original;
+		}
+	}
+
+	[AvaloniaTest]
+	public void NonNumeric_Input_Is_Ignored()
+	{
+		var (page, settings) = CreatePage();
+		var original = settings.SelectedFontSize;
+		try
+		{
+			settings.SelectedFontSize = 16;
+			page.SelectedFontSizePoints = "abc";
+			settings.SelectedFontSize.Should().Be(16,
+				"transient garbage while typing must not move the stored size");
+		}
+		finally
+		{
+			settings.SelectedFontSize = original;
+		}
+	}
+
+	[AvaloniaTest]
+	public void Typed_Sizes_Are_Clamped_To_The_6_To_72_Point_Range()
+	{
+		var (page, settings) = CreatePage();
+		var original = settings.SelectedFontSize;
+		try
+		{
+			page.SelectedFontSizePoints = "1";
+			settings.SelectedFontSize.Should().BeApproximately(6.0 * 4 / 3, 1e-9);
+
+			page.SelectedFontSizePoints = "500";
+			settings.SelectedFontSize.Should().BeApproximately(72.0 * 4 / 3, 1e-9);
+		}
+		finally
+		{
+			settings.SelectedFontSize = original;
+		}
+	}
+
+	[AvaloniaTest]
+	public void Size_List_Offers_6_Through_24_Points_Like_The_WPF_Host()
+	{
+		var (page, _) = CreatePage();
+		page.FontSizes.Should().Equal(Enumerable.Range(6, 24 - 6 + 1));
+	}
+
+	[AvaloniaTest]
+	public async Task Display_Panel_Size_Box_Is_An_Editable_ComboBox_Showing_Points()
+	{
+		var window = AppComposition.Current.GetExport<MainWindow>();
+		window.Show();
+		var settings = AppComposition.Current.GetExport<SettingsService>();
+		var original = settings.DisplaySettings.SelectedFontSize;
+		try
+		{
+			settings.DisplaySettings.SelectedFontSize = 10.0 * 4 / 3;
+
+			AppComposition.Current.GetExport<MainMenuCommandRegistry>()
+				.GetCommand(nameof(Resources._Options)).Execute(null);
+			var vm = (MainWindowViewModel)window.DataContext!;
+			var model = (OptionsPageModel)vm.DockWorkspace.Documents!.VisibleDockables!
+				.OfType<ContentTabPage>().First(t => t.Content is OptionsPageModel).Content!;
+			model.SelectedPage = model.Pages.OfType<DisplaySettingsViewModel>().Single();
+			TestCapture.Step("display-page-selected");
+
+			var panel = await window.WaitForComponent<DisplaySettingsPanel>();
+			var box = panel.FindControl<ComboBox>("fontSizeComboBox");
+			((object?)box).Should().NotBeNull("the size box must be the named editable ComboBox");
+			Dispatcher.UIThread.RunJobs();
+
+			box!.IsEditable.Should().BeTrue("custom sizes must be typeable, like Notepad's font page");
+			box.Text.Should().Be("10", "the box shows points, not device-independent pixels");
+		}
+		finally
+		{
+			settings.DisplaySettings.SelectedFontSize = original;
+		}
+	}
+}
--- a/ILSpy.Tests/Options/DisplayFontSizeTests.cs
+++ b/ILSpy.Tests/Options/DisplayFontSizeTests.cs
@@ -23,6 +23,7 @@ using System.Threading.Tasks;
 using Avalonia.Controls;
 using Avalonia.Headless.NUnit;
 using Avalonia.Threading;
+using Avalonia.VisualTree;
 
 using AwesomeAssertions;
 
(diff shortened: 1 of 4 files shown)


christophwille:
## Code review

Verified locally: `ILSpy/ILSpy.csproj` builds clean on the branch (0 warnings), the Simple theme's `ComboBox` template does ship a `PART_EditableTextBox` (checked the compiled `Avalonia.Themes.Simple` 12.1.1 resource strings), so `IsEditable="True"` is really typeable and not just a no-op property; the `Text` two-way + `ItemsSource` wiring round-trips correctly through Avalonia's `ComboBox.TextChanged`/`UpdateInputTextFromSelection` (dropdown pick -> `SelectedItem` -> `SetCurrentValue(TextProperty, ...)` -> binding write-back). The initialization order is also safe: `ComboBox.OnPropertyChanged` for `ItemsSourceProperty` re-publishes `Text` (which is `string.Empty` before the `Text` binding lands), and the setter's "ignore non-numeric" rule is what keeps that from wiping the setting -- worth a comment, because that guard is load-bearing, not just typing UX.

Four things below.

---

### 1. `Math.Clamp(NaN, 6, 72)` returns `NaN`, and `NaN` gets persisted -- medium

`ILSpy/Options/DisplaySettingsViewModel.cs:75-77`

```csharp
if (Settings == null || !double.TryParse(value, NumberStyles.Float, CultureInfo.CurrentCulture, out double points))
    return;
points = Math.Clamp(points, 6, 72);
```

`double.TryParse` with `NumberStyles.Float` accepts the culture's NaN symbol on .NET Core 3.0+ (verified: `"NaN"` and `"nan"` both parse, `Math.Clamp(NaN, 6, 72) == NaN` -- `Clamp` compares with `<`/`>`, both false for NaN, so it falls through to `return value`). Infinity is fine (`+inf` -> 72, `-inf` -> 6, `"1e400"` -> 72); only NaN slips through.

Scenario: user types or pastes `NaN` into the size box.

* `DisplaySettings.SelectedFontSize = double.NaN`.
* `SaveToXml` writes `FontSize="NaN"`; `LoadFromXml`'s `(double?)section.Attribute("FontSize")` reads it straight back, so it survives restart.
* `DecompilerTextEditor.ApplyFontSettings` guards with `if (displaySettings.SelectedFontSize > 0)` -- false for NaN -- so from then on the editor **never applies a font size again**, on this run or any future one.
* The preview `TextBlock` binding is rejected by Avalonia's `TextElement.FontSizeProperty` validate func (`fontSize > 0 && !IsNaN && !IsInfinity`), which surfaces as a logged binding error rather than a crash.
* Recovery requires Reset-to-defaults or hand-editing `ILSpy.xml`.

The `NumericUpDown` this replaces could not produce a non-finite value. One line fixes it:

```csharp
if (!double.IsFinite(points))
    return;
```

(The old WPF `FontSizeConverter.ConvertBack` had the same hole, so this isn't a WPF-parity regression -- but it is a regression against what ships today.)

---

### 2. macOS: the 4/3 factor is a Windows-DIP assumption -- medium

`ILSpy/Options/DisplaySettingsViewModel.cs:63-73`

The conversion is correct on Windows and Linux, and wrong-by-convention on macOS.

* On Windows, a WPF/Avalonia device-independent pixel is defined as 1/96 in, and the Windows font dialogs report points where `pt = DIP * 3/4`. 4/3 is exactly right.
* On Linux/X11, Avalonia derives `RenderScaling` from `Xft.dpi/96`, so the logical unit is again ~1/96 in and GTK-style point sizes line up.
* On macOS, `Avalonia.Native`'s `RenderScaling` is the `NSWindow`/`NSScreen` `backingScaleFactor` and `DesktopScaling => 1.0` (decompiled `Avalonia.Native` 12.1.1). That makes one Avalonia logical unit **identical to one Cocoa point** -- and a Cocoa point is exactly the number every native macOS font UI puts in its "size" field (`NSFontPanel`, Xcode's editor font, VS Code's `editor.fontSize`).

Concretely on a Mac: Xcode's default SF Mono at "11" and ILSpy at `SelectedFontSize = 11` render identically, but after this PR ILSpy's box would show `8` for that size. Conversely a Mac user who types `11` to match Xcode gets 14.67 logical units -- ~33% larger than every other app at the same stated number.

So on macOS this change makes the displayed number *less* comparable to the rest of the desktop than the raw pixel value was. Nothing crashes and nothing round-trips wrong, but the doc comment's premise ("points, like the Windows font dialogs") is a Windows premise living in a cross-platform view model, and #3996's complaint ("a fresh profile shows 13.333") has a different right answer per platform. Options, roughly in order of laziness:

1. Ship as-is and accept the macOS mismatch (defensible -- the stored value still round-trips, and this matches the 9.x host).
2. Label the unit in the UI (`Size (pt):`) so the number isn't silently compared against native Mac apps.
3. Apply 4/3 only where the platform's logical unit is 1/96 in (i.e. skip it on `OperatingSystem.IsMacOS()`), which then makes the *stored* value platform-dependent -- probably not worth it.

Worth an explicit decision either way; right now the comment asserts a universal that only holds on two of the three targets.

---

### 3. A clamped value never makes it back into the box -- low/medium

`ILSpy/Options/DisplaySettingsViewModel.cs:78-88`

The echo suppression is unconditional, and there is no commit-on-lost-focus normalization, so whenever the typed text and the stored value disagree the dialog keeps lying:

* Type `3` and tab away -> `SelectedFontSize` becomes 8 px (6 pt), the box still reads `3`, and the preview under it renders 6 pt. Reopening the Options page is the only way to resync.
* Same for `500` -> stored 96 px, box still reads `500`.

`NumericUpDown` handled this: `OnLostFocus` -> `CommitInput(forceTextUpdate: true)` rewrote the text from the clamped value.

The suppression is only *needed* while the round-trip text is unchanged, so making it conditional fixes both without a focus handler:

```csharp
var text = points.ToString(CultureInfo.CurrentCulture);
updatingFontSizeFromText = text == value;   // only suppress when the box already shows the truth
try { Settings.SelectedFontSize = points * 4 / 3; }
finally { updatingFontSizeFromText = false; }
```

---

### 4. The headless test doesn't actually exercise the editable box -- low

`ILSpy.Tests/Options/DisplayFontSizeTests.cs:206-210`

```csharp
box!.IsEditable.Should().BeTrue(...);
box.Text.Should().Be("10", ...);
```

Both are plain styled-property reads on the `ComboBox`; they pass whether or not the applied `ControlTheme` actually realizes an editable text box. (I confirmed by hand that the Simple theme does -- but the test wouldn't have caught it if it didn't, which is the interesting failure this test is meant to guard.) `box.ApplyTemplate()` plus asserting a `TextBox` named `PART_EditableTextBox` exists in the visual tree would make the assertion mean what its "custom sizes must be typeable" message claims.

Two smaller notes on the same file:

* `Default_Pixel_Size_Displays_As_10_Points` sets `SelectedFontSize = 10.0 * 4 / 3` itself rather than reading the actual default, so it doesn't test the default. `ResetAppStateAttribute` rebuilds the MEF container per test, so a fresh `DisplaySettings` is already there -- the assignment (and the `try`/`finally` restores throughout the fixture) can just go.
* `NonNumeric_Input_Is_Ignored` is the test that pins down the load-bearing behaviour from the intro paragraph (empty `Text` published during `ItemsSource` initialization must not clobber the setting). Worth saying so in the comment, otherwise someone will "improve" the setter into falling back to a default like the WPF converter did (`return 11.0 * 4 / 3`) and silently reset user font sizes on every page load.

---

### Cosmetic

`ILSpy/Options/DisplaySettingsPanel.axaml:30` -- `Resources.Size` is `"Size:"` (with colon) while the neighbouring `Resources.Font` is `"Font"` (without), so the row renders `Font [ ] Size: [ ]`. The WPF panel had the same asymmetry, so this is parity, not a regression -- just noting it since the label is new here.

ILSpy/Options/DisplaySettingsViewModel.cs
```diff
@@ -55,6 +57,40 @@ public sealed partial class DisplaySettingsViewModel : ObservableObject, IOption
...
+		/// (it is usually a transient typing state), numeric input is clamped to 6-72 pt.
+		/// </summary>
+		public string SelectedFontSizePoints {
+			get => Settings == null
+				? string.Empty
+				: Math.Round(Settings.SelectedFontSize * 3 / 4).ToString(CultureInfo.CurrentCulture);
+			set {
+				if (Settings == null || !double.TryParse(value, NumberStyles.Float, CultureInfo.CurrentCulture, out double points))
+					return;
+				points = Math.Clamp(points, 6, 72);
```
christophwille, line 89: **`Math.Clamp(NaN, 6, 72)` returns `NaN`.** `double.TryParse` with `NumberStyles.Float` accepts the culture's NaN symbol (verified empirically), and `Clamp` compares with `<`/`>` -- both false for NaN -- so it passes straight through.

Typing or pasting `NaN` stores `SelectedFontSize = double.NaN`. `SaveToXml`/`LoadFromXml` persist it verbatim, and `DecompilerTextEditor.ApplyFontSettings`'s `if (SelectedFontSize > 0)` guard is then false forever: the editor never applies a font size again, across restarts. Recovery needs Reset-to-defaults or hand-editing `ILSpy.xml`. The replaced `NumericUpDown` could not produce a non-finite value.

```csharp
if (!double.IsFinite(points))
    return;
```

ILSpy/Options/DisplaySettingsViewModel.cs
```diff
@@ -55,6 +57,40 @@ public sealed partial class DisplaySettingsViewModel : ObservableObject, IOption
 			.OrderBy(n => n, System.StringComparer.OrdinalIgnoreCase)
 			.ToArray();
 
+		/// <summary>Point sizes offered in the size dropdown; same list as the WPF host.</summary>
+		public IReadOnlyList<int> FontSizes { get; } = Enumerable.Range(6, 24 - 6 + 1).ToArray();
+
+		/// <summary>
+		/// The font size as the user sees and edits it: points, like the Windows font dialogs.
+		/// <see cref="DisplaySettings.SelectedFontSize"/> itself stays in device-independent
+		/// pixels (1 pt = 4/3 px) so persisted settings round-trip with the WPF host; the
```
christophwille: **macOS: `1 pt = 4/3 px` is a Windows-DIP assumption, not a universal one.**

On Windows a DIP is 1/96 in, so `pt = DIP * 3/4` is exact; Linux/X11 derives `RenderScaling` from `Xft.dpi/96`, so it matches there too. On macOS, decompiled `Avalonia.Native` shows `RenderScaling` = `NSWindow.backingScaleFactor` and `DesktopScaling => 1.0`, which makes one Avalonia logical unit **identical to one Cocoa point** -- the exact number `NSFontPanel`, Xcode and VS Code put in their size fields.

So on a Mac this box shows `8` for text rendered at what Xcode calls `11`, and a user typing `11` to match their editor gets ~33% larger text. Nothing crashes and the stored value still round-trips -- it is a per-platform convention decision, but the doc comment currently asserts the ratio as universal.

ILSpy/Options/DisplaySettingsViewModel.cs
```diff
@@ -55,6 +57,40 @@ public sealed partial class DisplaySettingsViewModel : ObservableObject, IOption
...
+			get => Settings == null
+				? string.Empty
+				: Math.Round(Settings.SelectedFontSize * 3 / 4).ToString(CultureInfo.CurrentCulture);
+			set {
+				if (Settings == null || !double.TryParse(value, NumberStyles.Float, CultureInfo.CurrentCulture, out double points))
+					return;
+				points = Math.Clamp(points, 6, 72);
+				// Suppress the PropertyChanged echo for this property: the binding would
+				// immediately rewrite the size box with the rounded value mid-keystroke.
+				updatingFontSizeFromText = true;
```
christophwille, line 92: **A clamped value never makes it back into the box.** The echo suppression is unconditional and there is no commit-on-lost-focus normalization, so typing `3` and tabbing away leaves the setting at 6 pt while the box still reads `3` (likewise `500` -> 72 pt, box shows `500`). `NumericUpDown` handled this via `OnLostFocus -> CommitInput(forceTextUpdate: true)`.

Suppression is only actually needed when the round-trip text is unchanged:

```csharp
updatingFontSizeFromText = points.ToString(CultureInfo.CurrentCulture) == value;
```

ILSpy.Tests/Options/DisplayFontSizeTests.cs
```diff
@@ -0,0 +1,216 @@
...
+				.OfType<ContentTabPage>().First(t => t.Content is OptionsPageModel).Content!;
+			model.SelectedPage = model.Pages.OfType<DisplaySettingsViewModel>().Single();
+			TestCapture.Step("display-page-selected");
+
+			var panel = await window.WaitForComponent<DisplaySettingsPanel>();
+			var box = panel.FindControl<ComboBox>("fontSizeComboBox");
+			((object?)box).Should().NotBeNull("the size box must be the named editable ComboBox");
+			Dispatcher.UIThread.RunJobs();
+
+			box!.IsEditable.Should().BeTrue("custom sizes must be typeable, like Notepad's font page");
```
christophwille: **This does not exercise the editable box.** `box.IsEditable` and `box.Text` are plain styled-property reads; they pass even if the applied `ControlTheme` realized no editable text box at all -- which is precisely the failure the message ("custom sizes must be typeable") claims to guard.

`ApplyTemplate()` plus asserting a `TextBox` named `PART_EditableTextBox` in the visual tree would make the test mean what it says. (SimpleTheme does ship that part -- I checked the compiled `Avalonia.Themes.Simple` 12.1.1 string heap -- so the assertion will pass; it just is not currently being checked.)

christophwille (commented):
Findings from the review inline below (summary is in [the comment above](https://github.com/icsharpcode/ILSpy/pull/3998#issuecomment-5293403713)).

ILSpy/Options/DisplaySettingsViewModel.cs
christophwille: Fixed in 1bcb33b: `if (!double.IsFinite(points)) return;` before the clamp, with a comment explaining the persistence consequence. Covered by the new `NaN_Input_Is_Ignored` test (also asserts the stored value stays finite).

ILSpy/Options/DisplaySettingsViewModel.cs
christophwille: Decision: ship the WPF-host convention as-is (option 1) - keeping `SelectedFontSize` in logical pixels with one fixed conversion is what lets the same settings file round-trip across hosts and with 9.x, and a platform-dependent stored value isn't worth the mismatch. The doc comment no longer asserts the ratio as universal (1bcb33b): it now states it's exact where the logical unit is 1/96 in (Windows DIPs, X11 via Xft.dpi) and that on macOS the shown number is deliberately the Windows-style value, not the Cocoa-point size.

ILSpy/Options/DisplaySettingsViewModel.cs
christophwille: Fixed in 1bcb33b, but via commit-on-LostFocus rather than the conditional suppression: making suppression conditional on `points.ToString() == value` re-breaks the mid-typing cases the suppression exists for - typing `10.` would round-trip as `"10" != "10."` and snap the dot away, and a leading `1` (heading for `12`) would clamp to 6 and rewrite the box to `6` under the user's fingers. Instead the panel wires the box's `LostFocus` to a `CommitFontSizeText()` on the viewmodel that re-raises `PropertyChanged` for `SelectedFontSizePoints`, i.e. the same commit point `NumericUpDown.CommitInput(forceTextUpdate: true)` used. Covered by `Clamped_Value_Is_Written_Back_Into_The_Box_On_Focus_Loss` (types `3`, raises `LostFocus`, asserts the box reads `6`).

ILSpy.Tests/Options/DisplayFontSizeTests.cs
christophwille: Fixed in 1bcb33b: the test now calls `ApplyTemplate()` and asserts a `TextBox` named `PART_EditableTextBox` exists in the ComboBox's visual tree. Also picked up the two smaller notes: the default-size test no longer arranges its own value (it asserts the genuine built-in default, relying on the per-test `ResetAppState` container/settings rebuild, and the fixture-wide try/finally restores are gone), and the non-numeric test now covers the empty string with a comment marking the ItemsSource-initialization guard as load-bearing - plus a matching "do not improve this into a fall-back-to-default" comment in the setter itself.

christophwille:
All four addressed in 1bcb33b; replies with details on each thread. Short version: NaN is rejected via `double.IsFinite` before the clamp; the macOS 4/3 question is decided as "keep the WPF-host convention, document it honestly" (stored value stays host-independent); the stale-clamped-text issue is fixed with NumericUpDown-style commit-on-LostFocus instead of conditional echo suppression (the conditional variant re-breaks the `10.` and leading-`1` typing cases); and the headless test now asserts `PART_EditableTextBox` is realized, tests the genuine default, and pins the load-bearing empty-Text guard. The cosmetic `Size:`/`Font` label asymmetry is left as WPF parity. Options tests all green (10/10 new + 10/10 OptionsTabTests).

christophwille:
The Ubuntu leg failed on the new `Clamped_Value_Is_Written_Back_Into_The_Box_On_Focus_Loss` test: it raised `LostFocus` synthetically with a plain `RoutedEventArgs`, but the event is typed (`FocusChangedEventArgs`), so the first typed subscriber on the route - present on the Linux headless run, absent on Windows - hit an `InvalidCastException` in the handler adapter. Fixed by performing a real focus traversal instead: focus the box's `PART_EditableTextBox`, then focus a checkbox elsewhere on the panel, which raises the genuine event and exercises the actual commit path. (`ComboBox.Focus()` itself returns false when `IsEditable` - focus is delegated to the inner text box - which the test now documents.) Squashed into the hardening commit per branch convention, now `9fbebb8`.

ILSpy/Options/DisplaySettingsViewModel.cs
```diff
@@ -55,6 +57,58 @@ public sealed partial class DisplaySettingsViewModel : ObservableObject, IOption
 			.OrderBy(n => n, System.StringComparer.OrdinalIgnoreCase)
 			.ToArray();
 
+		/// <summary>Point sizes offered in the size dropdown; same list as the WPF host.</summary>
```
siegfriedpammer: don't mention WPF

siegfriedpammer (changes requested):

ILSpy/Options/DisplaySettingsViewModel.cs
christophwille: Done - removed the WPF references this PR added (dropdown-list doc, pt/px conversion doc, converter fallback note, and the test name/comments). The pre-existing `Fonts.SystemFontFamilies` comment on `AvailableFonts` is untouched since it predates this PR. Squashed into the hardening commit, now 6b0b49ab2.

siegfriedpammer (approved):

PR #4000, opened 2026-08-14
Fix #3729: recognize value-type ctor call on unresolved declaring type

### LINKED ISSUE #3729: Decompilation error for Struct constructors defined in missing libraries
### Input code

Create a new dotnet project called `App.csproj` with this code:
```cs
public class Program
{
	public void TestConstructor()
	{
		Library1.MyStruct c = new Library1.MyStruct(4);
		System.Console.WriteLine("Test new struct: " + c);
	}
}
```
Make a separate library project with this struct:
```cs
namespace Library1
{
	public struct MyStruct
	{
		public int x;
		public MyStruct(int x)
		{
			this.x = x;
		}
	}
}
```

Add a reference from the main `App` project to the `Library1` project, 
then compile the solution, 
then copy ONLY the `App.dll` without the `Library1.dll` to another folder
(this is very important, otherwise the error won't reproduce).
Load the `App.dll` in ILSpy and ensure that `Library1.dll` is not listed in ILSpy's assembly list 
even after expanding the `App.Program` class in the tree view, 
if it does show `Library1`, remove it from the list 
and ensure that dll is not next to the `App.dll` in the same folder.

Then open the `App.dll > Program > TestConstructor` in ILSpy to see the generated code.
It fails in both the desktop app or the vscode extension.

### Erroneous output
```c#
public class Program
{
	public unsafe void TestConstructor()
	{
		MyStruct val = default(MyStruct);
		((MyStruct)(ref val))..ctor(4);
		// ((MyStruct)(ref val))._002Ector(4);  // <<< NOTE: If you save it as code instead of just previewing the code, it will show this line instead.
		Console.WriteLine("Test new struct: " + ((object)(*(MyStruct*)(&val))/*cast due to constrained. prefix*/).ToString());
	}
}
```
Compiler error message:
```
Invalid expression term 'ref'
```
The result is almost right, 
it knows it should use a constructor, 
it knows the type of the variable is `MyStruct`, 
it just fails to write it down correctly.
And it adds the `unsafe` keyword instead of using the struct normally.
This is what was expected:
```cs

public class Program
{
	public void TestConstructor()
	{
		MyStruct val = new MyStruct(4);
		Console.WriteLine("Test new struct: " + val);
	}
}
```

### Details
* Product in use: 
ILSpy version 10.0.1.8346+aad16c66e96eb887eb05887d6b5a9e0522637906
.NET version 10.0.7-servicing.26217.108+b16286c2284fecf303dbc12a0bb152476d662e44
 vscode extension "icsharpcode.ilspy-vscode" Version 0.24.0

If you have problems reproducing this, let me know and I can share the full project and more details.

siegfriedpammer:
Fixes #3729.

A `call` to a value-type constructor is rewritten early in `ILReader.DecodeCall` into `stobj Struct(target, newobj Struct.ctor(...))`, because `Struct.ctor(target, ...)` has no C# equivalent. That rewrite only fired for `TypeKind.Struct`, so when the declaring type comes from an assembly that is not available, the type resolves as `TypeKind.Unknown` and the call fell through to the ordinary call path - the decompiled method then failed with a stack-type mismatch instead of producing `new MyStruct(4)`.

The declaring type's kind is unavailable in that situation, but the IL shape is not: a value-type constructor call is the only case where the receiver of a `call` to a `.ctor` is a managed reference. `TypeKind.Unknown` now takes the same branch when the receiver at stack depth `method.Parameters.Count` is a `ref`, which keeps the rewrite off unresolved *reference*-type constructors (they arrive as `newobj`, and their receiver is not a `ref` anyway).

The check needs to inspect the receiver without consuming it, since `PrepareArguments` runs afterwards, so `PeekStackTypeAtDepth` looks through the expression stack first and then walks the immutable stack behind it, reporting `StackType.Unknown` rather than throwing if the stack is shorter than expected.

New IL-pretty fixture `Issue3729` covers a single-argument ctor, a parameterless one, a multi-argument one, and a ctor called on a field target, all on structs from a missing assembly. Full ICSharpCode.Decompiler.Tests sweep: 3378 total, 0 failed, 46 skipped.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
+++ b/ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
@@ -202,6 +202,9 @@
     <None Include="TestCases\ILPretty\TruncatedAccessorBody.cs" />
     <Compile Remove="TestCases\ILPretty\SealedRecordProtectedCopyCtor.cs" />
     <None Include="TestCases\ILPretty\SealedRecordProtectedCopyCtor.cs" />
+    <Compile Remove="TestCases\ILPretty\Issue3729.cs" />
+    <None Include="TestCases\ILPretty\Issue3729.cs" />
+    <None Include="TestCases\ILPretty\Issue3729.il" />
     <Compile Remove="TestCases\ILPretty\FSharpLoops_Debug.cs" />
     <None Include="TestCases\ILPretty\FSharpLoops_Debug.cs" />
     <Compile Remove="TestCases\ILPretty\FSharpLoops_Release.cs" />
--- a/ICSharpCode.Decompiler.Tests/TestCases/ILPretty/Issue3729.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/ILPretty/Issue3729.cs
@@ -1,0 +1,80 @@
+using System;
+using System.Runtime.CompilerServices;
+using Library1;
+
+namespace ICSharpCode.Decompiler.Tests.TestCases.ILPretty
+{
+	public class Issue3729
+	{
+		private MyStruct structField;
+
+		public void TestSingleArgCtor()
+		{
+			//IL_0003: Unknown result type (might be due to invalid IL or missing references)
+			Console.WriteLine("Test new struct: " + ((object)new MyStruct(4)/*cast due to constrained. prefix*/).ToString());
+		}
+
+		public void TestParameterlessCtor()
+		{
+			//IL_0002: Unknown result type (might be due to invalid IL or missing references)
+			//IL_0007: Unknown result type (might be due to invalid IL or missing references)
+			MyEmptyStruct val = new MyEmptyStruct();
+			Console.WriteLine(val);
+		}
+
+		public void TestMultiArgCtor()
+		{
+			//IL_0011: Unknown result type (might be due to invalid IL or missing references)
+			//IL_0016: Unknown result type (might be due to invalid IL or missing references)
+			MyBigStruct val = new MyBigStruct(1, "hello", 3.14);
+			Console.WriteLine(val);
+		}
+
+		public void TestFieldCtor()
+		{
+			//IL_0007: Unknown result type (might be due to invalid IL or missing references)
+			ref MyStruct reference = ref structField;
+			reference = new MyStruct(5);
+		}
+
+		public unsafe static void TestPointerCtor(void* ptr)
+		{
+			//IL_0002: Unknown result type (might be due to invalid IL or missing references)
+			System.Runtime.CompilerServices.Unsafe.Write(ptr, new MyStruct(6));
+		}
+
+		public void TestArrayElemCtor()
+		{
+			//IL_000f: Unknown result type (might be due to invalid IL or missing references)
+			MyStruct[] array = (MyStruct[])(object)new MyStruct[1] {
+				new MyStruct(7)
+			};
+		}
+
+		public void TestGenericStructCtor()
+		{
+			//IL_0003: Unknown result type (might be due to invalid IL or missing references)
+			//IL_0008: Unknown result type (might be due to invalid IL or missing references)
+			MyGenericStruct<int> val = new MyGenericStruct<int>(4);
+			Console.WriteLine(val);
+		}
+
+		public void TestRefTypeNewobj()
+		{
+			//IL_0000: Unknown result type (might be due to invalid IL or missing references)
+			//IL_0006: Expected O, but got Unknown
+			MyClass value = new MyClass();
+			Console.WriteLine(value);
+		}
+	}
+	public class Issue3729_DerivedFromUnknown : MissingBase
+	{
+	}
+	public class Issue3729_DerivedFromUnknownWithArgs : MissingBase
+	{
+		public Issue3729_DerivedFromUnknownWithArgs()
+			: base(42)
+		{
+		}
+	}
+}
--- a/ICSharpCode.Decompiler.Tests/TestCases/ILPretty/Issue3729.il
+++ b/ICSharpCode.Decompiler.Tests/TestCases/ILPretty/Issue3729.il
@@ -1,0 +1,228 @@
+// Reproducer for issue #3729:
+// Decompilation of a struct constructor call where the struct's defining
+// assembly (Library1) cannot be resolved at decompile time.
+// Mirrors what csc emits for `new Library1.MyStruct(4)` followed by
+// string concat with the struct value.
+
+.assembly extern Library1
+{
+}
+.assembly extern System.Runtime
+{
+  .publickeytoken = (B0 3F 5F 7F 11 D5 0A 3A )
+  .ver 4:0:0:0
+}
+.assembly extern System.Console
+{
+  .publickeytoken = (B0 3F 5F 7F 11 D5 0A 3A )
+  .ver 4:0:0:0
+}
+.assembly Issue3729
+{
+  .ver 1:0:0:0
+}
+.module Issue3729.dll
+
+.class public auto ansi beforefieldinit ICSharpCode.Decompiler.Tests.TestCases.ILPretty.Issue3729
+    extends [System.Runtime]System.Object
+{
+    .field private valuetype [Library1]Library1.MyStruct structField
+
+    // single-int-arg struct ctor with subsequent constrained.callvirt ToString()
+    .method public hidebysig
+        instance void TestSingleArgCtor () cil managed
+    {
+        .maxstack 2
+        .locals init (
+            [0] valuetype [Library1]Library1.MyStruct
+        )
+
+        IL_0000: ldloca.s 0
+        IL_0002: ldc.i4.4
+        IL_0003: call instance void [Library1]Library1.MyStruct::.ctor(int32)
+        IL_0008: ldstr "Test new struct: "
+        IL_000d: ldloca.s 0
+        IL_000f: constrained. [Library1]Library1.MyStruct
+        IL_0015: callvirt instance string [System.Runtime]System.Object::ToString()
+        IL_001a: call string [System.Runtime]System.String::Concat(string, string)
+        IL_001f: call void [System.Console]System.Console::WriteLine(string)
+        IL_0024: ret
+    }
+
+    // explicit parameterless struct ctor (C# 10+ feature in source)
+    // Single-argument ctor: the receiver sits directly below the argument.
+    .method public hidebysig
+        instance void TestParameterlessCtor () cil managed
+    {
+        .maxstack 1
+        .locals init (
+            [0] valuetype [Library1]Library1.MyEmptyStruct
+        )
+
+        IL_0000: ldloca.s 0
+        IL_0002: call instance void [Library1]Library1.MyEmptyStruct::.ctor()
+        IL_0007: ldloc.0
+        IL_0008: box [Library1]Library1.MyEmptyStruct
+        IL_000d: call void [System.Console]System.Console::WriteLine(object)
+        IL_0012: ret
+    }
+
+    // Multi-argument ctor with mixed parameter types: the receiver sits below three arguments
+    .method public hidebysig
+        instance void TestMultiArgCtor () cil managed
+    {
+        .maxstack 5
+        .locals init (
+            [0] valuetype [Library1]Library1.MyBigStruct
+        )
+
+        IL_0000: ldloca.s 0
+        IL_0002: ldc.i4.1
+        IL_0003: ldstr "hello"
+        IL_0008: ldc.r8 3.14
+        IL_0011: call instance void [Library1]Library1.MyBigStruct::.ctor(int32, string, float64)
+        IL_0016: ldloc.0
+        IL_0017: box [Library1]Library1.MyBigStruct
+        IL_001c: call void [System.Console]System.Console::WriteLine(object)
+        IL_0021: ret
+    }
+
+    // struct field init via `ldflda + call .ctor`. Receiver is the field
+    // address (StackType.Ref) - same shape as `ldloca`, just sourced from
+    // an instance field instead of a local.
+    .method public hidebysig
+        instance void TestFieldCtor () cil managed
+    {
+        .maxstack 2
+
+        IL_0000: ldarg.0
+        IL_0001: ldflda valuetype [Library1]Library1.MyStruct ICSharpCode.Decompiler.Tests.TestCases.ILPretty.Issue3729::structField
+        IL_0006: ldc.i4.5
+        IL_0007: call instance void [Library1]Library1.MyStruct::.ctor(int32)
+        IL_000c: ret
+    }
+
+    // Unsafe code passes the target as a pointer, so the receiver is a native
+    // pointer (StackType.I) rather than a managed reference.
+    .method public hidebysig static
+        void TestPointerCtor (
+            void* ptr
+        ) cil managed
+    {
+        .maxstack 2
+
+        IL_0000: ldarg.0
+        IL_0001: ldc.i4.6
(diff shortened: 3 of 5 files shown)


ICSharpCode.Decompiler/IL/ILReader.cs
```diff
@@ -1767,12 +1785,17 @@ DecodedInstruction DecodeCall(OpCode opCode)
 				}
 				case TypeKind.Struct when method.IsConstructor && !method.IsStatic && opCode == OpCode.Call
 					&& method.ReturnType.Kind == TypeKind.Void:
+				case TypeKind.Unknown when method.IsConstructor && !method.IsStatic && opCode == OpCode.Call
```
christophwille: **[PLAUSIBLE] correctness** - The new `TypeKind.Unknown` arm never consults `method.DeclaringType.IsReferenceType`, so a ctor call on an unknown type that metadata marks as a class (TypeSpec with `class` prefix, `IsReferenceType == true`) is silently rewritten into a struct-init StObj whenever the receiver slot reports `StackType.Ref`.

Scenario: obfuscated/merged IL where a Ref-typed slot carries an object reference (`ImportedBlock.MergeStackTypes` takes the higher StackType: Unknown=0 < O=6 < Ref=7, so O-vs-Ref and Unknown-vs-Ref merges yield Ref): `call instance void class [Missing]Base::.ctor()` is rewritten to `*target = new Base()`, the base-ctor rendering is lost, and no Warn fires because PopPointer's Ref path is warning-free.

One-clause hardening: `&& method.DeclaringType.IsReferenceType != true`. The existing `PopFieldTarget`/`ExpectedTypeForThisPointer` machinery already encodes this question and could be reused.

ICSharpCode.Decompiler/IL/ILReader.cs
```diff
@@ -1767,12 +1785,17 @@ DecodedInstruction DecodeCall(OpCode opCode)
...
+					&& method.ReturnType.Kind == TypeKind.Void
+					&& PeekStackTypeAtDepth(method.Parameters.Count) == StackType.Ref:
```
christophwille: **[CONFIRMED] correctness** - The `== StackType.Ref` guard excludes `StackType.I` receivers, which the resolved `TypeKind.Struct` arm accepts (`PopStObjTarget` -> `PopPointer` returns I unchanged), so #3729 remains unfixed for pointer-target value-type ctor calls.

Scenario: C++/CLI or unsafe-code IL doing `ldloc ptr; ldc.i4.4; call instance void [Missing]MyStruct::.ctor(int32)` (receiver is a native pointer, `StackType.I`): the guard is false, the call falls to the default path, and the same broken `.ctor`-as-member output the PR fixes reappears - while the identical IL with a *resolved* struct decompiles correctly.

Fix: `is StackType.Ref or StackType.I` (optionally combined with `IsReferenceType == false`).

ICSharpCode.Decompiler/IL/ILReader.cs
```diff
@@ -1767,12 +1785,17 @@ DecodedInstruction DecodeCall(OpCode opCode)
 				}
 				case TypeKind.Struct when method.IsConstructor && !method.IsStatic && opCode == OpCode.Call
 					&& method.ReturnType.Kind == TypeKind.Void:
+				case TypeKind.Unknown when method.IsConstructor && !method.IsStatic && opCode == OpCode.Call
+					&& method.ReturnType.Kind == TypeKind.Void
```
christophwille: **[CONFIRMED] simplification** - The new case label copy-pastes all four conditions of the `TypeKind.Struct` arm (`method.IsConstructor && !method.IsStatic && opCode == OpCode.Call && method.ReturnType.Kind == TypeKind.Void`), differing only in the trailing peek clause. A future tightening of the ctor-detection predicate applied to one label and not the other makes the unresolved path silently more permissive than the resolved one, with no test covering the difference. Consider factoring the shared predicate into one local function next to `PrepareArguments`.

ICSharpCode.Decompiler/IL/ILReader.cs
```diff
@@ -1286,6 +1286,24 @@ StackType PeekStackType()
 				return currentStack.Peek().StackType;
 		}
 
+		StackType PeekStackTypeAtDepth(int depth)
```
christophwille, line 1281: **[CONFIRMED] simplification** - `PeekStackType()` is now a byte-for-byte duplicate of `PeekStackTypeAtDepth(0)` (verified branch-by-branch, including both empty-stack Unknown fallbacks), leaving two hand-maintained copies of the expressionStack/currentStack layering logic ten lines apart. `PeekStackType` has 6 call sites (including `PopFieldTarget`, which makes decisions about the same stack slot the new ctor guard reads); a future fix landing in only one copy makes them disagree about slot 0 on the same instruction. Suggest: `StackType PeekStackType() => PeekStackTypeAtDepth(0);`

ICSharpCode.Decompiler.Tests/TestCases/ILPretty/Issue3729.cs
```diff
@@ -0,0 +1,62 @@
+using System;
```
christophwille, line 1: **[CONFIRMED] test-coverage** - Every instruction produced by the new arm has `ResultType` Unknown and therefore triggers ILReader's Warn ("Unknown result type"), which surfaces as `//IL_xxxx: Unknown result type ...` comments in real output (cf. `GuessAccessors.cs`) - but this expected file omits them and only passes because `CodeAssert.NormalizeLine` strips `//` comments. Anyone regenerating the file gets a diff full of warning comments, and the file conceals that the fix newly emits a spurious warning above every `new MyStruct(4)` that real ILSpy users will see. Either suppress the warning for this deliberate rewrite or check in the comments as `GuessAccessors.cs` does.

ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
```diff
@@ -202,6 +202,8 @@
     <None Include="TestCases\ILPretty\TruncatedAccessorBody.cs" />
     <Compile Remove="TestCases\ILPretty\SealedRecordProtectedCopyCtor.cs" />
     <None Include="TestCases\ILPretty\SealedRecordProtectedCopyCtor.cs" />
+    <Compile Remove="TestCases\ILPretty\Issue3729.cs" />
+    <None Include="TestCases\ILPretty\Issue3729.cs" />
```
christophwille, line 206: **[CONFIRMED] project-hygiene** - `Issue3729.il` is never added to the project: the PR adds the `Compile Remove`/`None Include` pair for `Issue3729.cs` but no `<None Include="TestCases\ILPretty\Issue3729.il" />`, although `EnableDefaultNoneItems=false` deliberately keeps all `.il` inputs explicit and the other ILPretty `.il` files are listed. The test passes anyway (the runner reads the `.il` from disk), so nothing catches the omission - the sole input file of the new test is invisible in Solution Explorer and to project-item-driven tooling.

ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
christophwille: **[CONFIRMED] hygiene** - An accidental UTF-8 BOM is added to the first line, contradicting `.editorconfig` (standardize on no BOM) and the repo state (this file and all main csproj siblings are BOM-less on master). It causes unrelated line-1 churn: git blame on line 1 permanently points at this PR, and any concurrent PR touching this frequently-edited csproj conflicts on the encoding flip. Re-save without BOM.

ICSharpCode.Decompiler.Tests/TestCases/ILPretty/Issue3729.il
```diff
@@ -0,0 +1,213 @@
...
+        IL_000d: ldloca.s 0
+        IL_000f: constrained. [Library1]Library1.MyStruct
+        IL_0015: callvirt instance string [System.Runtime]System.Object::ToString()
+        IL_001a: call string [System.Runtime]System.String::Concat(string, string)
+        IL_001f: call void [System.Console]System.Console::WriteLine(string)
+        IL_0024: ret
+    }
+
+    // explicit parameterless struct ctor (C# 10+ feature in source)
+    // exercises the fix at PeekStackTypeAtDepth(0) — receiver right at top.
```
christophwille: **[CONFIRMED] conventions** - Fixture comments violate two CLAUDE.md rules: nine lines contain U+2014 em-dashes ("ASCII-only in code and comments"; all 58 existing ILPretty `.il` files are pure ASCII) and three comments are diff-relative ("exercises the fix at...", "the new arm should fire here too") against "Comments must stand on their own, with no memory of how the code was written" - "the fix"/"the new arm" are meaningless to a reader once the PR context is gone. Replace em-dashes with ASCII and reword the comments to describe present behavior.

christophwille (commented):
## Review summary

**What the PR does.** In `ILReader.cs`, `DecodeCall` already rewrote `call Struct::.ctor(target, ...)` into `stobj(target, newobj ...)` for resolved structs. The PR adds a `TypeKind.Unknown` arm gated on the receiver's stack slot being `StackType.Ref`, via a new non-consuming `PeekStackTypeAtDepth` helper, plus an ILPretty fixture (`Issue3729.il`/`.cs`) covering ctor arity 0/1/3, `ldloca`/`ldflda`/`ldelema` receivers, a generic TypeSpec case, and negative cases for unresolved base-class ctors.

**What holds up.** The core is sound: the helper's index/skip arithmetic is correct; `depth = method.Parameters.Count` does address the receiver; the generic TypeSpec case genuinely reaches the new arm (`ParameterizedType.Kind` delegates to `UnknownType.Kind`); the `StackType.O` fall-through keeps `: base(...)` rendering intact; no measurable hot-path cost or allocation; `CallBuilder.HandleConstructorCall` handles unresolved types without `GetDefinition()` derefs. The commit-message rationale (TypeRef parents carry no valuetype bit, so a type-system fix is unavailable for the primary case) is correct.

**Main weaknesses.** (1) The guard trusts only the stack slot and never the tri-state `method.DeclaringType.IsReferenceType`, which metadata does populate for TypeSpec-sourced references - so it both misses provable cases (`IsReferenceType == false` with a pointer receiver) and can silently misfire on `IsReferenceType == true` (inline comments below). (2) The new shape carries `ResultType == StackType.Unknown` downstream, which trips the reader's "Unknown result type" warning on every rewritten call and opens a debug-assert path in `StObjToStLoc` (see below). (3) A deeper root cause is worth a follow-up issue rather than this PR: `TypeUtils.GetStackType` returns `StackType.Unknown` for `IsReferenceType == false` unknowns, which is also why the expected output enshrines artifacts like the spurious `ref` local in `TestFieldCtor` and the `(MyStruct[])(object)` casts.

### Findings outside the diff

**`ICSharpCode.Decompiler/IL/Transforms/EarlyExpressionTransforms.cs:113` [PLAUSIBLE]** - The new rewrite feeds `StObjToStLoc` a `stobj Unknown(ldloca v, newobj ...)` that converts to `stloc v(newobj)` even when local `v` has a resolved type, tripping StLoc's `Debug.Assert(value.ResultType == variable.StackType)` (StLoc.cs:44), since `NewObj.ResultType` is Unknown. Repro shape: hand-written/obfuscated IL with `ldloca.s 0` on an `int32` local followed by `call instance void [Missing]MyStruct::.ctor(...)` - the new arm fires (slot type is Ref regardless of the local's type), `IsCompatibleTypeForMemoryAccess` short-circuits true on Unknown (TypeUtils.cs:257), the compensating Conv is skipped, and `function.CheckInvariant` after EarlyExpressionTransforms asserts in debug builds (silently bogus `stloc int_local(newobj ...)` in release). Pre-PR this IL produced a plain Call and never reached StObjToStLoc.

**Commit conventions [CONFIRMED]** - Both PR commits are AI-assisted (PR body says "Generated with Claude Code") but carry no `Assisted-by:` trailer required by CLAUDE.md; the test commit's subject is 75 chars (target <= 72) and both bodies contain em-dashes plus a per-test enumeration CLAUDE.md tells authors to omit.

**Minor (below the finding cap)** - the `TestRefTypeNewobj` control case duplicates `UnknownTypes.il`; the `DerivedFromUnknown` classes largely duplicate `Issue2443`.

ICSharpCode.Decompiler/IL/ILReader.cs
siegfriedpammer: Added: the unresolved arm now also requires `method.DeclaringType.IsReferenceType != true`, so metadata that does state the type is a reference type keeps the plain call.

_Reply written by an AI agent (Claude) on Siegfried's behalf._

ICSharpCode.Decompiler/IL/ILReader.cs
siegfriedpammer: Fixed: the receiver check is now `is StackType.Ref or StackType.I`, and the fixture gained `TestPointerCtor` with a `void*` receiver. It decompiles to `Unsafe.Write(ptr, new MyStruct(6))` instead of the broken `.ctor` member call.

_Reply written by an AI agent (Claude) on Siegfried's behalf._

ICSharpCode.Decompiler/IL/ILReader.cs
siegfriedpammer: Factored into `IsValueTypeCtorCall()` next to `PrepareArguments`; both case labels call it, and the unresolved one adds `ReceiverLooksLikeValueTypeTarget()`.

_Reply written by an AI agent (Claude) on Siegfried's behalf._

ICSharpCode.Decompiler/IL/ILReader.cs
siegfriedpammer: Fixed: `PeekStackType() => PeekStackTypeAtDepth(0);`.

_Reply written by an AI agent (Claude) on Siegfried's behalf._

ICSharpCode.Decompiler.Tests/TestCases/ILPretty/Issue3729.cs
siegfriedpammer: Right about the omission: the expected file now carries the `//IL_xxxx: Unknown result type` comments, 11 of them, as the real output emits. Suppressing the warning would be a behaviour change wider than this fix, so the `StackType.Unknown` modelling behind it is left for a separate issue.

_Reply written by an AI agent (Claude) on Siegfried's behalf._

ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
siegfriedpammer: Added `<None Include="TestCases\ILPretty\Issue3729.il" />`.

_Reply written by an AI agent (Claude) on Siegfried's behalf._

ICSharpCode.Decompiler.Tests/ICSharpCode.Decompiler.Tests.csproj
siegfriedpammer: Removed; the file starts with `3c 3f 78` again. It came from resolving a rebase conflict with a script that wrote UTF-8 with BOM.

_Reply written by an AI agent (Claude) on Siegfried's behalf._

ICSharpCode.Decompiler.Tests/TestCases/ILPretty/Issue3729.il
siegfriedpammer: Fixed: no non-ASCII characters left in the fixture, and the diff-relative comments now describe the IL shapes themselves.

_Reply written by an AI agent (Claude) on Siegfried's behalf._

dgrunwald (approved):

PR #4002, opened 2026-08-14
Give dark-theme syntax colors a perceptual contrast floor

christophwille:
Addresses the dark-mode highlighting reports in #3986 -- [the XML screenshots](https://github.com/icsharpcode/ILSpy/issues/3986#issuecomment-5277881451) and [the About-page link](https://github.com/icsharpcode/ILSpy/issues/3986#issuecomment-5290216567). The other topics in that issue are untouched, so it is referenced rather than closed.

## How dark syntax colors are derived today

`ThemeManager.ApplyHighlightingColors` is the single decision point:

- `definition.Name == "C#"` gets the hand-authored `SyntaxColorPalettes.CSharpDark`.
- **Everything else** -- `XML-Mode.xshd`, `ILAsm-Mode.xshd`, `Asm-Mode.xshd`, and every AvaloniaEdit built-in definition (JSON, Markdown, JS, HTML, CSS, Python) -- falls back to the algorithmic conversion `GetColorForDarkTheme` -> `AdjustForDarkTheme`, which inverts HSL *lightness* (`l = 1 - l^1.2`) and then softens saturation.

The same math is also applied per paint by `ThemeAwareHighlightingColorizer` for definitions that never register, so both routes share one function.

## Problem 1: HSL lightness is not perceptual luminance

The inversion produces a nominally "light" color, but relative luminance is hue-weighted (0.2126 R / 0.7152 G / 0.0722 B). Greens, cyans and yellows came out fine; blue -- weight 0.0722 -- did not. That is why only the blue tokens were reported.

A second, worse case: `1 - l^1.2` maps `l = 0.75` to `0.29`, so a source color that is *already light* inverts **downwards**, into the background.

| xshd source | used by | derived before | contrast vs `#1E1E1E` |
|---|---|---|---|
| `Blue` `#0000FF` | XML `AttributeValue` / `XmlDeclaration` / `CData` / `DocType`, IL `Instructions` + `Keywords`, Asm `Instructions` | `#6F6FEA` | **4.08** -- the reported bug |
| `#8080FF` | Asm `Registers` | `#16169B` | **1.29** -- effectively invisible |

## Problem 2: the desaturation guard misses colors that invert to light

`if (s > 0.75f && l < 0.75f)` only softened saturation when the *inverted* lightness stayed dark. A dark, fully saturated source inverts to a light, still fully saturated color -- exactly the neon the softening exists to prevent. XML's `XmlTag` (`DarkMagenta`, inverted lightness ~0.79) came back as `#FF93FF`.

## Problem 3: hyperlinks were never themed at all

The About page renders links through `ResourceLinkGenerator` -> AvaloniaEdit `VisualLineLinkText`, whose foreground comes from `TextView.LinkTextForegroundBrush`. Nothing in ILSpy ever set it, so it stayed at AvaloniaEdit's registered default of pure blue: **1.94:1** on `#1E1E1E`. `MetadataColumnBuilder`'s token cells had the same problem via a hardcoded `Brushes.Blue`.

## The fix

**`ThemeManager.AdjustForDarkTheme`** (dark-only path; the light theme is unreachable from here and byte-identical after the change):

1. Make the saturation softening unconditional (the paired lightness lift stays scoped to over-saturated colors, where it was: it is not monotone across its 0.75 boundary, so applying it to greys would reorder neighboring ones).
2. After the inversion, bisect HSL lightness to a **5.5:1 WCAG floor**, preserving hue and saturation so the token keeps its identity.

5.5 rather than the 4.5 AA threshold: 4.5 was measured and only moves the reported blue to 4.51. 5.5 is where the hand-authored `CSharpDark` values already sit (5.0-8.8) and matches VS Code's own dark keyword blue (5.65).

The floor is **foreground-only**. `AdjustForDarkTheme` is called for both `Foreground` and `Background`; forcing a span *background* to contrast with the canvas would repaint `Asm-Mode.xshd`'s `#EEEEEE` `Registers` background as a bright block and bury the text drawn on top of it.

It is also measured against **the surface the foreground actually lands on**: a `HighlightingColor` that declares its own background is converted background-first, and that converted background's luminance becomes the reference; only a color with no background of its own is measured against the editor canvas. Because a light span background needs the foreground to move *down*, the search heads for whichever lightness extreme offers more contrast (white above the dark canvas, black under a light background) rather than only lifting. A `foreground="White" background="Navy"` span converts to `#000000` on `#AAAAF2` (9.74:1); measuring both against the canvas independently would have produced a mid grey on light blue (1.39:1).

No per-language dark palette was added: the floor fixes XML, IL, Asm **and** every AvaloniaEdit built-in in the one place they all route through, which an `XmlDark` palette would not -- the built-ins have no palette to author.

### Before / after

Every distinct foreground shipped by the repo's own `.xshd` files, contrast measured against `#1E1E1E`:

| token | source | before | CR | after | CR |
|---|---|---|---|---|---|
| `Blue` | `#0000FF` | `#6F6FEA` | 4.08 | `#8A8AEE` | **5.52** |
| `#8080FF` | `#8080FF` | `#16169B` | **1.29** | `#8A8AEE` | **5.52** |
| `Brown` | `#A52A2A` | `#DB7575` | 5.41 | `#DC7777` | **5.52** |
| `DarkBlue` | `#00008B` | `#9393FF` | 6.24 | `#A1A1F1` | 7.05 |
| `DarkMagenta` | `#8B008B` | `#FF93FF` | 8.63 | `#F1A1F1` | 8.82 |
| `Green` | `#008000` | `#9DFF9D` | 13.67 | `#AAF2AA` | 12.69 |
| `Teal` | `#008080` | `#9DFFFF` | 14.43 | `#AAF2F2` | 13.25 |
| `Olive` | `#808000` | `#FFFF9D` | 15.91 | `#F2F2AA` | 14.33 |
| `Red` | `#FF0000` | `#EA6F6F` | 5.56 | `#EA6F6F` | 5.56 |
| `Magenta` | `#FF00FF` | `#EA6FEA` | 6.32 | `#EA6FEA` | 6.32 |
| `Orange` | `#FFA500` | `#EABE6F` | 9.60 | `#EABE6F` | 9.60 |
| `#0080C0` | `#0080C0` | `#B2DEF4` | 11.63 | `#B2DEF4` | 11.63 |
| `Black` | `#000000` | `#FFFFFF` | 16.67 | `#FFFFFF` | 16.67 |
| *background* `#EEEEEE` | Asm `Registers` bg | `#141414` | 1.11 | `#141414` | 1.11 (floor not applied) |

The two broken rows are fixed; `Brown` clears the floor it was just under; the fully saturated rows lose their neon edge; the rest, backgrounds included, are untouched.

**Link color** (`App.axaml`, `MetadataColumnBuilder.cs`):

- New `ILSpy.LinkForeground` in both theme dictionaries: Light `#0066CC` (5.47 on the `#FFFFE1` canvas -- the value `DocumentationRenderer` had already picked), Dark `#4DA6FF` (6.52 on `#1E1E1E`).
- Applied via a `aeRendering|TextView` style, since `LinkTextForegroundBrush` is a `StyledProperty`. No C# needed, and it covers the About page plus every other hyperlink in a decompiler view.
- `MetadataColumnBuilder`'s token cells carry a `link` class instead of `Brushes.Blue`, with `DataGridRow Button.link` / `DataGridRow:selected Button.link` styles in `App.axaml`. `Foreground` is inherited, so a local brush on the button would outrank the existing `DataGridRow:selected` white override -- and cell navigation selects the row, where neither link color has usable contrast on the accent fill.

## Tests

`ILSpy.Tests/Themes/ThemeManagerTests.cs`:

- Data-driven over every distinct *named* foreground shipped in the four repo `.xshd` files, asserting the converted foreground clears the floor. Red before the change on `Blue` (4.08), `#8080FF` (1.29) and `Brown` (5.41).
- A background brush is *not* pushed to the floor (guards the `Registers` case above).
- A foreground+background pair is measured against its own background, not the canvas (red at 1.39:1 against the canvas reference).
- `DarkMagenta` must not come back with a fully saturated channel.
- `#505050` must not invert brighter than the lighter `#525252` -- pins the lightness lift to over-saturated colors.
- The `DarkEditorBackground` constant the floor measures against must equal the Dark `ILSpy.EditorBackground` resource in `App.axaml`, so retuning the canvas cannot leave the floor measuring a stale value with every test still green.

The contrast math in the test computes WCAG luminance independently of the production code, so the tests do not validate themselves.

`ILSpy.Tests/Themes/LinkColorTests.cs` (new, headless): the hosted editor's `TextView.LinkTextForegroundBrush` is not AvaloniaEdit's default blue, and it changes between the Light and Dark variants. Both were confirmed red by removing the `App.axaml` style -- a typo in the selector or the `xmlns` would otherwise fall back to the unreadable default silently.

`ILSpy.Tests/Metadata/MetadataColumnBuilderTests.cs`: a mounted grid's token cell takes the themed link color, and goes white once its row is selected. Red against the previous local-`Foreground` version.

Full `ILSpy.Tests` run: 1200 passed, 3 skipped (pre-existing opt-in visual/perf cases).

## Deliberately out of scope

`OpenFromNuGetFeedDialog.axaml` hardcodes a `#1B6EC2` link (~3.1:1 on the dark window) and is a candidate for `ILSpy.LinkForeground` in its own right.

`DocumentationRenderer.cs` hardcodes both its link brush (`#0066CC`) and a light popup background (`#FCFCFC` / `#AAAAAA`). Its link is *correct* today precisely because that popup stays light in dark mode; fixing it properly means fixing the popup's chrome, which is a separate dark-mode job worth its own issue. It paints the definition's own (dark-converted) named colors inside that light popup, so the floor makes IL/ReadyToRun tooltips slightly worse there -- the same condition C# tooltips already have with `CSharpDark`.

### DIFF
--- a/ILSpy.Tests/Themes/ThemeManagerTests.cs
+++ b/ILSpy.Tests/Themes/ThemeManagerTests.cs
@@ -19,6 +19,10 @@
 using System;
 using System.Collections.Generic;
 
+// ResourceNodeExtensions.TryFindResource lives in Avalonia.Controls.
+using Avalonia.Controls;
+using Avalonia.Headless.NUnit;
+
 using AvaloniaEdit.Highlighting;
 
 using AwesomeAssertions;
@@ -135,6 +139,134 @@ public class ThemeManagerTests
 
 		remapped.FontWeight.Should().Be((AvFontWeight?)AvFontWeight.Bold);
 		remapped.FontStyle.Should().Be((AvFontStyle?)AvFontStyle.Italic);
+	}
+
+	// Every distinct *named* foreground shipped by the repo's own XSHDs (ILSpy/TextView/*.xshd).
+	// Colours declared inline on a rule are anonymous, never land in NamedHighlightingColors and
+	// so are never converted at all. None of the definitions declares ILSpy.IsThemeAware and only
+	// "C#" has a hand-authored dark palette, so all of these reach the algorithmic conversion.
+	static readonly string[] ShippedXshdForegrounds = {
+		// XML-Mode.xshd
+		"Green", "Blue", "DarkMagenta", "Red", "Teal", "Olive",
+		// ILAsm-Mode.xshd
+		"Magenta",
+		// Asm-Mode.xshd
+		"Orange", "#0080C0", "Brown", "#8080FF", "DarkBlue",
+	};
+
+	[Test]
+	[TestCaseSource(nameof(ShippedXshdForegrounds))]
+	public void GetColorForDarkTheme_Lifts_Every_Shipped_Foreground_To_The_Contrast_Floor(string light)
+	{
+		// The reported symptom in #3986: XSHD colours surviving the HSL inversion with too
+		// little perceptual contrast against the dark editor canvas. HSL lightness is not
+		// luminance, so without a floor the hue decides the outcome: plain Blue lands at
+		// 4.08:1, and the already-light #8080FF inverts *downwards* to 1.29:1 (invisible).
+		var color = new HighlightingColor { Foreground = new SimpleHighlightingBrush(AvColor.Parse(light)) };
+
+		var remapped = ThemeManager.GetColorForDarkTheme(color);
+
+		ContrastAgainstDarkEditor(ColorOf(remapped.Foreground!))
+			.Should().BeGreaterThanOrEqualTo(ThemeManager.MinimumDarkContrastRatio,
+				$"'{light}' must stay readable on the dark editor background");
+	}
+
+	[Test]
+	public void GetColorForDarkTheme_Leaves_Backgrounds_Below_The_Contrast_Floor()
+	{
+		// The floor is a *foreground* rule. Asm-Mode.xshd gives the Registers token a light
+		// #EEEEEE background; forcing that to contrast with the canvas would repaint it as a
+		// bright block and bury the foreground drawn on top of it.
+		var color = new HighlightingColor { Background = new SimpleHighlightingBrush(AvColor.Parse("#EEEEEE")) };
+
+		var remapped = ThemeManager.GetColorForDarkTheme(color);
+
+		ContrastAgainstDarkEditor(ColorOf(remapped.Background!))
+			.Should().BeLessThan(ThemeManager.MinimumDarkContrastRatio);
+	}
+
+	[Test]
+	public void GetColorForDarkTheme_Softens_Saturation_Of_Colors_That_Invert_To_Light()
+	{
+		// DarkMagenta inverts to lightness ~0.79. Saturation softening has to apply wherever the
+		// inverted lightness lands, or a dark fully saturated source comes back as light and
+		// still fully saturated: neon #FF93FF, exactly what the softening exists to prevent.
+		var color = new HighlightingColor { Foreground = new SimpleHighlightingBrush(AvColors.DarkMagenta) };
+
+		var remapped = ThemeManager.GetColorForDarkTheme(color);
+
+		var fg = ColorOf(remapped.Foreground!);
+		Math.Max(fg.R, Math.Max(fg.G, fg.B)).Should().BeLessThan(255,
+			"a fully saturated channel means the softening was skipped");
+	}
+
+	[Test]
+	public void GetColorForDarkTheme_Measures_The_Foreground_Against_Its_Own_Span_Background()
+	{
+		// A colour that carries both is painted on its own background, not on the editor canvas.
+		// White-on-Navy is the pathological case: both invert across the canvas, so measuring the
+		// foreground against the canvas would push it to a mid grey sitting on a light blue block.
+		var color = new HighlightingColor {
+			Foreground = new SimpleHighlightingBrush(AvColors.White),
+			Background = new SimpleHighlightingBrush(AvColors.Navy),
+		};
+
+		var remapped = ThemeManager.GetColorForDarkTheme(color);
+
+		Contrast(ColorOf(remapped.Foreground!), ColorOf(remapped.Background!))
+			.Should().BeGreaterThanOrEqualTo(ThemeManager.MinimumDarkContrastRatio,
+				"the span background is the surface the foreground has to read against");
+	}
+
+	[Test]
+	public void GetColorForDarkTheme_Keeps_Neighboring_Greys_In_Order()
+	{
+		// The lightness lift that pairs with the saturation softening is not monotone across its
+		// 0.75 boundary, so it must stay scoped to over-saturated colours. Applied to greys it
+		// would map #505050 brighter than the lighter #525252.
+		var darker = ColorOf(ThemeManager.GetColorForDarkTheme(Foreground("#505050")).Foreground!);
+		var lighter = ColorOf(ThemeManager.GetColorForDarkTheme(Foreground("#525252")).Foreground!);
+
+		darker.R.Should().BeGreaterThanOrEqualTo(lighter.R,
+			"inversion has to preserve the ordering of the source colours");
+	}
+
+	[AvaloniaTest]
+	public void DarkEditorBackground_Constant_Tracks_The_Theme_Resource()
+	{
+		// The floor is measured against a constant copy of the dark canvas so the conversion
+		// stays pure static math. Retuning ILSpy.EditorBackground in App.axaml without updating
+		// the constant would leave every floor test green while the shipped colours drift below
+		// the real floor -- Blue clears it by 0.02.
+		var window = new global::Avalonia.Controls.Window();
+
+		window.TryFindResource("ILSpy.EditorBackground", global::Avalonia.Styling.ThemeVariant.Dark, out var resource)
+			.Should().BeTrue("the dark theme dictionary defines the editor canvas");
+		(resource as global::Avalonia.Media.ISolidColorBrush)?.Color
+			.Should().Be(ThemeManager.DarkEditorBackground);
+	}
+
+	static HighlightingColor Foreground(string color)
+		=> new() { Foreground = new SimpleHighlightingBrush(AvColor.Parse(color)) };
+
+	static double ContrastAgainstDarkEditor(AvColor color)
+		=> Contrast(color, ThemeManager.DarkEditorBackground);
+
+	// WCAG relative-luminance contrast, computed independently of the production code so the
+	// tests don't validate themselves.
+	static double Contrast(AvColor color, AvColor surface)
+	{
+		var (a, b) = (Luminance(color) + 0.05, Luminance(surface) + 0.05);
+		return a > b ? a / b : b / a;
+
+		static double Luminance(AvColor c)
+			=> 0.2126 * Channel(c.R) + 0.7152 * Channel(c.G) + 0.0722 * Channel(c.B);
+
+		static double Channel(byte value)
+		{
+			var v = value / 255.0;
+			return v <= 0.03928 ? v / 12.92 : Math.Pow((v + 0.055) / 1.055, 2.4);
+		}
 	}
 
 	static AvColor ColorOf(HighlightingBrush brush)
--- a/ILSpy/Metadata/MetadataColumnBuilder.cs
+++ b/ILSpy/Metadata/MetadataColumnBuilder.cs
@@ -434,9 +434,12 @@ namespace ICSharpCode.ILSpy.Metadata
 					Padding = new Thickness(2, 0),
 					HorizontalAlignment = HorizontalAlignment.Left,
 					VerticalAlignment = VerticalAlignment.Center,
-					Foreground = Brushes.Blue,
 					Cursor = new global::Avalonia.Input.Cursor(global::Avalonia.Input.StandardCursorType.Hand),
 					Content = FormatTokenValue(row, prop, format),
+					// Token cells are links. The colour comes from the "Button.link" styles in
+					// App.axaml rather than a local Foreground, so the selected row's white
+					// override still wins over it -- neither link colour reads on the accent fill.
+					Classes = { "link" },
 				};
 				btn.Click += (_, _) => {
 					if (btn.FindAncestorOfType<Views.MetadataTablePage>()?.DataContext is MetadataTablePageModel page)
--- a/ILSpy/Themes/ThemeManager.cs
+++ b/ILSpy/Themes/ThemeManager.cs
@@ -41,6 +41,25 @@ namespace ICSharpCode.ILSpy.Themes
 		// declares its own dark palette in two variants).
 		const string IsThemeAwareKey = "ILSpy.IsThemeAware";
 
+		/// <summary>
+		/// Minimum WCAG contrast ratio a dark-converted FOREGROUND must reach against the surface
+		/// it is painted on. 5.5 sits where the hand-authored <see cref="SyntaxColorPalettes.CSharpDark"/>
+		/// values already live (5.0-8.8) and matches VS Code's own dark keyword blue (5.65).
+		/// The 4.5 WCAG AA threshold is not enough here: it leaves plain Blue at 4.51, which
+		/// still reads as too dark against the editor background.
+		/// </summary>
+		internal const double MinimumDarkContrastRatio = 5.5;
+
+		// The dark editor canvas a foreground is measured against when its colour declares no
+		// span background of its own. Mirrors ILSpy.EditorBackground in the Dark theme dictionary
+		// of App.axaml, which ThemeManagerTests.DarkEditorBackground_Constant_Tracks_The_Theme_Resource
+		// pins it to -- kept as a constant so the conversion stays pure static math, usable
+		// without a running Application.
+		internal static readonly Color DarkEditorBackground = Color.FromRgb(0x1E, 0x1E, 0x1E);
+
+		static readonly double DarkEditorBackgroundLuminance =
+			RelativeLuminance(DarkEditorBackground.R, DarkEditorBackground.G, DarkEditorBackground.B);
+
 		// Highlighting definitions whose named colours we re-theme on every theme switch.
 		readonly List<IHighlightingDefinition> themableDefinitions = new();
 
@@ -193,10 +212,17 @@ namespace ICSharpCode.ILSpy.Themes
 		/// Clones <paramref name="lightColor"/> with its foreground/background brushes flipped
 		/// for a dark-theme background. Lightness inverts with a small curve adjustment;
 		/// over-saturated colours are softened so they don't burn through the dark editor
-		/// background. Non-colour style attributes (bold/italic/underline) pass through
-		/// unchanged. When the input has no colour brushes at all, returns it as-is so the
-		/// caller's cache can short-circuit.
+		/// background, and the foreground is then moved to <see cref="MinimumDarkContrastRatio"/>
+		/// against the surface it lands on. Non-colour style attributes (bold/italic/underline)
+		/// pass through unchanged. When the input has no colour brushes at all, returns it as-is
+		/// so the caller's cache can short-circuit.
 		/// </summary>
+		/// <remarks>
+		/// The contrast guarantee assumes opaque colours: the ratio is computed on the RGB
+		/// channels and the source alpha is carried over untouched, so a translucent foreground
+		/// composites to less contrast than the floor promises. Every named XSHD colour that
+		/// reaches this path today is opaque.
+		/// </remarks>
 		public static HighlightingColor GetColorForDarkTheme(HighlightingColor lightColor)
 		{
 			ArgumentNullException.ThrowIfNull(lightColor);
@@ -204,12 +230,26 @@ namespace ICSharpCode.ILSpy.Themes
 				return lightColor;
 
 			var darkColor = (HighlightingColor)lightColor.Clone();
-			darkColor.Foreground = AdjustForDarkTheme(darkColor.Foreground);
-			darkColor.Background = AdjustForDarkTheme(darkColor.Background);
+			// The background converts first: when a colour declares one, that -- not the editor
+			// canvas -- is the surface its own foreground is painted on, so it is what the
+			// foreground's contrast has to be measured against.
+			darkColor.Background = AdjustForDarkTheme(darkColor.Background, contrastReference: null);
+			darkColor.Foreground = AdjustForDarkTheme(darkColor.Foreground,
+				contrastReference: LuminanceOf(darkColor.Background) ?? DarkEditorBackgroundLuminance);
 			return darkColor;
 		}
 
-		static HighlightingBrush? AdjustForDarkTheme(HighlightingBrush? lightBrush)
+		static double? LuminanceOf(HighlightingBrush? brush)
+		{
+			var color = brush?.GetColor(null!);
+			return color is null ? null : RelativeLuminance(color.Value.R, color.Value.G, color.Value.B);
+		}
+
+		// contrastReference is the luminance of the surface the colour will be painted on, or
+		// null for a colour that IS a surface: backgrounds are left where the inversion put them,
+		// or a light XSHD span background would be repainted as a bright block that buries the
+		// text drawn on top of it.
+		static HighlightingBrush? AdjustForDarkTheme(HighlightingBrush? lightBrush, double? contrastReference)
 		{
 			if (lightBrush is null)
 				return null;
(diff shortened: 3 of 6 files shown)


ILSpy/Themes/ThemeManager.cs
```diff
@@ -231,17 +249,71 @@ static Color AdjustForDarkTheme(Color color)
...
 				l *= 1.2f;
-			}
+
+			// HSL lightness is not perceptual luminance, so inversion alone leaves hues with a
+			// low luminance weight (blue above all) too dim to read, and drags an already-light
+			// source colour *downwards* into the background. Foregrounds therefore get a
+			// contrast floor; backgrounds must not, or a light XSHD span background would be
+			// repainted as a bright block that buries the text drawn on top of it.
+			if (isForeground)
+				l = LiftToContrastFloor(h, s, l);
```
christophwille: **Floor is measured against the canvas, not the span's own (also converted) background.** `GetColorForDarkTheme` converts `Foreground` and `Background` independently, so a `HighlightingColor` that carries both never has its foreground checked against the surface it is actually drawn on. For a light-on-dark span (e.g. `foreground="White" background="Navy"`) the old inversion gave black on #9D9DFF (8.7:1); now the foreground is lifted to ~#959595 while the background still inverts to #AAAAF2 -> ~1.4:1, invisible. Same story for a mid-tone grey span background. Nothing shipped hits this today (Asm `Registers` #8080FF/#EEEEEE happens to work because #EEEEEE inverts to ~#181818), but `GetColorForDarkTheme` is public and every AvaloniaEdit built-in / plugin XSHD routes through it.

Suggestion: convert the background first and pass its luminance (or the canvas luminance when there is no background) as the reference for the lift -- a single reference-luminance parameter would replace the `isForeground` bool through the three overloads, and the existing `Leaves_Backgrounds_Below_The_Contrast_Floor` test still holds. Worth one test with a fg+bg pair either way.

christophwille (commented):
Reviewed the dark-conversion math (re-derived old vs. new results for every shipped and AvaloniaEdit built-in XSHD colour with an independent script; the PR's before/after table checks out), the App.axaml style against AvaloniaEdit's `TextView.LinkTextForegroundBrush` registration (decompiled 12.0.0: `Register<TextView, IBrush>(..., Brushes.Blue)`, `VisualLineLinkText.CreateTextRun` reads it per run, a property change triggers `Redraw`), and ran the `ICSharpCode.ILSpy.Tests.Themes` fixtures locally on the branch (29 passed). CI is green.

The change does what it says for the reported cases and the link fix is the right mechanism (same pattern as the existing `aeFolding|FoldingMargin` style). Nothing blocking; the inline comments cover:

1. **Floor reference** -- the foreground floor is measured against the canvas even when the colour carries its own span background, so light-on-dark spans regress (nothing shipped today; plugin/built-in XSHDs can hit it). Suggest lifting against the converted background's luminance when present.
2. **Widened `l *= 1.2` lift** -- now applies to unsaturated colours and span backgrounds, and introduces a non-monotone step at l = 0.75 for greys (`#505050` -> #BFBFBF, `#525252` -> #E3E3E3). Probably tie the lift back to `s > 0.75` (only the desaturation needed to become unconditional).
3. **Alpha** ignored by the floor (opaque-only guarantee; named colours are all opaque today).
4. **`DarkEditorBackground` drift** -- three copies of #1E1E1E with no test tying the constant to the App.axaml resource.
5. **Metadata token links on the selected row** -- local Foreground beats the row's white override; light theme drops from ~4.4:1 to ~2.9:1 on the accent fill.
6. Test coverage claim / stale hand-copied colour list; CLAUDE.md conventions (em-dashes, `Colour` identifiers, comments narrating the old guard).

Two things outside the diff worth an issue or a follow-up rather than this PR:
- `DocumentationRenderer` paints the *definition's* named colours (dark-converted in place) inside the light `#FCFCFC` hover popup. For IL / ReadyToRun tooltips the floor makes this slightly worse (Blue instructions #6F6FEA -> #8A8AEE, ~4.0:1 -> ~2.9:1 on the popup); C# tooltips already had this with `CSharpDark`. Reinforces the PR's own note that the popup needs a dark-mode pass.
- `OpenFromNuGetFeedDialog.axaml` still hardcodes link `#1B6EC2` (~3.1:1 on the dark window) -- another candidate for `ILSpy.LinkForeground`.

ILSpy/Themes/ThemeManager.cs
```diff
@@ -231,17 +249,71 @@ static Color AdjustForDarkTheme(Color color)
...
-			if (s > 0.75f && l < 0.75f)
-			{
+			// a dark editor background. This has to happen whatever the inverted lightness is:
+			// a dark, fully saturated source (DarkMagenta) inverts to a *light*, still fully
+			// saturated colour, i.e. exactly the neon the softening exists to prevent. The
+			// extra lift is only for results that landed dark.
+			if (s > 0.75f)
 				s *= 0.75f;
+			if (l < 0.75f)
 				l *= 1.2f;
```
christophwille: **Splitting the guard widens the `l *= 1.2` lift beyond what the PR describes.** Before, the lift only ran for over-saturated colours (`s > 0.75 && l < 0.75`); now it runs for *every* colour whose inverted lightness lands below 0.75 -- greys, pastels, and span **backgrounds** (the PR frames backgrounds as untouched, but e.g. bg `#999999` moves #747474 -> #8C8C8C, LightGray #333333 -> #3E3E3E). Since the floor covers foregrounds anyway, the extra lift is mostly redundant there, but for unsaturated colours the mapping is no longer monotone: `#505050` -> #BFBFBF while the *lighter* `#525252` -> #E3E3E3 (inverted 0.744 < 0.75 gets x1.2, 0.751 does not), so neighbouring greys can swap order. Either keep the lift tied to `s > 0.75` (only the desaturation needed to become unconditional for the DarkMagenta case) or state/test the new rule explicitly.

ILSpy/Themes/ThemeManager.cs
```diff
@@ -231,17 +249,71 @@ static Color AdjustForDarkTheme(Color color)
...
+			// HSL lightness is not perceptual luminance, so inversion alone leaves hues with a
+			// low luminance weight (blue above all) too dim to read, and drags an already-light
+			// source colour *downwards* into the background. Foregrounds therefore get a
+			// contrast floor; backgrounds must not, or a light XSHD span background would be
+			// repainted as a bright block that buries the text drawn on top of it.
+			if (isForeground)
+				l = LiftToContrastFloor(h, s, l);
 
 			var (r, g, b) = HslToRgb(h, s, l);
 			return Color.FromArgb(color.A, r, g, b);
```
christophwille, line 294: Minor: the contrast floor is evaluated on the opaque RGB and the original alpha is re-attached afterwards, so a translucent foreground (`#80xxxxxx`) can pass the floor while its composited colour on #1E1E1E is well under 5.5:1 (50% white lifted to #959595 composites to ~#595959, 2.4:1). All named colours today are opaque; the shipped alpha foregrounds (`#EEE0E000` in ILAsm/Asm-Mode) are anonymous rule colours that never reach this path -- so this is only a gap in the guarantee the doc comment promises. Either composite over the canvas before measuring, or note the opaque-only assumption.

ILSpy/Themes/ThemeManager.cs
```diff
@@ -41,6 +41,23 @@ public sealed class ThemeManager
...
+		/// values already live (5.0-8.8) and matches VS Code's own dark keyword blue (5.65).
+		/// The 4.5 WCAG AA threshold is not enough here: it leaves plain Blue at 4.51, which
+		/// still reads as too dark against the editor background.
+		/// </summary>
+		public const double MinimumDarkContrastRatio = 5.5;
+
+		// The dark editor canvas the foreground floor is measured against. Mirrors
+		// ILSpy.EditorBackground in the Dark theme dictionary of App.axaml -- kept as a constant
+		// so the conversion stays pure static math (and unit-testable without an Application).
+		static readonly Color DarkEditorBackground = Color.FromRgb(0x1E, 0x1E, 0x1E);
```
christophwille: `DarkEditorBackground` is a hand copy of the Dark `ILSpy.EditorBackground` in App.axaml, and `ThemeManagerTests.ContrastAgainstDarkEditor` hardcodes `#1E1E1E` a third time. If someone retunes the dark canvas in App.axaml, the floor keeps measuring against #1E1E1E and every floor test stays green (Blue sits at 5.52, right at the edge). Cheapest guard: one `[AvaloniaTest]` asserting `Application.Current.TryFindResource("ILSpy.EditorBackground", ThemeVariant.Dark, ...)` equals `ThemeManager.DarkEditorBackground` (and have the test helper reference the constant instead of re-parsing the hex). Also: `MinimumDarkContrastRatio` is `public` only for the tests; `InternalsVisibleTo("ILSpy.Tests")` already exists and `ThemeAwareHighlightingColorizer.GetEffectiveColor` uses `internal` for the same purpose.

ILSpy/Metadata/MetadataColumnBuilder.cs
```diff
@@ -434,10 +435,12 @@ static DataGridTemplateColumn BuildTokenColumn(PropertyInfo prop, ColumnInfoAttr
 					Padding = new Thickness(2, 0),
 					HorizontalAlignment = HorizontalAlignment.Left,
 					VerticalAlignment = VerticalAlignment.Center,
-					Foreground = Brushes.Blue,
 					Cursor = new global::Avalonia.Input.Cursor(global::Avalonia.Input.StandardCursorType.Hand),
 					Content = FormatTokenValue(row, prop, format),
 				};
+				// Token cells are links; share the theme's link colour rather than a fixed blue,
+				// which is unreadable on the dark grid.
+				btn[!Button.ForegroundProperty] = new DynamicResourceExtension("ILSpy.LinkForeground");
```
christophwille: Local `Foreground` on the button beats the App.axaml `DataGridRow:selected { Foreground = White }` override (same as the old `Brushes.Blue` did), and `MetadataTablePage.ApplyScrollTarget` selects the row you navigate to -- so the row a token click lands on is the one where the link colour is measured against the accent fill (Simple `ThemeAccentColor2`, `#99119EDA`, ~#70C5E9 over the light pane). Light theme: `#0066CC` on that fill is ~2.9:1, down from ~4.4:1 with pure Blue; dark: `#4DA6FF` on ~#166B8F is ~2.3:1. Not a blocker, but if you want the selected-row text readable, a `Classes="link"` style in App.axaml with a `DataGridRow:selected Button.link` override would let the row's white win. (Nit: code-behind elsewhere binds resources via `control.Bind(Prop, control.GetResourceObservable(key))` -- DecompilerTextEditor.cs:70 -- rather than `DynamicResourceExtension`; either works.)

ILSpy.Tests/Themes/ThemeManagerTests.cs
```diff
@@ -137,6 +137,82 @@ public void GetColorForDarkTheme_Preserves_Non_Colour_Style_Attributes()
 		remapped.FontStyle.Should().Be((AvFontStyle?)AvFontStyle.Italic);
 	}
 
+	// Every distinct foreground shipped by the repo's own XSHDs (ILSpy/TextView/*.xshd).
```
christophwille: The "every distinct foreground shipped by the repo's own XSHDs" claim is slightly too strong: ILAsm-Mode/Asm-Mode ship inline `<Keywords foreground="#FFFF0000">` / `#EEE0E000` (and CSharp-Mode `#E0E000`) that are anonymous rule colours, never in `NamedHighlightingColors`, and therefore never converted at all (ThemeManager only rewrites named colours and the per-paint colorizer is off once the definition is registered). "Every distinct *named* foreground" would be accurate. The hand-copied list also goes stale silently when an XSHD colour changes; enumerating `HighlightingService.GetByExtension(...).NamedHighlightingColors` (under Light) would keep the claim true by construction -- your call whether the determinism of a fixed list is worth more.

ILSpy.Tests/Themes/ThemeManagerTests.cs
```diff
@@ -137,6 +137,82 @@ public void GetColorForDarkTheme_Preserves_Non_Colour_Style_Attributes()
...
+		var remapped = ThemeManager.GetColorForDarkTheme(color);
+
+		ContrastAgainstDarkEditor(ColorOf(remapped.Background!))
+			.Should().BeLessThan(ThemeManager.MinimumDarkContrastRatio);
+	}
+
+	[Test]
+	public void GetColorForDarkTheme_Softens_Saturation_Of_Colors_That_Invert_To_Light()
+	{
+		// DarkMagenta inverts to lightness ~0.79, which the old `s > 0.75 && l < 0.75` guard
```
christophwille: CLAUDE.md conventions on the added lines (low severity, but the rules are explicit): (1) "Comments must stand on their own, with no memory of how the code was written" -- "the old `s > 0.75 && l < 0.75` guard skipped" here and "used to land at 4.08:1" at line 158 only make sense to someone who read this diff; phrase as the invariant ("softening must apply whatever the inverted lightness is, or DarkMagenta comes back as neon #FF93FF"). (2) "ASCII-only in code and comments" -- em-dashes on lines 158 and 187 here, LinkColorTests.cs:39-40, App.axaml:421 (ThemeManager.cs correctly uses `--`). (3) "en-US English in ... identifiers" -- new test names `..._Link_Colour_...` / `Link_Colour_Follows_...` and "not a solid colour brush" in LinkColorTests.cs, mixed with the en-US class name `LinkColorTests`. Pre-existing text in these files is British already, so this is precedent vs. rule; only the new lines are in question.

ILSpy/Themes/ThemeManager.cs
christophwille: Fixed. `GetColorForDarkTheme` now converts the background **first** and passes its luminance as the foreground's contrast reference, falling back to the canvas when the colour declares no background. The `isForeground` bool is gone: a single `double? contrastReference` threads through the overloads, where `null` means "this colour *is* a surface" (no floor).

Your White-on-Navy case needed one more thing than a reference swap: under a light span background the foreground has to move **down**, and a pure lift can only go up. `MoveToContrastFloor` now heads for whichever end of the lightness range has more contrast against the reference -- white above the dark canvas, black under a light background -- and bisects for the smallest move that clears the floor; if even that end misses (a mid-tone surface no colour can clear) it returns it as the best available. Luminance is monotone in HSL lightness at fixed hue/saturation, so the bisection is still sound.

`foreground="White" background="Navy"` now converts to `#000000` on `#AAAAF2`, **9.74:1** (was 1.39:1 measuring against the canvas). Asm `Registers` `#8080FF`/`#EEEEEE` converts to `#8181ED` on `#141414`, 5.52:1 against the surface it is actually painted on.

Covered by the new `GetColorForDarkTheme_Measures_The_Foreground_Against_Its_Own_Span_Background` (red before the change at 1.39). `Leaves_Backgrounds_Below_The_Contrast_Floor` still holds unchanged.

ILSpy/Themes/ThemeManager.cs
christophwille: Fixed -- the lift is nested back inside `s > 0.75`, so only the desaturation became unconditional (which was all the DarkMagenta case needed):

```csharp
if (s > 0.75f)
{
    s *= 0.75f;
    if (l < 0.75f)
        l *= 1.2f;
}
```

Backgrounds are now byte-identical to before this PR again (`#EEEEEE` -> `#141414`, `#999999` -> `#747474`, LightGray -> `#333333`), and the grey ordering is restored: `#505050` -> `#BFBFBF`, `#525252` -> `#BDBDBD`. `GetColorForDarkTheme_Keeps_Neighboring_Greys_In_Order` pins that (it fails on the split guard).

The only shipped foreground that moves is Brown: `#E9AAAA` 8.57 -> `#DC7777` 5.52, i.e. the floor now does the work the stray lift was doing, and stays closer to the source colour. Table in the PR body updated.

ILSpy/Themes/ThemeManager.cs
christophwille: Noted rather than composited, in a `<remarks>` on `GetColorForDarkTheme`: the guarantee is opaque-only, the ratio is computed on the RGB channels and the source alpha is carried over untouched.

Compositing properly needs the surface *colour*, not just its luminance, which would mean threading a `Color` through instead of a `double` for a case that no colour reaching this path hits today -- the shipped `#EEE0E000` foregrounds are anonymous rule colours that never enter `NamedHighlightingColors`. Happy to do it if a translucent named colour ever shows up.

ILSpy/Themes/ThemeManager.cs
christophwille: Both done.

- `MinimumDarkContrastRatio` and `DarkEditorBackground` are `internal` now, matching `ThemeAwareHighlightingColorizer.GetEffectiveColor`.
- `ThemeManagerTests.DarkEditorBackground_Constant_Tracks_The_Theme_Resource` resolves `ILSpy.EditorBackground` under `ThemeVariant.Dark` and asserts it equals the constant, so retuning the canvas in App.axaml fails a test instead of silently invalidating the floor.
- `ContrastAgainstDarkEditor` in the tests measures against `ThemeManager.DarkEditorBackground` instead of re-parsing `#1E1E1E`, so the third copy is gone. The contrast math itself is still duplicated in the test on purpose -- it is what keeps the tests from validating the production formula against itself.

ILSpy/Metadata/MetadataColumnBuilder.cs
christophwille: Fixed, and thanks for catching the direction of the regression -- light theme going 4.4 -> 2.9 on the row navigation lands on is worse than the bug being fixed.

The button now carries `Classes = { "link" }` instead of any local `Foreground`, and App.axaml owns both states:

```xml
<Style Selector="DataGridRow Button.link">
    <Setter Property="Foreground" Value="{DynamicResource ILSpy.LinkForeground}" />
</Style>
<Style Selector="DataGridRow:selected Button.link">
    <Setter Property="Foreground" Value="White" />
</Style>
```

`MetadataColumnBuilderTests.Token_Cell_Link_Color_Comes_From_The_Theme_And_Yields_To_The_Selected_Row` mounts a real grid, checks the unselected cell against the `ILSpy.LinkForeground` resource, then selects the row and checks it goes white. It fails against the previous local-`Foreground` version, so the descendant selector really is doing the work.

The `DynamicResourceExtension` (and its `using`) is gone with it, so the `GetResourceObservable` nit no longer applies.

ILSpy.Tests/Themes/ThemeManagerTests.cs
christophwille: Reworded to "every distinct **named** foreground", with a note that inline rule colours are anonymous, never land in `NamedHighlightingColors` and so are never converted at all.

Kept the fixed list rather than enumerating: it doubles as the source table for the before/after numbers in the PR body, and it stays meaningful when read as data. An XSHD colour change is a deliberate edit, and adding its new colour here is a one-line follow-up -- whereas an enumerated version would quietly stop asserting anything if the definition failed to load in the headless host.

ILSpy/Themes/ThemeManager.cs
christophwille: Fixed. `GetColorForDarkTheme` now converts the background **first** and passes its luminance as the foreground's contrast reference, falling back to the canvas when the colour declares no background. The `isForeground` bool is gone: a single `double? contrastReference` threads through the overloads, where `null` means "this colour *is* a surface" (no floor).

Your White-on-Navy case needed one more thing than a reference swap: under a light span background the foreground has to move **down**, and a pure lift can only go up. `MoveToContrastFloor` now heads for whichever end of the lightness range has more contrast against the reference -- white above the dark canvas, black under a light background -- and bisects for the smallest move that clears the floor; if even that end misses (a mid-tone surface no colour can clear) it returns it as the best available. Luminance is monotone in HSL lightness at fixed hue/saturation, so the bisection is still sound.

`foreground="White" background="Navy"` now converts to `#000000` on `#AAAAF2`, **9.74:1** (was 1.39:1 measuring against the canvas). Asm `Registers` `#8080FF`/`#EEEEEE` converts to `#8181ED` on `#141414`, 5.52:1 against the surface it is actually painted on.

Covered by the new `GetColorForDarkTheme_Measures_The_Foreground_Against_Its_Own_Span_Background` (red before the change at 1.39). `Leaves_Backgrounds_Below_The_Contrast_Floor` still holds unchanged.

christophwille:
All seven inline points addressed and answered in-thread; the branch is amended and force-pushed (`2936f330b` -> `fe25b0244`, per the repo's squash-back convention), and the PR body's before/after table is updated for the numbers that moved.

Behaviour changes from the review:

1. **Contrast reference** -- the background is converted first and its luminance becomes the foreground's reference; only a colour with no background of its own is measured against the canvas. `isForeground` is replaced by a `double? contrastReference` (null = "this colour *is* a surface"). The search now heads for whichever lightness extreme has more contrast, so a light span background pushes its foreground *down* instead of up. White-on-Navy: 1.39:1 -> 9.74:1.
2. **Lift scope** -- the `l *= 1.2` lift is back inside `s > 0.75`; only the desaturation is unconditional. Backgrounds and greys are byte-identical to pre-PR again, and the 0.75 non-monotonicity is gone. `Brown` moves 8.57 -> 5.52 as a result (the floor now does that work).
3. **Alpha** -- opaque-only assumption documented in a `<remarks>` rather than composited.
4. **Constant drift** -- `MinimumDarkContrastRatio` / `DarkEditorBackground` are `internal`, a test pins the constant to the Dark `ILSpy.EditorBackground` resource, and the tests' third copy of `#1E1E1E` is gone.
5. **Metadata token links** -- `Classes = { "link" }` plus `DataGridRow Button.link` / `DataGridRow:selected Button.link` styles, so the selected row's white wins over the accent fill instead of the link colour. Covered by a mounted-grid test.
6. **Test claim** -- "every distinct *named* foreground", with the anonymous-rule-colour caveat spelled out. Fixed list kept deliberately.
7. **Conventions** -- comments rephrased as invariants, em-dashes replaced with ASCII, new identifiers en-US.

New tests: span-background reference, grey ordering, canvas-constant drift, metadata link colour + selected row. Each was confirmed red against the previous implementation. `ILSpy.Tests`: 1200 passed, 3 skipped.

The two out-of-diff items (`DocumentationRenderer`'s light popup chrome, `OpenFromNuGetFeedDialog`'s hardcoded `#1B6EC2`) are noted in the PR description as follow-ups rather than folded in here.

siegfriedpammer (approved):

PR #4004, opened 2026-08-15
Decompile natural types for lambdas and method groups

siegfriedpammer:
Reconstructs the C# 10/12/13 "function type" features, so anonymous functions and method groups
come back as the source wrote them instead of as delegate constructions with unspeakable type
names.

**What is decompiled now**

- **Explicit lambda return types** (C# 10), including `ref` and `ref readonly` returns. The
  syntax tree gained `LambdaExpression.ReturnType`; the return type is written wherever the
  delegate's return type is not what C# would re-infer from the body.
- **Natural-typed lambdas for synthesized delegate types.** `<>A{...}` / `<>F{...}` (and the
  unbraced `` <>F`18 `` shapes for >16 parameters) cannot be named, so those sites now print
  `var x = (ref int y) => ++y;` rather than leaking the compiler-generated name. Conversions to
  `object`, `Delegate` and `Expression` targets no longer emit a cast to it either.
- **Method groups with a natural type** (C# 10) - `Func<int, int> f = M;` instead of
  `new Func<int, int>(M)` - gated by the existing `NaturalTypeForLambdaAndMethodGroup` setting.
  `ICSharpCode.Decompiler/CSharp/Resolver/MethodGroupNaturalType.cs` is the oracle that decides
  when the natural type of an emitted method group really is the delegate type in the IL.
- **C# 13 method-group improvements** behind a new `MethodGroupNaturalTypeImprovements` setting
  (scope-by-scope pruning vs. the C# 10 all-scopes rules), so pre-C# 13 output stays valid.
- **Lambda parameter defaults and `params`** (C# 12) where they are legal, downgraded to
  `[Optional]` / `[DefaultParameterValue]` / `[ParamArray]` attributes below C# 12
  (`LambdaOptionalAndParamsParameters`, with an `Ugly` fixture for the disabled state).
- **Statement-expression lambdas** render as expression bodies (`() => Console.WriteLine()`)
  where the expression is void-typed, or where an explicit return type spells the discarded
  value out. A non-void value keeps its block body: C# would otherwise read the expression's
  type as the lambda's return type and pick a different overload.

**Testing**

New fixtures: `Pretty/LambdaReturnTypes`, `Pretty/LambdaNaturalTypeConversions`,
`Pretty/LambdaOptionalAndParamsParameters` (+ `Ugly/NoLambdaOptionalAndParamsParameters`),
`Pretty/MethodGroupNaturalType`, `Pretty/MethodGroupNaturalTypeImprovementsDisabled`,
`Pretty/MethodGroupSynthesizedDelegates`, and the round-tripping
`Correctness/LambdaNaturalType` / `Correctness/MethodGroupNaturalType` - the correctness pair
matters most here, because a lost return type or parameter modifier silently rebinds an
overload instead of failing to compile.

Full `ICSharpCode.Decompiler.Tests` (3372 passed / 0 failed, roundtrip excluded) and
`ILSpy.Tests` (1181 passed / 0 failed) are green.

**Known gaps**: `static` lambdas are not emitted (see #829), and statement-bodied lambdas in
`Func`/`Action` positions still print as `delegate(int x) { ... }`.

**Specifications used**

Language proposals (dotnet/csharplang) - the three #829 entries this covers:

- [Lambda improvements](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/lambda-improvements.md) (C# 10) - attributes, explicit return types, and the natural type of a lambda or method group
- [Lambda method group defaults](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/lambda-method-group-defaults.md) (C# 12) - parameter default values and `params` on lambdas
- [Method group natural type improvements](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-13.0/method-group-natural-type-improvements.md) (C# 13) - the scope-by-scope candidate pruning behind the new setting

Normative rules (dotnet/csharpstandard, branch `draft-v12`) - what the oracle in
`MethodGroupNaturalType.cs` and the conversion handling are checked against:

- [§12.22.8 Anonymous function type](https://github.com/dotnet/csharpstandard/blob/draft-v12/standard/expressions.md#12228-anonymous-function-type) - when a natural type exists, when a synthesized delegate is used instead of `Func`/`Action` (any non-by-value parameter or return, more than 16 parameters, or a type that is not a valid type argument), and the mandated `arg1..argn` parameter names
- [§10.2.21 Anonymous function type conversion](https://github.com/dotnet/csharpstandard/blob/draft-v12/standard/conversions.md#10221-anonymous-function-type-conversion) - which targets an anonymous-function-typed value converts to (`object`, `Delegate`, base classes, `Expression`), which is why those sites no longer need a cast

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
@@ -2662,11 +2662,17 @@ namespace ICSharpCode.Decompiler.CSharp
 					inferredReturnType = lambda.Body!.GetResolveResult().Type;
 					naturalReturnType = inferredReturnType;
 				}
-				else if (isAnonymousDelegate && body.Statements.Count == 1
-					&& body.Statements.Single() is ExpressionStatement exprStmt)
+				else if (body.Statements.Count == 1 && body.Statements.Single() is ExpressionStatement exprStmt
+					&& !NeedsDeclarationInBody(function, exprStmt)
+					&& (isAnonymousDelegate || exprStmt.Expression.GetResolveResult().Type.Kind == TypeKind.Void))
 				{
 					// A single statement-expression can be the expression body of a void-returning
 					// lambda; its value (if any) is discarded, so the return type stays void.
+					// Discarding a value is only safe where the void-ness survives into the output:
+					// C# reads the expression's type as the lambda's return type, which picks a
+					// different natural type and a different overload. An anonymous delegate spells
+					// the return type out below; everywhere else the block body, which has no return
+					// statement, is what says the value is dropped.
 					lambda.Body = exprStmt.Expression.Detach();
 					inferredReturnType = compilation.FindType(KnownTypeCode.Void);
 					naturalReturnType = lambda.Body!.GetResolveResult().Type;
@@ -2752,6 +2758,51 @@ namespace ICSharpCode.Decompiler.CSharp
 						}
 						return true;
 				}
+			}
+		}
+
+		/// <summary>
+		/// Whether the statement uses a variable that DeclareVariables may still have to declare,
+		/// which only a block body can host. Collapsing such a body to an expression pushes the
+		/// declaration inside the lambda when the block is rebuilt for it - turning a captured
+		/// variable into per-invocation state, which no longer round-trips.
+		/// </summary>
+		static bool NeedsDeclarationInBody(ILFunction function, ExpressionStatement statement)
+		{
+			// A deconstruction assignment produces the tuple it assigned from. Compiled as an
+			// expression body its value is discarded differently than in a block, and the IL that
+			// comes back no longer looks like one statement - so the shape would flip on every
+			// round trip. Keep the block, which is stable.
+			if (statement.Expression is AssignmentExpression { Left: TupleExpression })
+				return true;
+			foreach (var identifier in statement.DescendantsAndSelf.OfType<IdentifierExpression>())
+			{
+				if (identifier.GetResolveResult() is not ILVariableResolveResult { Variable: var variable })
+					continue;
+				// Parameters and this-references are never declared.
+				if (variable.Kind is not (VariableKind.Local or VariableKind.StackSlot))
+					continue;
+				// Not closure state: a declaration inside the lambda is correct.
+				if (variable.CaptureScope is null)
+					continue;
+				// Read (or written) elsewhere too, so the declaration lands in the shared scope
+				// no matter what this body looks like, and collapsing is safe.
+				if (!AllUsesAreInside(variable, function))
+					continue;
+				return true;
+			}
+			return false;
+
+			static bool AllUsesAreInside(ILVariable variable, ILFunction function)
+			{
+				foreach (var use in variable.LoadInstructions.Cast<ILInstruction>()
+					.Concat(variable.StoreInstructions.Cast<ILInstruction>())
+					.Concat(variable.AddressInstructions))
+				{
+					if (use.Ancestors.OfType<ILFunction>().FirstOrDefault() != function)
+						return false;
+				}
+				return true;
 			}
 		}
 
--- a/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
@@ -2587,7 +2587,7 @@ namespace ICSharpCode.Decompiler.CSharp
 				let v = ident.GetILVariable()
 				where v != null && v.Function == function && v.Kind == VariableKind.Parameter
 				select ident).Any();
-			bool isAnonymousDelegate = delegateType.IsAnonymousDelegate();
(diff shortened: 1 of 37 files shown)


christophwille:
### CI failure analysis (Windows Debug/Release)

The only failing test in both Windows jobs was `RoundtripAssembly.NewtonsoftJson_net45` (Linux/macOS pass because the round-trip suite is Windows/MSBuild-only). The decompiled `Newtonsoft.Json.csproj` (LangVersion 8.0) no longer compiled; every lambda bound to a `params` delegate (`ObjectConstructor<T>`, `MethodCall<T,R>`) failed with the same pair:

```
ReflectionObject.cs(53,26): error CS8400: Feature 'lambda attributes' is not available in C# 8.0.
ReflectionObject.cs(53,18): error CS0674: Do not use 'System.ParamArrayAttribute'. Use the 'params' keyword instead.
```

Master emitted `creator2 = (object[] args) => ctor();`; the branch emitted `creator2 = ([ParamArray] object[] args) => ctor();`.

**Root cause**, two parts in `ExpressionBuilder.TranslateFunction`:

1. `params`/defaults were copied from *every* target delegate's `Invoke` onto the lambda (`pd.IsParams |= invokeParameter.IsParams`). Newtonsoft's closure `<Create>b__0(object[])` carries no `ParamArrayAttribute` in IL - the source was a plain `(object[] args) => ...` - so this invented a `params` that was never written.
2. With `LambdaOptionalAndParamsParameters` off (C# < 12), the fallback rewrote it as `[ParamArray]` / `[Optional][DefaultParameterValue]` attributes. `[ParamArray]` cannot be written explicitly in any C# version (CS0674), and attributes on lambda parameters need C# 10 (CS8400) - so under C# 8 it was doubly uncompilable. The `Ugly/NoLambdaOptionalAndParamsParameters` fixture enshrined that output (Ugly fixtures aren't recompiled, which is why it slipped through).

I checked the premise in the code comment ("Roslyn 4.14 emits no ParamArrayAttribute there") against both compilers in the test matrix. It's half right: Roslyn 4.14 puts the **default value** on the closure method but **not** `ParamArrayAttribute`; the current compiler (SDK 11 preview) puts both there, exactly when the source spelled them and never otherwise.

### Fix (a4ef71bb7)

- Consult `Invoke` only for compiler-synthesized (anonymous) delegate types - that's the one place the lambda declaration is the only way to spell the type, and where 4.14 leaves no other trace. Named delegates keep the lambda's own metadata: faithful on current Roslyn, and where 4.14 omits the attribute the plain parameter list is equivalent anyway. (Restricting it to anonymous delegates was necessary: dropping the copy entirely broke `Correctness/LambdaNaturalType` on 4.14.)
- Below C# 12, drop `params`/defaults instead of downgrading to attributes (no legal spelling exists); the `RequiredNamespaceCollector` hook that only served that fallback is gone.
- `Ugly/NoLambdaOptionalAndParamsParameters.Expected.cs` updated; `Pretty/LambdaOptionalAndParamsParameters.cs` got a `#if ROSLYN5 || !EXPECTED_OUTPUT` split for the named-delegate `params` cases.

Verified locally: `NewtonsoftJson_net45` passes again, and the 152 Lambda/MethodGroup/DelegateConstruction/Optional fixtures pass across the Roslyn 4.14 + latest matrix.

Note: the PR description still says the modifiers are "downgraded to `[Optional]` / `[DefaultParameterValue]` / `[ParamArray]` attributes below C# 12" - that should now read "dropped below C# 12".

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -2629,23 +2684,60 @@ from ident in body.Descendants.OfType<IdentifierExpression>()
 				{
 					lambda.Body = returnStmt.Expression.Detach();
 					inferredReturnType = lambda.Body!.GetResolveResult().Type;
+					naturalReturnType = inferredReturnType;
+				}
+				else if (body.Statements.Count == 1 && body.Statements.Single() is ExpressionStatement exprStmt
+					&& !NeedsDeclarationInBody(function, exprStmt)
+					&& (isAnonymousDelegate || exprStmt.Expression.GetResolveResult().Type.Kind == TypeKind.Void))
```
dgrunwald, line 2691: This new code path is questionable -- why is there a difference between `inferredReturnType` and `naturalReturnType` here? If we remove the braces, I would think that both variable's usages need to take the expression's type.

Why did we even need to remove the braces? How does this change relate to natural types?

dgrunwald:
I recommend splitting unrelated bugfixes out of this PR.
I'm not sure we really need the pretty syntax sugar for rare lambda expressions using natural types with ref/out params.

siegfriedpammer:
Closing: the natural-type work is not planned for now.

The two bugfixes that were found while building this branch do not depend on the feature and are extracted into #4057 (attributes forcing a parenthesized lambda parameter list; `params`/default values printed where the language forbids them). Both were reproduced against master and the emitted code confirmed uncompilable, with CS8916 and CS1670 respectively.

Not carried over: the pre-C# 12 downgrade of `params`/defaults to `[ParamArray]`/`[Optional]`, and the void statement-expression collapse. The branch `natural-type-lambdas-methods` stays on origin, so all of it remains available if the feature is picked up again.

PR #4005, opened 2026-08-15
Merge tuple element names during type inference

siegfriedpammer:
When fixing a type parameter, Roslyn merges the tuple element names of bounds that are
identical apart from those names: a name is kept where all bounds agree and dropped where
they conflict (`MergeTupleNames` / `MergeEquivalentTypes` in Roslyn's `MethodTypeInference.cs`).
The C# standard's type-inference section does not describe this step, so `TypeInference` did
not implement it.

Without it, fixing either kept the *first* bound's element names verbatim, or - with two exact
bounds that differ only in names - failed outright, because `AddExactBound` compares bounds
with name-sensitive equality. Either way, inferred tuple types could carry element names that
csc would never produce.

This adds the merge in the three places a bound set is reduced to a single type:

* `AddExactBound` - two exact bounds that differ only in element names are merged instead of
  being recorded as conflicting.
* `Fix` (exact-bound path) - the exact bound is merged with each lower/upper bound of the same
  shape.
* `Fix` (bounds path) - shape-equivalent bounds are collapsed before `FindTypesInBounds`, so
  they do not survive as distinct candidates.

The merge recurses through tuple elements, generic type arguments, and array element types,
matching Roslyn; anything not equal modulo element names is left alone, so non-tuple inference
is unaffected.

No decompiler-visible misdecompilation was demonstrated for this - the change aligns the
resolver's inference with Roslyn rather than fixing a reported output bug.

All expectations in the new tests are verified against csc.

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/Resolver/TypeInference.cs
+++ b/ICSharpCode.Decompiler/CSharp/Resolver/TypeInference.cs
@@ -18,6 +18,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Linq;
 
@@ -242,8 +243,18 @@ namespace ICSharpCode.Decompiler.CSharp.Resolver
 				// Exact bounds need to stored separately, not just as Lower+Upper bounds,
 				// due to TypeInferenceTests.GenericArgumentImplicitlyConvertibleToAndFromAnotherTypeList (see #281)
 				if (ExactBound == null)
+				{
 					ExactBound = type;
-				else if (!ExactBound.Equals(type))
+					return;
+				}
+				if (ExactBound.Equals(type))
+					return;
+				// Two exact bounds that differ only in tuple element names are not conflicting;
+				// their names are merged instead (kept where both agree, dropped otherwise).
+				IType merged = MergeTupleNames(ExactBound, type);
+				if (merged != null)
+					ExactBound = merged;
+				else
 					MultipleDifferentExactBounds = true;
 			}
 
@@ -968,8 +979,15 @@ namespace ICSharpCode.Decompiler.CSharp.Resolver
 			Debug.Assert(!tp.IsFixed);
 			if (tp.ExactBound != null)
 			{
-				// the exact bound will always be the result
-				tp.FixedTo = tp.ExactBound;
+				// Roslyn behavior (not in the C# standard): when a lower/upper bound has the
+				// same shape as the exact bound except for tuple element names, the names are
+				// merged - kept where both sides agree, dropped where they conflict. See
+				// MergeTupleNames in Roslyn's MethodTypeInference.cs.
+				IType fixedTo = tp.ExactBound;
+				foreach (var b in tp.LowerBounds.Concat(tp.UpperBounds))
+					fixedTo = MergeTupleNames(fixedTo, b) ?? fixedTo;
+				// the exact bound determines the result, up to the merged element names
+				tp.FixedTo = fixedTo;
 				// check validity
 				if (tp.MultipleDifferentExactBounds)
 					return false;
@@ -977,7 +995,11 @@ namespace ICSharpCode.Decompiler.CSharp.Resolver
 					&& tp.UpperBounds.All(b => conversions.ImplicitConversion(tp.FixedTo, b).IsValid);
 			}
 			Log.Indent();
-			var types = CreateNestedInstance().FindTypesInBounds(tp.LowerBounds.ToArray(), tp.UpperBounds.ToArray());
+			// Same Roslyn-style merge, over lower and upper bounds together as Roslyn's Fix does,
+			// so bounds that differ only in tuple element names don't survive into
+			// FindTypesInBounds as distinct candidates.
+			var (lowerBounds, upperBounds) = MergeShapeEquivalentBounds(tp.LowerBounds, tp.UpperBounds);
+			var types = CreateNestedInstance().FindTypesInBounds(lowerBounds, upperBounds);
 			Log.Unindent();
 			if (algorithm == TypeInferenceAlgorithm.ImprovedReturnAllResults)
 			{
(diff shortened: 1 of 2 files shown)


ICSharpCode.Decompiler/CSharp/Resolver/TypeInference.cs
```diff
@@ -968,16 +979,29 @@ bool Fix(TP tp)
...
 				if (tp.MultipleDifferentExactBounds)
 					return false;
 				return tp.LowerBounds.All(b => conversions.ImplicitConversion(b, tp.FixedTo).IsValid)
 					&& tp.UpperBounds.All(b => conversions.ImplicitConversion(tp.FixedTo, b).IsValid);
 			}
 			Log.Indent();
-			var types = CreateNestedInstance().FindTypesInBounds(tp.LowerBounds.ToArray(), tp.UpperBounds.ToArray());
+			// Same Roslyn-style merge across the bound sets so two bounds that differ only
+			// in tuple element names don't survive into FindTypesInBounds as distinct candidates.
+			var lowerBounds = MergeShapeEquivalentBounds(tp.LowerBounds);
```
christophwille: Lower and upper bounds are merged separately here, so a lower/upper pair that differs only in tuple names still reaches `FindTypesInBounds` as two candidates. Roslyn's `Fix` merges exact + lower + upper into one candidate dictionary keyed by `IgnoreDynamicAndTupleNames`.

Probe on this branch: `M<T>(T x, Action<T> y)` with `IList<(int a, string b)>` and `Action<IList<(int a, string c)>>` (reference-type argument, so `Action<in T>` yields an upper bound instead of an exact one) -> lower `IList<(int a, string b)>`, upper `IList<(int a, string c)>`, `Union` keeps both (name-sensitive `Equals`), both are identity-convertible to each other, `candidateTypes.Count == 2`, `Fix` returns false; csc infers `IList<(int a, string)>`.

Not a regression (this failed before too), but the PR description says shape-equivalent bounds are collapsed before `FindTypesInBounds`, and this is the shape it misses. Merging once over `tp.LowerBounds.Concat(tp.UpperBounds)` (or seeding the upper pass with the merged lower list) covers it and would also let the exact path above reuse the same helper instead of the two copy-pasted `foreach` loops.

ICSharpCode.Decompiler/CSharp/Resolver/TypeInference.cs
```diff
@@ -992,6 +1016,91 @@ bool Fix(TP tp)
...
+					mergedArgs[i] = merged;
+				}
+				return new ParameterizedType(pa.GenericType, mergedArgs);
+			}
+			if (a is ArrayType arrA && b is ArrayType arrB && arrA.Dimensions == arrB.Dimensions)
+			{
+				var mergedElem = MergeTupleNames(arrA.ElementType, arrB.ElementType);
+				if (mergedElem == null)
+					return null;
+				return new ArrayType(arrA.Compilation, mergedElem, arrA.Dimensions);
```
christophwille: This rebuilds the array with the ctor default `Nullability.Oblivious`, and the guard above only checks `Dimensions`, while `ArrayType.Equals` compares nullability. Two consequences:

1. Nullability is dropped. Probe on this branch: `ref (int a, string b)[]?` + `ref (int a, string c)[]?` fixes to `(int a, string)[]` with `Nullability.Oblivious`.
2. The branch fires for arrays that contain no tuple at all: `string[]?` + `string[]` (bounds keep their annotation whenever `U.Nullability != V.Nullability`) fail `a.Equals(b)`, the element types compare equal, and both collapse into one oblivious `string[]`. Inference flips from failure to success for a case the doc comment says is untouched.

Pass `arrA.Nullability` through (as `ArrayType.VisitChildren` does), and either require equal nullability in the guard or merge it deliberately (Roslyn merges nullability with the variance kind, which is a bigger change).

ICSharpCode.Decompiler/CSharp/Resolver/TypeInference.cs
```diff
@@ -992,6 +1016,91 @@ bool Fix(TP tp)
...
+		/// Merges tuple element names from two types that are equal modulo tuple element names.
+		/// Returns the merged type - names kept where both sides agree, null where they conflict -
+		/// or <c>null</c> if <paramref name="a"/> and <paramref name="b"/> have different shapes.
+		///
+		/// Mirrors Roslyn's <c>MergeTupleNames</c> / <c>MergeEquivalentTypes</c> behavior. The C#
+		/// standard (§12.6.3) does not describe this merging step; it's a compiler-specific
+		/// extension that nonetheless governs the inferred type when an exact bound from an
+		/// explicitly-typed lambda parameter meets a lower bound from a nameless source sequence.
+		/// </summary>
+		static IType MergeTupleNames(IType a, IType b)
```
christophwille, line 1023: `NullabilityAnnotatedType` wrappers are not unwrapped, so any annotation on either side skips the merge entirely (`a is ParameterizedType` / `a is TupleType` is false for the wrapper).

Probe on this branch: `M<T>(T x, T y)` with `IList<(int a, string b)>?` and `IList<(int a, string c)>` -> no merge, two mutually convertible candidates, `Fix` fails; csc infers `IList<(int a, string)>?` (it compares with `IgnoreNullableModifiersForReferenceTypes` before merging).

Cheapest fix: gate on `NormalizeTypeVisitor.IgnoreNullabilityAndTuples.EquivalentTypes(a, b)` (already in the type system, walks every `IType` shape) and unwrap `NullabilityAnnotatedType` in the recursion, re-applying the annotation on the result.

ICSharpCode.Decompiler/CSharp/Resolver/TypeInference.cs
```diff
@@ -992,6 +1016,91 @@ bool Fix(TP tp)
...
+					if (merged == null)
+						return null;
+					mergedElements.Add(merged);
+				}
+				var mergedNames = ImmutableArray.CreateBuilder<string>(ta.ElementNames.Length);
+				for (int i = 0; i < ta.ElementNames.Length; i++)
+				{
+					mergedNames.Add(ta.ElementNames[i] == tb.ElementNames[i] ? ta.ElementNames[i] : null);
+				}
+				return new TupleType(ta.Compilation, mergedElements.MoveToImmutable(), mergedNames.MoveToImmutable());
```
christophwille: This is the only `TupleType` construction site in the repo that does not pass `valueTupleAssembly` (`TupleType.VisitChildren`, `TupleType.FromUnderlyingType`, `ApplyAttributeTypeVisitor`, `CallBuilder`, `TranslatedExpression`, `TupleResolveResult` all thread it through). Without it `CreateUnderlyingType` re-resolves `System.ValueTuple`N` via `compilation.FindType`, which returns the first module's definition; when more than one loaded module defines `ValueTuple` (net461-era `System.ValueTuple.dll` next to a newer mscorlib, or an embedded copy) the merged tuple gets a different `ITypeDefinition`, and `TupleType.Equals` no longer matches the bounds it was merged from. `ta.GetDefinition()?.ParentModule` keeps it on the original definition.

ICSharpCode.Decompiler/CSharp/Resolver/TypeInference.cs
```diff
@@ -968,16 +979,29 @@ bool Fix(TP tp)
 			Debug.Assert(!tp.IsFixed);
 			if (tp.ExactBound != null)
 			{
+				// Roslyn behavior (not in the C# standard): when a lower/upper bound has the
+				// same shape as the exact bound except for tuple element names, the names are
+				// merged - kept where both sides agree, dropped where they conflict. See
+				// MergeTupleNames in Roslyn's MethodTypeInference.cs.
+				IType fixedTo = tp.ExactBound;
```
christophwille, line 986: Structure nit: one rule now lives in three places with two different strategies (pairwise fold into `ExactBound` in `AddExactBound` and here; absorb-into-list in `MergeShapeEquivalentBounds` below). Doing the merge once at fix time over exact + lower + upper -- and keeping only the `MultipleDifferentExactBounds` decision in `AddExactBound` -- would mirror Roslyn's `Fix` and remove the two duplicated loops. Also: the `// the exact bound will always be the result` comment two lines down is no longer literally true; `if (a == null || b == null)` in `MergeTupleNames` is unreachable; and the doc comment on `MergeTupleNames` restates the paragraph that already sits at all three call sites (and cites a lambda-parameter scenario none of the tests exercise) -- one sentence on the helper is enough.

christophwille (commented):
## Review summary

The change is sound in the direction it takes: the three merge points do line up with Roslyn's `Fix` (exact candidates merged with each other, then with lower/upper bounds; lower/upper bounds merged among themselves), the name-merge rule (`n1 == n2 ? n1 : null`, all-null == default) matches `MergeTupleNames`, and it fixes a real pre-existing miscompile: exact `(int a, string b)` (from `IList<T>`) + lower `(int a, string c)` used to fix to `(int a, string b)` with no explicit type args, and the emitted `.b` access would then not compile against csc's `(int a, string)`.

I ran the four new tests plus three probes against this branch (worktree at `2a63710`). What I found, most important first:

1. **Lower and upper bounds are merged separately, so a lower/upper pair that differs only in names still fails to fix.** Roslyn merges everything into one candidate dictionary. Probe: `M<T>(T x, Action<T> y)` with `IList<(int a, string b)>` and `Action<IList<(int a, string c)>>` (reference-type argument, so `Action<in T>` gives an upper bound rather than an exact one) -> `FindTypesInBounds` sees two mutually identity-convertible candidates, `Fix` returns false, csc infers `IList<(int a, string)>`. Not a regression (this failed before too), but it is the one shape the PR description says it collapses that it does not. Fix: run `MergeShapeEquivalentBounds` once over `LowerBounds.Concat(UpperBounds)` (or seed the upper pass with the merged lower list), which also gets rid of the two copy-pasted `foreach` loops in the exact path. Inline comment on the code.
2. **`MergeTupleNames` merges more than tuple names in the array branch and drops nullability.** `ArrayType.Equals` compares nullability but the branch only checks `Dimensions` and rebuilds with the default `Oblivious`. Probe: two `ref` bounds `(int a, string b)[]?` / `(int a, string c)[]?` fix to `(int a, string)[]` with `Nullability.Oblivious`. And with no tuple in sight, `string[]?` + `string[]` (bounds keep their annotation whenever `U.Nullability != V.Nullability`) now collapse to an oblivious `string[]`. Pass `arrA.Nullability` through (as `ArrayType.VisitChildren` does), and decide whether nullability-only differences should merge at all.
3. **`NullabilityAnnotatedType` wrappers are not unwrapped, so the merge is skipped whenever one bound carries an annotation.** Probe: `M<T>(T x, T y)` with `IList<(int a, string b)>?` and `IList<(int a, string c)>` -> no merge, two candidates, `Fix` fails; csc infers `IList<(int a, string)>?`. Roslyn compares with `IgnoreNullableModifiersForReferenceTypes` before merging. Cheapest fix: gate on `NormalizeTypeVisitor.IgnoreNullabilityAndTuples.EquivalentTypes(a, b)` (already in the type system) and unwrap `NullabilityAnnotatedType` in the walk. Inline comment.
4. **Merged `TupleType` is rebuilt without `valueTupleAssembly`.** Every other construction site threads it through (`TupleType.VisitChildren`, `ApplyAttributeTypeVisitor`, `CallBuilder`, `TranslatedExpression`); here the underlying `ValueTuple` is re-resolved via `compilation.FindType`, which walks modules in order and can bind to a different definition when two real `ValueTuple` definitions are loaded (net461-era non-facade `System.ValueTuple.dll` next to a 4.7+ mscorlib, or a main module embedding its own copy; facades are fine, forwarders are followed). Then `TupleType.Equals` no longer matches the original bound, and `IsAppropriateCallTarget` (which compares the erased *underlying* type) rejects the candidate and forces explicit type arguments. Rare, one-argument fix: `ta.GetDefinition()?.ParentModule`. Inline comment.
5. Structure/altitude: one rule, three patch points, two merge strategies (pairwise fold in `AddExactBound` and the exact path of `Fix`; absorb-into-list in the bounds path). Doing the merge once at `Fix` time over exact + lower + upper (and keeping only the `MultipleDifferentExactBounds` decision in `AddExactBound`) matches Roslyn's structure and removes the duplicated loops. Nit-level: `if (a == null || b == null)` in `MergeTupleNames` is unreachable (no caller passes null), and the doc comment on `MergeTupleNames` restates the same paragraph that already sits at all three call sites (and cites a lambda-parameter scenario none of the tests exercise) -- one sentence on the helper is enough.

Checked and not flagged: the `dynamic`/`object` half of Roslyn's `MergeEquivalentTypes` and pointer/function-pointer/modopt wrappers (can't be type arguments, or outside this PR's scope); the `out var` name-sensitive `Equals` check in `CallBuilder.IsUnambiguousCall` -- with a merged T it degrades `out var x` to an explicitly typed `out (int a, string b) x`, which is exactly what keeps a later `x.b` compiling, so that is an improvement, not a regression. CI is still pending at the time of writing.

PR #4010, opened 2026-08-15
Drop the fixup/autosquash rule from the commit workflow

siegfriedpammer:
Removes the CLAUDE.md bullet mandating that small follow-up fixes be squashed back into the commit they amend via `git commit --fixup` + `git rebase --autosquash`.

Squashing back rewrites already-pushed branch history, which churns review threads on open PRs. How to sequence commits on a branch is left to whoever owns the branch.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -82,7 +82,6 @@ Solutions & filters: `ILSpy.sln` builds everything; `ILSpy.XPlat.slnf` is the de
 - **Subject** is a succinct phrase describing the change. Target <= 72 chars. No area prefix; the subject itself should make the change clear.
 - **Body explains the *why*** and the non-diff context only: the constraint, the prior incident, the decision, what was tried and rejected, the invariant that motivated the change. Keep it short — one short paragraph is usually enough. The diff already shows the *what* — don't restate it, and don't enumerate per-file changes.
 - `Fix #NNNN: ...` closes an issue. `#NNNN` references one without closing.
-- **Small follow-up fixes against a commit still on the branch get squashed back** via `git commit --fixup` + `git rebase --autosquash`, not appended as "fix X" commits.
 - **en-US English** in subject and body. ASCII-only unless a non-ASCII character is genuinely required for what the message describes.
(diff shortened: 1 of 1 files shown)


christophwille (commented):
Reviewed the one-line CLAUDE.md change (removal of the fixup/autosquash bullet) for correctness, removed-behavior, cross-references, and conventions.

- Nothing else in the repo references the removed rule (`grep -ri 'fixup|autosquash|squash'` over `*.md`/`*.yml`/`.github` hits only this line), so the deletion leaves no dangling references.
- Commit subject/body/trailer follow the remaining Commit workflow rules (`Assisted-by:` trailer, why-only body).

One low-severity note inline: the deletion drops the rule wholesale, but the stated motivation (don't rewrite already-pushed history) is not written down anywhere, and the local-only case the rule also covered now has no guidance either. Consider narrowing instead of deleting. Not blocking.

CLAUDE.md
```diff
@@ -82,7 +82,6 @@ Solutions & filters: `ILSpy.sln` builds everything; `ILSpy.XPlat.slnf` is the de
 - **Subject** is a succinct phrase describing the change. Target <= 72 chars. No area prefix; the subject itself should make the change clear.
 - **Body explains the *why*** and the non-diff context only: the constraint, the prior incident, the decision, what was tried and rejected, the invariant that motivated the change. Keep it short — one short paragraph is usually enough. The diff already shows the *what* — don't restate it, and don't enumerate per-file changes.
 - `Fix #NNNN: ...` closes an issue. `#NNNN` references one without closing.
-- **Small follow-up fixes against a commit still on the branch get squashed back** via `git commit --fixup` + `git rebase --autosquash`, not appended as "fix X" commits.
```
christophwille, line 85: The motivation for dropping this (rewriting already-pushed history churns review threads) lives only in the PR/commit body, not in CLAUDE.md, so after this merge an agent is neither told to squash locally nor told *not* to autosquash + force-push an open PR branch -- the harmful case this PR is trying to avoid is now simply unspecified. The old rule was also scoped to "a commit still on the branch", which includes unpushed local commits, where autosquash is harmless and keeps "fix typo" noise out of the branch.

Suggestion (optional): replace rather than delete, e.g. "Squash small follow-up fixes into their target commit only while the branch is unpushed; once a PR is open, append follow-up commits and never rewrite/force-push the branch." That states the invariant you actually care about.

PR #4013, opened 2026-08-16
Add ILAst output to ilspycmd debug builds

siegfriedpammer:
The decompiler's intermediate representation was only reachable through the GUI's ILAst
language, so debugging a transform or a matcher meant driving the UI by hand. `--ilast` writes
the same representation to stdout, and `--after-transform <name-or-index>` truncates the pipeline
at a chosen point, which makes the effect of a single transform diffable and scriptable:

    ilspycmd --ilast -m M:Some.Type.Method Assembly.dll
    ilspycmd --ilast --after-transform ILInlining -t Some.Type Assembly.dll

What is dumped follows the usual selection options: `-m` a single method, `-t` every method of a
type, neither one every method of the assembly. Methods are enumerated over the metadata handles
rather than `ITypeDefinition.Methods`, so property and event accessors are included. `-usepdb`
reaches the `ILReader`, so the dump carries the same local names the real pipeline sees, and `-o`
writes one `<assembly>.ilast` per input file.

A method whose body cannot be read, transformed or written does not abort the run: the partial
function is printed with the exception, and the failure travels through the same
decompilation-error path as the other commands, so it shows up on stderr and in the exit code
unless `--ignore-decompilation-errors` is passed.

`--after-transform` takes a transform name or a 1-based pipeline index, and prints the pipeline
when it cannot resolve one. Entries that are a `BlockILTransform` list the transforms they run,
which is both what tells two such entries apart and what makes the nested transforms visible;
asking for one of those by name reports which entry runs it. Stopping after a nested transform
individually still needs the `Stepper`, which is compiled out of release builds anyway.

The writing options match the UI's ILAst pane (field and logic-operation sugar), so output from
the two can be diffed directly.

Both options are `#if DEBUG`-only, like the UI language they mirror: ILAst serves ILSpy's own
development, not users of the released tool, so they stay out of the shipped NuGet package and
out of the README's option list.

Covered by 9 tests in `ICSharpCode.ILSpyCmd.Tests/ILAstOptionTests.cs` (31 in the project); full
solution builds clean.

### DIFF
--- a/ICSharpCode.ILSpyCmd.Tests/ILAstOptionTests.cs
+++ b/ICSharpCode.ILSpyCmd.Tests/ILAstOptionTests.cs
@@ -1,0 +1,130 @@
+// Copyright (c) 2026 Siegfried Pammer
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this
+// software and associated documentation files (the "Software"), to deal in the Software
+// without restriction, including without limitation the rights to use, copy, modify, merge,
+// publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons
+// to whom the Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all copies or
+// substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+// INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+// PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+// FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+// The --ilast options exist in debug builds only, so this fixture compiles away with them.
+#if DEBUG
+
+using System.Threading.Tasks;
+
+using NUnit.Framework;
+
+using static ICSharpCode.ILSpyCmd.Tests.CliTestRunner;
+
+namespace ICSharpCode.ILSpyCmd.Tests
+{
+	[TestFixture]
+	public class ILAstOptionTests
+	{
+		static readonly string testAssemblyPath = typeof(ILAstOptionTests).Assembly.Location;
+
+		const string sumLoopId = "M:ICSharpCode.ILSpyCmd.Tests.ILAstSample.SumLoop(System.Int32)";
+
+		static Task<(int ExitCode, string Output, string Error)> RunILAstAsync(params string[] args)
+		{
+			string[] common = { testAssemblyPath, "--disable-updatecheck", "-m", sumLoopId };
+			string[] all = new string[common.Length + args.Length];
+			common.CopyTo(all, 0);
+			args.CopyTo(all, common.Length);
+			return RunAsync(all);
+		}
+
+		[Test]
+		public async Task ILAstOfSelectedMethodIsWritten()
+		{
+			var result = await RunILAstAsync("--ilast");
+
+			Assert.That(result.ExitCode, Is.EqualTo(0), result.Error);
+			Assert.That(result.Output, Does.Contain(nameof(ILAstSample.SumLoop)));
+			Assert.That(result.Output, Does.Contain("ILFunction"));
+		}
+
+		[Test]
+		public async Task StoppingAfterATransformYieldsDifferentILAst()
+		{
+			var full = await RunILAstAsync("--ilast");
+			// ILInlining is the third transform of the pipeline; stopping there leaves the
+			// method as unstructured blocks, while the full run has loops and expressions.
+			var partial = await RunILAstAsync("--after-transform", "ILInlining");
+
+			Assert.That(full.ExitCode, Is.EqualTo(0), full.Error);
+			Assert.That(partial.ExitCode, Is.EqualTo(0), partial.Error);
+			Assert.That(partial.Output, Does.Contain(nameof(ILAstSample.SumLoop)));
+			Assert.That(partial.Output, Is.Not.EqualTo(full.Output));
+		}
+
+		[Test]
+		public async Task TransformCanBeSelectedByIndex()
+		{
+			var partial = await RunILAstAsync("--after-transform", "1");
+			var full = await RunILAstAsync("--ilast");
+
+			Assert.That(partial.ExitCode, Is.EqualTo(0), partial.Error);
+			Assert.That(partial.Output, Is.Not.EqualTo(full.Output));
+		}
+
+		[Test]
+		public async Task UnknownTransformNameListsThePipeline()
+		{
+			var result = await RunILAstAsync("--after-transform", "NoSuchTransform");
+
+			Assert.That(result.ExitCode, Is.EqualTo(ProgramExitCodes.EX_USAGE));
+			// the error must be actionable on its own: it lists the pipeline in run order
+			Assert.That(result.Error, Does.Contain("ControlFlowSimplification"));
+			Assert.That(result.Error, Does.Contain("AssignVariableNames"));
+		}
+
+		[Test]
+		public async Task AmbiguousTransformNameReportsItsOccurrences()
+		{
+			// SplitVariables runs three times; the name alone cannot identify a stop point
+			var result = await RunILAstAsync("--after-transform", "SplitVariables");
+
+			Assert.That(result.ExitCode, Is.EqualTo(ProgramExitCodes.EX_USAGE));
+			Assert.That(result.Error, Does.Contain("SplitVariables"));
+			Assert.That(result.Error, Does.Contain("2"));
+		}
+
+		[Test]
+		public async Task WholeTypeCanBeDumped()
+		{
+			var result = await RunAsync(testAssemblyPath, "--disable-updatecheck",
+				"-t", "ICSharpCode.ILSpyCmd.Tests.ILAstSample", "--ilast");
+
+			Assert.That(result.ExitCode, Is.EqualTo(0), result.Error);
+			Assert.That(result.Output, Does.Contain(nameof(ILAstSample.SumLoop)));
+			Assert.That(result.Output, Does.Contain(nameof(ILAstSample.Identity)));
+		}
+	}
+
+	public static class ILAstSample
+	{
+		public static int SumLoop(int n)
+		{
+			int sum = 0;
+			for (int i = 0; i < n; i++)
+			{
+				sum += i;
+			}
+			return sum;
+		}
+
+		public static string Identity(string value) => value;
+	}
+}
+
+#endif
--- a/ICSharpCode.ILSpyCmd.Tests/ILAstOptionTests.cs
+++ b/ICSharpCode.ILSpyCmd.Tests/ILAstOptionTests.cs
@@ -19,6 +19,7 @@
 // The --ilast options exist in debug builds only, so this fixture compiles away with them.
 #if DEBUG
 
+using System.IO;
 using System.Threading.Tasks;
 
 using NUnit.Framework;
@@ -53,6 +54,11 @@ namespace ICSharpCode.ILSpyCmd.Tests
 			Assert.That(result.Output, Does.Contain("ILFunction"));
 		}
 
+		// the loop of SumLoop is only recognised by HighLevelLoopTransform, at the very end of the
+		// pipeline: its presence tells a full run from a truncated one, while the header line
+		// (which names the requested transform count) would differ either way
+		const string structuredLoop = "BlockContainer (for)";
+
 		[Test]
 		public async Task StoppingAfterATransformYieldsDifferentILAst()
 		{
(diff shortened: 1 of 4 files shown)


ICSharpCode.ILSpyCmd/IlspyCmdProgram.cs
```diff
@@ -627,6 +650,65 @@ int ShowIL(string assemblyFileName, TextWriter output)
...
+				methods = new[] { decompiler.TypeSystem.MainModule.GetDefinition((MethodDefinitionHandle)handle) };
+			}
+			else if (TypeName != null)
+			{
+				if (!TryResolveType(decompiler.TypeSystem, TypeName, out ITypeDefinition typeDefinition, out string error))
+				{
+					Console.Error.WriteLine(error);
+					return ProgramExitCodes.EX_DATAERR;
+				}
+				methods = typeDefinition.Methods;
```
christophwille: `ITypeDefinition.Methods` (and `type.Methods` in the whole-assembly branch below) is built by `MetadataTypeDefinition` dropping every method that has `MethodSemantics`, i.e. property get/set and event add/remove accessors never appear here. So `-t` / whole-assembly `--ilast` silently skips accessor bodies even though the help text promises every method.

Verified: `ilspycmd ilspycmd.dll -t ICSharpCode.ILSpyCmd.ILSpyCmdProgram --ilast` prints 30 methods and no `get_*`/`set_*` bodies; a transform bug in a getter is only reachable via `-m M:...get_X`.

Iterating the metadata handles (`GetMethods()` on the type handle / `metadata.MethodDefinitions` for the assembly case) and resolving each via `MainModule.GetDefinition` covers them.

ICSharpCode.ILSpyCmd/ILAstDumper.cs
```diff
@@ -0,0 +1,157 @@
...
+			var handle = (MethodDefinitionHandle)method.MetadataToken;
+			var methodDefinition = metadataFile.Metadata.GetMethodDefinition(handle);
+			if (!methodDefinition.HasBody())
+				return;
+
+			output.WriteLine($"// {method.FullName}");
+			output.WriteLine($"// ILAst after {transformCount} of {TransformCount} transforms ({TransformNames[transformCount - 1]})");
+
+			var reader = new ILReader(decompiler.TypeSystem.MainModule) {
+				UseDebugSymbols = settings.UseDebugSymbols,
```
christophwille: `UseDebugSymbols` alone does nothing: `ILReader` only consults the PDB when `DebugInfo` is set (see `CSharpDecompiler.DecompileBody`, which also passes `DebugInfo = DebugInfoProvider`). Without it, `-usepdb` is inert for the ILAst dump and local names / PDB extra type info never reach the ILAst.

Verified: `--ilast -usepdb -m M:...SumLoop` prints `V_0`/`V_1` while the C# output of the same command uses `sum`/`i`, so the dumped ILAst isn't what the real pipeline is fed.

```cs
var reader = new ILReader(decompiler.TypeSystem.MainModule) {
	UseDebugSymbols = settings.UseDebugSymbols,
	DebugInfo = decompiler.DebugInfoProvider,
	...
};
```

ICSharpCode.ILSpyCmd/IlspyCmdProgram.cs
```diff
@@ -347,6 +358,18 @@ int PerformPerFileAction(string fileName)
 
 					return ShowIL(fileName, output);
 				}
+#if DEBUG
+				else if (ShowILAstFlag || AfterTransformName != null)
+				{
+					if (outputDirectory != null)
+					{
+						string outputName = Path.GetFileNameWithoutExtension(fileName);
+						output = File.CreateText(Path.Combine(outputDirectory, outputName) + ".ilast");
```
christophwille: With `-o` this reassigns the shared captured `output` writer per input file without disposing the previous one; only the last writer is closed by `OnExecuteAsync`'s `finally`, so earlier `.ilast` files lose their buffered tail.

Verified: `ilspycmd A.dll B.dll --ilast -o out` wrote `out/A.ilast` at 318464 bytes ending mid-line vs 318957 bytes when `A.dll` is dumped alone.

The `DumpTable` branch in this same method already handles this with a per-file `using var writer = File.CreateText(...)`; same pattern here.

ICSharpCode.ILSpyCmd/ILAstDumper.cs
```diff
@@ -0,0 +1,157 @@
...
+				UseRefLocalsForAccurateOrderOfEvaluation = settings.UseRefLocalsForAccurateOrderOfEvaluation,
+			};
+			var body = metadataFile.GetMethodBody(methodDefinition.RelativeVirtualAddress);
```
christophwille: Only `RunTransforms` is inside the `try`; `GetMethodBody`, `ReadIL`, `CreateILTransformContext` and `WriteTo` are unguarded and `ShowILAst` has no per-method catch, so a single malformed body escapes to `OnExecuteAsync`'s catch-all and aborts the whole dump with `EX_SOFTWARE`. `CSharpDecompiler.DecompileBody` isolates `BadImageFormatException` per method for exactly this reason.

Scenario: whole-assembly `--ilast` on an obfuscated assembly with one bad body RVA stops right after that method's header line was written; everything after it is missing and the truncated file looks complete.

ICSharpCode.ILSpyCmd/ILAstDumper.cs
```diff
@@ -0,0 +1,157 @@
...
+			try
+			{
+				function.RunTransforms(transforms.Take(transformCount), context);
+			}
+			catch (Exception ex)
+			{
+				// Showing how far the pipeline got is the point of this command, so a crashing
+				// transform prints its exception and then the partially transformed function
+				// rather than aborting the whole dump.
+				output.WriteLine(ex.ToString());
```
christophwille, line 189: Printing the partial function is useful, but the crash is written into the normal output stream and the method returns normally: `ShowILAst` always returns 0 and nothing goes through `ReportDecompilationErrors` / `ExitCodeForDecompilationErrors`, so `--ignore-decompilation-errors` has no meaning for this command and scripts can't detect the failure.

Scenario: a CI step runs `ilspycmd lib.dll --ilast -o out` after a transform change; a `NullReferenceException` in a transform is buried in `out/lib.ilast`, stderr is empty, exit code is 0.

Suggest also collecting the exception (or at least a count) and feeding it through the existing error-reporting path so the exit code reflects it.

ICSharpCode.ILSpyCmd/IlspyCmdProgram.cs
```diff
@@ -627,6 +650,65 @@ int ShowIL(string assemblyFileName, TextWriter output)
...
+
+			int transformCount = ILAstDumper.TransformCount;
+			if (AfterTransformName != null
+				&& !ILAstDumper.TryResolveTransformCount(AfterTransformName, out transformCount, out string transformError))
+			{
+				app.Error.WriteLine(transformError);
+				return ProgramExitCodes.EX_USAGE;
+			}
+
+			var settings = GetSettings(new PEFile(assemblyFileName));
```
christophwille: `GetDecompiler` already calls `GetSettings`, so this runs the settings pipeline a second time (opening another `PEFile`, repeating its stderr side effects).

Verified: `--ilast -ds Bogus=1 -m ...` prints "Decompiler setting 'Bogus' is unknown." twice; a broken `--ilspy-settingsfile` is reported twice as well.

Build the settings once and hand them to the decompiler, or read them back from `decompiler.Settings`.

ICSharpCode.ILSpyCmd.Tests/ILAstOptionTests.cs
```diff
@@ -0,0 +1,130 @@
...
+		{
+			var full = await RunILAstAsync("--ilast");
+			// ILInlining is the third transform of the pipeline; stopping there leaves the
+			// method as unstructured blocks, while the full run has loops and expressions.
+			var partial = await RunILAstAsync("--after-transform", "ILInlining");
+
+			Assert.That(full.ExitCode, Is.EqualTo(0), full.Error);
+			Assert.That(partial.ExitCode, Is.EqualTo(0), partial.Error);
+			Assert.That(partial.Output, Does.Contain(nameof(ILAstSample.SumLoop)));
+			Assert.That(partial.Output, Is.Not.EqualTo(full.Output));
```
christophwille: This (and `TransformCanBeSelectedByIndex`) only asserts `partial.Output != full.Output`, but `WriteMethod`'s header line embeds the requested count (`// ILAst after 3 of 38 transforms (ILInlining)`), so the outputs differ even if `--after-transform` stopped nothing. If `transforms.Take(transformCount)` regressed to running the full pipeline, both tests would still pass.

Assert on body content instead, e.g. that the partial output has no `loop`/structured block after ILInlining while the full one does, or on the header text plus a body property.

ICSharpCode.ILSpyCmd.Tests/ILAstOptionTests.cs
```diff
@@ -0,0 +1,130 @@
...
+
+		[Test]
+		public async Task AmbiguousTransformNameReportsItsOccurrences()
+		{
+			// SplitVariables runs three times; the name alone cannot identify a stop point
+			var result = await RunILAstAsync("--after-transform", "SplitVariables");
+
+			Assert.That(result.ExitCode, Is.EqualTo(ProgramExitCodes.EX_USAGE));
+			Assert.That(result.Error, Does.Contain("SplitVariables"));
+			Assert.That(result.Error, Does.Contain("2"));
```
christophwille: The unknown-name and out-of-range branches also return `EX_USAGE` and print the pipeline listing, which contains "SplitVariables" and the digit "2", so this test can't distinguish the ambiguity message from the others. Dropping the `occurrences.Length > 1` branch (falling through to "Unknown transform") keeps it green.

Assert on something specific to the ambiguity path, e.g. `Does.Contain("runs 3 times")` or the exact index list.

ICSharpCode.ILSpyCmd/ILAstDumper.cs
```diff
@@ -0,0 +1,157 @@
...
+		readonly ILAstWritingOptions writingOptions = new ILAstWritingOptions();
+
+		/// <summary>
+		/// Names of the IL transforms, in the order they run. Transforms nested inside a
+		/// BlockILTransform (LoopDetection, ConditionDetection, the statement transforms, ...)
+		/// are not listed: they run as part of their containing entry and cannot be stopped
+		/// after individually.
+		/// </summary>
+		public static IReadOnlyList<string> TransformNames { get; } =
+			CSharpDecompiler.GetILTransforms().Select(t => t.GetType().Name).ToArray();
```
christophwille: `GetType().Name` makes both `BlockILTransform` entries appear as identical bare `BlockILTransform` rows in `DescribePipeline`, in the ambiguity error and in the dump header, and `LoopDetection`/`ConditionDetection` are never mentioned anywhere.

Verified: `--after-transform LoopDetection` -> "Unknown transform" with a listing that never contains `LoopDetection`; `--after-transform BlockILTransform` -> "runs 2 times, at index 22, 25" with no way to tell which one holds `ConditionDetection`.

`ILFunction.RunTransforms` already labels these entries with `BlockILTransform.ToString()` (which lists the nested names); reusing `t is BlockILTransform b ? b.ToString() : t.GetType().Name` for display fixes all three places.

ICSharpCode.ILSpyCmd/ILAstDumper.cs
```diff
@@ -0,0 +1,157 @@
...
+{
+	/// <summary>
+	/// Writes the decompiler's intermediate representation (ILAst) of a method body, optionally
+	/// stopping the IL transform pipeline after a chosen transform. This is the command-line
+	/// counterpart of the UI's "ILAst" language, and makes transform output diffable.
+	/// </summary>
+	class ILAstDumper
+	{
+		readonly IReadOnlyList<IILTransform> transforms = CSharpDecompiler.GetILTransforms();
+		readonly ILAstWritingOptions writingOptions = new ILAstWritingOptions();
```
christophwille: The UI's ILAst language writes with `UseFieldSugar = true, UseLogicOperationSugar = true` (`DebugStepsPaneModel.WritingOptions`), while this uses the defaults (both false). So the same method renders `ldobj(ldflda ...)` / nested if-blocks on the CLI vs `ldfld` / `logic.and` in the UI pane; the "same representation as the UI" claim in the class doc doesn't hold and UI-vs-CLI diffs are noisy.

One-line fix: initialize the options with both sugars true (or expose switches).

christophwille (commented):
Automated review (Claude Code, `/code-review`) of `--ilast` / `--after-transform`.

Built the branch and ran the `ILSpyCmd.Tests` suite locally (28/28 pass), then probed the built tool by hand. Conventions (headers, ASCII, TDD, `Assisted-by:` trailer) look clean. Ten inline findings below: the first six are correctness issues reproduced on the branch, the last four are test-strength and UI-parity/reuse.

**Correctness**
1. `-t` / whole-assembly `--ilast` never dumps property/event accessors (`ITypeDefinition.Methods` filters them out).
2. `ILReader.DebugInfo` isn't set, so `-usepdb` is inert for the ILAst dump (`V_0` vs `sum` in the C# output of the same command).
3. `-o` with multiple inputs leaks/truncates all but the last `.ilast` file (verified: 318464 vs 318957 bytes).
4. Only `RunTransforms` is guarded; a bad method body aborts the whole dump.
5. Transform crashes go to stdout with exit code 0, so `--ignore-decompilation-errors` has no meaning here.
6. Settings pipeline runs twice (duplicate diagnostics on stderr).

**Tests / parity**
7-8. Two tests pass even if the feature under test regresses (header line / EX_USAGE ambiguity).
9. Both `BlockILTransform` entries display identically; `LoopDetection` is unfindable.
10. Writing options differ from the UI's ILAst pane (no field/logic sugar).

PR #4014, opened 2026-08-16
Emit C# 7.1 default literals where the context supplies the type

christophwille:
Alternative implementation of #3913: same feature, same tests, different mechanism.

## Approach

Shortening `default(T)` is the same problem as removing the redundant cast around a lambda whose delegate type the context already fixes, so it uses the same mechanism instead of a dedicated AST transform: `default(T)` is emitted as before, and `TranslatedExpression.ConvertTo(targetType, allowImplicitConversion: true)` makes the explicit type implicit when the conversion is an identity conversion - a few lines above the existing lambda/`ConversionResolveResult` case in the very same block.

- **Shorten**: the `DefaultValueExpression` loses its type and is annotated with `DefaultLiteralResolveResult`, which remembers the type it was shortened from.
- **Restore**: any later `ConvertTo` that is not implicit, or that supplies a different type, spells `default(T)` out again and converts that. This is what keeps `object o = default(SomeStruct)` boxing (the bare literal would be `null`), and what `CastArguments` uses when an overload resolution recheck rejects the untyped literal.
- Because the shortened literal resolves to `DefaultLiteralResolveResult`, `CallBuilder`'s existing `IsUnambiguousCall` recheck sees a real default literal and rejects ambiguous calls on its own - no bookkeeping about which arguments may stay untyped, no argument annotations.

Only the contexts that supply no target type at all restore the explicit form via `RestoreDefaultLiteralType`: an awaited expression (`await default` does not compile) and arguments of operator methods (`ReplaceMethodCallsWithOperators` turns those into operator or cast syntax). Two further spots opt *in* to the shortening because their declaration always spells out the type: `DeclareVariables`' synthesized initializer and `using` resource declarations.

Compared to #3913 this drops the `IntroduceDefaultLiterals` transform, `UseImplicitlyTypedDefaultAnnotation` and the `ArgumentList.UseImplicitlyTypedDefault` plumbing in `CallBuilder`. Kept from it unchanged: optional `DefaultValueExpression.Type` plus output visitor, `DefaultLiteralResolveResult` / `Conversion.DefaultLiteralConversion` / the `CSharpConversions` hook, and the `DefaultLiterals` setting (C# 7.1, on by default).

## Tests

The tests are the ones from #3913. Three expectations differ, because `ConvertTo` covers different ground than the transform did - each fixture is also the compiled input, so csc validates every one of them:

- `this[default]` - indexer arguments shorten (#3913 pinned them typed); the fixture method is renamed `IndexerArgumentStaysTyped` -> `IndexerArgument`.
- `() => default` in `GuidExpressionTree` - expression-tree lambda bodies shorten.
- `Params(default(int), default(int), default(int))` - params-array elements synthesized by the params expansion never pass through `ConvertTo`, so they keep their type.

These fixtures needed updating on top of #3913's set, because `ConvertTo` reaches contexts the transform did not: `RefFields`, `RefStructInterfaces`, `FirstClassSpanTypes`, `FirstClassSpanConversions`, `QueryExpressions`, `DeconstructionTests`, `SpanConversionOperatorMismatch`. In `Issue2260SwitchString` the `decimal x = default;` lines from #3913 are back to `0m`, which is what master emits there today.

Full `ICSharpCode.Decompiler.Tests` suite passes (the Windows-only test projects were not run - this was developed on macOS).

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/CompoundAssignmentTest.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/CompoundAssignmentTest.cs
@@ -5024,7 +5024,11 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 		}
 		public static void CustomStructAddTest(CustomStruct p, CustomClass c, CustomStruct2 s)
 		{
+#if CS71
+			CustomStruct num = default;
+#else
 			CustomStruct num = default(CustomStruct);
+#endif
 			p += default(CustomStruct);
 			num += default(CustomStruct);
 			Use(ref num);
@@ -5051,7 +5055,11 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 
 		public static void CustomStructSubtractTest(CustomStruct p, CustomClass c, CustomStruct2 s)
 		{
+#if CS71
+			CustomStruct num = default;
+#else
 			CustomStruct num = default(CustomStruct);
+#endif
 			p -= default(CustomStruct);
 			num -= default(CustomStruct);
 			Use(ref num);
@@ -5078,7 +5086,11 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 
 		public static void CustomStructMultiplyTest(CustomStruct p, CustomClass c, CustomStruct2 s)
 		{
+#if CS71
+			CustomStruct num = default;
+#else
 			CustomStruct num = default(CustomStruct);
+#endif
 			p *= default(CustomStruct);
 			num *= default(CustomStruct);
 			Use(ref num);
@@ -5105,7 +5117,11 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 
 		public static void CustomStructDivideTest(CustomStruct p, CustomClass c, CustomStruct2 s)
 		{
+#if CS71
+			CustomStruct num = default;
+#else
 			CustomStruct num = default(CustomStruct);
+#endif
 			p /= default(CustomStruct);
 			num /= default(CustomStruct);
 			Use(ref num);
@@ -5132,7 +5148,11 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 
 		public static void CustomStructModulusTest(CustomStruct p, CustomClass c, CustomStruct2 s)
 		{
+#if CS71
+			CustomStruct num = default;
+#else
 			CustomStruct num = default(CustomStruct);
+#endif
 			p %= default(CustomStruct);
 			num %= default(CustomStruct);
 			Use(ref num);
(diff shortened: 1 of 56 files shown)


ICSharpCode.Decompiler/CSharp/TranslatedExpression.cs
```diff
@@ -195,11 +209,37 @@ public TranslatedExpression UnwrapChild(Expression descendant)
...
 			{
 				// Make explicit conversion implicit, if possible
 				if (allowImplicitConversion)
 				{
+					if (Expression is DefaultValueExpression { Type: not null }
+						&& expressionBuilder.settings.DefaultLiterals)
+					{
+						// The target type is supplied by the context, so "default(T)" can be
+						// shortened to the C# 7.1 default literal.
+						var shortened = new DefaultValueExpression();
```
dgrunwald: Cloning the node is unnecessary, this could have been an in-place modification.

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -4425,6 +4425,8 @@ protected internal override TranslatedExpression VisitAwait(Await inst, Translat
 			if (expectedType != null)
 			{
 				value = value.ConvertTo(expectedType, this, allowImplicitConversion: true);
+				// The awaited expression is not target-typed: "await default" does not compile.
+				value = value.RestoreDefaultLiteralType(this);
```
dgrunwald: This is a workaround for a bug we'll fix separately on another branch -- really `await` shouldn't be using allowImplicitConversion
.

ICSharpCode.Decompiler.Tests/TestCases/Pretty/CompoundAssignmentTest.cs
```diff
@@ -5400,7 +5452,11 @@ public static void CustomStructPostDecTest(CustomStruct p, CustomClass c, Custom
 
 		public static void CustomStructPreDecTest(CustomStruct p, CustomClass c, CustomStruct2 s)
 		{
+#if CS71
```
dgrunwald, line 5455: I'm not sure if it's such a great idea to add so many `#if` to so many tests -- this seems like it might make maintenance of the tests harder in the future.
Maybe we instead should disable this feature for the majority of the impacted tests?

ICSharpCode.Decompiler.Tests/TestCases/Pretty/TargetTypedDefault.cs
```diff
@@ -0,0 +1,227 @@
...
+
+		public void OptGenericStruct<T>(T value = default(T)) where T : struct
+		{
+		}
+
+		public void OptTuple((int, string) pair = default((int, string)))
+		{
+		}
+
+		public void CallsWithExplicitDefaults()
```
dgrunwald, line 96: Method name doesn't seem to fit the body?

dgrunwald (approved):

PR #4022, opened 2026-08-16
Improve floating-point fraction and negative constant reconstruction

### LINKED ISSUE #1459: RFE: Fraction decompilation - more weight to common divisors?
I really liked the introduction of fractions instead of numbers with long decimal sequences. It cought me by surprise and put a smile on my face. That feature is the very definition of _quality_. Really good job!

But, there's always room for improvement. :-)

I think it would make sense to give preference to the following divisors:
2^n for n in (8,10,12-16, 20)
2^n-1 for n in (8,10,12,15,16)

Some rationale:
255 (2^8-1) is very common when converting (u)byte (f.ex. RGB) <-> normalized float.
2^10 and 2^20 is common when converting larger numbers to computer-K and -M.
I saw a long color initialization sequence where the decompiler preferred 51 as divisor, where 255 would have been more obvious (example below).

Coming to think of it, should negative numbers become handled, maybe even 2^7 and 2^7-1 should be considered.

I have also seen code normalizing values in f.ex the ranges +/- 4K, 16K, 32K and 64K, so maybe negative numbers should be considered too (at least for a/this small subset of potential divisors)?

I believe that would allow better/nicer decompilation of some code than the following example of current decompiler output (all three should preferrably be divisor 255f):
`0.882352948f, 40f / 51f, 0.5882353f`.

Again, good job!

SychicBoy:
Fixes #1459

## Summary

Improve floating-point constant reconstruction by preferring exact, recognizable normalization fractions where appropriate, while preserving simpler existing representations.

This also extends well-known constant detection to recognize negated constants such as `-float.Epsilon` and `-double.Epsilon`.

## Details

Some normalized floating-point values are currently reconstructed into less recognizable forms. For example:

```csharp
200f / 255f
```

may become:

```csharp
40f / 51f
```

Although equivalent, the `/255f` representation better communicates common byte-normalization patterns.

This change adds a restricted set of preferred denominators based on powers of two and `2^n - 1` scales. Candidates are only selected when they reproduce the exact original `float` or `double` value.

The existing fraction approximation logic remains unchanged, and simpler fractions are preserved rather than expanded into larger equivalent representations.

Negative well-known constants are also recognized, allowing values such as:

```csharp
-float.Epsilon
-double.Epsilon
```

to be emitted symbolically instead of as numeric literals.

## Examples

Before:

```csharp
0.88235295f
40f / 51f
0.5882353f

-1E-45f
-5E-324
```

After:

```csharp
225f / 255f
200f / 255f
150f / 255f

-float.Epsilon
-double.Epsilon
```

## Implementation notes

* Preserves exact `float` and `double` values and existing simpler fraction representations.
* Keeps the existing fraction approximation limits unchanged.
* Restricts preferred fractions to a small set of common normalization scales.
* Keeps existing canonical special constants authoritative.

## Tests

Regression coverage has been added to the existing `WellKnownConstants` Pretty test fixture for:

* `/255f` byte-normalized values
* negative normalized values
* `double` fractions
* power-of-two K/M-scale denominators
* preservation of simpler fractions
* `-float.Epsilon`
* `-double.Epsilon`

The added cases verify both the new behavior and that less-readable equivalent representations are not introduced.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/WellKnownConstants.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/WellKnownConstants.cs
@@ -95,6 +95,14 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 		public const double Double_ByteScale = 200.0 / 255.0;
 		public const float Float_KScale = 123f / 1024f;
 		public const float Float_MScale = 123f / 1048576f;
+		public const float Float_UnitKScale = 1f / 1024f;
+		public const float Float_SmallKScale = 17f / 1024f;
+		public const float Float_UnreducedKScale = 126f / 1024f;
+		public const float Float_ShortBinaryFraction = 13f / 256f;
+		public const float Float_NearOnePowerOfTwo = 225f / 256f;
+		public const float Float_TinyPowerOfTwoScale = 1f / 32768f;
+		public const float Float_NotAPreferredScale = 0.123456f;
+		public const double Double_NotAPreferredScale = 0.123456789;
 
 #if ROSLYN2 && !NET40
 		public const float Float_PI = MathF.PI;
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/WellKnownConstants.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/WellKnownConstants.cs
@@ -84,6 +84,17 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 		public const double Double_Sixth = 1.0 / 6.0;
 		public const float Float_Tenth = 0.1f;
 		public const double Double_Tenth = 0.1;
+		public const float Float_Third = 1f / 3f;
+		public const float Float_PowerOfTwoFraction = 21f / 32f;
+		public const float Float_SmoothFraction = 2f / 15f;
+		public const float Float_UnitFraction = 1f / 85f;
+		public const float Float_ByteScale_225 = 225f / 255f;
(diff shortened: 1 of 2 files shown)


siegfriedpammer:
This PR looks like it changes a lot of stuff... can you split all the changes into multiple commits, so that there is one small commit per change? Thanks!

christophwille:
https://github.com/icsharpcode/ILSpy/blob/master/CONTRIBUTING.md that awfully looks like a drive-by AI PR. Could you please elaborate why you picked that specific issue and given your Github timeline, how experienced you are with the internal of our decompiler engine? Letting your clanker implement an issue and post a PR is not how we work with contributors.

SychicBoy:
> https://github.com/icsharpcode/ILSpy/blob/master/CONTRIBUTING.md that awfully looks like a drive-by AI PR. Could you please elaborate why you picked that specific issue and given your Github timeline, how experienced you are with the internal of our decompiler engine? Letting your clanker implement an issue and post a PR is not how we work with contributors.

Hi Christoph,

No, this is not a “drive-by AI PR.” English is not my native language, so I sometimes use AI to help me phrase my messages more clearly and in a more formal tone. That may be why the wording came across that way.

To clarify my approach to this PR:

I picked #1459 because it was marked Help Wanted and looked like something I could solve.

I am still new to ILSpy’s internals, but I have made an effort to understand the relevant code paths related to this change. I’m open to technical discussion and feedback on the implementation.

I appreciate maintainers being careful about contributions, and I hope this PR can be evaluated based on the actual code and discussion around it.

dgrunwald (approved):

siegfriedpammer:
@SychicBoy Thank you for your contribution!

PR #4029, opened 2026-08-18
Walk the whole decompiler pipeline in the Debug Steps pane

siegfriedpammer:
The pane used to split the pipeline across two languages: the ILAst language stepped the IL
transforms, the C# language stepped the AST transforms, and nothing showed the seam between
them. A step index therefore meant a different thing depending on which language happened to be
selected.

Recording both halves into one `Stepper` makes an index replayable across the whole pipeline. A
limit that lands in the IL phase has no C# to print, so the halted function is rendered as ILAst
instead.

What is left of the ILAst language is its typed-IL dump, which runs no transforms at all. That
stays, as `TypedILLanguage`. `IDebugStepProvider` was down to a single implementation and is
removed.

## Which function a halt is attributed to

A member group's `EndStep` is the next member's first step, so stopping after a member's last IL
step lands on the *next* member's `StepStartGroup`. The halt is therefore attributed to the last
member whose IL phase completed, preferring the function that actually holds the halted step's
instruction - the outermost `ILFunction` ancestor of `LimitReachedStep.Position`. That also
covers functions that are not yet attached to their parent when the limit hits, such as
`ProxyCallReplacer`'s proxy function.

A transform that throws never reaches the limit, so the crashing member used to fall out of the
recording entirely. It is now recorded when the crash site is already at the step limit, which
brings back the retired ILAst language's "ILAst after the crash" view.

## Cost

Retention is opt-in, and now gated on the pane being open (`DebugStepsPaneModel.IsRecording`,
flipped by the view's logical-tree attach/detach) rather than on every Debug-build decompile.
Every kept step pins the ILAst it captured, which is affordable for the single type the pane
shows but not for a whole-module decompile. Measured on `System.Linq.Enumerable` with recording
on vs off: 84,712 step nodes vs 2,776, 67 MB vs 33 MB live, 8.6 s vs 5.0 s.

## Verification

Rebased onto master (`76ec7588e`). `ILSpy.Desktop.slnf` builds clean, and both suites pass on
Linux:

- `ILSpy.Tests` - 1237 total, 0 failed, 3 skipped
- `ICSharpCode.Decompiler.Tests` - 3499 total, 0 failed, 45 skipped

26 of those cover this change: 17 pane tests in `ILSpy.Tests/Views/DebugStepsTests.cs` and 9
recording tests in `ICSharpCode.Decompiler.Tests/DebugStepRecordingTests.cs`.


ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
```diff
@@ -2333,6 +2390,10 @@ void DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun de
 						decompileRun,
 						CancellationToken
 					);
+					// The seam between the two halves of the pipeline. Besides marking where the IL steps
+					// end and the C# AST steps begin, it is the only handle on the fully transformed
+					// ILAst: every IL step shows the state before some transform, never after the last.
+					context.Step("Convert ILAst to C#", function.Body);
```
christophwille, line 2433: **Bug: `Show state after` on this seam step (or on the member group) of a non-last member shows the wrong member.**

This Step records index N and increments to N+1; `StepEndGroup` then sets the member group's `EndStep = N+1`. The pane replays `state after` with `StepLimit = EndStep = N+1`, and the next `Step` call in the replay is the *next* member's `context.StepStartGroup(method.FullName)` (after its `ReadIL`), which throws. `StepLimitHaltedFunction` becomes the next member's raw, untransformed ILFunction, `TryWriteILAst` replaces the whole document with it, and `HighlightStep = N` (this member's `function.Body`) is not in the rendered NodeLookup, so there is no highlight either. Only for the last member of the type does `after` land in the AST phase and print C#.

A cheap fix: treat a halt that lands exactly on a member's opening group step as `state after the previous member` (e.g. remember the previous member's transformed function before opening the next group and render that when the halted step is a group opener).

ILSpy/Languages/CSharpLanguage.cs
```diff
@@ -327,6 +327,10 @@ CSharpDecompiler CreateDecompiler(MetadataFile module, DecompilationOptions opti
 			};
 			decompiler.Stepper.StepLimit = options.StepLimit;
 			decompiler.Stepper.IsDebug = options.IsDebug;
+			// The Debug Steps pane walks the whole pipeline, so the IL transforms have to record into
+			// the same stepper as the C# AST transforms. Inert unless the decompiler was built with the
+			// STEP symbol, which only its debug builds define.
+			decompiler.RecordILTransformSteps = true;
```
christophwille: **Debug-build regression: every C# decompile now retains all IL steps and eagerly wraps them on the UI thread, whether or not the pane is open.**

`RecordILTransformSteps = true` is unconditional, `OnCSharpDecompiled` keeps `decompiler.Stepper` on the MEF-shared `CSharpLanguage.stepper` until the next full decompile, and `DebugStepsPaneModel` (materialised by ToolPaneRegistry regardless of `IsVisibleByDefault=false`) reacts to `StepperUpdated` with `Dispatcher.UIThread.Post(() => SetStepsSource(...))` -> `StepNodeViewModel.Wrap`, which recursively allocates one VM per node. Pre-PR the IL steps went to a throwaway per-member stepper and only the (small) AST steps were retained/wrapped. For a large type (e.g. `System.Linq.Enumerable` incl. nested iterators) that is tens of thousands of `Stepper.Node`s (each with three lists + ancestor chain, pinning removed ILAst subtrees) plus as many VMs, on every selection in Debug builds.

Cheaper: opt in only while the Debug Steps pane is visible (flip a flag from the pane's show/hide and re-decompile on open), and/or drop `stepper` on `DetachFromLanguage`, and wrap children lazily on expansion.

ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
```diff
@@ -2349,14 +2410,29 @@ void DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun de
...
+			}
 			catch (Exception innerException) when (!(innerException is OperationCanceledException))
 			{
 				// One method the decompiler cannot handle must not cost the user the type or, when
 				// exporting a project, the assembly around it: keep the signature, put the error in
 				// front of it, and let the remaining members decompile.
 				errors.Add(innerException as DecompilerException ?? new DecompilerException(module, method, innerException));
+				// The unwind left this member's step groups open; close them so the members after it are
+				// recorded as its siblings instead of disappearing into the group that failed.
+				Stepper.EndOpenGroups();
```
christophwille: **Lost debugging capability: the ILAst *at the point a transform crashed* is no longer reachable.**

The deleted ILAst language printed the exception plus `ILAst after the crash:` + the half-transformed function. Here the crash becomes a C# error comment and `EndOpenGroups()` sets the group's `EndStep` to the counter at unwind. Replaying that group's `state after` (`StepLimit = EndStep`) re-runs the transform, which throws again before any `Step` reaches the limit, so `StepLimitHaltedFunction` stays null for that member; the *next* member's opening group step then hits the limit and its raw IL is shown instead (or, for the last member, the C# error comment again). Only `state before` the crashing transform is reachable.

If this is intended, fine, but consider recording the crashed function (e.g. set `StepLimitHaltedFunction = function` here too when a step limit is set, or expose a `CrashedFunction` the UI can render) so a developer debugging a throwing transform can still see what it had already mutated.

ICSharpCode.Decompiler.Tests/DebugStepRecordingTests.cs
```diff
@@ -0,0 +1,192 @@
...
+			decompiler.RecordILTransformSteps = true;
+			decompiler.ILTransforms.Add(new ThrowingILTransform("CleanUpFileName"));
+
+			decompiler.DecompileTypeAsString(SampleType);
+
+			var crashed = decompiler.Stepper.Steps.Single(n => n.Description.Contains("CleanUpFileName"));
+			using (Assert.EnterMultipleScope())
+			{
+				Assert.That(crashed.EndStep, Is.GreaterThan(crashed.BeginStep + 1), "the abandoned group was closed at the step it stopped on");
+				Assert.That(decompiler.Stepper.Steps, Has.Some.Matches<Stepper.Node>(n => n.Description.Contains("DecompileProject")),
```
christophwille: **Vacuous assertion.** `WholeProjectDecompiler` is a single non-partial class and members are decompiled in metadata (= source) order; both `DecompileProject` overloads (~L230/239) precede `CleanUpFileName` (~L843), so their groups are top-level whether or not `Stepper.EndOpenGroups()` runs. Removing the `EndOpenGroups()` call from the general catch keeps this assertion green; only the `EndStep` assertion above detects the fix. Assert on a member declared *after* the crash, e.g. `CleanUpDirectoryName` or `CleanUpPath`.

ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
```diff
@@ -2349,14 +2410,29 @@ void DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun de
...
 				CleanUpMethodDeclaration(entityDecl, body, function, localSettings.DecompileMemberBodies);
 			}
+			catch (StepLimitReachedException)
+			{
+				// A step limit reached in the IL phase stops the pipeline for good: this member and every
+				// member after it stay body-less, and the caller renders StepLimitHaltedFunction as ILAst
+				// rather than the C# it would otherwise print. This clause has to stay above the general
+				// handler below, which would turn the halt into an error comment and report the member as
+				// a decompilation failure.
+				StepLimitHaltedFunction = function;
```
christophwille: Note (pre-existing in the ILAst language, but now advertised as a whole-pipeline replay): when the limit is hit inside transforms run on a *detached* helper function that shares this Stepper (`ProxyCallReplacer`'s `proxyFunction`, `LocalFunctionDecompiler`/`DelegateConstruction` nested functions before they are attached), `function` is the enclosing top-level ILFunction, whose `WriteTo` never prints the halted instruction, so `TryResolve` finds no candidate/seam/ancestor and the view shows a different tree with no highlight. Recording the halted function at the exception site (e.g. `context.Function` carried in a `StepLimitReachedException` field) would let the UI render the tree the step actually mutated.

ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
```diff
@@ -2309,11 +2357,20 @@ void DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun de
...
 				};
+				if (RecordILTransformSteps)
+					context.Stepper = Stepper;
+				context.StepStartGroup(method.FullName);
 				foreach (var transform in ilTransforms)
 				{
 					CancellationToken.ThrowIfCancellationRequested();
+					// Named the way ILFunction.RunTransforms names them, so the same transform reads the
+					// same in every step tree.
+					context.StepStartGroup(transform is BlockILTransform blockTransform
```
christophwille: Reuse: this loop body (group naming `BlockILTransform ? ToString() : GetType().Name`, `Run`, `CheckInvariant`, `StepEndGroup(keepIfEmpty: true)`) is a copy of `ILFunction.RunTransforms` (ILFunction.cs ~L402), kept in sync only by the comment. `RunTransforms` also emits `DecompilerEventSource.ILTransformExecuted` tracing that this loop lacks (pre-existing gap, but now the two loops share the group contract). Consider `function.RunTransforms(localSettings.DecompileMemberBodies ? ilTransforms : ilTransforms.TakeWhile(...up to and including AsyncAwaitDecompiler), context)`, or at least a shared `StepStartGroup(IILTransform)` helper used by both.

ILSpy/Views/DebugSteps.axaml
```diff
@@ -13,29 +12,23 @@
...
-							          IsChecked="{Binding UseLogicOperationSugar, Mode=TwoWay}" />
-							<CheckBox Margin="3" Content="Show IL ranges"
-							          IsChecked="{Binding ShowILRanges, Mode=TwoWay}" />
-							<CheckBox Margin="3" Content="Show child index in block"
-							          IsChecked="{Binding ShowChildIndexInBlock, Mode=TwoWay}" />
-						</StackPanel>
-					</DataTemplate>
-				</ContentControl.DataTemplates>
-			</ContentControl>
+			<StackPanel Orientation="Horizontal" DataContext="{Binding Options}" x:CompileBindings="False">
```
christophwille: Simplification: `Options` is now statically typed (`ILAstWritingOptions`, INotifyPropertyChanged), so the four checkboxes can stay on compiled bindings from the pane's existing `x:DataType`: `IsChecked="{Binding Options.UseFieldSugar, Mode=TwoWay}"` etc. (or `x:DataType="il:ILAstWritingOptions"` on this StackPanel). That drops `x:CompileBindings="False"`, the DataContext swap, and the need for the runtime `Writing_Option_CheckBoxes_Are_Bound_To_The_ILAst_Writing_Options` test that only exists to catch renames the compiler would otherwise catch.

ILSpy.Tests/Views/DebugStepsTests.cs
```diff
@@ -217,25 +214,42 @@ public async Task ILAst_DebugStep_Replay_Highlights_Changed_Instruction()
...
-			"a step-limited ILAst replay must not replace the full step tree shown by the pane");
+			"a step-limited replay must not replace the full step tree shown by the pane");
 
 		await tab.RestartDecompileWithStepLimit(replayStep.EndStep, isDebug: false, replayStep.BeginStep);
-		tab.Text.Should().NotBeNullOrWhiteSpace("ILAst replay after a selected step must still emit IL");
-		AssertPreciseHighlight(tab, "ILAst replay after a selected step must locate the changed instruction");
+		tab.Text.Should().NotBeNullOrWhiteSpace("replaying the state after an IL-phase step must still emit output");
+		AssertPreciseHighlight(tab, "an IL-phase replay must locate the changed instruction");
+
+		static string StripStepNumber(string description)
```
christophwille: Duplicate of the identical `StripStepNumber` local function at L187; hoist one to a fixture-level static next to `AssertPreciseHighlight`. Similarly, `DebugStepRecordingTests.CreateDecompiler` and `ThrowingILTransform` are byte-for-byte copies of the helpers in `DecompilationErrorRecoveryTests.cs` (~L128-143) - make those internal and reuse them.

ICSharpCode.Decompiler.Tests/DebugStepRecordingTests.cs
```diff
@@ -0,0 +1,192 @@
...
+using ICSharpCode.Decompiler.TypeSystem;
+
+using NUnit.Framework;
+
+namespace ICSharpCode.Decompiler.Tests
+{
+	/// <summary>
+	/// The Debug Steps view replays a decompilation by index, so a single <see cref="Stepper"/> has to
+	/// span the whole pipeline: the IL transforms, the ILAst-to-C# conversion, and the C# AST
+	/// transforms. These tests pin the IL half, which the C# path used to discard.
```
christophwille: CLAUDE.md: `Comments must stand on their own, with no memory of how the code was written ... never reference the previous version`. `which the C# path used to discard` (and `used to live on a separate ILAst language ... now belongs to C#` in DebugStepsTests.cs ~L277) describe the change rather than the code; reword to what the C# path does now. Also `behaviour` (L113) -> `behavior` (en-US per CLAUDE.md).

christophwille (commented):
Review summary (recall-biased pass, findings verified individually).

The core idea is sound: one Stepper across the IL and AST halves keeps step indices replayable, and the halted-IL rendering path (StepLimitHaltedFunction / TryWriteILAst) is wired consistently. Step numbering is identical between the full run and the step-limited replay as far as I can trace. Findings, most severe first (inline comments on the respective lines):

1. **Bug** - `Show state after` on a non-last member's `Convert ILAst to C#` step or member group halts at the *next* member's `StepStartGroup(method.FullName)` and renders that member's raw, untransformed IL with no highlight (CSharpDecompiler.cs:2396).
2. **Debug-build regression** - `RecordILTransformSteps = true` on every UI C# decompile retains every IL step of every member (each Node pins ILAst incl. removed instructions) on the shared `CSharpLanguage.stepper`, and `StepNodeViewModel.Wrap` eagerly builds a VM per node on the UI thread even with the pane hidden (CSharpLanguage.cs:333).
3. **Lost capability** - the deleted ILAst language's `ILAst after the crash` view is unreachable via the pane: replaying the crashed group's `state after` re-throws before any step reaches the limit (CSharpDecompiler.cs:2435).
4. **Test** - `AFailingTransformDoesNotNestLaterMembersUnderIt` asserts on `DecompileProject`, which is declared *before* `CleanUpFileName` in metadata order, so the sibling assertion passes with or without `EndOpenGroups` (DebugStepRecordingTests.cs:145).
5. **Pre-existing limitation now advertised as replayable** - a step limit hit in a detached helper function (ProxyCallReplacer's proxyFunction, nested functions before attachment) attributes the halt to the top-level function, whose ILAst does not contain the halted instruction (CSharpDecompiler.cs:2424).
6. **Reuse** - the per-transform loop duplicates `ILFunction.RunTransforms` naming/invariant logic (CSharpDecompiler.cs:2368).
7. **Simplification** - `x:CompileBindings="False"` + `Options` forwarder + runtime binding test could be `{Binding Options.UseFieldSugar}` etc. with compiled bindings (DebugSteps.axaml:22).
8. **Test duplication** - `StripStepNumber` twice in DebugStepsTests.cs; `CreateDecompiler`/`ThrowingILTransform` copied from DecompilationErrorRecoveryTests.cs.
9. **Post-merge stale docs (not in this diff)** - master's CLAUDE.md:94 (`like the UI's ILAst language`), ICSharpCode.ILSpyCmd/ILAstDumper.cs:40/47 and IlspyCmdProgram.cs:122 still point at the retired ILAst language; worth rewording on rebase.
10. **Conventions** - comments describing the change relative to the previous version (`which the C# path used to discard`, `used to live on a separate ILAst language`), and `behaviour` (en-US per CLAUDE.md) in DebugStepRecordingTests.cs.

Minor, not commented inline: `Should().Equal(astTransformNames)` weakened to `EndWith`, so no test pins the exact top-level tree shape; the `ref bool handled` partial in WriteCode could be an inline `StepLimitHaltedFunction is { } halted` branch; `EndOpenGroups()` popping to zero relies on nothing ever opening a group around DecompileBody (a depth-restoring close or a `finally` would be more robust).

_Review by an AI agent (Claude) on Christoph's behalf._

christophwille (commented):
## Review summary

High-effort pass: line-by-line, removed-behavior and cross-file correctness angles plus reuse, efficiency and altitude angles, each candidate verified independently against the base tree. The simplification and CLAUDE.md-conventions angles did not complete in time; the one conventions item found on the way (TypedILLanguage.cs header) is noted inline.

The core decompiler change verified clean: every exit path of `DecompileBody` closes its groups, `HaltedStepFunction` cannot resolve to an unrelated member's function (every separately read function is either attached before a counted step or transformed on a private stepper), the reset in `CreateDecompileRun` is reached from every public entry point, and startup with a persisted-open pane is safe (`DecompileAsync` no-ops with no nodes).

What did not hold up is the UI-side wiring of the recording flag.

**Correctness (confirmed)**
1. `CSharpDecompiler.cs:2477` - crash-at-limit check reads the shared `Stepper` even when IL steps went to the throwaway stepper; `StepLimit == 0` plus any throwing IL transform renders an unrelated member's ILAst.
2. `CSharpLanguage.DebugSteps.cs:78` - replay numbering is decided by the static `IsRecording` at `CreateDecompiler` time; the AST-only tree stays clickable after pane-open flips the flag, so a click in that window replays an AST index under IL-inclusive numbering.
3. `CSharpLanguage.DebugSteps.cs:65` - multi-node tabs: each node halts at `StepLimit` in its own numbering; the first ILAst dump sets `.il` for the whole document and later nodes' C# is highlighted as IL.
4. `DebugStepsPaneModel.cs:411` - closing the pane does not stop the in-flight recording run; `OnCSharpDecompiled` re-pins the full tree after `ReleaseSteps()`; and `ReleaseSteps` goes through `activeLanguage`, null whenever the language is not C#.
5. `DebugStepsPaneModel.cs:406` - enable path re-decompiles unconditionally, even for non-C# languages.
6. `StepNodeViewModel.cs:51` - lazy `Children` is fully materialised by `ApplyFilter`/`SnapshotExpansion` on the first filter keystroke.

**Plausible**
7. `DebugSteps.axaml.cs:90` - keying recording to logical-tree attach/detach means any dock relayout (sibling tab switch, float, re-dock) drops the tree and re-decompiles.
8. `CSharpDecompiler.cs:2341` - after an IL-phase halt every remaining member still gets its full AST declaration built and then discarded (bounded, but it is the hot path of step selection).

**Cleanup**
9. `TreeTraversal.PreOrder` replaces the three hand-rolled node walkers in the test files; `StepperTesting.CreateDecompiler` is the existing file-name `CSharpDecompiler` ctor.
10. `TypedILLanguage.cs` is a new file per git with an AlphaSierraPapa header; CLAUDE.md wants the contributor's name on new files.

Items 1, 2, 4, 5 and 7 share a root: the recording flag lives in a view-flipped static rather than travelling in `DecompilationOptions` next to `StepLimit`/`IsDebug`. Carrying it there (set by `RestartDecompileWithStepLimit` from the mode the displayed tree was recorded in) makes full run and replay agree by construction and removes the cross-thread static read; item 4 additionally needs the VM to ignore `StepperUpdated` while the pane is closed.

ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
```diff
@@ -2349,14 +2439,43 @@ void DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun de
...
 				// front of it, and let the remaining members decompile.
 				errors.Add(innerException as DecompilerException ?? new DecompilerException(module, method, innerException));
+				// The unwind left this member's step groups open; close them so the members after it are
+				// recorded as its siblings instead of disappearing into the group that failed.
+				Stepper.EndOpenGroups(groupDepth);
+				// A replay aiming at the state after the crashed group stops exactly here: the transform
+				// throws before any step can reach the limit, so without this the halt would be attributed
+				// to the next member and the half-transformed ILAst the crash left behind - the one thing
+				// worth looking at when debugging a throwing transform - would be unreachable.
+				if (function != null && Stepper.CurrentStep == Stepper.StepLimit)
```
christophwille: **Bug.** `Stepper` here is the shared stepper, but when `RecordILTransformSteps` is false the IL phase recorded into the throwaway `context.Stepper`, so `Stepper.CurrentStep` is still 0 for the entire IL phase. With `StepLimit == 0` ("show state before" on the first tree node, or `lastSelectedStep` replayed by `OnWritingOptionsChanged` after the pane's flag flipped) any member whose IL transform throws satisfies `0 == 0`, is attributed as the halted function, `RunTransforms(AstNode)` bails, and the document becomes that unrelated member's half-transformed ILAst.

The same equality also holds when a member completes all its steps and then crashes in a transform that recorded nothing while the limit is aimed at the *next* group's opener: the crash is reported as the halt of the wrong index.

Minimal guard: `RecordILTransformSteps && function != null && ...`. Deeper: let the Stepper own the predicate (set `LimitReachedStep = LastStep` when `step == StepLimit` on crash) so both catch clauses use the same `HaltedStepFunction()` line and `CurrentStep` need not be public.

ILSpy/Languages/CSharpLanguage.DebugSteps.cs
```diff
@@ -32,26 +32,64 @@
...
+			handled = true;
+		}
+
+		/// <summary>
+		/// Points the decompiler's IL phase at the shared stepper, but only while the Debug Steps pane
+		/// is there to display what it records - see <see cref="DebugStepsPaneModel.IsRecording"/>.
+		/// </summary>
+		static partial void ConfigureStepRecording(CSharpDecompiler decompiler)
+		{
+			decompiler.RecordILTransformSteps = DebugStepsPaneModel.IsRecording;
```
christophwille: **Bug (numbering race).** `IsRecording` is sampled here inside `Task.Run` at `CreateDecompiler` time, not when the replay was requested. `OnCSharpDecompiled` publishes the stepper on any full run regardless of recording mode, so an AST-only tree is displayed and clickable while the pane is closed. Opening the pane flips `IsRecording` and requests a refresh, but nothing invalidates the old tree until `StepperUpdated` lands. A "show state before" click on AST step k in that window cancels the refresh (`DecompileAsync` -> `activeCts.Cancel()`) and replays k under IL-inclusive numbering: it halts on an unrelated IL step and renders ILAst. The same window is entered via the "Show Steps" button. The mirror case (pane closed in the window between click and task start) replays an IL+AST index against AST-only numbering.

The `RecordILTransformSteps` doc names the invariant ("requires the same value on the full run and on the step-limited re-run"); the code does not enforce it. Carrying the flag in `DecompilationOptions` next to `StepLimit`/`IsDebug`, set in `RestartDecompileWithStepLimit` from the mode the displayed tree was recorded in, makes it hold by construction.

ILSpy/Languages/CSharpLanguage.DebugSteps.cs
```diff
@@ -32,26 +32,64 @@
...
+		/// not the user's selection.
+		/// </summary>
+		static partial void TryWriteILAst(ITextOutput output, DecompilationOptions options, CSharpDecompiler decompiler, ref bool handled)
+		{
+			if (decompiler.StepLimitHaltedFunction is not { } function)
+				return;
+			if (output is AvaloniaEditTextOutput avaloniaOutput)
+			{
+				// The dump is IL, not C#; without this the editor would highlight it as C#.
+				avaloniaOutput.SyntaxExtensionOverride = ".il";
```
christophwille, line 66: **Bug (multi-node tabs).** `DecompileAsync` writes all `CurrentNodes` into one `AvaloniaEditTextOutput` with one `options.StepLimit`, and each node gets its own `CSharpDecompiler`/`Stepper`. Select method A (long IL phase) plus a field or short method B: A halts in its IL phase and sets `SyntaxExtensionOverride = ".il"` for the whole document; B halts in its AST phase (or never) and appends C# that is then highlighted as IL, with `DebugStepHighlighter` resolving against whichever node's ranges matched. The doc comment only considers members within one node. Simplest fix: restrict step replay to single-node tabs.

ILSpy/ViewModels/DebugStepsPaneModel.cs
```diff
@@ -385,6 +391,27 @@ void OnWritingOptionsChanged(object? sender, PropertyChangedEventArgs e)
...
+				return;
+			IsRecording = enabled;
+			if (enabled)
+			{
+				RequestRedecompile(int.MaxValue, isDebug: false);
+			}
+			else
+			{
+				SetStepsSource(null);
+				activeLanguage?.ReleaseSteps();
```
christophwille: **Bug (retention after close).** Two issues:

1. `activeLanguage` is null whenever the current language is not C# (`DetachFromLanguage`). Open the pane on C# with a large type, switch to IL, close the pane: `ReleaseSteps()` never runs and the MEF-shared `CSharpLanguage` keeps the full IL tree until the next C# full run. `DetachFromLanguage` does not release either. Resolve `CSharpLanguage` via `languageService.Languages.OfType<CSharpLanguage>()` (as the tests do).

2. Closing the pane does not cancel an in-flight run created with `RecordILTransformSteps = true`. When it finishes, `OnCSharpDecompiled` does `stepper = decompiler.Stepper` and raises `StepperUpdated`; the VM's `OnStepperUpdated` (still attached, language unchanged) re-pins the ~35 MB tree into `Steps` of a closed pane. `OnStepperUpdated` should drop updates while `!IsRecording`.

ILSpy/ViewModels/DebugStepsPaneModel.cs
```diff
@@ -385,6 +391,27 @@ void OnWritingOptionsChanged(object? sender, PropertyChangedEventArgs e)
...
+		/// tree the language is still holding, which is where the retained ILAst lives.
+		/// </summary>
+		internal void SetRecordingEnabled(bool enabled)
+		{
+			if (IsRecording == enabled)
+				return;
+			IsRecording = enabled;
+			if (enabled)
+			{
+				RequestRedecompile(int.MaxValue, isDebug: false);
```
christophwille: `RequestRedecompile` runs unconditionally, even when the current language is not C# (the only language that reads `IsRecording`) or the active tab shows a step-limited replay. Opening the pane on the IL language re-decompiles the tab for nothing and discards the current view. Guard on `activeLanguage != null`.

ILSpy/ViewModels/StepNodeViewModel.cs
```diff
@@ -39,9 +39,17 @@ public sealed partial class StepNodeViewModel : ObservableObject
...
 		public string Description => Step.Description;
 
+		IReadOnlyList<StepNodeViewModel>? children;
+
+		/// <summary>
+		/// Wrappers for this step's children, built on first access. A recorded type runs to tens of
+		/// thousands of steps, so materialising the whole tree up front would put that many view-models
+		/// on the UI thread for the handful of rows an expanded path actually shows.
+		/// </summary>
+		public IReadOnlyList<StepNodeViewModel> Children => children ??= Wrap(Step.Children, this);
```
christophwille, line 80: The laziness is defeated for the filter case: `ApplyFilterToNode`, `SnapshotExpansion` and `RestoreExpansion` in `DebugStepsPaneModel` recurse through `Children` for every root, so the first character typed into "Filter steps" (and every `SetStepsSource` while filtering) materialises all wrappers synchronously on the UI thread (85k for Enumerable), then re-walks them per keystroke. Either filter over the raw `Stepper.Node` tree and only touch `children` where it is already non-null, or drop the doc comment's claim.

ILSpy/Views/DebugSteps.axaml.cs
```diff
@@ -83,14 +83,20 @@ void AttachModel()
 			DetachModel();
 			attachedModel = model;
 			if (model != null)
+			{
 				model.SelectionRevealRequested += OnSelectionRevealRequested;
+				// The view is in the tree exactly while the pane is open, which is the only time the
+				// recorded IL steps are worth their memory.
+				model.SetRecordingEnabled(true);
```
christophwille, line 90: Recording is keyed to logical-tree attach/detach, not pane open/close. The Debug Steps view shares the bottom ToolDock with Search/Analyzer as tabs; switching to a sibling tab, floating, re-docking or a layout reset fires detach (tree nulled, `ReleaseSteps`) then attach (`RequestRedecompile(int.MaxValue)`): selection and expansion state are lost and the whole type re-decompiles with recording on. The base VM comment still says "it doesn't matter when the matching View materialises". Consider keying on the dockable's visibility/close instead, and skipping the redecompile when the language's current stepper already contains IL groups.

ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
```diff
@@ -2257,8 +2308,45 @@ internal static bool IsWindowsFormsInitializeComponentMethod(IMethod method)
...
+				if (!decompileMemberBodies && transform is AsyncAwaitDecompiler)
+					yield break;
+			}
+		}
+
 		void DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun decompileRun, ITypeResolveContext decompilationContext, ExtensionInfo? extensionInfo)
 		{
+			// An earlier member's IL phase already hit the step limit, so the pipeline is stopped for
+			// good: reading IL for every remaining member only to throw on its first step is wasted work.
+			if (StepLimitHaltedFunction != null)
```
christophwille, line 2351: Efficiency (bounded): this skips `ILReader` for later members, but `DoDecompile(IMethod/IProperty/IEvent)` and `DecompileTypeDefinition` still build every remaining member's `TypeSystemAstBuilder` declaration, attributes and accessors, and `Decompile()` assembles a `SyntaxTree` that `TryWriteILAst` discards. Every step selection in the pane pays this. A `break` in the member loops when `StepLimitHaltedFunction != null` cuts it. (Verified nothing downstream dereferences the missing body/annotation; body-less members are an already-supported state.)

ICSharpCode.Decompiler.Tests/DebugStepRecordingTests.cs
```diff
@@ -0,0 +1,311 @@
...
+		/// <summary>
+		/// Runs the decompiler once and returns the top-level group recording the named member's IL phase.
+		/// </summary>
+		static Stepper.Node MemberGroup(string methodName, CSharpDecompiler decompiler)
+		{
+			decompiler.DecompileTypeAsString(SampleType);
+			return decompiler.Stepper.Steps.Single(n => n.Description.EndsWith("." + methodName, StringComparison.Ordinal));
+		}
+
+		static IEnumerable<Stepper.Node> AllNodes(IEnumerable<Stepper.Node> nodes)
```
christophwille: Reuse: `AllNodes` here, `Descriptions` below it, and `AllDescriptions` in `ILSpy.Tests/Views/DebugStepsTests.cs` are three copies of the same pre-order walk. `ICSharpCode.Decompiler.Util.TreeTraversal.PreOrder(decompiler.Stepper.Steps, n => n.Children)` (already used by `ILSpy/Commands/ExtractPackageEntryContextMenuEntry.cs`) replaces all three; `.Select(n => n.Description)` covers the description variants. Likewise `StepperTesting.CreateDecompiler` is `new CSharpDecompiler("ICSharpCode.Decompiler.dll", new UniversalAssemblyResolver(null, false, null), new DecompilerSettings())` via the existing file-name ctor (only difference: `LoadInMemory`/prefetch, harmless here).

ILSpy/Languages/TypedILLanguage.cs
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) 2026 AlphaSierraPapa for the SharpDevelop Team
```
christophwille: Conventions (CLAUDE.md, "File headers"): this is a new file per git (`new file mode`), and the rule says line 1 carries "the name of the human contributing the change ... never AlphaSierraPapa". The content was moved out of `ILAstLanguage.cs`, which carried this header, so keeping the original holder is defensible, but the PR's other new file (`StepperTesting.cs`) uses `Siegfried Pammer`; pick one convention for the PR.

christophwille (commented):
Follow-up: the simplification and CLAUDE.md-conventions angles that were missing from the review above are now complete. Five additional low-severity items inline (no new correctness bugs): the halted-function fallback chain is mostly derivable from Stepper state, the `ref bool handled` partial can be a `partial bool`, `EndOpenGroups`'s unused default, duplicated boot sequence in DebugStepsTests, and two conventions nits (Assert.Ignore bypass under Release, one en-GB spelling).

ICSharpCode.Decompiler/CSharp/CSharpDecompiler.cs
```diff
@@ -2257,8 +2308,45 @@ internal static bool IsWindowsFormsInitializeComponentMethod(IMethod method)
...
+				return;
+			// Declared out here so the StepLimitReachedException handler below can report the function
+			// its transforms were halted in, and tell a halt inside this member from one on its boundary.
+			ILFunction? function = null;
+			bool haltBelongsToThisMember = false;
```
christophwille: Simplification: `haltBelongsToThisMember` + `lastCompletedFunction` (reset in `CreateDecompileRun`) exist to reconstruct what the Stepper already knows. When the limit hits a member's opener, `Stepper.LastStep` is the previous member's last recorded step, so `FunctionOf(Stepper.LimitReachedStep) ?? FunctionOf(Stepper.LastStep) ?? function` covers the boundary case, the detached-function case and the first-member case, and drops the flag, the field and its reset. One caveat to keep in mind: with `DecompileMemberBodies == false` there is no seam step, so `LastStep` is a transform group opener whose `Position` is null; the fallback then lands on `function` (the untouched next member) instead of the previous one. If that path matters, give group openers in `ILFunction.RunTransforms` a `near: this` so their Position resolves.

ILSpy/Languages/CSharpLanguage.cs
```diff
@@ -732,38 +736,55 @@ public void Run(AstNode rootNode, TransformContext context)
...
-			// single formatting pass (see Bookmarks.BookmarkDebugInfoCollector). The collector is the
-			// outermost writer so its StartNode sees each node's start line before any token is written.
-			// Other outputs (IL view, ilspycmd's plain text) are not AvaloniaEditTextOutput and are unaffected.
-			Bookmarks.BookmarkDebugInfoCollector? bookmarkCollector = null;
-			if (output is TextView.AvaloniaEditTextOutput bookmarkOutput)
-				tokenWriter = bookmarkCollector = new Bookmarks.BookmarkDebugInfoCollector(tokenWriter, bookmarkOutput);
-
-			syntaxTree.AcceptVisitor(new CSharpOutputVisitor(tokenWriter, settings.CSharpFormattingOptions));
-			bookmarkCollector?.Publish();
+			bool handled = false;
```
christophwille: Simplification: the `ref bool handled` out-channel is only needed for `partial void`. Extended partial methods (C# 9+) may return a value when declared with an access modifier: `private static partial bool TryWriteILAst(ITextOutput, DecompilationOptions, CSharpDecompiler);` with a Release stub `=> false`, then `if (!TryWriteILAst(...)) { ...C# path... }`. Or drop the partial entirely: `StepLimitHaltedFunction` is public and always null in Release, so `if (decompiler.StepLimitHaltedFunction is { } fn) WriteILAst(...) else WriteCSharp(...)` reads the same in both configurations and only `WriteILAst` needs `#if DEBUG`.

ICSharpCode.Decompiler/DebugSteps/Stepper.cs
```diff
@@ -256,6 +263,29 @@ public Node StartGroup(string description, DebugStepNodeInfo? modifiedNode = nul
...
+		/// that started them. Without this, everything recorded afterwards is filed as a child of the
+		/// group that failed, and that group's <see cref="Node.EndStep"/> still points at its own start,
+		/// so replaying "the state after this step" lands on the wrong step.
+		/// The groups are kept even when empty: a group that recorded nothing before it was abandoned is
+		/// precisely the one worth seeing, and <see cref="EndGroup"/>'s removal path expects the group to
+		/// still be the last entry of its parent, which an unwind cannot guarantee.
+		/// Closing stops at <paramref name="targetDepth"/>: a group that was already open before the
+		/// unwinding code ran belongs to whoever opened it, not to the unwind.
+		/// </summary>
+		public void EndOpenGroups(int targetDepth = 0)
```
christophwille: The `targetDepth = 0` default has no caller (both sites pass `groupDepth`) and is the one value that closes groups the caller does not own - the case `AFailingTransformLeavesAGroupOpenedAroundItAlone` guards against. Make it required.

ILSpy.Tests/Views/DebugStepsTests.cs
```diff
@@ -186,25 +184,80 @@ await Waiters.WaitForAsync(
...
+		{
+			await CoverILTransformsAndReplay();
+		}
+		finally
+		{
+			debugStepsVm.SetRecordingEnabled(false);
+		}
+	}
+
+	static async Task CoverILTransformsAndReplay()
```
christophwille, line 364: `CoverILTransformsAndReplay` is a single-use extraction so the caller can wrap it in try/finally; inlining the body into the test (with the same try/finally) is shorter. The boot sequence (GetExport<MainWindow>, Show, WaitForAssembliesAsync, LanguageService, find the Enumerable node, WaitForDecompiledTextAsync) now appears four times in this file; one shared `BootAndSelectEnumerableAsync()` helper would cover all four.

ICSharpCode.Decompiler.Tests/DebugStepRecordingTests.cs
```diff
@@ -0,0 +1,311 @@
...
+
+		/// <summary>
+		/// A member whose transform throws must not swallow the members after it: their steps belong
+		/// beside it in the tree, not inside the group it abandoned.
+		/// </summary>
+		[Test]
+		public void AFailingTransformDoesNotNestLaterMembersUnderIt()
+		{
+			if (!Stepper.SteppingAvailable)
+				Assert.Ignore("Transform stepping is compiled out without the STEP symbol, so there are no groups to check.");
```
christophwille, line 136: Conventions (CLAUDE.md, "No silent returns in tests. When an expected component is missing, assert and fail"): `Assert.Ignore` when `!Stepper.SteppingAvailable` (here and at lines 158, 184, 203) makes a Release-configuration run report green with zero step-recording coverage. The repo does use `Assert.Ignore` for the missing ILSpy-tests submodule, so this may be an accepted exception; if so, a single `[SetUp]` guard (or `Assume.That`) beats repeating it in four tests. Also en-US: `ILSpy.Tests/Views/DebugStepsTests.cs:237` has "realises" in a new comment ("realizes").

PR #4034, opened 2026-08-19
Overlap the sequential waits inside the decompiler test infrastructure

christophwille:
The "Round 2: sequential awaits inside the test infra" commits of #3940, split out on their own. They do not depend on #3940's NUnit worker-count / fixture-parallelism change (which is not being pursued): `RoundtripAssembly` and the matrix runners are already `Parallelizable(ParallelScope.All)` on master, and the overlaps below are per-test or per-process, independent of how many workers NUnit runs. All seven apply to master without conflicts and are cherry-picked verbatim; only `ICSharpCode.Decompiler.Tests` is touched (`Helpers/Tester.cs`, `Helpers/RoslynToolset.cs`, `CorrectnessTestRunner.cs`, `RoundtripAssembly.cs`).

| commit | what | why |
|---|---|---|
| Synchronize toolset registration dictionaries | `lock` around the two `Dictionary.Add` in `RoslynToolset`/`RefAssembliesToolset.Fetch`; lookups stay lock-free (they only happen after `Initialize` completed) | prerequisite of the next one |
| Overlap toolset downloads and TestRunner builds in `Tester.Initialize` | the nine NuGet fetches and the TestRunner builds start eagerly and are awaited in one `Task.WhenAll` (every failure surfaced, not just the first); only the two Windows RID TestRunner builds stay sequential with each other (shared `obj/`, implicit restores would race on `project.assets.json`) | `Initialize` gates every test in the suite; on a cold machine / CI it serialized all downloads and builds |
| Cache the vswhere-based MSBuild lookup | `Lazy<Task<string>>` (`ExecutionAndPublication`); the non-Windows `Assert.Ignore` stays outside the cache so it is raised per test, not memoized as a faulted task | the parallel roundtrip fixture spawned one `vswhere.exe` per test for a process-invariant answer |
| Run original and decompiled executables concurrently | `Tester.StartRun` returns the in-flight run (fault pre-observed, so an abandoned run after an upstream failure cannot raise `UnobservedTaskException`); `RunAndCompareOutput` gets an overload taking that task and awaits both with a plain `WhenAll` (keeps NUnit Ignore semantics); exit codes are still asserted in the original order, so failure output is unchanged | the two runs are independent processes with separately buffered output |
| Start the original executable before decompiling in correctness tests | `RunCS`/`RunVB`/`RunIL` start the original exe right after the first compile; for mcs configs the `.exe.config` write moves ahead of the run start (the runtime reads it at launch) while the compiler-option mutation stays after the decompile, which must see the original options | the decompile/recompile stages only read the original binary |
| Make the roundtrip `testAction` asynchronous | `Func<string, Task>` instead of `GetAwaiter().GetResult()` on NUnit worker threads | cleanup; enables the next one |
| Overlap the pristine-executable run with the roundtrip pipeline | `RunWithOutput` starts the reference exe from the `ILSpy-tests` checkout before the whole-project decompile + MSBuild rebuild (nothing in that pipeline writes to the input dir); the submodule-missing guard is hoisted into `EnsureTestDirAvailable()` and runs before the early start, so those tests still report Ignored | the pristine run used to wait for the multi-minute decompile it does not depend on |

## Verification

Local, 24-thread Windows 11 box, Debug, `ILSpy-tests` checked out (roundtrips and the mcs matrix run: 140 mcs cases passed, which exercises the `.exe.config`-before-run reordering), workstation GC (this branch is off master; #4033 is separate):

| run | result | wall | sum of per-test durations (TRX) |
|---|---|---|---|
| master, warm (same day, same box) | 4254 / 0 / 20 | 553 s | 11,532 s |
| this branch, **cold** (`bin/.../{roslyn,vswhere,netfx}` deleted, exercises the concurrent `Initialize`) | 4254 / 0 / 20 | 554 s | 10,819 s |
| this branch, warm | 4254 / 0 / 20 | 512 s | 10,663 s |

The 20 skips are the usual environment-gated ones. The wall-time difference is within a few percent and is not the point; the per-test overlap shows up as ~7% less summed test time for the same work. The local "cold" `Initialize` is served from `ILSpy-tests/nuget` and the already-built TestRunner, so it is only a correctness check of the concurrent path here (4 s); CI is where the downloads actually happen.

### CI

Run https://github.com/icsharpcode/ILSpy/actions/runs/32240841471: all four jobs green (Windows Debug/Release, Linux, macOS).

| decompiler test host | master (run 32177067040) | this branch |
|---|---|---|
| Windows Debug | 1393 s | 1373 s |
| Windows Release | 664 s | 766 s |
| Linux (own step) | 6m19 | 6m25 |

No measurable CI wall-time effect either way: the Windows numbers sit inside the run-to-run spread of that shared 4-core runner (master's own Release test step has ranged 11m10-15m26 across recent runs), and on Linux the suite is CPU-bound with most of its matrix skipped. The value of these commits is structural (no blocking waits on worker threads, no per-test `vswhere.exe`, setup downloads and builds overlapped, original/decompiled runs concurrent), not a headline number.

Not taken from #3940: the NUnit `LevelOfParallelism`/`Parallelizable(Fixtures)`/`Order` commit and the Defender doc (dropped), the block-invariant change (reverted there), server GC (#4033), the fixture-host hardening (#4032), the concurrent .NET Framework scan and the diagnostics sampler (reverted there).

### DIFF
--- a/ICSharpCode.Decompiler.Tests/CorrectnessTestRunner.cs
+++ b/ICSharpCode.Decompiler.Tests/CorrectnessTestRunner.cs
@@ -436,10 +436,30 @@ namespace ICSharpCode.Decompiler.Tests
 			string testOutputFileName = TestsAssemblyOutput.GetFilePath(TestCasePath, testName, Tester.GetSuffix(options) + ".exe");
 			Helpers.CompilerResults outputFile = null, decompiledOutputFile = null;
 
+			// The mcs mutation below never touches these flags, so they can be captured here
+			// and used for the original run started before the mutation happens.
+			bool useTestRunner = (options & CompilerOptions.UseTestRunner) != 0;
+			bool force32Bit = (options & CompilerOptions.Force32Bit) != 0;
+
 			try
 			{
 				outputFile = await Tester.CompileCSharp(Path.Combine(TestCasePath, testFileName), options,
 					outputFileName: testOutputFileName).ConfigureAwait(false);
+				if ((options & CompilerOptions.UseMcsMask) != 0)
+				{
+					// Add an .exe.config so that we consistently use the .NET 4.x runtime.
+					// Written before the original executable starts below, because the runtime
+					// reads it at process start.
+					File.WriteAllText(outputFile.PathToAssembly + ".config", @"<?xml version=""1.0"" encoding=""utf-8""?>
+<configuration>
+	<startup>
+		<supportedRuntime version=""v4.0"" sku="".NETFramework,Version=v4.0,Profile=Client"" />
+	</startup>
+</configuration>");
+				}
+				// The original executable is complete at this point; its run overlaps the
+				// decompile and recompile of the same assembly, which only read it.
+				var originalRun = Tester.StartRun(outputFile.PathToAssembly, useTestRunner, force32Bit);
 				string decompiledCodeFile = await Tester.DecompileCSharp(outputFile.PathToAssembly, Tester.GetSettings(options)).ConfigureAwait(false);
 				if ((options & CompilerOptions.UseMcsMask) != 0)
 				{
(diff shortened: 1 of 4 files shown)


christophwille:
I am closing this as there are no wall-clock gains on CI (and parallel downloading of toolsets would need resilience as well)

PR #4036, opened 2026-08-19
Restore keyboard activation for selected tree items

### LINKED ISSUE #4030: "Enter" key now does nothing in Analyzer pane
### Steps to reproduce
1. Analyze something.
2. Select a node in the Analyzer pane that represents a member.
3. Press the "Enter" key.

### Expected behavior
The member becomes selected in the Assemblies pane and is decompiled.  (This was the behavior in ILSpy 10.1.1.8388.)

### Actual behavior
Nothing happens.

### Details
* Product in use: ILSpy
* Version in use: 11.0.0.9375

SychicBoy:
Fixes #4030

### Problem
In ILSpy 10.x, Enter and Space activated the selected item in `SharpTreeView`.
After the Avalonia migration, `ListBoxItem` handles these keys as selection input before they reach `SharpTreeView`, so the existing activation logic is not invoked.

### Solution
* Override `ShouldTriggerSelection` so Enter and Space can be handled by `SharpTreeView` instead of being intercepted by `ListBoxItem` selection handling.

### DIFF
--- a/ILSpy/Controls/TreeView/SharpTreeView.cs
+++ b/ILSpy/Controls/TreeView/SharpTreeView.cs
@@ -295,6 +295,23 @@ namespace ICSharpCode.ILSpy.Controls.TreeView
 			scrollViewer.Offset = new Vector(scrollViewer.Offset.X, newOffsetY);
 		}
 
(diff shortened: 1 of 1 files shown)


christophwille:
Superseded by #4039

PR #4037, opened 2026-08-19
Fix derived type filtering

### LINKED ISSUE #4028: Derived types do not shown in tree.
### Steps to reproduce
1. Open ILSpy
2. Expand "Derived Types" node

Image

### Details
* Product in use: ILSpy
* ILSpy version: 11.0.0.9375+cdeae656fee6e184ff76cdc549edc7f9031dba09
* .NET version: 10.0.11-servicing.26373.116+e2f47b0110ed922f21a1522da67279133ce28f32

SychicBoy:
Fixes #4028

### Problem
`DerivedTypesEntryNode.Filter()` returns `Recurse` for visible entries, causing the filter cascade to recursively filter their lazy children. This can hide derived-type entries and modify the child collection during visibility updates.

### Solution
* Return `Match` from `DerivedTypesEntryNode.Filter()` for visible entries so their children remain lazy until the entry is expanded.

### DIFF
--- a/ILSpy/TreeNodes/DerivedTypesEntryNode.cs
+++ b/ILSpy/TreeNodes/DerivedTypesEntryNode.cs
@@ -69,8 +69,8 @@ namespace ICSharpCode.ILSpy.TreeNodes
 		};
 
 		/// <summary>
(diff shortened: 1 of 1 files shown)


christophwille:
Superseded by #4039

PR #4043, opened 2026-08-23
Fix #3282: leave out trailing optional arguments in indexer accesses

### LINKED ISSUE #3060: Assert at ICSharpCode.Decompiler.CSharp.CallBuilder.ArgumentList.CheckNoNamedOrOptionalArguments() in CallBuilder.cs:line 174
### Input code
Multiple similar asserts when browsing assembly, see example below
@060001AF
and others

### Erroneous output
```c#
Error decompiling @060001AF  -----.Enumerator.MoveNext
in assembly "c:\tmp\err131.dll"
 ---> ICSharpCode.ILSpy.AssertionFailedException: 
   at ICSharpCode.ILSpy.ILSpyTraceListener.Fail(String message, String detailMessage) in offset 379
   at System.Diagnostics.TraceInternal.Fail(String message, String detailMessage) in offset 48
   at System.Diagnostics.TraceInternal.TraceProvider.Fail(String message, String detailMessage) in offset 0
   at System.Diagnostics.Debug.Fail(String message, String detailMessage) in offset 0
   at ICSharpCode.Decompiler.CSharp.CallBuilder.ArgumentList.CheckNoNamedOrOptionalArguments() in CallBuilder.cs:line 174
   at ICSharpCode.Decompiler.CSharp.CallBuilder.Build(OpCode callOpCode, IMethod method, IReadOnlyList`1 callArguments, IReadOnlyList`1 argumentToParameterMap, IType constrainedTo) in CallBuilder.cs:line 406
   at ICSharpCode.Decompiler.CSharp.CallBuilder.Build(CallInstruction inst) in CallBuilder.cs:line 217
   at ICSharpCode.Decompiler.CSharp.ExpressionBuilder.VisitCallVirt(CallVirt inst, TranslationContext context) in ExpressionBuilder.cs:line 2328
   at ICSharpCode.Decompiler.IL.CallVirt.AcceptVisitor[C,T](ILVisitor`2 visitor, C context) in Instructions.cs:line 2196
   at ICSharpCode.Decompiler.CSharp.ExpressionBuilder.Translate(ILInstruction inst, IType typeHint) in ExpressionBuilder.cs:line 164
   at ICSharpCode.Decompiler.CSharp.CallBuilder.BuildArgumentList(ExpectedTargetDetails expectedTargetDetails, ResolveResult target, IMethod method, Int32 firstParamIndex, IReadOnlyList`1 callArguments, IReadOnlyList`1 argumentToParameterMap) in CallBuilder.cs:line 793
   at ICSharpCode.Decompiler.CSharp.CallBuilder.Build(OpCode callOpCode, IMethod method, IReadOnlyList`1 callArguments, IReadOnlyList`1 argumentToParameterMap, IType constrainedTo) in CallBuilder.cs:line 289
   at ICSharpCode.Decompiler.CSharp.CallBuilder.Build(CallInstruction inst) in CallBuilder.cs:line 217
   at ICSharpCode.Decompiler.CSharp.ExpressionBuilder.VisitCall(Call inst, TranslationContext context) in ExpressionBuilder.cs:line 2323
   at ICSharpCode.Decompiler.IL.Call.AcceptVisitor[C,T](ILVisitor`2 visitor, C context) in Instructions.cs:line 2173
   at ICSharpCode.Decompiler.CSharp.ExpressionBuilder.Translate(ILInstruction inst, IType typeHint) in ExpressionBuilder.cs:line 164
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.Default(ILInstruction inst) in StatementBuilder.cs:line 89
   at ICSharpCode.Decompiler.IL.ILVisitor`1.VisitCall(Call inst) in Instructions.cs:line 7338
   at ICSharpCode.Decompiler.IL.Call.AcceptVisitor[T](ILVisitor`1 visitor) in Instructions.cs:line 2169
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.Convert(ILInstruction inst) in StatementBuilder.cs:line 78
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.VisitBlock(Block block) in StatementBuilder.cs:line 1207
   at ICSharpCode.Decompiler.IL.Block.AcceptVisitor[T](ILVisitor`1 visitor) in Instructions.cs:line 875
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.Convert(ILInstruction inst) in StatementBuilder.cs:line 78
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.VisitIfInstruction(IfInstruction inst) in StatementBuilder.cs:line 147
   at ICSharpCode.Decompiler.IL.IfInstruction.AcceptVisitor[T](ILVisitor`1 visitor) in Instructions.cs:line 1427
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.Convert(ILInstruction inst) in StatementBuilder.cs:line 78
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertBlockContainer(BlockStatement blockStatement, BlockContainer container, IEnumerable`1 blocks, Boolean isLoop) in StatementBuilder.cs:line 1418
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertBlockContainer(BlockContainer container, Boolean isLoop) in StatementBuilder.cs:line 1348
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.VisitBlockContainer(BlockContainer container) in StatementBuilder.cs:line 1234
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.Convert(ILInstruction inst) in StatementBuilder.cs:line 78
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertAsBlock(ILInstruction inst) in StatementBuilder.cs:line 83
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun decompileRun, ITypeResolveContext decompilationContext) in CSharpDecompiler.cs:line 1741
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun decompileRun, ITypeResolveContext decompilationContext) in CSharpDecompiler.cs:line 1756
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DoDecompile(IMethod method, DecompileRun decompileRun, ITypeResolveContext decompilationContext) in CSharpDecompiler.cs:line 1605
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.<>c__DisplayClass74_0.<DoDecompile>g__DoDecompileMember|0(IEntity entity, RecordDecompiler recordDecompiler, PartialTypeInfo partialType) in CSharpDecompiler.cs:line 1480
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DoDecompile(ITypeDefinition typeDef, DecompileRun decompileRun, ITypeResolveContext decompilationContext) in CSharpDecompiler.cs:line 1369
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.<>c__DisplayClass74_0.<DoDecompile>g__DoDecompileMember|0(IEntity entity, RecordDecompiler recordDecompiler, PartialTypeInfo partialType) in CSharpDecompiler.cs:line 1492
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DoDecompile(ITypeDefinition typeDef, DecompileRun decompileRun, ITypeResolveContext decompilationContext) in CSharpDecompiler.cs:line 1369
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.Decompile(IEnumerable`1 definitions) in CSharpDecompiler.cs:line 1034
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.Decompile(EntityHandle[] definitions) in CSharpDecompiler.cs:line 1006
   at ICSharpCode.ILSpy.CSharpLanguage.DecompileType(ITypeDefinition type, ITextOutput output, DecompilationOptions options) in offset 93
   at ICSharpCode.ILSpy.TreeNodes.TypeTreeNode.Decompile(Language language, ITextOutput output, DecompilationOptions options) in offset 8
   at ICSharpCode.ILSpy.TextView.DecompilerTextView.DecompileNodes(DecompilationContext context, ITextOutput textOutput) in offset 122
   at ICSharpCode.ILSpy.TextView.DecompilerTextView.<>c__DisplayClass53_0.<DecompileAsync>b__0() in offset 20
```
If the output has the wrong behavior, explain how it differs from the expected behavior.

### Details
ILSpy version 8.1.0.7481
.NET version 6.0.20-servicing.23320.17+a08d9ce2caf02455c0b825bcdc32974bdf769a80
[err131.zip](https://github.com/icsharpcode/ILSpy/files/12366039/err131.zip)

### LINKED ISSUE #3282: Indexer invocations should allow optional arguments
### Input code
Save project

### Erroneous output
```c#
 ---> ICSharpCode.ILSpy.AssertionFailedException: 
   at ICSharpCode.ILSpy.ILSpyTraceListener.Fail(String message, String detailMessage) in offset 379
   at System.Diagnostics.TraceInternal.Fail(String message, String detailMessage) in offset 48
   at System.Diagnostics.Debug.Fail(String message, String detailMessage) in offset 0
   at ICSharpCode.Decompiler.CSharp.CallBuilder.ArgumentList.CheckNoNamedOrOptionalArguments() in CallBuilder.cs:line 174
   at ICSharpCode.Decompiler.CSharp.CallBuilder.Build(OpCode callOpCode, IMethod method, IReadOnlyList`1 callArguments, IReadOnlyList`1 argumentToParameterMap, IType constrainedTo) in CallBuilder.cs:line 441
   at ICSharpCode.Decompiler.CSharp.CallBuilder.Build(CallInstruction inst, IType typeHint) in CallBuilder.cs:line 230
   at ICSharpCode.Decompiler.CSharp.ExpressionBuilder.VisitCallVirt(CallVirt inst, TranslationContext context) in ExpressionBuilder.cs:line 2350
   at ICSharpCode.Decompiler.IL.CallVirt.AcceptVisitor[C,T](ILVisitor`2 visitor, C context) in Instructions.cs:line 2196
   at ICSharpCode.Decompiler.CSharp.ExpressionBuilder.VisitConv(Conv inst, TranslationContext context) in ExpressionBuilder.cs:line 2122
   at ICSharpCode.Decompiler.IL.Conv.AcceptVisitor[C,T](ILVisitor`2 visitor, C context) in Instructions.cs:line 2278
   at ICSharpCode.Decompiler.CSharp.ExpressionBuilder.VisitStLoc(StLoc inst, TranslationContext context) in ExpressionBuilder.cs:line 752
   at ICSharpCode.Decompiler.IL.StLoc.AcceptVisitor[C,T](ILVisitor`2 visitor, C context) in Instructions.cs:line 2567
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.VisitStLoc(StLoc inst) in StatementBuilder.cs:line 116
   at ICSharpCode.Decompiler.IL.StLoc.AcceptVisitor[T](ILVisitor`1 visitor) in Instructions.cs:line 2563
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.Convert(ILInstruction inst) in StatementBuilder.cs:line 80
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.VisitBlock(Block block) in StatementBuilder.cs:line 1229
   at ICSharpCode.Decompiler.IL.Block.AcceptVisitor[T](ILVisitor`1 visitor) in Instructions.cs:line 875
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.Convert(ILInstruction inst) in StatementBuilder.cs:line 80
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertAsBlock(ILInstruction inst) in StatementBuilder.cs:line 85
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertLoop(BlockContainer container) in StatementBuilder.cs:line 1289
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.VisitBlockContainer(BlockContainer container) in StatementBuilder.cs:line 1243
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.Convert(ILInstruction inst) in StatementBuilder.cs:line 80
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertBlockContainer(BlockStatement blockStatement, BlockContainer container, IEnumerable`1 blocks, Boolean isLoop) in StatementBuilder.cs:line 1466
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertBlockContainer(BlockContainer container, Boolean isLoop) in StatementBuilder.cs:line 1370
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.VisitBlockContainer(BlockContainer container) in StatementBuilder.cs:line 1256
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.Convert(ILInstruction inst) in StatementBuilder.cs:line 80
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertAsBlock(ILInstruction inst) in StatementBuilder.cs:line 85
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.TransformToForeach(UsingInstruction inst, Expression resource) in StatementBuilder.cs:line 780
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.VisitUsingInstruction(UsingInstruction inst) in StatementBuilder.cs:line 552
   at ICSharpCode.Decompiler.IL.UsingInstruction.AcceptVisitor[T](ILVisitor`1 visitor) in Instructions.cs:line 2072
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.Convert(ILInstruction inst) in StatementBuilder.cs:line 80
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertBlockContainer(BlockStatement blockStatement, BlockContainer container, IEnumerable`1 blocks, Boolean isLoop) in StatementBuilder.cs:line 1466
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertBlockContainer(BlockContainer container, Boolean isLoop) in StatementBuilder.cs:line 1370
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.VisitBlockContainer(BlockContainer container) in StatementBuilder.cs:line 1256
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.Convert(ILInstruction inst) in StatementBuilder.cs:line 80
   at ICSharpCode.Decompiler.CSharp.StatementBuilder.ConvertAsBlock(ILInstruction inst) in StatementBuilder.cs:line 85
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun decompileRun, ITypeResolveContext decompilationContext) in CSharpDecompiler.cs:line 1780
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DecompileBody(IMethod method, EntityDeclaration entityDecl, DecompileRun decompileRun, ITypeResolveContext decompilationContext) in CSharpDecompiler.cs:line 1795
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.DoDecompile(IProperty property, DecompileRun decompileRun, ITypeResolveContext decompilationContext) in CSharpDecompiler.cs:line 2070
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.Decompile(IEnumerable`1 definitions) in CSharpDecompiler.cs:line 1067
   at ICSharpCode.Decompiler.CSharp.CSharpDecompiler.Decompile(EntityHandle[] definitions) in CSharpDecompiler.cs:line 1010
   at ICSharpCode.ILSpy.CSharpLanguage.DecompileProperty(IProperty property, ITextOutput output, DecompilationOptions options) in offset 98
   at ICSharpCode.ILSpy.TreeNodes.PropertyTreeNode.Decompile(Language language, ITextOutput output, DecompilationOptions options) in offset 8
   at ICSharpCode.ILSpy.TextView.DecompilerTextView.DecompileNodes(DecompilationContext context, ITextOutput textOutput) in offset 122
   at ICSharpCode.ILSpy.TextView.DecompilerTextView.<>c__DisplayClass53_0.<DecompileAsync>b__0() in offset 20

```

### Details
* ILSpy version 9.0.0.7777-preview2+bf2cf936e143384221d3277ac2a2960c7e2f0005
* .NET version 8.0.8+08338fcaa5c9b9a8190abb99222fed12aaba956c

[err131.zip](https://github.com/user-attachments/files/17083342/err131.zip)

siegfriedpammer:
> _Written by Claude (agent), posted on behalf of @siegfriedpammer._

Fixes #3282.
Fixes #3060 - the same assert reached from a different assembly. Verified against the reporter's `err131.dll`: `ilspycmd err131.dll -m 0x060001AF` hits `CheckNoNamedOrOptionalArguments` on master and decompiles to `list[num]` on this branch, where master's Release build wrote `list[num, false]`.

`HandleAccessorCall` had no way to express an omitted argument, so `CallBuilder` asserted that no optional argument had been detected before it got there. Any assembly that indexes through an indexer with an optional parameter hits that assert in a Debug build; Release builds silently wrote the defaults back out.

**Let an indexer access leave out trailing optional arguments**

Accessor calls now go through the same `ArgumentList` helpers as an ordinary call. Two things had to reach them: the assigned value of a setter is the last argument of the call but not an argument of the access - the standard adds it only for the invocation of the accessor (§12.6.2.1) - so it neither ends the run of optional arguments nor is written out with them; and the argument names have to stop wherever the argument list does.

Whether the shortened access still binds to the same member is left to `IsUnambiguousAccess`. If it does not, the omitted arguments are written out again before any cast is tried, since restoring them cannot change what the access means. A type declaring both `this[int]` and `this[int, int = 10]` therefore keeps both arguments; the fixture pins that.

**Write named arguments for indexer accesses**

C# allows named arguments in an element access, but `NamedArgumentTransform` refused to introduce one for any accessor, so an access whose arguments the compiler reordered came out as a temporary variable assigned before the access. Only indexers gain this: a property access has no argument list, an operator cannot take names either, and a setter's last argument stays unnamed on the right-hand side.

**Tests**

Indexer cases in `OptionalArguments` (get, set, compound assignment, increment, struct receiver, object initializer, and the overload that must keep its argument), in `OptionalArgumentsDisabled` (with the setting off the arguments stay explicit), and in `NamedArguments`.

**Not covered**

Still written out explicitly, each for its own reason: `params` indexers, `[Optional]` without a constant and `[DateTimeConstant]`-style defaults (nothing in the signature to compare against), `default(T)` at a value-type instantiation, and omitting a *middle* optional argument - the last of which plain calls do not do either (`M(0, z: 9)` decompiles to `M(0, 1, 9)`).

**Interaction with #3972**

Checked: all three commits of #3972 cherry-pick onto this branch without conflict, the combined tree builds, and its full decompiler suite is green (3467 tests, 0 failures), including `c[1] += 5` on an indexer, which goes through both changes.


ICSharpCode.Decompiler/IL/Transforms/NamedArgumentTransform.cs
```diff
@@ -33,8 +33,19 @@ internal static FindResult CanIntroduceNamedArgument(CallInstruction call, ILIns
...
 				return FindResult.Stop; // cannot use named arg to move expressionBeingMoved before this pointer
-			if (call.Method.IsOperator || call.Method.IsAccessor)
-				return FindResult.Stop; // cannot use named arg for operators or accessors
+			if (call.Method.IsOperator)
+				return FindResult.Stop; // cannot use named arg for operators
+			bool isIndexerSetter = false;
+			if (call.Method.IsAccessor)
+			{
+				// Only an indexer access has an argument list that can carry names.
+				if (call.Method.AccessorOwner is not IProperty { IsIndexer: true })
```
christophwille: **Bug (default settings): indexer setter inside a `CallInlineAssign` block gets replaced by a `CallWithNamedArgs` block.**

```csharp
Check(this[y: Get(1), x: Get(2)] = Get(3));
// or
return this[y: Get(1), x: Get(2)] = this[y: Get(3), x: Get(4)];
```

`TransformAssignment` builds `Block CallInlineAssign { call set_Item(this, ldloc t2, ldloc t1, stloc i(Get(3))); final: ldloc i }`. Inlining `t1` stops at `Get(2)`, `CanIntroduceNamedArgument` (indexer setter, index 1 != 3, load at index 2) returns `NamedArgument`, `NonAggressiveInlineInto` passes (`t1` was stored with `this` on the IL stack), and `IntroduceNamedArgument` does `call.ReplaceWith(namedArgBlock)` *inside* the `CallInlineAssign` block, breaking `MatchInlineAssignBlock`. Reproduced: Debug build terminates with `Assertion failed` (Block.cs:120); Release emits `Error: MatchInlineAssignBlock() returned false`. Master emits `int y = Get(1); int i = (this[Get(2), y] = Get(3));`.

**Same root cause, needs `AggressiveInlining` (or any aggressive context like a catch-when filter):** `this[y: Get(1), x: Get(2)] += 5;` -- the indexer *getter* is the `Target` of a `CompoundAssignmentInstruction`, which `CheckValidTarget` requires to be a `Call`/`CallVirt`; it now becomes a `CallWithNamedArgs` block -> assert at CompoundAssignmentInstruction.cs:97. The non-aggressive default is only saved by the `ILStackWasEmpty` heuristic.

Suggest returning `FindResult.Stop` here when `call.Parent` is a `CallInlineAssign` block or a `CompoundAssignmentInstruction` target (i.e. only allow it where `IntroduceNamedArgument` may legally replace the call by a block).

ICSharpCode.Decompiler/IL/Transforms/NamedArgumentTransform.cs
```diff
@@ -45,7 +56,8 @@ internal static FindResult CanIntroduceNamedArgument(CallInstruction call, ILIns
 			}
 			if (call.Method.Parameters.Any(p => string.IsNullOrEmpty(p.Name)))
 				return FindResult.Stop; // cannot use named arguments
-			for (int i = child.ChildIndex; i < call.Arguments.Count; i++)
+			int lastNameableArgument = isIndexerSetter ? call.Arguments.Count - 1 : call.Arguments.Count;
```
christophwille: **Bug (`AggressiveInlining`): `CanExtendNamedArgument` (unchanged, below) has no equivalent value guard.** It still hands out any `ldloc v` call argument as a named argument, including an indexer setter's value. `CallWithNamedArgs` then reorders the arguments so the value is no longer last, but `BuildArgumentList` (`isSetter && i + 1 == callArguments.Count`) and `HandleAccessorCall` (`Arguments[Length - 1]`) identify the value by position.

```csharp
int s = Get(0);
this[y: Get(1), x: Get(2)] = s;
```

After the index temp is named, inlining `s` hits `FindLoadInNext(Block CallWithNamedArgs)` -> `CanExtendNamedArgument` -> the foreach loop returns `NamedArgument` for the value. PR build emits `this[value: Get(0), y: Get(1)] = Get(2);` (wrong and uncompilable). Non-aggressive mode is only saved by `NonAggressiveInlineInto`'s `ILStackWasEmpty` heuristic.

Fix at the root: identify the setter value by parameter index (`argumentToParameterMap[i] == Parameters.Count - 1`) in `CallBuilder`, and/or add the same guard to `CanExtendNamedArgument`.

ICSharpCode.Decompiler/CSharp/CallBuilder.cs
```diff
@@ -1845,6 +1871,9 @@ ExpressionWithResolveResult HandleAccessorCall(ExpectedTargetDetails expectedTar
 				}
 			}
 
+			var arguments = argumentList.GetArgumentExpressions().ToList();
```
christophwille, line 1986: **Bug (default settings): an all-optional indexer yields zero arguments and falls into the property branches.**

```csharp
public int this[int x = 10] { get; set; }
...
this[10] = 5; this[10] += 5; Console.WriteLine(this[10]); new Probe4 { [10] = 3 };
```

decompiles (PR build, default settings) to `this.Item = 5; this.Item += 5; Console.WriteLine(this.Item); new Probe4 { (initializedObject.Item = 3) }` -- none of which compiles. When every argument is an omitted trailing optional, `GetArgumentExpressions()` is empty and the `arguments.Count == 0` branches below emit a property access. The all-optional case must keep at least one argument (or `FirstOptionalArgumentIndex` must never reach 0 for indexers). `OptionalArguments.cs` only covers `this[int x, int y = 10]`; please add an all-optional fixture.

ICSharpCode.Decompiler/CSharp/CallBuilder.cs
```diff
@@ -1808,24 +1832,26 @@ ExpressionWithResolveResult HandleAccessorCall(ExpectedTargetDetails expectedTar
...
+					// what the access means, so try that first.
+					argumentList.FirstOptionalArgumentIndex = -1;
+				}
+				else if (!argumentsCasted)
 				{
 					argumentsCasted = true;
-					CastArguments(arguments, method.Parameters.ToList());
+					CastArguments(
+						new ArraySegment<TranslatedExpression>(argumentList.Arguments, 0, argumentList.GetActualArgumentCount()),
+						method.Parameters.ToList());
```
christophwille: **Bug (default settings): the arguments are in argument order, `method.Parameters` in declaration order.** With reordered named indexer arguments (now possible since `CheckNoNamedOrOptionalArguments` was dropped for indexers) each argument is cast to the wrong parameter's type.

```csharp
// this[int i, object o] and this[int i, string o]
var v = this[o: (Get(1) == 1) ? "a" : (object)"b", i: Get(2)];
```

Overload resolution is ambiguous once the `(object)` cast is dropped, the loop casts, and the PR build emits `this[o: (int)((Get(1) == 1) ? "a" : "b"), i: (object)Get(2)]` (setter form too) -- uncompilable. Pass `argumentList.ExpectedParameters` (argument order) instead, as `GetRequiredTransformationsForCall` does via `paramTypesInArgumentOrder` (~line 1445).

ICSharpCode.Decompiler/CSharp/CallBuilder.cs
```diff
@@ -1808,24 +1832,26 @@ ExpressionWithResolveResult HandleAccessorCall(ExpectedTargetDetails expectedTar
...
-			if (isSetter)
-			{
-				value = arguments.Last();
-				arguments.Remove(value);
-			}
-
 			IMember? foundMember;
-			while (!IsUnambiguousAccess(expectedTargetDetails, targetResolveResult, method, arguments, argumentNames, out foundMember))
+			while (!IsUnambiguousAccess(expectedTargetDetails, targetResolveResult, method,
+				argumentList.GetArgumentResolveResults(), argumentList.GetArgumentNames(), out foundMember))
```
christophwille: **Default-output change, untested: `AddNamesToPrimitiveValues` now applies to every indexer access, and the retry loop never turns it off.** `GetRequiredTransformationsForCall` tries `AddNamesToPrimitiveValues = false` before casting arguments/target; this loop does not, so names win over casts.

PR build, default settings: `dictionary[true]` -> `dictionary[key: true]`, `this[true]` -> `this[flag: true]` (master: `dictionary[true]`). Worse, with `class Base { virtual this[bool flag] }`, `class Derived : Base { override this[bool f] }` and `d[true]` on a `Derived d` (IL: `callvirt Base::get_Item` on a `Derived` receiver), the names `['flag']` fail against the override's parameter name and the loop ends in a target cast: `((Base)d)[flag: true]` where master emitted `d[true]`.

Either don't apply primitive-value names to accessor calls, or add the `AddNamesToPrimitiveValues = false` step first in this loop -- and add a bool-indexer fixture either way (the PR description does not mention this change).

ICSharpCode.Decompiler/CSharp/CallBuilder.cs
```diff
@@ -1111,9 +1131,18 @@ private ArgumentList BuildArgumentList(ExpectedTargetDetails expectedTargetDetai
...
 			list.UseImplicitlyTypedOut = true;
 			list.AddNamesToPrimitiveValues = expressionBuilder.settings.NamedArguments && expressionBuilder.settings.NonTrailingNamedArguments;
 			return list;
+
+			static bool IsSetterAccessorWrittenAsAssignment(IMethod method)
+			{
+				if (!method.IsAccessor || !method.ReturnType.IsKnownType(KnownTypeCode.Void))
+					return false;
+				// Same shapes that Build() hands to HandleAccessorCall.
+				return method.AccessorOwner!.SymbolKind == SymbolKind.Indexer || method.Parameters.Count == 1;
```
christophwille: Low severity / contrived: this predicate (`method.Parameters.Count == 1`) and `Build()`'s routing condition (`argumentList.ExpectedParameters.Length == allowedParamCount`) can disagree after params expansion. An IL-level void accessor `set_Foo([ParamArray] int[] value)` with `ExpandParamsArguments` is expanded to N arguments by `TransformParamsArgument`, `Build` skips `HandleAccessorCall`, but `IsSetter` is already true, so `GetArgumentExpressions()` drops the last argument in the ordinary invocation path: `set_Foo(1)` for `set_Foo(1, 2)`. Not producible from C#, but the comment "Same shapes that Build() hands to HandleAccessorCall" is not literally true.

christophwille (commented):
## Review summary

The optional-argument part works for the fixtures added, but opening indexer accessors to the named-argument machinery (`NamedArgumentTransform` + `HandleAccessorCall` now consuming `GetArgumentNames()`/`GetArgumentExpressions()`) re-exposes several assumptions elsewhere that were only safe because accessor calls never carried names. All findings below were reproduced against a build of this branch with small probe assemblies; `master` decompiles the same probes correctly.

**Default settings, plain Roslyn output (crash or uncompilable code):**
1. `NamedArgumentTransform`: an indexer setter inside a `BlockKind.CallInlineAssign` block gets replaced by a `CallWithNamedArgs` block -> `Block.CheckInvariant` assert (Debug) / `MatchInlineAssignBlock() returned false` (Release).
2. `HandleAccessorCall`: when every indexer argument is an omitted trailing optional, zero arguments remain and the code falls into the property branches, emitting `this.Item = 5` / `initializedObject.Item`.
3. `HandleAccessorCall`: `CastArguments` pairs argument-order arguments with declaration-order `method.Parameters`; with reordered named indexer arguments each argument is cast to the wrong parameter's type.
4. `HandleAccessorCall`: `AddNamesToPrimitiveValues` now applies to every indexer access (`dictionary[true]` -> `dictionary[key: true]`) and, unlike `GetRequiredTransformationsForCall`, the retry loop never tries turning it off before casting -> `((Base)d)[flag: true]` on overridden indexers. Untested default-output change.

**Need `AggressiveInlining` (or an aggressive context: catch-when, ctor initializer, expression tree):**
5. `CanExtendNamedArgument` still names an indexer setter's value argument; `CallWithNamedArgs` then reorders it away from the last position, which `BuildArgumentList`/`HandleAccessorCall` rely on -> `this[value: Get(0), y: Get(1)] = Get(2);`.
6. An indexer getter that is the `Target` of a `CompoundAssignmentInstruction` can now become a `CallWithNamedArgs` block -> `CompoundAssignmentInstruction.CheckValidTarget` assert.

**Low severity:**
7. `IsSetterAccessorWrittenAsAssignment` and `Build()`'s routing condition can disagree after params expansion (contrived, not producible from C#).

Common root cause for 1, 5, 6: accessor calls can now be wrapped in `CallWithNamedArgs` blocks, but several IL consumers (`CallInlineAssign` invariant, `CompoundAssignmentInstruction.CheckValidTarget`, `CanExtendNamedArgument`) and the position-based "last argument is the value" logic in `CallBuilder` still assume they cannot. Guarding `CanIntroduceNamedArgument` against a `call.Parent` that is a `CallInlineAssign` block / compound-assignment target, plus identifying the setter value by parameter index rather than position, closes all three.

Minor cleanups (no separate comments): `GetArgumentExpressions` still does `argumentNames.Take(argumentCount)` after `GetArgumentNames` already truncates (dead); `lastNameableArgument` is an exclusive end (misnamed); the setter predicate is duplicated in three places (`Build`, `IsSetterAccessorWrittenAsAssignment`, `NamedArgumentTransform`).

Details and repro snippets are in the inline comments.

christophwille (commented):
## Re-review (head d5c17cd)

All seven findings of the previous review (id 5002832803) are fixed at this head; each was re-run against a Debug build of the branch with the original probe snippets:

1. Indexer setter inside a `CallInlineAssign` block -> `CanIntroduceNamedArgument` now stops on that parent; `Check(this[y: Get(1), x: Get(2)] = Get(3))` decompiles to `int y = Get(1); int x = (this[Get(2), y] = Get(3)); Check(x);`.
2. All-optional indexer -> `FirstOptionalArgumentIndex` is bumped to 1 in `HandleAccessorCall`; `this[10] = 5; this[10] += 5; Console.WriteLine(this[10]); new Probe4 { [10] = 3 }` all come back as written.
3. `CastArguments` now takes `ExpectedParameters` (argument order); the `this[o: ..., i: ...]` overload pair casts the right argument.
4. `AddNamesToPrimitiveValues` is off for accessor calls; `dictionary[true]`, `this[true]` and `d[true]` on an overriding indexer are unchanged.
5. `CanExtendNamedArgument` stops at `NameableArgumentCount`; `int s = Get(0); this[y: Get(1), x: Get(2)] = s;` keeps the value last (also with `AggressiveInlining=true`).
6. Compound-assignment target guard in `CanIntroduceNamedArgument`; `this[y: Get(1), x: Get(2)] += 5` and `++` are fine in both inlining modes.
7. `params` setter: `IsWrittenAsMemberAccess` + `!isSetter` guard and the `ParamsPropertySetter` fixture.

Also probed without regressions: `OptionalArguments=false`, `NamedArguments=false`, `AlwaysQualifyMemberReferences=true`, `AggressiveInlining=true`, null-conditional, deconstruction targets, ref-returning indexers (assignment, `++`, `ref` local), struct receivers (field, `ref` parameter, array element), interface indexers with optional parameters, `base[x]`, generic indexers, string/enum/nullable defaults, lambdas, object/collection initializers, `in` parameters, and the constructor path that now shares the ladder.

**One new regression (inline comment):** `OmittedArgumentsAreDefaultsOf` refuses any argument list that has an `ArgumentToParameterMap`, so plain method calls that combine named arguments with omitted trailing optional arguments now write the defaults back out. `M(b: Get(2), a: Get(1))` for `void M(int a, int b, int c = 3)` decompiles to `M(b: Get(2), a: Get(1), c: 3)` where `master` gives `M(b: Get(2), a: Get(1))`. Compilable, but a step back for ordinary calls, and not covered by the `NamedArguments`/`OptionalArguments` fixtures (which is why the suite stays green).

**Low-severity observation (inline comment):** a named indexer access on a receiver whose static type overrides the indexer with different parameter names now comes out as a target cast, `((Base2)d)[y: Get(1), x: Get(2)]`, where `master` emitted `int y = Get(1); d[Get(2), y]`. Correct, and consistent with what the method path already does, so no action required unless you want to keep the temporary in that case.

All findings were reproduced against a Debug build of this branch (`ilspycmd`, default settings unless stated) and cross-checked against `master`.

ICSharpCode.Decompiler/CSharp/CallBuilder.cs
```diff
@@ -1193,6 +1243,31 @@ bool ExtractArguments([NotNullWhen(true)] out IType? elementType, [NotNullWhen(t
...
+		{
+			int argumentCount = argumentList.IsSetter ? argumentList.Length - 1 : argumentList.Length;
+			int omittedFrom = argumentList.GetActualArgumentCount();
+			if (omittedFrom >= argumentCount)
+				return true;
+			if (foundMember is not IParameterizedMember foundParameterizedMember)
+				return false;
+			// A name says nothing about which arguments the declaration considers trailing.
+			if (argumentList.ArgumentToParameterMap != null)
+				return false;
```
christophwille: **Regression (plain method calls, default settings): named arguments and omitted trailing optional arguments no longer combine.**

This early `return false` fires for every argument list that has an `ArgumentToParameterMap`, and the caller then sets `FirstOptionalArgumentIndex = -1`, so the defaults are written back out whenever a call also has a reordered named argument. `master` dropped them (its `IsUnambiguousCall` truncated the names to `firstOptionalArgumentIndex` and let overload resolution decide).

```csharp
public static int Get(int x) => x;
public void M(int a, int b, int c = 3) { }
public void N(int x, int y = 10, int z = 20) { }
public void Call()
{
	M(b: Get(2), a: Get(1));
	N(y: Get(1), x: Get(2));
	N(z: Get(1), x: Get(2));
}
```

PR build:
```csharp
M(b: Get(2), a: Get(1), c: 3);
N(y: Get(1), x: Get(2), z: 20);
N(z: Get(1), x: Get(2), y: 10);
```
master:
```csharp
M(b: Get(2), a: Get(1));
N(y: Get(1), x: Get(2));
N(z: Get(1), x: Get(2));
```

The comment above is not quite right: with names in play, an omitted argument does not have to be a *trailing parameter* at all (`N(z: Get(1), x: Get(2))` legitimately omits the middle `y`), it only has to be a trailing *argument* whose parameter is optional in the member found. That is exactly what this loop can check by mapping the argument index through the map, e.g.

```csharp
var map = argumentList.ArgumentToParameterMap;
int offset = map == null ? 0 : map.Count - argumentList.Length; // skips the 'this' slot
for (int i = omittedFrom; i < argumentCount; i++)
{
	int p = map == null ? i : map[offset + i];
	if (p < 0 || p >= parameters.Count || !IsOptionalArgument(parameters[p], argumentList.Arguments[i]))
		return false;
}
```

and then drop the `ArgumentToParameterMap != null` bail-out. A fixture line for `M(b: Get(2), a: Get(1))` in `NamedArguments.cs` or `OptionalArguments.cs` would pin it; none of the existing ones exercise the combination.

ICSharpCode.Decompiler/CSharp/CallBuilder.cs
```diff
@@ -1013,6 +1037,22 @@ private ArgumentList BuildArgumentList(ExpectedTargetDetails expectedTargetDetai
...
+				// An access writes the value from the last argument; in any other order there is
+				// no access syntax for it.
+				writtenAsMemberAccess = false;
+			}
+			bool isSetter = writtenAsMemberAccess && TakesAssignedValueLast(method);
+			// A name in an element access names a parameter of the indexer, which the type system
+			// takes from the getter; the accessor being called may name them differently.
+			IReadOnlyList<IParameter> namedParameters = method.AccessorOwner is IProperty { IsIndexer: true } indexer
+				? indexer.Parameters
+				: method.Parameters;
```
christophwille, line 1055: Low severity / output quality: the names come from the indexer the IL names, which for a virtual call is the base declaration. When the receiver's static type overrides it with different parameter names, `IsUnambiguousAccess` fails on the names and the ladder ends in a target cast.

```csharp
public class Base2 { public virtual int this[int x, int y] { get => x; set { } } }
public class Derived2 : Base2 { public override int this[int a, int b] { get => a; set { } } }
...
public void Call(Derived2 d) { Console.WriteLine(d[b: Get(1), a: Get(2)]); }
```

PR build: `Console.WriteLine(((Base2)d)[y: Get(1), x: Get(2)]);` -- master: `int y = Get(1); Console.WriteLine(d[Get(2), y]);`

Both compile and mean the same thing, and the method path on master already produces `((Base2)d).M(y: Get(1), x: Get(2))`, so this is consistent rather than wrong. Mentioning it only because the previous round flagged the `((Base)d)[flag: true]` cast for primitive names; here the names are genuinely needed, so a cast is the honest fallback. No change required.

dgrunwald:
"Let an indexer access leave out arguments and name them" -- do we really need this feature?
It took a lot of complexity to get there (too much to review in one evening).
Please don't let AI transform the CallBuilder into something no one understands anymore.

siegfriedpammer:
The main complexity comes from the fact that I asked the LLM to refactor the disambiguation algorithm into something we can reuse. At least that was my intent for a later refactoring, currently it seems there is a lot of duplication going on.

Honestly, I think CallBuilder has way too many responsibilities already. Originally we extracted the "call" specific things out of ExpressionBuilder and it seems that now CallBuilder suffers from the same disease:

- Calls to ordinary methods
- Calls to ctors
- Dictionary initializers???
- string.Format parsing???
- Calls to operators
- Calls to indexers
- Range support???
- Delegate construction???

all the items marked ??? should be discussed and moved somewhere else, IMHO. At least we should think about it.

And there is of course the big machinery that repeatedly asks OR/MemberLookup: Would this call still resolve to the method called in IL? ... And, what about now? ... Now? ... Still not? That is something we should probably make reusable too... if it were reusable, adding optional args support for indexers would be almost free and not this monster of a change.

That aside: A cheap fix would be to just remove the assertion?

PR #4044, opened 2026-08-24
Fix #4040: Compose accessibility and static overlays onto search result icons

### LINKED ISSUE #4040: ILSpy does not show mini icons (for `static`, `private` or `internal` ) in search list.
### Steps to reproduce
1. Open ILSpy
2. Type something (`ICollection` for example) in search field

Image
ILSpy version: 10.1.1.8388+1377eb6e7351b21112858c8c1df39848f40181ec
Image

### Details
* Product in use: ILSpy
* ILSpy version: 11.0.0.9375+cdeae656fee6e184ff76cdc549edc7f9031dba09
* .NET version: 10.0.11-servicing.26373.116+e2f47b0110ed922f21a1522da67279133ce28f32

christophwille:
Fixes #4040.

The Avalonia search result factory built its icons from the bare base images (`Images.Class`, `Images.Field`, ...) and never went through the `Images.GetIcon` overlay compositor the tree nodes use. Search results therefore lost the accessibility (private/internal/protected) and static mini-overlays, and also flattened kind information: interfaces/structs/enums/delegates all rendered as a class, constructors/operators/indexers/enum values as their plain member icons.

This mirrors the ILSpy 10 (WPF) `SearchResultFactory`, which delegates per entity kind to the tree nodes' static `GetIcon` helpers so search icons match the tree by construction:

- `TypeTreeNode` and `EventTreeNode` get the same `public static GetIcon(...)` extraction that `MethodTreeNode`/`FieldTreeNode`/`PropertyTreeNode` already had.
- `AvaloniaSearchResultFactory.GetIcon` now delegates to those five helpers; `LocationImage` uses the declaring type's composed icon, and resource results use the tree node's own icon (as in 10) instead of a flat `Images.Library`.
- New headless tests (`ILSpy.Tests/Search/SearchResultFactoryIconTests.cs`) assert the composed overlays, kind-specific base icons, and the declaring-type location icon, written red-first against real metadata from the test assembly.

### DIFF
--- a/ILSpy.Tests/Search/SearchResultFactoryIconTests.cs
+++ b/ILSpy.Tests/Search/SearchResultFactoryIconTests.cs
@@ -117,9 +117,49 @@ public class SearchResultFactoryIconTests
 		location.BaseImage.Should().BeSameAs(Images.Class);
 		location.Overlays.Should().Equal(Images.OverlayInternal);
 	}
+
+	[AvaloniaTest]
+	public void Internal_Static_Method_Result_Composes_Static_And_Internal_Overlays()
+	{
+		var method = fixtureType.Methods.Single(m => m.Name == "InternalStaticMethod");
+
+		var icon = (LayeredImage)factory.Create(method).Image;
+
+		icon.BaseImage.Should().BeSameAs(Images.Method);
+		icon.Overlays.Should().Equal(Images.OverlayStatic, Images.OverlayInternal);
+	}
+
+	[AvaloniaTest]
+	public void Protected_Property_Result_Composes_Protected_Overlay()
+	{
+		var property = fixtureType.Properties.Single(p => p.Name == "ProtectedProperty");
+
+		var icon = (LayeredImage)factory.Create(property).Image;
+
+		icon.BaseImage.Should().BeSameAs(Images.Property);
+		icon.Overlays.Should().Equal(Images.OverlayProtected);
+	}
+
+	[AvaloniaTest]
+	public void Private_Event_Result_Composes_Private_Overlay()
+	{
+		var @event = fixtureType.Events.Single(e => e.Name == "privateEvent");
+
+		var icon = (LayeredImage)factory.Create(@event).Image;
+
+		icon.BaseImage.Should().BeSameAs(Images.Event);
+		icon.Overlays.Should().Equal(Images.OverlayPrivate);
+	}
+
+	[AvaloniaTest]
+	public void Top_Level_Type_Result_Falls_Back_To_Namespace_Location_Image()
+	{
+		// A top-level type has no declaring type, so the location column shows its namespace.
+		factory.Create(fixtureType).LocationImage.Should().BeSameAs(Images.Namespace);
+	}
 }
 
-#pragma warning disable CS0169 // fields exist only as metadata probes for the tests above
+#pragma warning disable CS0169, CS0067 // members exist only as metadata probes for the tests above
 class SearchIconFixture
 {
 	static int privateStaticField;
@@ -129,5 +169,11 @@ class SearchIconFixture
 	enum NestedEnum { None }
 
 	protected internal class NestedProtectedInternal { }
+
+	internal static void InternalStaticMethod() { }
+
+	protected int ProtectedProperty { get; }
+
+	event System.EventHandler? privateEvent;
 }
-#pragma warning restore CS0169
+#pragma warning restore CS0169, CS0067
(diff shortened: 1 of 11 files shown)


siegfriedpammer (commented):
Checked this against `release/10.1` and ran it locally.

**Verified**

- The member/type path matches 10.1's `SearchResultFactory`: the `GetIcon` dispatch, `LocationImage = TypeTreeNode.GetIcon(declaringType)`, and the extracted `TypeTreeNode.GetIcon` / `EventTreeNode.GetIcon` helpers are equivalent to the WPF originals.
- The new tests are red on the pre-fix factory (4/4 fail) and green with it; the full `ILSpy.Tests` suite is 1223 passed / 3 skipped / 0 failed.

**Behaviour differences vs 10.1 (all pre-existing on master, not introduced here)**

- `protected internal` types get a different badge. 10.1 has a separate type-only overlay mapper (`TypeTreeNode.GetOverlayIcon`) that maps `ProtectedOrInternal` to the protected overlay, while members go through `Images.GetOverlayIcon` and get protected-internal. The Avalonia `Images.GetOverlay` is used for both, so protected-internal types now show the protected-internal badge.
- Static overlay: 10.1's `GetTypeIcon` reports `isStatic` only in the class branch; here `type.IsStatic` is passed for every kind. No observable difference, since only static classes report `IsStatic`.
- The resource-result change is currently unreachable: `RunningSearch.GetStrategy` returns `null` for `SearchMode.Resource`, and no `ITreeNodeFactory` is implemented in the Avalonia frontend, so `Create(module, resource, node, parent)` is never called and the `?? Images.Resource` / `?? Images.Library` fallbacks are not exercised.

**Same icon composition is still missing elsewhere; 10.1 routes these through the helpers this PR now exposes**

- `ILSpy/TreeNodes/DerivedTypesEntryNode.cs:56` - bare `Interface`/`Class`; 10.1 uses `TypeTreeNode.GetIcon(type)`. (`BaseTypesEntryNode` matches 10.1 as-is.)
- `ILSpy/ViewModels/ComparisonEntryTreeNode.cs:85-101` - flat image per kind; 10.1's `CompareViewModel` uses all five `GetIcon` helpers.
- `ILSpy/Analyzers/AnalyzedTypeTreeNode.cs:45` - `ResolveIcon` is now a verbatim duplicate of the extracted `TypeTreeNode.GetIcon`.
- `ILSpy/Analyzers/AnalyzedFieldTreeNode.cs:47` - flat `Images.Field`; loses `EnumValue`, `Literal`, `FieldReadOnly`.
- `ILSpy/Analyzers/AnalyzedPropertyTreeNode.cs:49` - loses the indexer glyph and the extension overlay.
- `ILSpy/Analyzers/AnalyzedMethodTreeNode.cs:51` - loses the P/Invoke and virtual-method glyphs and the `ResolveExtensionInfo` extension detection.
- `ILSpy/Analyzers/AnalyzedEventTreeNode.cs:50` - equivalent to `EventTreeNode.GetIcon`, duplicated.

**Test coverage**

Only the `ITypeDefinition` and `IField` arms are covered; the `IMethod`, `IProperty` and `IEvent` delegations and the namespace `LocationImage` fallback are untested.

christophwille:
Thanks for the thorough check. Addressed in two follow-up commits:

**Behaviour differences**

- `protected internal` types: restored the WPF type-only overlay mapping — `TypeTreeNode.GetIcon` now maps `ProtectedOrInternal` to the plain protected badge (members keep the combined badge via `Images.GetOverlay`). Covered by a new red-first test (`Protected_Internal_Nested_Type_Gets_Plain_Protected_Overlay`). Since the mapping lives in the helper, the tree, search, and every other caller pick it up together.
- Static overlay for non-class kinds: left as-is. Only static classes report `IsStatic` (abstract+sealed), so passing `type.IsStatic` unconditionally is observably identical to 10.1's class-branch-only check, and the code is simpler.
- Resource-result path: left in place. It matches 10.1's factory and becomes live once a resource search strategy / `ITreeNodeFactory` lands in the Avalonia frontend; the `??` fallbacks are harmless until then.

**Missing icon composition elsewhere**

All seven sites now delegate to the tree nodes' helpers, matching 10.1:

- `DerivedTypesEntryNode` -> `TypeTreeNode.GetIcon`
- `ComparisonEntryTreeNode` -> all five helpers (namespace/module arms unchanged)
- `AnalyzedTypeTreeNode` / `AnalyzedMethodTreeNode`: the private `ResolveIcon` duplicates are deleted; `AnalyzedFieldTreeNode`, `AnalyzedPropertyTreeNode`, `AnalyzedEventTreeNode` delegate directly. The analyzer pane thereby gains the enum-value/literal/readonly, indexer, and P/Invoke/virtual/extension glyphs it was missing.

Full `ILSpy.Tests` suite green after both commits.

_Posted by an AI agent (Claude) on Christoph's behalf._

siegfriedpammer (approved):

PR #4053, opened 2026-08-26
Handle unresolved value-type call targets

sonyps5201314:
Related to #4049.

### Problem

When an instance method belongs to a value type from a missing assembly, its declaring type has unknown stack semantics. ILReader inserts an invalid `Ref -> Unknown` conversion around the managed-reference receiver.

`ExpressionBuilder.TranslateTarget` then treats the conversion as a value receiver and emits invalid C# such as:

```csharp
((MyEnumerator)(ref value)).MoveNext();
((IDisposable)(*(MyEnumerator*)(&value))).Dispose();
```

### Solution

For the narrowly defined `Invalid`, `Ref -> Unknown` conversion on an unresolved call target:

- use the existing `UnwrapConv` helper to recover the original managed-reference receiver;
- infer managed-reference target semantics from the IL argument when the unresolved declaring type cannot provide them;
- treat same-named unresolved by-reference element types as the same receiver type.

This PR contains only this call-target fix and its `Issue3729` regression case. It does not include constructor-initializer, primary-constructor, collection-expression, resolver, project export, or backing-field changes from #4049.

### Test

The ILPretty case was added first. Before the implementation it failed with an `unsafe` method, a `(MyEnumerator)(ref value)` cast, and a pointer dereference for the constrained `IDisposable.Dispose` call.

After the implementation:

- `ILPrettyTestRunner.Issue3729`: passed.
- Complete `ILPrettyTestRunner` fixture: 63 total, 62 passed, 1 pre-existing ignored, 0 failed.
- Full `Release|Any CPU` solution build: 16 succeeded, 0 failed.
- `BuildTools/format.ps1`: clean.


dgrunwald (approved):

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -2891,13 +2902,16 @@ internal TranslatedExpression TranslateTarget(ILInstruction? target, bool nonVir
...
-					if (CallInstruction.ExpectedTypeForThisPointer(memberDeclaringType, constrainedTo) == StackType.Ref)
+					if (requiresManagedReference)
 					{
 						// When accessing members on value types, ensure we use a reference of the correct type,
 						// and not a pointer or a reference to a different type (issue #1333)
-						if (!(translatedTarget.Type is ByReferenceType brt && NormalizeTypeVisitor.TypeErasure.EquivalentTypes(brt.ElementType, constrainedTo ?? memberDeclaringType)))
+						IType expectedTargetType = constrainedTo ?? memberDeclaringType;
+						if (!(translatedTarget.Type is ByReferenceType brt
+							&& (NormalizeTypeVisitor.TypeErasure.EquivalentTypes(brt.ElementType, expectedTargetType)
+								|| (expectedTargetType.Kind == TypeKind.Unknown && brt.ElementType.ReflectionName == expectedTargetType.ReflectionName))))
```
dgrunwald: The ReflectionName string comparison is now unnecessary with the change on #4070.

sonyps5201314: Agreed. I removed the `ReflectionName` fallback and now rely solely on `TypeErasure.EquivalentTypes` after #4070. The single commit was force-pushed as `d730ec0a2`.

I also applied it without conflicts on current `master` (`05d274ff7`) and verified:
- `BuildTools/format.ps1`: clean;
- full `Release|Any CPU` solution build: 16 succeeded, 0 failed;
- complete `ILPrettyTestRunner`: 64 total, 63 passed, 1 pre-existing ignored, 0 failed.

PR #4057, opened 2026-08-26
Fix uncompilable lambda parameter lists

siegfriedpammer:
Two bugs that produce uncompilable C# for lambda parameter lists. Both were found and fixed
while building #4004, but neither depends on that feature, so they are extracted here and
#4004 is closed - the natural-type work is not planned for now.

## Attributes force a parenthesized parameter list

`LambdaNeedsParenthesis` predates lambda attributes and only looked at the single parameter's
type and modifiers. `MakeParameters` erases parameter types when one of them is an anonymous
type (it cannot be named), and an attribute on the *lambda* does not block that erasure - so the
parameter loses its type, the parenthesis check says none is needed, and the output does not
parse:

```csharp
// ILSpy today
new[] { new { X = 1 } }.Select([My] a => a.X).Sum();
// error CS8916: Attributes on lambda expressions require a parenthesized parameter list
```

The guard now also triggers on attributes, on the lambda and on its single parameter. The second
case is defence in depth: `MakeParameters` already refuses to erase types when a parameter itself
carries attributes.

## 'params' and default values only where the language allows them

`ConvertParameter` takes `IsParams` and the default value from the `IParameter` and they were
printed on whichever syntax the renderer had chosen. Anonymous methods can declare neither
(CS1670, CS1065), and lambdas only since C# 12:

```csharp
// ILSpy today
return delegate(params int[] values) {
    return values.Length + seed;
};
// error CS1670: params is not valid in this context
```

Lambda syntax is now forced when a parameter carries either shape, gated by a new
`LambdaOptionalAndParamsParameters` setting (C# 12). Below that version, and whenever the
parameter list is about to be dropped anyway (`delegate { }` for unused parameters), the
modifiers are stripped instead - the delegate type still provides both, so on the anonymous
function they are decorative.

Only what the anonymous function's own metadata declares is written; nothing is taken from the
delegate's `Invoke`. A lambda may state a different default than its target delegate, or none
where the delegate has one, and reflection over the lambda's method reports what the lambda
declared - copying from `Invoke` would change what the recompiled assembly says. The fixture
covers the disagreeing cases (`OptionalDifferentDefault`, `OptionalOnlyInDelegate`,
`OptionalOnlyInLambda`, `ParamsOnlyInDelegate`, `ParamsOnlyInLambda`).

One consequence: Roslyn 4.x does not emit `ParamArrayAttribute` on the synthesized method, so a
`params` lambda compiled by those versions decompiles without the modifier. Roslyn 5 records it,
so the `params` cases in the fixture are guarded on `ROSLYN5`. The output still compiles either
way - the delegate type declares `params`, and that is what call sites bind against.

## Known gap

Below C# 12 the modifiers are dropped rather than preserved in attribute form, so a C# 11
decompile no longer shows that a parameter was `params` or had a default. Tracked in #4058.
Only half of it is repairable: `([Optional][DefaultParameterValue(5)] int x) => ...` compiles at
C# 11, but `[ParamArray]` is rejected by CS0674 on any parameter and anonymous methods cannot
declare `params` either (CS1670), so below C# 12 there is no legal way to spell it on the
anonymous function.

## Verification

Rebased onto master. `ICSharpCode.Decompiler.Tests`: full suite green on the previous head; the
touched fixtures (`LambdaOptionalAndParamsParameters`, `DelegateConstruction`, `LambdaAttributes`,
Pretty + Correctness) re-run after the rebase. Both defects above were reproduced against a
master build and the emitted code confirmed uncompilable with the quoted compiler errors.

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/LambdaOptionalAndParamsParameters.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/LambdaOptionalAndParamsParameters.cs
@@ -1,0 +1,99 @@
+#pragma warning disable CS9099, CS9100
+using System;
+
+namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
+{
+	internal class LambdaOptionalAndParamsParameters
+	{
+		public delegate void ParamsAction(params int[] xs);
+
+		public delegate int OptionalFunc(int x = 5);
+
+		public delegate int PlainFunc(int x);
+
+		// A lambda states 'params' and defaults on its own account: what the delegate declares
+		// is the delegate's, and reflection over the lambda's method reports only the lambda's.
+		// Roslyn records the modifier on the anonymous function's method from version 5 on, so
+		// the cases that need it back are guarded on ROSLYN5.
+
+		private int total;
+
+#if ROSLYN5
+		public ParamsAction ParamsStatementBody()
+		{
+			return (params int[] xs) => {
+				total += xs.Length;
+				Console.WriteLine(xs.Length);
+			};
+		}
+#endif
+
+#if ROSLYN5
+		public ParamsAction ParamsSingleStatementBody()
+		{
+			return (params int[] xs) => {
+				Console.WriteLine(xs.Length);
+			};
+		}
+#endif
+
+		public OptionalFunc OptionalStatementBody()
+		{
+			return (int x = 5) => {
+				total += x;
+				return x * 2;
+			};
+		}
+
+		public OptionalFunc OptionalExpressionBody()
+		{
+			return (int x = 5) => x * 2;
+		}
+
+		public ParamsAction ParamsWithoutParameterList()
+		{
+			return delegate {
+			};
+		}
+
+		public OptionalFunc OptionalWithoutParameterList()
+		{
+			return delegate {
+				return 1;
+			};
+		}
+
+		// The lambda's own defaults and params modifier can differ from the delegate's; the metadata
+		// records the lambda's, so that is what must round-trip.
+		public OptionalFunc OptionalDifferentDefault()
+		{
+			return (int x = 7) => x * 2;
+		}
+
+		public OptionalFunc OptionalOnlyInDelegate()
+		{
+			return (int x) => x * 2;
+		}
+
+		public PlainFunc OptionalOnlyInLambda()
+		{
+			return (int x = 3) => x * 2;
+		}
+
+		public ParamsAction ParamsOnlyInDelegate()
+		{
+			return (int[] xs) => {
+				Console.WriteLine(xs.Length);
+			};
+		}
+
+#if ROSLYN5
+		public Action<int[]> ParamsOnlyInLambda()
+		{
+			return (params int[] xs) => {
+				Console.WriteLine(xs.Length);
+			};
+		}
+#endif
+	}
+}
--- a/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
@@ -2627,7 +2627,6 @@ namespace ICSharpCode.Decompiler.CSharp
 				let v = ident.GetILVariable()
 				where v != null && v.Function == function && v.Kind == VariableKind.Parameter
 				select ident).Any();
-
 			bool isLambda = false;
 			if (ame.Parameters.Any(p => p.Type is null))
 			{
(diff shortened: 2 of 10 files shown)


ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -2621,6 +2652,12 @@ from ident in body.Descendants.OfType<IdentifierExpression>()
 				// C# 10 lambdas can have attributes, but anonymous methods cannot
 				isLambda = true;
 			}
+			else if (ame.Parameters.Any(p => p.IsParams || p.DefaultExpression is not null))
```
christophwille, line 2655: **Regression:** this branch sits above the existing `parametersAreUsed || (ParameterTypesAreAccessible && ParametersAreNameable)` guard, so any Invoke-derived `params`/default forces lambda syntax even when master would (correctly) emit `delegate { }`.

Minimal repro (C# 12 settings, `LambdaOptionalAndParamsParameters = true`):

```csharp
public delegate void ParamsAction(params int[] xs);

public ParamsAction NoParamList()
{
    return delegate { };
}
```

Expected (master):
```csharp
return delegate { };
```
Actual (this branch):
```csharp
return (params int[] array) => { };
```

The name `array` is synthesized from the type; the source had no parameter list at all. Worse, the same path with a `params` delegate whose element type is not accessible/nameable from the current type (`ParameterTypesAreAccessible(function) == false`, e.g. obfuscated or IL-only assemblies) now emits `(params Inaccessible[] p) => { }` -> CS0122, where `delegate { }` was always legal.

Suggested fix: when `!parametersAreUsed` (or the types are not accessible/nameable), strip `IsParams`/`DefaultExpression` and let it fall through to the `delegate {}` path instead of forcing `isLambda`. As the strip comment above already says, the decorations are purely cosmetic there.

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -2621,6 +2652,12 @@ from ident in body.Descendants.OfType<IdentifierExpression>()
...
+			{
+				// 'params' and parameter default values cannot be declared on an anonymous method
+				// in any language version; only the lambda form is legal.
+				isLambda = true;
+			}
 			else if (settings.UseLambdaSyntax && ame.Parameters.All(p => p.ParameterModifier == ReferenceKind.None && !p.IsParams)
```
christophwille, line 2661: Nit: `!p.IsParams` here can never be false anymore - any `IsParams` parameter is already caught by the new `else if` above. The comment inside this branch ("params parameters ... conservatively left alone") now describes behaviour this PR removed. Drop the predicate and the two comment sentences about params.

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -2565,6 +2565,26 @@ internal ExpressionWithResolveResult TranslateFunction(IType delegateType, ILFun
...
+			// there - so take them from Invoke, keeping declaration and call sites consistent.
+			var invokeParameters = delegateType.GetDelegateInvokeMethod()?.Parameters;
+			if (invokeParameters?.Count == ame.Parameters.Count)
+			{
+				int parameterIndex = 0;
+				foreach (var pd in ame.Parameters)
+				{
+					var invokeParameter = invokeParameters[parameterIndex++];
+					pd.IsParams |= invokeParameter.IsParams;
+					if (pd.DefaultExpression is null
```
christophwille, line 2580: Nit/perf: `astBuilder.ConvertParameter(invokeParameter)` builds a full `ParameterDeclaration` (type conversion, attribute conversion, name) per parameter of every anonymous function only to read `DefaultExpression` and discard the rest. Cheaper and equivalent to what `ConvertParameter` does internally:

```csharp
if (pd.DefaultExpression is null && invokeParameter.IsDefaultValueAssignmentAllowed())
{
    pd.DefaultExpression = astBuilder.ConvertConstantValue(invokeParameter.Type, invokeParameter.GetConstantValue(throwOnInvalidMetadata: true));
}
```
(with the same `BadImageFormatException` handling `ConvertParameter` has).

christophwille (commented):
## Review summary

Read every hunk plus the full `TranslateFunction`, `MakeParameters`, `ConvertParameter`, `MetadataParameter.IsParams`, `AssignVariableNames` and `TransformExpressionTrees`. Built `ilspycmd` from this branch, decompiled a 10-case sample (params / optional / generic default / decimal default / multi-param / expression tree / `delegate {}` / differing default) at C# 12 and C# 11 with both this branch and master, and recompiled the outputs with csc at `-langversion 12` / `11`. Both compile clean.

**What holds up**
- Both bugs in the PR description are real and the fixes are correct: attribute-on-lambda now forces parentheses; `params`/defaults are only emitted on explicitly typed lambda lists at C# 12+, stripped otherwise.
- Expression trees are unaffected (the Invoke lookup happens while `DelegateType` is still `Expression<T>`, so nothing is applied) - output stays as before.
- Generic delegate `T x = default` renders as `(int x = default(int))` / `(string x = null)` - legal and matches the delegate.
- A lambda's own C# 12 default (`(int x = 7)` against a delegate default of 5) is preserved, not overwritten by Invoke's.
- `params ReadOnlySpan<T>` is gated by `TypeSystemOptions.ParamsCollections`, so C# 12 settings never emit a C# 13 form.

**Findings** (inline): one regression - the new forced-lambda branch bypasses the existing `delegate {}` guard, which regresses parameter-list-less anonymous methods and can produce uncompilable output for unnameable/inaccessible parameter types - plus two cleanups.

**Aside, pre-existing on master (not this PR):** for `delegate void MultiOpt(int a, string s = "hi", params object[] rest)`, the call `c.Multi()(1)` decompiles as `c.Multi()(1, "hi")`. The "call sites are already trimmed" statement in the PR body only holds when the optional parameter is last.

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -2609,7 +2629,18 @@ from ident in body.Descendants.OfType<IdentifierExpression>()
 				let v = ident.GetILVariable()
 				where v != null && v.Function == function && v.Kind == VariableKind.Parameter
 				select ident).Any();
-
+			if (!settings.LambdaOptionalAndParamsParameters || ame.Parameters.Any(p => p.Type is null))
```
dgrunwald: If this condition is true, the whole params/default-adding logic above was unnecessary.
Move this block up and use an `else if` to skip adding the stuff when we'd delete it again anyways.

dgrunwald (changes requested):

ICSharpCode.Decompiler.Tests/TestCases/Pretty/LambdaOptionalAndParamsParameters.cs
```diff
@@ -0,0 +1,54 @@
...
+		{
+			return (int x = 5) => {
+				total += x;
+				return x * 2;
+			};
+		}
+
+		public OptionalFunc OptionalExpressionBody()
+		{
+			return (int x = 5) => x * 2;
```
dgrunwald, line 50: This test only shows happy cases where the default value in the lambda agrees with the default value in the delegate. But C# doesn't require these to match, and the metadata can distinguish other cases, so please add more tests.

In particular, add a test where the delegate has a default, but the lambda does not. The whole "copy default from delegate type" affair feels unmotivated to me -- and arguably wrong if anyone inspects the default via reflection.

dgrunwald (changes requested):

dgrunwald (approved):

Issue #4054, opened 2026-08-26
Drag & Drop from Non-Explorer Windows no longer works


sonyps5201314:
https://github.com/icsharpcode/ILSpy/issues/3760
> 2. Drag and drop is not supported
> 
> The new cross-platform version of ILSpy doesn't seem to support dragging and dropping DLLs directly into the assembly list view. This feature was fully supported and worked well in previous versions.

This issue hasn't been completely resolved. Dragging and dropping managed DLLs directly from the search results of third-party programs like *FileLocator Pro* still doesn't work. However, dragging those exact same files into VS Code opens them perfectly fine.

christophwille:
I tried with https://files.community/docs/getting-started/install - and that works.

But indeed the software you mentioned (installed in Lite mode) allows me to dnd into v10, but I get the no-drop icon for v11.

sonyps5201314:
Yes, this is definitely a regression, but I'm not sure if it's an underlying bug in Avalonia.

PR #4062, opened 2026-08-27
Detect local functions renamed by obfuscators

siegfriedpammer:
Closes half of #3202. The other half - the `Class148.GetItem` async pattern in the same report -
already decompiles cleanly on master and needs nothing.

## What the report shows

`Class148.DefaultCheckCopyed` keeps its closure struct in the output:

```csharp
<>c__DisplayClass29_0 <>c__DisplayClass29_0_ = default;
<>c__DisplayClass29_0_.iid = 0L;
...
smethod_1(null, ref <>c__DisplayClass29_0_);
```

## Why

The 2024 analysis on the issue blamed `TransformDisplayClassUsage` for requiring the display
struct's first field store to be the first instruction of its init block. That is no longer what
blocks this file: a clean Roslyn build of the same shape - struct closure, field stores preceded
by unrelated statements, captured by a local function, inside an async method - decompiles
correctly on master today.

The obfuscator destroyed both markers `LocalFunctionDecompiler.IsLocalFunctionMethod` relies on:

```
.method assembly hidebysig static
    void smethod_1 (string string_0, valuetype Class148/'<>c__DisplayClass29_0'& ...)
```

No `[CompilerGenerated]`, and the name is `smethod_1` rather than
`<DefaultCheckCopyed>g__...|29_0`. The method therefore stays an ordinary static method, the
display struct escapes by reference into a plain call, and
`TransformDisplayClassUsage.ValidateDisplayClassUses` correctly refuses to scalar-replace it.
Everything downstream behaves as designed; only the detection gate is wrong.

## The fix

When the attribute-and-name test fails, fall back to a signature test: an assembly-visible method
that takes a by-ref, compiler-generated, generated-named nested value type **of the same module**
can only be a local function. Roslyn emits struct closures exclusively for local functions, and no
hand-written C# can name a `<>c__DisplayClass` type.

It reuses the existing `FindRefStructParameters` visitor, which already drops cross-module handles
- so BCL signatures like `AwaitUnsafeOnCompleted(ref TAwaiter, ref TStateMachine)` cannot match.
`LocalFunctionNeedsAccessibilityChange` had its own copy of that loop and now shares the helper.

The second commit stops `AssignVariableNames` renaming the recovered function to `f`: an
obfuscated name has no synthetic `|x_y` suffix to split off, so its trailing digits belong to the
name and it is taken verbatim.

## Verification

- `DefaultCheckCopyed` decompiles to plain `long iid` / `IEnumerable<string> hist` locals plus
  `void smethod_1(string)` as a local function.
- `ICSharpCode.Decompiler.Tests`: 3490 total, 0 failed, 45 skipped.
- Whole-assembly decompile of `ICSharpCode.Decompiler.dll` (161,864 lines): **byte-identical**
  before and after, and identical again at `--languageversion CSharp6`, which is the only setting
  that exercises `LocalFunctionNeedsAccessibilityChange`.
- No measurable cost: three timed runs of that decompile, medians 69.1 s before and after.

### DIFF
--- a/ICSharpCode.Decompiler/IL/Transforms/AssignVariableNames.cs
+++ b/ICSharpCode.Decompiler/IL/Transforms/AssignVariableNames.cs
@@ -505,13 +505,21 @@ namespace ICSharpCode.Decompiler.IL.Transforms
 
 		private static void AssignNameToLocalFunction(ILFunction function, VariableScope parentContext)
 		{
-			if (!LocalFunctionDecompiler.ParseLocalFunctionName(function.Name, out _, out var newName) || !IsValidName(newName))
-				newName = null;
 			string nameWithoutNumber;
 			int number;
-			if (!string.IsNullOrEmpty(newName))
+			string newName;
+			if (LocalFunctionDecompiler.ParseLocalFunctionName(function.Name, out _, out newName) && IsValidName(newName))
 			{
 				nameWithoutNumber = SplitName(newName, out number);
+			}
+			else if (IsValidName(function.Name))
+			{
+				// An obfuscated assembly carries an arbitrary name instead of
+				// "<caller>g__name|x_y". Take it verbatim: it has no synthetic suffix to
+				// split off, so its trailing digits belong to the name itself
+				// ("smethod_1", not "smethod_" numbered 1).
+				nameWithoutNumber = function.Name;
+				number = 1;
 			}
 			else
 			{
--- a/ICSharpCode.Decompiler/IL/Transforms/LocalFunctionDecompiler.cs
+++ b/ICSharpCode.Decompiler/IL/Transforms/LocalFunctionDecompiler.cs
@@ -781,34 +781,44 @@ namespace ICSharpCode.Decompiler.IL.Transforms
 			var method = metadata.GetMethodDefinition(methodHandle);
 			var declaringType = method.GetDeclaringType();
 
-			if ((method.Attributes & MethodAttributes.Assembly) == 0 || !(method.IsCompilerGenerated(metadata) || declaringType.IsCompilerGenerated(metadata)))
+			if ((method.Attributes & MethodAttributes.Assembly) == 0)
 				return false;
 
-			if (!ParseLocalFunctionName(metadata.GetString(method.Name), out _, out _))
-				return false;
+			if ((method.IsCompilerGenerated(metadata) || declaringType.IsCompilerGenerated(metadata))
+				&& ParseLocalFunctionName(metadata.GetString(method.Name), out _, out _))
+			{
+				return true;
+			}
 
-			return true;
+			// Obfuscators strip the CompilerGeneratedAttribute and rewrite the
+			// "<caller>g__name|x_y" name, but they cannot remove the by-ref display-struct
+			// parameter: a compiler-generated struct closure is only ever passed by reference
+			// to the local functions that capture it.
+			return HasDisplayStructParameter(module, methodHandle);
+		}
+
+		/// <summary>
+		/// True if any parameter is a by-ref compiler-generated closure struct of this module.
+		/// </summary>
+		static bool HasDisplayStructParameter(MetadataFile module, MethodDefinitionHandle methodHandle)
+		{
+			var metadata = module.Metadata;
+			var method = metadata.GetMethodDefinition(methodHandle);
+			FindRefStructParameters visitor = new FindRefStructParameters();
+			method.DecodeSignature(visitor, default);
+			foreach (var h in visitor.RefStructTypes)
+			{
+				var td = metadata.GetTypeDefinition(h);
+				if (td.IsCompilerGenerated(metadata) && td.IsValueType(metadata) && td.HasGeneratedName(metadata))
+					return true;
+			}
+			return false;
 		}
 
 		public static bool LocalFunctionNeedsAccessibilityChange(MetadataFile module, MethodDefinitionHandle methodHandle)
 		{
-			if (!IsLocalFunctionMethod(module, methodHandle))
-				return false;
-
-			var metadata = module.Metadata;
-			var method = metadata.GetMethodDefinition(methodHandle);
-
-			FindRefStructParameters visitor = new FindRefStructParameters();
-			method.DecodeSignature(visitor, default);
-
-			foreach (var h in visitor.RefStructTypes)
-			{
-				var td = metadata.GetTypeDefinition(h);
-				if (td.IsCompilerGenerated(metadata) && td.IsValueType(metadata))
-					return true;
-			}
-
-			return false;
+			return IsLocalFunctionMethod(module, methodHandle)
+				&& HasDisplayStructParameter(module, methodHandle);
 		}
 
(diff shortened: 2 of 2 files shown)


ICSharpCode.Decompiler/IL/Transforms/LocalFunctionDecompiler.cs
```diff
@@ -781,36 +781,46 @@ public static bool IsLocalFunctionMethod(MetadataFile module, MethodDefinitionHa
...
+			if ((method.IsCompilerGenerated(metadata) || declaringType.IsCompilerGenerated(metadata))
+				&& ParseLocalFunctionName(metadata.GetString(method.Name), out _, out _))
+			{
+				return true;
+			}
 
-			return true;
+			// Obfuscators strip the CompilerGeneratedAttribute and rewrite the
+			// "<caller>g__name|x_y" name, but they cannot remove the by-ref display-struct
+			// parameter: a compiler-generated struct closure is only ever passed by reference
```
christophwille, line 795: `HasDisplayStructParameter` never sees the closure struct when the declaring type or method is generic: `FindRefStructParameters.GetGenericInstantiation` (further down, not in this diff) returns `default`, so a `ref <>c__DisplayClass0_0<T>` parameter is a TypeSpec that gets dropped before it reaches `RefStructTypes`. Result: an obfuscated local function inside `class Generic<T>` is byte-identical to master (struct closure, `ref A_`, `internal static smethod_1(..., ref <>c__DisplayClass0_0)` all leak into the output). Full sample + expected/actual in the review summary.

The same visitor gap already affects `LocalFunctionNeedsAccessibilityChange` on clean Roslyn output under `--languageversion CSharp6` (generic sibling stays `internal` with a `private` struct parameter, CS0051).

Verified fix, one line:

```csharp
public TypeDefinitionHandle GetGenericInstantiation(TypeDefinitionHandle genericType, ImmutableArray<TypeDefinitionHandle> typeArguments) => genericType;
```

christophwille (commented):
## Review summary

Method: built `ilspycmd` from this branch, compiled a 3-case C# sample, "obfuscated" it via ildasm/ilasm (renamed the `<...>g__...|x_y` methods to `smethod_N`, stripped `[CompilerGenerated]` from the methods but kept it on the closure structs), and diffed master vs. PR output.

The fix works for the issue's exact shape (local function in a non-generic type) and the reasoning behind the fallback is sound - I could not construct a Roslyn-emitted false positive. Two real gaps, one cosmetic:

1. **Generic declaring type / generic method is silently missed** (inline comment). `FindRefStructParameters.GetGenericInstantiation` returns `default`, so a `ref <>c__DisplayClass0_0<T>` parameter (a TypeSpec) never reaches `RefStructTypes`. The same visitor gap already affects `LocalFunctionNeedsAccessibilityChange` on clean Roslyn output. One-line fix verified.
2. **No regression test.** The change loosens a detection gate and alters `AssignVariableNames`, with no fixture. CLAUDE.md asks for the rejected fixture to be captured first; `ICSharpCode.Decompiler.Tests/CLAUDE.md` describes the ILPretty `.il` + expected `.cs` mechanism this needs. Happy to share the `.il` I used (plain / generic / colliding-name cases) if useful.
3. Cosmetic: name collisions between obfuscated local functions render as `smethod_0` / `smethod_02` (inline comment).

Also: the branch carries `a980639f9` (deconstruction fixtures) and `27b45a445` (Fix #4054 drag/drop), which are unrelated to this PR's subject - probably worth rebasing them out so the merge stays focused.

### Repro for (1)

```csharp
using System;
using System.Threading.Tasks;

public class Generic<T>
{
    public async Task M(string s)
    {
        long iid = 0;
        IEnumerable<string> hist = null;
        Local(null);
        await Task.Delay(1);
        Local(s);
        void Local(string x) => Console.WriteLine(x + iid + hist);
    }
}
```

Compile with Roslyn, then in IL rename `<M>g__Local|0_0` to `smethod_1` and remove `[CompilerGenerated]` from that method only (keep it on the struct). Same sample in a non-generic `class Plain` for comparison.

**Expected** (what this PR produces for `Plain`):

```csharp
long iid = 0L;
IEnumerable<string> hist = null;
smethod_1(null);
await Task.Delay(1);
smethod_1(s);
void smethod_1(string x)
{
    Console.WriteLine(x + iid + hist);
}
```

**Actual** for `Generic<T>` (byte-identical to master):

```csharp
<>c__DisplayClass0_0 A_ = default(<>c__DisplayClass0_0);
A_.iid = 0L;
A_.hist = null;
smethod_1(null, ref A_);
await Task.Delay(1);
smethod_1(s, ref A_);
// plus, at type level:
internal static void smethod_1(string x, ref <>c__DisplayClass0_0 A_1) { ... }
// and the <>c__DisplayClass0_0 struct itself
```

Same root cause on *clean* Roslyn output with `--languageversion CSharp6`: `Generic<T>.<M>g__Local|0_0` stays `internal static` with a `private` struct parameter (CS0051 on recompile), while `Plain`'s is correctly rewritten to `private static`.

Verified fix: `GetGenericInstantiation(genericType, typeArguments) => genericType` in `FindRefStructParameters` makes both the obfuscated-generic case decompile to a local function and flips the CS6 output to `private`.

ICSharpCode.Decompiler/IL/Transforms/AssignVariableNames.cs
```diff
@@ -505,14 +505,22 @@ public void Run(ILFunction function, ILTransformContext context)
...
 				nameWithoutNumber = SplitName(newName, out number);
 			}
+			else if (IsValidName(function.Name))
+			{
+				// An obfuscated assembly carries an arbitrary name instead of
+				// "<caller>g__name|x_y". Take it verbatim: it has no synthetic suffix to
+				// split off, so its trailing digits belong to the name itself
+				// ("smethod_1", not "smethod_" numbered 1).
+				nameWithoutNumber = function.Name;
+				number = 1;
```
christophwille, line 522: Cosmetic: obfuscators legally give two local functions in the same method the same name (they're just overloads in IL). With the name taken verbatim and `number = 1`, the collision is resolved by gluing the counter onto the trailing digits:

```csharp
// class Collide { void M(int a) { First(1); Second("x"); void First(int i) => ...; void Second(string s) => ...; } }
// both renamed to smethod_0
smethod_0(1);
smethod_02("x");
void smethod_0(int i) { ... }
void smethod_02(string s) { ... }
```

Compiles and is correct, but `smethod_02` reads like a third obfuscator output. A separator (`smethod_0_2`) or only falling back to `SplitName` on collision would be clearer. Low priority.

dgrunwald:
Uhhh, local functions are a pure prettiness feature. Why do we want to have this for obfuscated code?
IMHO that issue should have been closed as won't fix.

siegfriedpammer:
I don't see much harm in reusing the existing infrastructure and make it less dependent on compiler-generated names...

dgrunwald:
I haven't looked at the diff yet in detail; I just saw a lot more added than removed lines and question if we want to increase the size of our codebase for this.

dgrunwald (approved):

Issue #3736, opened 2026-05-13
Cast lost when ternary with mismatched branch types is hoisted to `object` local


OpenSourcereR-dev:
When a `?:`'s two branches have different stack types and ILSpy can't find a common type tighter than `object`, it hoists the conditional into an `if/else` with an `object obj` local — but drops the explicit cast that was present in the source. The dropped cast can re-bind to a different overload on recompile.

### Repro

```csharp
using System.Collections.Generic;
using System.Linq;

public class Repro
{
    public IEnumerator<byte> Bug(byte[] data, bool reversed)
        => (reversed ? ((IEnumerable<byte>)data).Reverse() : (IEnumerable<byte>)data).GetEnumerator();
}
```

### ilspycmd 10.0.1.8346 output

```csharp
public IEnumerator<byte> Bug(byte[] data, bool reversed)
{
    object obj;
    if (!reversed)
    {
        obj = data;
    }
    else
    {
        obj = data.Reverse();  // (IEnumerable<byte>) cast lost
    }
    return ((IEnumerable<byte>)obj).GetEnumerator();
}
```

When a project-local `void Reverse(this byte[])` (or similar shadowing extension) is in scope, `data.Reverse()` re-binds to it on recompile → CS0029 ("Cannot implicitly convert type 'void' to 'object'"). Even without a shadow, the decompile is semantically lossy.

dotPeek emits the inline ternary form with both casts preserved.

### Related

[#1501](https://github.com/icsharpcode/ILSpy/issues/1501), [#1614](https://github.com/icsharpcode/ILSpy/issues/1614), [#3322](https://github.com/icsharpcode/ILSpy/issues/3322), [#3464](https://github.com/icsharpcode/ILSpy/issues/3464), [#793](https://github.com/icsharpcode/ILSpy/issues/793) are all in the same family (ILSpy dropping a needed cast); none is exactly this `?:`→`if/else` hoist variant.

dgrunwald:
The `object obj` local and the cast on `data` are unrelated.

For method calls, ILSpy will omit redundant casts if this doesn't change the method that will be called.
So if for some reason ILSpy cannot see the `void Reverse(this byte[])` overload, there won't be a cast, which is valid in this case. If you have a `void Reverse(this byte[])` overload, ILSpy will add the cast as it's needed in that case.

siegfriedpammer:
[...]

PR #4064, opened 2026-08-27
Expand nested conditional operators back into if-else

### LINKED ISSUE #2027: [Enhancement] Optionally disable ternary operator for long statements
[sample.zip](https://github.com/icsharpcode/ILSpy/files/4707275/sample.zip) (part of [Terraria](https://store.steampowered.com/app/105600))
The code was decompiled "correctly" but as a 2000-characters long line:
```
((float)num > (float)(58 + player.armor.Length + player.dye.Length + player.miscEquips.Length + player.miscDyes.Length + player.bank.item.Length + player.bank2.item.Length + player.bank3.item.Length + 1)) ? player.bank4.item[num - 58 - (player.armor.Length + player.dye.Length + player.miscEquips.Length + player.miscDyes.Length + player.bank.item.Length + player.bank2.item.Length + player.bank3.item.Length + 1) - 1] : (((float)num > (float)(58 + player.armor.Length + player.dye.Length + player.miscEquips.Length + player.miscDyes.Length + player.bank.item.Length + player.bank2.item.Length + 1)) ? player.bank3.item[num - 58 - (player.armor.Length + player.dye.Length + player.miscEquips.Length + player.miscDyes.Length + player.bank.item.Length + player.bank2.item.Length + 1) - 1] : (((float)num > (float)(58 + player.armor.Length + player.dye.Length + player.miscEquips.Length + player.miscDyes.Length + player.bank.item.Length + player.bank2.item.Length)) ? player.trashItem : (((float)num > (float)(58 + player.armor.Length + player.dye.Length + player.miscEquips.Length + player.miscDyes.Length + player.bank.item.Length)) ? player.bank2.item[num - 58 - (player.armor.Length + player.dye.Length + player.miscEquips.Length + player.miscDyes.Length + player.bank.item.Length) - 1] : (((float)num > (float)(58 + player.armor.Length + player.dye.Length + player.miscEquips.Length + player.miscDyes.Length)) ? player.bank.item[num - 58 - (player.armor.Length + player.dye.Length + player.miscEquips.Length + player.miscDyes.Length) - 1] : (((float)num > (float)(58 + player.armor.Length + player.dye.Length + player.miscEquips.Length)) ? player.miscDyes[num - 58 - (player.armor.Length + player.dye.Length + player.miscEquips.Length) - 1] : (((float)num > (float)(58 + player.armor.Length + player.dye.Length)) ? player.miscEquips[num - 58 - (player.armor.Length + player.dye.Length) - 1] : (((float)num > (float)(58 + player.armor.Length)) ? player.dye[num - 58 - player.armor.Length - 1] : (((float)num > 58f) ? player.armor[num - 58 - 1] : player.inventory[num]))))))))
```
Adding an option to disable ternary operator `?:` would be great, either globally or only for long conditions.

siegfriedpammer:
`ExpressionTransforms.HandleConditionalOperator` collapses `if (c) a = x; else a = y;` into a
conditional operator, innermost first, and keeps going for as long as the chain does. A source
`else if` ladder therefore comes back as one expression, however long it was. The sample attached
to #2027 decompiles to a single **2095-character** statement; one NLog method reaches **2230
characters nested 29 brackets deep**.

`ExpandNestedConditionals` undoes that past one level, so a statement keeps at most a single `?:`.

```csharp
// before
text2 = ((!text.StartsWith("aspnet", ...) && !text.StartsWith("iis", ...)) ? (text.StartsWith("HostAppName", ...) ? ... /* 20 more levels */ ));

// after
if (text.StartsWith("aspnet", StringComparison.OrdinalIgnoreCase) || text.StartsWith("iis", StringComparison.OrdinalIgnoreCase))
{
    text2 += " - Extension NLog.Web.AspNetCore not included?";
}
else if (text.StartsWith("HostAppName", StringComparison.OrdinalIgnoreCase))
{
    ...
```

Note the compound assignment comes back too: the conditional form forced `text2 = ... (text2 + "...")`.

## Where it runs, and why that is the whole design

At the **end** of the pipeline, immediately before `AssignVariableNames` - not inside
`ExpressionTransforms`.

Every transform that needs its input to be a single expression has to see the collapsed form first:
object and collection initializers, `with`, switch expressions, interpolated string handlers, and
the query lambdas the C# stage later rewrites into clauses. Cutting the chain earlier leaves an
`if`/`else` between the statements they pattern-match on, and they silently stop matching - an
object initializer assigning an init-only member then does not even compile.

This is the same reasoning `ReduceNestingTransform` already gives for walking back
`ConditionDetection`'s aggressive else-inlining once the structure is settled: an earlier transform
is deliberately aggressive because later stages depend on it, and a late pass chooses the final
shape.

Two consequences of running late:

- **No type question.** Only a store to a declared local is expanded. Such a local carries its own
  type from metadata, so splitting the branches cannot widen the result to a stack type and lose a
  cast - the defect filed as #3736. No type inference is involved.
- **No polarity question.** The pass builds the `if`/`else` itself instead of inheriting
  `ConditionDetection`'s IL ordering, so the ladder reads `else if (cond)` rather than
  `else if (!(cond))`. Negated `else if` occurrences across the corpus are unchanged from master.

## Tests

**`TestCases/Pretty/ConditionalOperatorNesting.cs`** covers what must **not** change - each member
is a construct that has to stay a single expression, and each was a compile error or a silent
degradation in the earlier version of this PR:

| member | guards | guard |
|---|---|---|
| `ArrayInitializer` | array initializer block | |
| `Query` | query lambda | |
| `RefLocal` | ref local, CS8174 | `#if CS70` |
| `SwitchExpression` | switch expression arm | `#if CS80` |
| `ObjectInitializer` | init-only member, CS8852 | `#if CS90` |
| `WithExpression` | `with` / record clone, CS1061 | `#if CS90` |
| `FilterKeepsExpression` + the two ctor classes | catch-when filter, `: base(...)`, `: this(...)` | |
| `ShallowChainStaysInline`, `MismatchedBranchTypesStayInline`, `LogicOperatorsAreNotCounted` | the one-level boundary, and `&&` / `\|\|` sharing the IfInstruction representation | |

**`TestCases/ILPretty/ConditionalChain.il`** holds the positive case. It is an ILPretty test rather
than a Pretty one because a Pretty fixture is its own input and expected output, and a chain that
round-trips through collapse and expansion has no fixed point there - whichever polarity is written
comes back inverted.

## Measurements

decompdiff over 20 popular NuGet packages, all 20 decompilable on both sides, against
`origin/master`:

| | old | new |
|---|---|---|
| types compared | 5927 | |
| types changed | **35** | |
| new errors | **0** | |
| lines | 667345 | +451 |
| gotos / `//IL_` warnings / `<>` name leaks | | **0 delta** |

Across the changed types, deepest line 2230 -> 618 characters, maximum bracket depth 29 -> 7,
lines at depth >= 5 down from 22 to 12.

Every declaration the diff introduces carries a name from metadata - `XPathNavigator`, `DbType`,
`Expression`, `Type`, `string`, `byte[]`, `NamingStrategy` - except four (`int num`, `long num`)
that are pre-existing locals whose declaration split off its assignment. No temporary is typed
from a stack type.

`ICSharpCode.Decompiler.Tests`: **3587 total, 0 failed**.

## Scope

The transform handles a nested conditional in two situations.

**Stored to a variable that already exists** - the expansion reuses that variable, so nothing has
to be decided: the variable carries its own type from metadata.

**Anywhere else** - an argument, a return value, a field store - there is nothing to expand into,
so `ILExtraction` provides one, with `PrepareExtract` vetting the move so the order of evaluation
is preserved. The temporary it creates is typed from the *stack* type though, and `I4` is `int`,
`bool`, `char` and every enum at once; extracting on that basis turned a `bool` into `int num` with
`if (num == 0)`, and an enum into `dbType = (IsFixedLength ? 22 : 0)`.

`InferExpectedType` (new, next to `InferType` in `ILTypeExtensions`) is what makes that position
usable. Where `InferType` asks what a value *is*, it asks what the position the value flows into
says it *should be* - a parameter, a return type, a field - all of which carry their type in
metadata, and it answers `UnknownType` where nothing does. Extraction happens only where that
question has an answer, and the temporary is typed from it:

```csharp
XPathNavigator xPathNavigator;                  // the receiver's declaring type, no cast back
if (root) { xPathNavigator = node.GetRootElement(); } else if (IsElement) { … }
xPathNavigator.CreateAttribute("xmlns", prefix, …);

DbType dbType;                                  // the field's type, real member names
dbType = (IsFixedLength ? DbType.AnsiStringFixedLength : DbType.AnsiString);
```

Chains whose position names nothing are left alone - `JToken.ReadFrom`'s, for instance, sits inside
a `!(…)`, so the rule declines rather than guessing.

Two widenings were also built and rejected as unmeasurable: admitting stack slots whose type is not
`object`, and retyping a stack slot from the types its branches agree on. Both produced output
byte-identical to this PR over 5006 types.

Refs #2027 (implements the "only for long conditions" half of that request; there is no global
switch to disable `?:`).

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/ConditionalOperatorNesting.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/ConditionalOperatorNesting.cs
@@ -1,0 +1,144 @@
+// Copyright (c) 2026 Siegfried Pammer
+// 
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this
+// software and associated documentation files (the "Software"), to deal in the Software
+// without restriction, including without limitation the rights to use, copy, modify, merge,
+// publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons
+// to whom the Software is furnished to do so, subject to the following conditions:
+// 
+// The above copyright notice and this permission notice shall be included in all copies or
+// substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+// INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+// PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+// FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
+{
+	public class ConditionalOperatorNesting
+	{
+		private static void Use(object value)
+		{
+		}
+
+		public int ShallowChainStaysInline(int n, int[] a, int[] b)
+		{
+			return ((n > 20) ? a[0] : b[0]) + 1;
+		}
+
+		// One conditional is under the limit and stays where it is, whatever the branch types.
+		public void MismatchedBranchTypesUnderTheLimitStayInline(int n, string s, object o)
+		{
+			Use((n > 10) ? s : o);
+		}
+
+		public void LogicOperatorsAreNotCounted(int n, int[] a)
+		{
+			Use(n > 40 && a[0] > 1 && a[1] > 2 && a[2] > 3 && a[3] > 4);
+		}
+
+		public int FilterKeepsExpression(int n, int[] a, int[] b)
+		{
+			try
+			{
+				throw new Exception();
+			}
+			catch (Exception) when (((n > 40) ? a[0] : ((n > 30) ? b[0] : ((n > 20) ? a[1] : ((n > 10) ? b[1] : a[2])))) > 5)
+			{
+				return 1;
+			}
+		}
+	}
+
+	public class ConditionalOperatorNestingBase
+	{
+		public ConditionalOperatorNestingBase(int x)
+		{
+		}
+	}
+
+	public class ConditionalOperatorNestingCtor : ConditionalOperatorNestingBase
+	{
+		public ConditionalOperatorNestingCtor(int n, int[] a, int[] b)
+			: base((n > 40) ? a[0] : ((n > 30) ? b[0] : ((n > 20) ? a[1] : ((n > 10) ? b[1] : a[2]))))
+		{
+		}
+
+		public ConditionalOperatorNestingCtor(int n, int[] a)
+			: this(n, a, a)
+		{
+		}
+	}
+
+#if CS90
+	public class ConditionalOperatorNestingInit
+	{
+		public int P { get; init; }
+	}
+
+	public record ConditionalOperatorNestingRecord
+	{
+		public int P { get; init; }
+	}
+#endif
+
+	// A nested conditional feeding a construct that has to stay a single expression must not be
+	// expanded: the transform that recognizes the construct matches on the expression, and an
+	// if-else between the statements silently stops it matching - or produces code that does not
+	// compile, as an object initializer assigning an init-only member would.
+	public class ConditionalOperatorNestingSingleExpression
+	{
+		public int[] ArrayInitializer(int n, int a, int b, int c)
+		{
+			return new int[2] {
+				(n > 2) ? a : ((n > 1) ? b : c),
+				5
+			};
+		}
+
+		public IEnumerable<int> Query(int[] xs, int n, int a, int b, int c)
+		{
+			return xs.Where((int x) => x > ((n > 2) ? a : ((n > 1) ? b : c)));
+		}
+#if CS70
+		public int RefLocal(int n, int[] a, int[] b)
+		{
+			ref int reference = ref n > 2 ? ref a[0] : ref n > 1 ? ref b[0] : ref a[1];
+			reference++;
+			return reference;
+		}
+#endif
+#if CS80
+		public int SwitchExpression(int k, int n, int a, int b, int c)
+		{
+			return k switch {
+				1 => (n > 2) ? a : ((n > 1) ? b : c),
+				2 => a,
+				_ => b,
+			};
+		}
+#endif
+#if CS90
+		public ConditionalOperatorNestingInit ObjectInitializer(int n, int a, int b, int c)
+		{
+			return new ConditionalOperatorNestingInit {
+				P = ((n > 2) ? a : ((n > 1) ? b : c))
+			};
+		}
+
+		public ConditionalOperatorNestingRecord WithExpression(ConditionalOperatorNestingRecord i, int n, int a, int b, int c)
+		{
+			return i with {
+				P = ((n > 2) ? a : ((n > 1) ? b : c))
+			};
+		}
+#endif
+	}
+}
--- a/ICSharpCode.Decompiler/IL/ILTypeExtensions.cs
+++ b/ICSharpCode.Decompiler/IL/ILTypeExtensions.cs
@@ -164,6 +164,45 @@ namespace ICSharpCode.Decompiler.IL
 		}
 
 		/// <summary>
+		/// Infers the C# type an instruction expects of the child in <paramref name="childIndex"/>,
+		/// i.e. the counterpart to <see cref="InferType"/>: that one asks what a value is, this one
+		/// asks what the position it flows into says it should be.
+		///
+		/// Returns SpecialType.UnknownType where the position names nothing.
+		/// </summary>
+		/// <remarks>
+		/// Where a value's own type is only its stack type - `I4` being `int`, `bool`, `char` and
+		/// every enum at once - the consumer often still knows, because a parameter, a return type
+		/// or a field carries its type in metadata.
+		/// </remarks>
+		public static IType InferExpectedType(this ILInstruction inst, int childIndex, ICompilation? compilation)
+		{
+			switch (inst)
+			{
+				case CallInstruction call:
+					if (childIndex == 0 && call.IsInstanceCall)
+						return call.ConstrainedTo ?? call.Method.DeclaringType;
+					return call.GetParameter(childIndex)?.Type ?? SpecialType.UnknownType;
+				case Leave leave when childIndex == 0:
+					// the value of a leave is a return value only where it leaves the function body
+					var function = leave.Ancestors.OfType<ILFunction>().FirstOrDefault();
+					if (function == null || leave.TargetContainer != function.Body)
+						return SpecialType.UnknownType;
+					return function.Method?.ReturnType ?? SpecialType.UnknownType;
+				case StObj stobj when childIndex == 1:
+					return stobj.Type;
+				case StLoc stloc when childIndex == 0:
+					return stloc.Variable.Type;
+				case IfInstruction ifInst when childIndex == 0:
+					return compilation?.FindType(KnownTypeCode.Boolean) ?? SpecialType.UnknownType;
+				case NewArr newArr:
+					return compilation?.FindType(KnownTypeCode.Int32) ?? SpecialType.UnknownType;
+				default:
+					return SpecialType.UnknownType;
+			}
+		}
+
+		/// <summary>
 		/// Infers the C# type for an IL instruction.
 		/// 
 		/// Returns SpecialType.UnknownType for unsupported instructions.
--- a/ICSharpCode.Decompiler.Tests/ILPrettyTestRunner.cs
+++ b/ICSharpCode.Decompiler.Tests/ILPrettyTestRunner.cs
@@ -52,6 +52,12 @@ namespace ICSharpCode.Decompiler.Tests
 		}
 
 		[Test]
+		public async Task ConditionalChain()
+		{
+			await Run();
+		}
+
+		[Test]
 		public async Task AnonymousMethodEdgeCases()
 		{
 			await Run();
--- a/ICSharpCode.Decompiler.Tests/PrettyTestRunner.cs
+++ b/ICSharpCode.Decompiler.Tests/PrettyTestRunner.cs
@@ -458,6 +458,12 @@ namespace ICSharpCode.Decompiler.Tests
 		}
 
 		[Test]
+		public async Task ConditionalOperatorNesting([ValueSource(nameof(roslynOnlyOptions))] CompilerOptions cscOptions)
+		{
+			await RunForLibrary(cscOptions: cscOptions);
+		}
+
+		[Test]
 		public async Task UnsafeCode([ValueSource(nameof(defaultOptions))] CompilerOptions cscOptions)
 		{
 			await RunForLibrary(cscOptions: cscOptions | CompilerOptions.ReferenceUnsafe);
--- a/ICSharpCode.Decompiler.Tests/TestCases/ILPretty/ConditionalChain.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/ILPretty/ConditionalChain.cs
@@ -1,0 +1,27 @@
+using System;
+
+public class ConditionalChain
+{
+	public int Ladder(int n, int[] a, int[] b)
+	{
+		int num;
+		if (n > 40)
+		{
+			num = a[0];
+		}
+		else if (n > 30)
+		{
+			num = b[0];
+		}
+		else if (n > 20)
+		{
+			num = a[1];
+		}
+		else
+		{
+			num = ((n > 10) ? b[1] : a[2]);
+		}
+		Console.WriteLine(num);
+		return num;
+	}
+}
--- a/ICSharpCode.Decompiler.Tests/TestCases/ILPretty/ConditionalChain.il
+++ b/ICSharpCode.Decompiler.Tests/TestCases/ILPretty/ConditionalChain.il
@@ -1,0 +1,119 @@
+
+//  .NET IL Disassembler.  Version 10.0.11
+
+
+
+// Metadata version: v4.0.30319
+.assembly extern System.Runtime
+{
+  .publickeytoken = (B0 3F 5F 7F 11 D5 0A 3A )                         // .?_....:
+  .ver 4:0:0:0
+}
+.assembly extern System.Console
+{
+  .publickeytoken = (B0 3F 5F 7F 11 D5 0A 3A )                         // .?_....:
+  .ver 4:0:0:0
+}
+.assembly ConditionalChain
+{
+  .custom instance void [System.Runtime]System.Runtime.CompilerServices.CompilationRelaxationsAttribute::.ctor(int32) = ( 01 00 08 00 00 00 00 00 ) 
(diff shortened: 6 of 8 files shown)


ICSharpCode.Decompiler/IL/Transforms/ExpressionTransforms.cs
```diff
@@ -602,6 +682,20 @@ IfInstruction HandleConditionalOperator(IfInstruction inst)
 			ILInstruction value1, value2;
 			if (trueInst.Instructions[0].MatchStLoc(out v, out value1) && falseInst.Instructions[0].MatchStLoc(v, out value2))
 			{
+				// An if-else-if chain collapses into one conditional operator per step, innermost first.
+				// Past a few levels the resulting `a ? x : (b ? y : (c ? z : w))` is harder to read than
+				// the if-else statements it came from, so stop collapsing and leave the rest as statements.
+				// Ordered cheapest and most selective first: only a chain that is already too deep
+				// is worth asking the more expensive questions about.
+				if (!inst.MatchLogicAnd(out _, out _) && !inst.MatchLogicOr(out _, out _)
```
christophwille: **Bug (uncompilable output): object initializer with init-only property.** The retained if-chain sits between `stloc tmp(newobj)` and the setter call, so `TransformCollectionAndObjectInitializers` no longer matches and the setter is emitted as a plain assignment.

```csharp
public class Init { public int P { get; init; } }
public Init M(int n, int a, int b, int c) => new Init { P = n > 2 ? a : n > 1 ? b : c };
```
Expected (master):
```csharp
return new Init { P = ((n > 2) ? a : ((n > 1) ? b : c)) };
```
Actual (this PR):
```csharp
Init init = new Init();
int p;
if (n <= 2) { p = ((n > 1) ? b : c); } else { p = a; }
init.P = p;   // error CS8852: Init-only property 'Init.P' can only be assigned in an object initializer
return init;
```
Same shape with `required`:
```csharp
public class Req { public required int P { get; set; } }
public Req M(int n, int a, int b, int c) => new Req { P = n > 2 ? a : n > 1 ? b : c };
// actual: Req req = new Req(); int p; if (...) ...; req.P = p; return req;
// error CS9035: Required member 'Req.P' must be set in the object initializer
```

ICSharpCode.Decompiler/IL/Transforms/ExpressionTransforms.cs
```diff
@@ -602,6 +682,20 @@ IfInstruction HandleConditionalOperator(IfInstruction inst)
...
+					&& (ReachesNestingDepth(value1, MaxConditionalOperatorNesting)
+						|| ReachesNestingDepth(value2, MaxConditionalOperatorNesting))
+					&& KeepsResultType(v, value1, context.TypeSystem)
+					&& KeepsResultType(v, value2, context.TypeSystem)
+					&& AllowsStatements(inst))
+				{
+					return inst;
```
christophwille: **Bug (uncompilable output): `with` expression is lost entirely.** The clone call is emitted as the mangled `_003CClone_003E_0024()` and the init-only setter as a statement.

```csharp
public record Rec2 { public int P { get; init; } }
public Rec2 M(Rec2 i, int n, int a, int b, int c) => i with { P = n > 2 ? a : n > 1 ? b : c };
```
Expected:
```csharp
return i with { P = ((n > 2) ? a : ((n > 1) ? b : c)) };
```
Actual:
```csharp
Rec2 rec = i._003CClone_003E_0024();   // error CS1061
int p;
if (n <= 2) { p = ((n > 1) ? b : c); } else { p = a; }
rec.P = p;                             // error CS8852
return rec;
```

ICSharpCode.Decompiler/IL/Transforms/ExpressionTransforms.cs
```diff
@@ -589,6 +589,86 @@ protected internal override void VisitIfInstruction(IfInstruction inst)
...
+
+		/// <summary>
+		/// Gets whether the target variable still describes both values once they are left in
+		/// separate statements. A stack slot has no type of its own and falls back to its stack
+		/// type, which loses the common type of the branches; the conditional operator infers it,
+		/// so in that case the collapsed form is the one that keeps the result well typed.
+		/// </summary>
+		static bool KeepsResultType(ILVariable v, ILInstruction value, ICompilation compilation)
+		{
+			if (v.Kind != VariableKind.StackSlot)
```
christophwille: **Bug (uncompilable output): ref locals.** `KeepsResultType` returns `true` for every non-stack-slot variable, so a ref local assigned from a nested ref conditional is split into a declaration without initializer. The guard has no by-reference awareness.

```csharp
public int M(int n, int[] a, int[] b)
{
    ref int r = ref n > 2 ? ref a[0] : ref n > 1 ? ref b[0] : ref a[1];
    r++;
    return r;
}
```
Expected:
```csharp
ref int reference = ref n > 2 ? ref a[0] : ref n > 1 ? ref b[0] : ref a[1];
```
Actual:
```csharp
ref int reference;   // error CS8174: A declaration of a by-reference variable must have an initializer
if (n <= 2) { reference = ref n > 1 ? ref b[0] : ref a[1]; } else { reference = ref a[0]; }
```

ICSharpCode.Decompiler/IL/Transforms/ExpressionTransforms.cs
```diff
@@ -589,6 +589,86 @@ protected internal override void VisitIfInstruction(IfInstruction inst)
...
+		/// lambdas of a query expression - and statements left behind there do not compile.
+		/// </summary>
+		bool AllowsStatements(IfInstruction inst)
+		{
+			// The innermost enclosing statement: an outer one would be the whole try instruction
+			// for a filter, whose block says nothing about the context the value is built in.
+			var statement = inst.Ancestors.FirstOrDefault(a => a.Parent is Block);
+			if (statement?.Parent is not Block block)
+				return false;
+			var options = ILInlining.OptionsForBlock(block, statement.ChildIndex, context);
```
christophwille: **Bug: record struct primary ctor and static field initializers are no longer recognized.** `IsInConstructorInitializer` is false for struct ctors (no chained base call), so the if-chain stays in the ctor body and `RecordDecompiler` no longer detects the primary constructor.

```csharp
public record struct RS(int N, int A, int B, int C)
{
    public int X { get; init; } = N > 2 ? A : N > 1 ? B : C;
}
```
Expected: the positional record struct as written, with `X { get; init; } = (N > 2) ? A : ((N > 1) ? B : C);`.

Actual: a non-positional `record struct RS` with explicit `N`/`A`/`B`/`C` properties, an explicit `public RS(int N, int A, int B, int C)` ctor containing `int num; if (N <= 2) {...} X = num;`, and a `[CompilerGenerated] public readonly void Deconstruct(...)`.

Same mechanism with static fields (drops `beforefieldinit` semantics):
```csharp
public class St { static int a = 1, b = 2; public static int X = a > 2 ? a : b > 1 ? b : a; }
// actual: public static int X; static St() { int x; if (...) ...; X = x; }
```

ICSharpCode.Decompiler/IL/Transforms/ExpressionTransforms.cs
```diff
@@ -589,6 +589,86 @@ protected internal override void VisitIfInstruction(IfInstruction inst)
...
+			return true;
+		}
+
+		/// <summary>
+		/// Gets whether the statement holding <paramref name="inst"/> may be split into several.
+		/// The pipeline forces aggressive inlining wherever the result has to stay a single
+		/// expression - constructor initializers, catch-when filters, expression trees and the
+		/// lambdas of a query expression - and statements left behind there do not compile.
+		/// </summary>
+		bool AllowsStatements(IfInstruction inst)
```
christophwille: **Silent degradation in every other single-expression consumer.** None of these are `Aggressive` inlining contexts, so `AllowsStatements` says yes and the later transform no longer matches:

```csharp
// switch expression -> becomes a switch statement with `int result; if ...`
int S(int k, int n, int a, int b, int c) => k switch { 1 => n > 2 ? a : n > 1 ? b : c, 2 => a, _ => b };

// interpolated string -> explicit DefaultInterpolatedStringHandler h = new(5, 2); h.AppendLiteral("v="); int value; if ...; h.AppendFormatted(value); ...
string I(int n, int a, int b, int c) => $"v={(n > 2 ? a : n > 1 ? b : c)} w={n}";

// query -> xs.Where((int x) => { int num; if ...; return x > num; })
IEnumerable<int> Q(int[] xs, int n, int a, int b, int c) => from x in xs where x > (n > 2 ? a : n > 1 ? b : c) select x;

// array initializer -> int[] array = new int[2]; int num; if ...; array[0] = num; array[1] = 5;
int[] A(int n, int a, int b, int c) => new[] { n > 2 ? a : n > 1 ? b : c, 5 };
```
Call arguments, tuple literals and ctor args similarly get a temp. This is why the cut belongs after those transforms, not in `ExpressionTransforms`.

ICSharpCode.Decompiler/IL/Transforms/ExpressionTransforms.cs
```diff
@@ -589,6 +589,86 @@ protected internal override void VisitIfInstruction(IfInstruction inst)
...
+		/// Gets whether the target variable still describes both values once they are left in
+		/// separate statements. A stack slot has no type of its own and falls back to its stack
+		/// type, which loses the common type of the branches; the conditional operator infers it,
+		/// so in that case the collapsed form is the one that keeps the result well typed.
+		/// </summary>
+		static bool KeepsResultType(ILVariable v, ILInstruction value, ICompilation compilation)
+		{
+			if (v.Kind != VariableKind.StackSlot)
+				return true;
+			if (value is not IfInstruction)
```
christophwille: `KeepsResultType` is inconsistent rather than a type-safety check. `ILTypeExtensions.InferType` returns `UnknownType` for `ldc.i4`, `conv`, arithmetic and most instructions, and the O stack slot is `object`, so only chains whose leaves are plain int/long/float/double loads or calls are ever split:

```csharp
int x = n > 2 ? a[0] : n > 1 ? a[1] : a[2];          // split
int y = n > 2 ? 1 : n > 1 ? 2 : 3;                   // stays nested
int z = n > 2 ? a[0] + 1 : n > 1 ? a[1] + 1 : a[2] + 1; // stays nested
string s = n > 2 ? p : n > 1 ? q : r;                // stays nested (object slot vs string)
```
`MismatchedBranchTypesStayInline` passes because `ldstr` infers `Unknown`, not because of the `object` leaf.

ICSharpCode.Decompiler/IL/Transforms/ExpressionTransforms.cs
```diff
@@ -602,6 +682,20 @@ IfInstruction HandleConditionalOperator(IfInstruction inst)
 			ILInstruction value1, value2;
 			if (trueInst.Instructions[0].MatchStLoc(out v, out value1) && falseInst.Instructions[0].MatchStLoc(v, out value2))
 			{
+				// An if-else-if chain collapses into one conditional operator per step, innermost first.
+				// Past a few levels the resulting `a ? x : (b ? y : (c ? z : w))` is harder to read than
+				// the if-else statements it came from, so stop collapsing and leave the rest as statements.
+				// Ordered cheapest and most selective first: only a chain that is already too deep
+				// is worth asking the more expensive questions about.
+				if (!inst.MatchLogicAnd(out _, out _) && !inst.MatchLogicOr(out _, out _)
+					&& (ReachesNestingDepth(value1, MaxConditionalOperatorNesting)
```
christophwille: Dead code: `!inst.MatchLogicAnd(...) && !inst.MatchLogicOr(...)` on `inst` can never be false here, because both branches were just verified to be `Block`s while `MatchLogicAnd`/`MatchLogicOr` (PatternMatching.cs) require an `ldc.i4` branch. Likewise `depth <= 0` in `ReachesNestingDepth` is unreachable with the constant `1`, and `KeepsResultType`'s `value is not IfInstruction` early return duplicates the worklist's leaf case. With `MaxConditionalOperatorNesting = 1` the whole worklist reduces to `value is IfInstruction i && !i.MatchLogicAnd(..) && !i.MatchLogicOr(..)`. The comment sells the dead check as the cheap selective filter, which will mislead the next reader.

ICSharpCode.Decompiler.Tests/TestCases/Pretty/ConditionalOperatorNesting.cs
```diff
@@ -0,0 +1,111 @@
...
+			{
+				num = b[0];
+			}
+			else if (n > 40)
+			{
+				num = a[1];
+			}
+			else if (n <= 30)
+			{
+				int num2;
```
christophwille: The fixture bakes in accidental shapes (this `num2` temporary, the stack-slot-only split, `roslynOnlyOptions` only) and has no negative case for any single-expression consumer: object initializer with init/required member, `with`, switch expression, ref local, query, interpolated string, positional record struct. Those are exactly where the change breaks (see the inline comments on `ExpressionTransforms.cs`); please add them here so the 5 compile errors above become red tests.

christophwille (dismissed):
## Review summary

I built `ilspycmd` from this branch and from master, decompiled a small fixture assembly with both, and recompiled the output. Master's output compiles cleanly; this branch's output has **5 compiler errors** (CS8852 x2, CS1061, CS9035, CS8174) plus several silent degradations.

**Root cause:** the guard in `HandleConditionalOperator` cuts the chain in `ExpressionTransforms`, i.e. *before* every later transform that requires its input to be a single expression: object/collection initializers, `with`, switch expressions, query lambdas, interpolated-string handlers, record primary-constructor detection. `AllowsStatements` only consults `ILInlining.OptionsForBlock`, which knows about ctor initializers/filters/expression trees but nothing about those transforms. Any nested conditional feeding one of them now leaves an `if` chain between the statements they pattern-match on.

The 20-package corpus did not catch this because `KeepsResultType` only lets the guard fire when every leaf is a plain int/long/float/double load (`InferType` returns `Unknown` for `ldc.i4`, `conv`, arithmetic; an `object` stack slot never matches a `string` leaf). `MismatchedBranchTypesStayInline` passes for that reason, not for the one it claims.

**Suggested fix:** do not cut in `ExpressionTransforms` at all. Add a small late pass (after `TransformCollectionAndObjectInitializers` / switch-expression / query / interpolation transforms) that expands a *statement-level* `stloc local(if (c) x else y)` whose branches still nest a `?:` back into `if/else`, restricted to `VariableKind.Local` targets whose type is not by-reference. At that point every consumer that needed the expression has already taken it, the `else if` ladder this PR wants still appears, and the type / aggressive-inlining questions go away (a statement-level store to a plain local can always be split).

Repro samples with expected/actual are in the inline comments (all compiled Release, `net10.0`, decompiled with default settings). Please also add them as pretty-test negative cases so the guard cannot regress them again.

ICSharpCode.Decompiler.Tests/TestCases/PdbGen/NestedConditional.residual.txt
```diff
@@ -0,0 +1,13 @@
+0x06000001 ICSharpCode.Decompiler.Tests.TestCases.PdbGen.NestedConditional.Ladder
```
dgrunwald: Why is this test useful?

ICSharpCode.Decompiler/IL/ILTypeExtensions.cs
```diff
@@ -161,6 +163,46 @@ public static bool IsFloatType(this PrimitiveType type)
...
+		/// Where a value's own type is only its stack type - `I4` being `int`, `bool`, `char` and
+		/// every enum at once - the consumer often still knows, because a parameter, a return type
+		/// or a field carries its type in metadata.
+		/// </remarks>
+		public static IType InferExpectedType(this ILInstruction inst, int childIndex, ICompilation? compilation)
+		{
+			switch (inst)
+			{
+				case CallInstruction call:
+					bool hasThis = call.OpCode != OpCode.NewObj && !call.Method.IsStatic;
```
dgrunwald: Use `call.IsInstanceCall` instead.

dgrunwald (approved):

PR #4065, opened 2026-08-27
Insert casts in switch expression to ensure a best common type exists

### LINKED ISSUE #3683: Casts not being applied to switch expression case bodies when type is ambiguous
When something is done with the result of the switch expression before it is assigned/returned (e.g. call a member function), and the type is ambiguous (e.g. case bodies have multiple child types of a parent type), we have to apply a cast to at least one of the case bodies to tell the compiler which type it is, or it's a compile error.

### Input code
```c#
public abstract class Animal
{
    public abstract string Name { get; }

    public Animal ToValidated()
    {
        if (Name is null)
        {
            throw new ArgumentNullException(nameof(Name));
        }
        return this;
    }
}

public class Dog : Animal
{
    public override string Name => "Dog";
}

public class Fish : Animal
{
    public override string Name => "Fish";
}

public class TestAnimals
{
    public Animal GetAnimal(int animalType)
    {
        return (animalType switch
        {
            0 => (Animal)new Dog(),
            1 => new Fish(),
            _ => throw new ArgumentException("Invalid animal type")
        }).ToValidated();
    }
}
```
### Erroneous output
```c#
[...]
public class TestAnimals
{
    public Animal GetAnimal(int animalType)
    {
        return (animalType switch
        {
            0 => new Dog(), // no cast, compile error
            1 => new Fish(),
            _ => throw new ArgumentException("Invalid animal type")
        }).ToValidated();
    }
}
```

For reference, the specific compiler error when attempting to compile the output:
```
No best type was found for the switch expression.[CS8506](https://msdn.microsoft.com/query/roslyn.query?appId=roslyn&k=k(CS8506))
```

### Details
* Product in use: IlspyCmd
* Version in use: current master

dgrunwald:
Fixes #3683.

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
@@ -4342,6 +4342,8 @@ namespace ICSharpCode.Decompiler.CSharp
 				resultType = compilation.FindType(inst.ResultType);
 			}
 
+			var expressionsForTypeInference = new List<TranslatedExpression>();
+
 			foreach (var section in inst.Sections)
 			{
 				if (section == defaultSection)
@@ -4370,12 +4372,51 @@ namespace ICSharpCode.Decompiler.CSharp
 				switchExpr.SwitchSections.Add(defaultSES);
 			}
 
-			return switchExpr.WithILInstruction(inst).WithRR(new ResolveResult(resultType));
+			var ti = new TypeInference(compilation, resolver.conversions);
+			IType commonType = ti.GetBestCommonType(
+				expressionsForTypeInference.SelectArray(e => e.ResolveResult),
+				out bool success);
+			if (success && commonType.GetStackType() == inst.ResultType)
+			{
+				return switchExpr.WithILInstruction(inst).WithRR(new ResolveResult(commonType));
+			}
+			else
+			{
+				// Try to help out the C# compiler by casting the first element to the expected type:
+				expressionsForTypeInference[0].Expression.ReplaceWith(
+					node => expressionsForTypeInference[0] = expressionsForTypeInference[0].ConvertTo(resultType, this)
+				);
+				commonType = ti.GetBestCommonType(
+					expressionsForTypeInference.SelectArray(e => e.ResolveResult),
+					out success);
+				if (success && commonType.GetStackType() == inst.ResultType)
+				{
+					return switchExpr.WithILInstruction(inst).WithRR(new ResolveResult(commonType));
+				}
+				else
+				{
+					// Cast all expressions
+					for (int i = 1; i < expressionsForTypeInference.Count; i++)
+					{
+						expressionsForTypeInference[i].Expression.ReplaceWith(
+							node => expressionsForTypeInference[i].ConvertTo(resultType, this)
+						);
+					}
+					return switchExpr.WithILInstruction(inst).WithRR(new ResolveResult(resultType));
+				}
+			}
 
 			Expression TranslateSectionBody(IL.SwitchSection section)
 			{
 				var body = Translate(section.Body, resultType);
-				return body.ConvertTo(resultType, this, allowImplicitConversion: true);
+				// Initially we allow implicit conversions for the body,
+				body = body.ConvertTo(resultType, this, allowImplicitConversion: true);
+				// but we may add explicit casts later if needed to satisfy the C# compiler.
+				if (body.Expression is not ThrowExpression)
+				{
+					expressionsForTypeInference.Add(body);
+				}
+				return body;
 			}
 		}
 
--- a/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
@@ -1272,7 +1272,8 @@ namespace ICSharpCode.Decompiler.CSharp
 
 		protected internal override TranslatedExpression VisitThrow(Throw inst, TranslationContext context)
 		{
-			return new ThrowExpression(Translate(inst.Argument))
+			var ex = Translate(inst.Argument, typeHint: compilation.FindType(KnownTypeCode.Exception));
+			return new ThrowExpression(ex)
 				.WithILInstruction(inst)
 				.WithRR(new ThrowResolveResult());
 		}
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/SwitchExpressions.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/SwitchExpressions.cs
@@ -37,10 +37,34 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 			Null
 		}
 
+		public abstract class Animal
+		{
+			public abstract string Name { get; }
+
+			public Animal ToValidated()
+			{
+				return this;
+			}
+		}
+
+		public class Dog : Animal
+		{
+			public override string Name => "Dog";
+		}
+
+		public class Fish : Animal
+		{
+			public override string Name => "Fish";
+		}
+
 		public static bool? SwitchOverNullableEnum(State? state)
 		{
+			// Note: we output a cast for the first element because
+			// {false,true,null} has no "best common type".
+			// (however C# doesn't require this cast due to target
+			//  typing, but the decompiler doesn't make use of that yet)
 			return state switch {
-				State.False => false,
+				State.False => (bool?)false,
 				State.True => true,
 				State.Null => null,
 				_ => throw new InvalidOperationException(),
@@ -209,6 +233,15 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 				StringComparison.OrdinalIgnoreCase => StringComparison.InvariantCulture,
 			};
 #endif
+		}
+
+		public static Animal Issue3683(int animalType)
+		{
+			return (animalType switch {
+				0 => (Animal)new Dog(),
+				1 => new Fish(),
+				_ => throw new ArgumentException("Invalid animal type"),
+			}).ToValidated();
 		}
 
 		public static int SwitchOnStringImplicitDefault(string s)
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/SwitchExpressions.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/SwitchExpressions.cs
@@ -244,6 +244,15 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 			}).ToValidated();
 		}
 
+		public static void ThrowDifferentExceptions(int i)
+		{
+			throw i switch {
+				0 => (Exception)new ArgumentException("Invalid argument"),
+				1 => new InvalidOperationException("Invalid operation"),
+				_ => new NotSupportedException(),
+			};
+		}
+
 		public static int SwitchOnStringImplicitDefault(string s)
 		{
 			return s switch {
--- a/ICSharpCode.Decompiler/CSharp/StatementBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/StatementBuilder.cs
@@ -434,7 +434,8 @@ namespace ICSharpCode.Decompiler.CSharp
 
 		protected internal override TranslatedStatement VisitThrow(Throw inst)
 		{
-			return new ThrowStatement(exprBuilder.Translate(inst.Argument)).WithILInstruction(inst);
+			var ex = exprBuilder.Translate(inst.Argument, typeHint: typeSystem.FindType(KnownTypeCode.Exception));
+			return new ThrowStatement(ex).WithILInstruction(inst);
 		}
 
 		protected internal override TranslatedStatement VisitRethrow(Rethrow inst)


ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -4370,12 +4372,51 @@ protected internal override TranslatedExpression VisitSwitchInstruction(SwitchIn
...
+				expressionsForTypeInference.SelectArray(e => e.ResolveResult),
+				out bool success);
+			if (success && commonType.GetStackType() == inst.ResultType)
+			{
+				return switchExpr.WithILInstruction(inst).WithRR(new ResolveResult(commonType));
+			}
+			else
+			{
+				// Try to help out the C# compiler by casting the first element to the expected type:
+				expressionsForTypeInference[0].Expression.ReplaceWith(
```
siegfriedpammer, line 4387: **`throw <switch>` no longer compiles.**

`throw` supplies no `context.TypeHint`, so `resultType` falls back to `compilation.FindType(inst.ResultType)` = `object`. Casting an arm to `object` gives the switch a natural type of `object`, which defeats the switch-expression conversion that `throw` was relying on.

```csharp
public enum Kind { A, B, C }

public static class ThrowCase
{
	public static void Verify(Kind k, bool ok)
	{
		if (!ok)
		{
			throw k switch {
				Kind.A => (Exception)new ArgumentException("a"),
				Kind.B => new InvalidOperationException("b"),
				_ => new NotSupportedException("c")
			};
		}
	}
}
```

master emits `Kind.A => new ArgumentException("a")` (compiles: target-typed to `System.Exception`). This PR emits `Kind.A => (object)new ArgumentException("a")`, and recompiling that gives:

```
error CS0029: Cannot implicitly convert type 'object' to 'System.Exception'
```

Same shape in the wild: `System.Configuration.ConfigurationManager` / `ClientConfigurationHost.VerifyDefinitionAllowed` (arms are `ConfigurationErrorsException` x3 plus `ExceptionUtil.UnexpectedError` returning `InvalidOperationException`, so no best common type exists).

The underlying issue is the cast *target*: `object` is the stack-type fallback, not the arms' common base type. `Exception` here would be both correct and readable.

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -4370,12 +4372,51 @@ protected internal override TranslatedExpression VisitSwitchInstruction(SwitchIn
...
 			}
 
-			return switchExpr.WithILInstruction(inst).WithRR(new ResolveResult(resultType));
+			var ti = new TypeInference(compilation, resolver.conversions);
+			IType commonType = ti.GetBestCommonType(
+				expressionsForTypeInference.SelectArray(e => e.ResolveResult),
+				out bool success);
+			if (success && commonType.GetStackType() == inst.ResultType)
+			{
+				return switchExpr.WithILInstruction(inst).WithRR(new ResolveResult(commonType));
```
siegfriedpammer: **Returning `commonType` inserts a narrowing cast around the whole switch.**

`GetStackType()` equality cannot distinguish `int` from `short` / `byte` / `ushort` / `sbyte`, so the guard passes with `commonType == int` while `resultType` is `short`, and the RR becomes `int`. Every consumer then has to narrow explicitly.

```csharp
public static short ShortSwitch(int i)
{
	return i switch { 0 => 1, 1 => 2, 2 => 4, _ => 3 };
}
```

```csharp
// master
return i switch { 0 => 1, 1 => 2, 2 => 4, _ => 3, };
// this PR
return (short)(i switch { 0 => 1, 1 => 2, 2 => 4, _ => 3, });
```

Real hit, `System.Private.Xml` / `System.Xml.BinXmlSqlDecimal..ctor`:

```csharp
// master
m_bLen = data[offset] switch { 7 => 1, 11 => 2, 15 => 3, 19 => 4, _ => throw ... };
// this PR
m_bLen = (byte)(data[offset] switch { 7 => 1, 11 => 2, 15 => 3, 19 => 4, _ => throw ... });
```

The sibling `?:` path runs `AdjustConstantExpressionToType(branch, otherType)` before inferring, which types the constant arms as `short` and avoids this.

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -4370,12 +4372,51 @@ protected internal override TranslatedExpression VisitSwitchInstruction(SwitchIn
 				switchExpr.SwitchSections.Add(defaultSES);
 			}
 
-			return switchExpr.WithILInstruction(inst).WithRR(new ResolveResult(resultType));
+			var ti = new TypeInference(compilation, resolver.conversions);
+			IType commonType = ti.GetBestCommonType(
```
siegfriedpammer, line 4377: **`GetBestCommonType` compares candidate `IType`s by identity, so types C# considers identical are counted as two candidates and `success` comes back `false`.**

`FindTypesInBounds` builds candidates from `lowerBounds.Union(upperBounds)`; two variants of the same type both survive the unique-candidate filter, and `Fix` returns `types.Count == 1` -> `false`. Two ways to hit it:

**1. Reference nullability.** Instrumented on `Microsoft.CodeAnalysis` / `ConstantValue.Value`, after stage 2 has cast arm 0:

```
resultType=System.Object common=System.Object success=False
arms=[Object/Nullable, null/Oblivious, Object/NotNullable x15, DateTime/Oblivious, String/Nullable]
```

The common type *is* `object`; only the annotations make it fail, and the run falls into stage 3, which casts four arms:

```csharp
public object? Value => Discriminator switch
{
	ConstantValueTypeDiscriminator.Bad => (object?)null,       // was: null
	ConstantValueTypeDiscriminator.Nothing => (object?)null,   // was: null
	...
	ConstantValueTypeDiscriminator.DateTime => (object?)DateTimeValue,  // was: DateTimeValue
	ConstantValueTypeDiscriminator.String => (object?)StringValue,      // was: StringValue
	_ => throw ExceptionUtilities.UnexpectedValue(Discriminator),
};
```

**2. `object` vs `dynamic`** - identity-convertible in C#, distinct `IType`s here:

```csharp
public static dynamic DynMix(int i, string s, int[] a, object o, dynamic d)
{
	return i switch { 0 => s, 1 => a, 2 => o, 3 => d, _ => null };
}
```

```csharp
// master
0 => s, 1 => a, 2 => o, 3 => d, _ => null,
// this PR
0 => (dynamic)s, 1 => (dynamic)a, 2 => (dynamic)o, 3 => d, _ => (dynamic)null,
```

That is a cast `ConvertTo` otherwise never produces - `(T)dynamicExpr` is a dynamic conversion, not a static one, so recompiling introduces dynamic conversions the IL did not contain.

Note this is the first `GetBestCommonType` caller whose output depends on `success`: the two existing callers (`ExpressionBuilder.cs:2736`, `CSharpResolver.cs:2834`) both discard it.

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -4370,12 +4372,51 @@ protected internal override TranslatedExpression VisitSwitchInstruction(SwitchIn
...
+				expressionsForTypeInference.SelectArray(e => e.ResolveResult),
+				out bool success);
+			if (success && commonType.GetStackType() == inst.ResultType)
```
siegfriedpammer, line 4380: **`success` is computed over a filtered arm set and is never re-validated, so `commonType` can be a type some arm does not convert to.**

`MakeOutputTypeInference` only adds a bound `if (IsValidType(e.Type))`, and `IsValidType` rejects `TypeKind.Null`. A `null` arm therefore contributes nothing, and for nullable *struct* arms the stack-type guard cannot catch it either (`S` and `S?` are both `StackType.O`):

```csharp
public struct S { public int V; public S(int v) { V = v; } }

public static bool HasVal(int i)
{
	return (i switch { 0 => new S(1), 1 => new S(2), _ => (S?)null }).HasValue;
}
```

Arms are `{S, S, null}`, `GetBestCommonType` returns `S` with `success == true`, the guard passes, and the switch is reported as `S` for an expression that evaluates to `S?`. No arm cast is inserted, so the CS8506 this PR targets is not prevented here - the output only happens to compile because the receiver conversion independently emits `(S?)`. `int?` / `bool?` / `DayOfWeek?` are caught only incidentally, by the `GetStackType()` mismatch.

`VisitNullCoalescingInstruction` (line 4076) and the `?:` path (line 4150) both guard `!x.Type.Equals(SpecialType.NullType)` explicitly before trusting an inferred type.

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -4370,12 +4372,51 @@ protected internal override TranslatedExpression VisitSwitchInstruction(SwitchIn
...
+			{
+				return switchExpr.WithILInstruction(inst).WithRR(new ResolveResult(commonType));
+			}
+			else
+			{
+				// Try to help out the C# compiler by casting the first element to the expected type:
```
siegfriedpammer: **The check is "does a best common type exist?", but C# applies the switch-expression conversion whenever every arm converts to the target - so casts are added in the target-typed contexts where C# needs none.**

Measured with `TestTools/decompdiff.cs`, master vs this PR over 202 real-world assemblies / 33006 types: 168 types changed, **0 new decompilation errors**, and **454 newly inserted arm casts**. Classified by the context the switch sits in:

| context | count |
|---|---|
| `return X switch` / `T x = X switch` / `=> X switch` (target-typed) | 417 |
| inside an existing cast, e.g. `return (Stream)(X switch ...)` | 31 |
| receiver / other | 6 |

C# target-types the first two groups, so those 448 casts are not needed. The second group also reads badly - the enclosing cast already supplies the target type:

```csharp
// Microsoft.CodeAnalysis.CSharp, QualifiedNameSyntax.GetSlot
return (GreenNode?)(index switch
{
	0 => (object)left,   // master: 0 => left
	1 => dotToken,
	2 => right,
	_ => null,
});
```

Same in `System.Formats.Asn1` (`(Encoding)(... => (object)s_utf8Encoding ...)`), `SharpCompress` (`(Stream)(... => (object)... )`), `RabbitMQ.Client` (`(Task<bool>)(... => (object)HandleBasicCancelAsync(...) ...)`).

The `#3683` shape (switch in receiver position) accounts for 6 of the 454. The fixture change at `SwitchExpressions.cs:67` is the same regression - `public static bool? SwitchOverNullableEnum(State? state)` compiles fine with `State.False => false`.

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -4370,12 +4372,51 @@ protected internal override TranslatedExpression VisitSwitchInstruction(SwitchIn
...
+				expressionsForTypeInference[0].Expression.ReplaceWith(
+					node => expressionsForTypeInference[0] = expressionsForTypeInference[0].ConvertTo(resultType, this)
```
siegfriedpammer: **The cast target is `resultType` verbatim, including its nullable-reference annotation, so `(object?)` lands in output that has no `#nullable` context.**

In a `<Nullable>disable</Nullable>` assembly the hint can still come from an annotated reference - `Console.WriteLine(object? value)`:

```csharp
public static void EnumIntMix(int i, DayOfWeek d)
{
	object o = i switch { 0 => (object)d, 1 => 1, _ => "x" };
	Console.WriteLine(o);
}
```

```csharp
// master
Console.WriteLine(i switch { 0 => d, 1 => 1, _ => "x", });
// this PR
Console.WriteLine(i switch { 0 => (object?)d, 1 => 1, _ => "x", });
```

Recompiling the decompiled file: `warning CS8632: The annotation for nullable reference types should only be used in code within a '#nullable' annotations context` - ILSpy emits no `#nullable` directive for a non-annotated assembly. 6 such warnings in my probe assembly; master produced none. Corpus hits include `(OpenXmlPart?)` and `(SyntaxNode?)`.

`NormalizeTypeVisitor.IgnoreNullability` on `resultType` before using it as a cast target would cover this (and feeds into the inference problem noted above).

siegfriedpammer (commented):
Review of the switch-expression cast insertion, reproduced against two locally built decompilers (this PR `ca24b8f3a` vs master `8997904b7`).

**This review has been cut down.** It originally carried 11 inline comments. Eight are removed: some were casts that only target typing would let you omit - which ILSpy does not implement, so they are expected and not defects of this change - and the rest were minor readability or allocation notes that did not belong in a PR discussion. @dgrunwald's replies to those threads are left in place. He was also right that the narrowing cast on `short`/`byte` results is correct as emitted, and that the nullable-annotation cast is not specific to switch expressions; both of those claims were mine and both were wrong.

Three findings remain, all of which produce output that is wrong rather than merely verbose:

1. **`throw <switch>` no longer compiles** (line 4386) - the cast target is the stack-type fallback `object`, which gives the switch a natural type and defeats the conversion `throw` needs. `error CS0029: Cannot implicitly convert type 'object' to 'System.Exception'`. Casting to the arms' common base instead would be correct without needing target typing.
2. **Inference counts identity-convertible types as distinct candidates** (line 4376) - reference nullability, and `object` vs `dynamic`. The `dynamic` case emits `(dynamic)x`, which is a dynamic conversion rather than a static one, so recompiling changes behaviour.
3. **`success` is not re-validated against the arms inference dropped** (line 4379) - a `null` arm contributes no bound, so `{S, S, null}` is typed `S`, and the #3683 class stays unfixed for nullable structs. Latent: I could not produce uncompilable output from it.

**Verification**: `ICSharpCode.Decompiler.Tests` on Linux is green (3502 tests, 0 failed, 45 skipped). `TestTools/decompdiff.cs` over 202 real-world assemblies / 33006 types: 168 types changed, **0 new decompilation errors**.

**What the PR fixes**, beyond #3683: `(i switch { 0 => 1, 1 => 2, _ => (int?)null }) ?? 9` currently trips a Debug assertion on master (`IL instructions of integer type must convert into C# expressions of integer type`, `StatementBuilder.VisitLeave`); a `List<Derived>` / `List<Base>` variance case where master's output fails with CS0029 now compiles; and a wrong overload selection with user-defined implicit conversions.

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
siegfriedpammer: **Stage 2 casts the arm that `TranslateSectionBody` already cast, so the two stack.**

`expressionsForTypeInference[0]` holds the *result* of `ConvertTo(resultType, this, allowImplicitConversion: true)` from line 4413, so the new cast wraps the existing one instead of replacing it.

```csharp
public static byte? NullableTarget(int i)
{
	return i switch { 0 => 1, 1 => 2, 2 => 4, _ => (byte?)null };
}
```

```csharp
// master
0 => (byte)1, 1 => (byte)2, 2 => (byte)4, _ => null,
// this PR
0 => (byte?)(byte)1, 1 => (byte)2, 2 => (byte)4, _ => null,
```

Also visible in `System.Linq.Expressions` as `TypeCode.Byte => (object)(byte)obj` (8 occurrences in one file) and `TypeCode.Boolean => (object)(bool)obj`.

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -4370,12 +4372,51 @@ protected internal override TranslatedExpression VisitSwitchInstruction(SwitchIn
...
+					expressionsForTypeInference.SelectArray(e => e.ResolveResult),
+					out success);
+				if (success && commonType.GetStackType() == inst.ResultType)
+				{
+					return switchExpr.WithILInstruction(inst).WithRR(new ResolveResult(commonType));
+				}
+				else
+				{
+					// Cast all expressions
+					for (int i = 1; i < expressionsForTypeInference.Count; i++)
```
siegfriedpammer: **The loop starts at `i = 1` on the assumption that stage 2 cast arm 0, but `ConvertTo` returns the node unchanged on several paths - and this whole branch has no coverage.**

`ConvertTo` is a no-op when `EquivalentTypes(type, targetType)`, and when `targetType.Kind` is `Void` / `None` / `Unknown`. `resultType` is `SpecialType.UnknownType` for `StackType.Unknown` (`ReflectionHelper.cs:47-57`), and `HandleSwitchExpression` accepts `StackType.Void` (it rejects only `StackType.Ref`). In those runs arms 1..n get cast and arm 0 does not, and line 4405 still reports `resultType` with no third inference run and no assert. `for (int i = 0; ...)` costs nothing - a second `ConvertTo` to the same type is a no-op.

Coverage: instrumented over the 202-assembly corpus, stage 1 fired 489 times, stage 2 78, stage 3 **103** (57 of those with `resultType=System.String`, 36 with an unknown type) - so it is live, but both new fixture cases (`Issue3683`, `SwitchOverNullableEnum`) are stage 2 and nothing in the suite reaches stage 3. The nullability and `dynamic` cases above both do.

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -4370,12 +4372,51 @@ protected internal override TranslatedExpression VisitSwitchInstruction(SwitchIn
 				switchExpr.SwitchSections.Add(defaultSES);
 			}
 
-			return switchExpr.WithILInstruction(inst).WithRR(new ResolveResult(resultType));
+			var ti = new TypeInference(compilation, resolver.conversions);
```
siegfriedpammer: **This relies on the default `TypeInferenceAlgorithm.CSharp4`, and nothing says so - the obvious cleanup reverts the fix with a green build.**

`ExpressionBuilder` already owns `internal readonly TypeInference typeInference` (line 84), constructed with `{ Algorithm = TypeInferenceAlgorithm.Improved }` (line 112), and the conditional-operator path uses it for the same question at line 4152. Reusing that field here changes the answer:

```
GetBestCommonType({Dog, Fish})   CSharp4:  success=False
                                 Improved: success=True -> Animal
```

With `Improved`, line 4379 takes the early return, no cast is emitted, and #3683 is unfixed - caught only by the new `Issue3683` fixture. `CSharp4` is the right choice here since it models what Roslyn accepts; worth stating it explicitly rather than inheriting it from a default.

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
dgrunwald, line 4382: The added cast is correct; only target typing (which we don't implement yet) allows omitting that.

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
dgrunwald, line 4388: that doesn't seem specific to switch-exprs

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -4370,12 +4372,51 @@ protected internal override TranslatedExpression VisitSwitchInstruction(SwitchIn
...
+				{
+					return switchExpr.WithILInstruction(inst).WithRR(new ResolveResult(commonType));
+				}
+				else
+				{
+					// Cast all expressions
+					for (int i = 1; i < expressionsForTypeInference.Count; i++)
+					{
+						expressionsForTypeInference[i].Expression.ReplaceWith(
+							node => expressionsForTypeInference[i].ConvertTo(resultType, this)
```
dgrunwald, line 4403: I hate how AI poisons every PR discussion with completely unrelated minor readability "bugs".
(not to mention that it's recommendations here are utter bullshit)

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
dgrunwald, line 4400: So what? If the first arm 0 cast didn't do anything; the second attempt won't do anything either. Comments like these are a waste of time.

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -4342,6 +4342,8 @@ protected internal override TranslatedExpression VisitSwitchInstruction(SwitchIn
 				resultType = compilation.FindType(inst.ResultType);
 			}
 
+			var expressionsForTypeInference = new List<TranslatedExpression>();
```
dgrunwald, line 4346: Of the millions of decompiler allocations, these 175 are what AI worries about... utterly irrelevant.

siegfriedpammer:
I deleted eight of my twelve inline comments on this PR. Five of them had replies, and deleting a parent comment orphans its replies — GitHub clears their `in_reply_to` and promotes them to standalone comments — so those five replies now sit on their lines with nothing to point at. That was my mistake, not theirs. Each was answering a specific comment of mine:

- **line 4381** — *"The added cast is correct; only target typing (which we don't implement yet) allows omitting that."* Answered a claim that the `(short)` cast around the whole switch was a regression. It was not: the previous output compiled only because C# target-types the switch expression.
- **line 4387** — *"that doesn't seem specific to switch-exprs"* Answered a claim about the nullable annotation on the emitted cast (`(object?)`, CS8632 in non-nullable output).
- **line 4399** — *"So what? If the first arm 0 cast didn't do anything; the second attempt won't do anything either."* Answered a claim that the `for (int i = 1; ...)` loop skipped a cast arm 0 needed.
- **line 4402** — *"I hate how AI poisons every PR discussion with completely unrelated minor readability 'bugs'."* Answered a claim that the fixup casts fragment `checked` regions.
- **line 4345** — *"Of the millions of decompiler allocations, these 175 are what AI worries about."* Answered a note about the closure allocation in `VisitSwitchInstruction`.

All five were right, and the comments they were answering are gone because of them. The other three deletions had no replies. The three inline comments still open are the ones I did not withdraw.

PR #4068, opened 2026-08-28
Type inference for switch-expressions

dgrunwald:
Follow-up to #4065.
This fixes the object/dynamic issue and the "commonType can be a type some arm does not convert to" issue, both found by Claude.

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
@@ -4377,7 +4377,9 @@ namespace ICSharpCode.Decompiler.CSharp
 			IType commonType = ti.GetBestCommonType(
 				expressionsForTypeInference.SelectArray(e => e.ResolveResult),
 				out bool success);
-			if (success && commonType.GetStackType() == inst.ResultType)
+			// Note: we need to ensure the compiler actually picked the type that we used for the
+			// implicit conversions.
+			if (success && NormalizeTypeVisitor.TypeErasure.EquivalentTypes(commonType, resultType))
 			{
 				return switchExpr.WithILInstruction(inst).WithRR(new ResolveResult(commonType));
 			}
@@ -4390,7 +4392,7 @@ namespace ICSharpCode.Decompiler.CSharp
 				commonType = ti.GetBestCommonType(
 					expressionsForTypeInference.SelectArray(e => e.ResolveResult),
 					out success);
-				if (success && commonType.GetStackType() == inst.ResultType)
+				if (success && NormalizeTypeVisitor.TypeErasure.EquivalentTypes(commonType, resultType))
 				{
 					return switchExpr.WithILInstruction(inst).WithRR(new ResolveResult(commonType));
 				}
--- a/ICSharpCode.Decompiler/CSharp/Resolver/TypeInference.cs
+++ b/ICSharpCode.Decompiler/CSharp/Resolver/TypeInference.cs
@@ -251,7 +251,7 @@ namespace ICSharpCode.Decompiler.CSharp.Resolver
 					return;
 				// Two exact bounds that differ only in tuple element names are not conflicting;
 				// their names are merged instead (kept where both agree, dropped otherwise).
-				IType merged = MergeTupleNames(ExactBound, type);
+				IType merged = MergeSimilarTypes(ExactBound, type);
 				if (merged != null)
 					ExactBound = merged;
 				else
@@ -987,7 +987,7 @@ namespace ICSharpCode.Decompiler.CSharp.Resolver
 				// MergeTupleNames in Roslyn's MethodTypeInference.cs.
 				IType fixedTo = tp.ExactBound;
 				foreach (var b in tp.LowerBounds.Concat(tp.UpperBounds))
-					fixedTo = MergeTupleNames(fixedTo, b) ?? fixedTo;
+					fixedTo = MergeSimilarTypes(fixedTo, b) ?? fixedTo;
 				// the exact bound determines the result, up to the merged element names
 				tp.FixedTo = fixedTo;
 				// check validity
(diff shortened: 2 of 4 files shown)


siegfriedpammer (approved):

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -4377,7 +4377,9 @@ protected internal override TranslatedExpression VisitSwitchInstruction(SwitchIn
 			IType commonType = ti.GetBestCommonType(
 				expressionsForTypeInference.SelectArray(e => e.ResolveResult),
 				out bool success);
-			if (success && commonType.GetStackType() == inst.ResultType)
+			// Note: we need to ensure the compiler actually picked the type that we used for the
+			// implicit conversions.
+			if (success && NormalizeTypeVisitor.TypeErasure.EquivalentTypes(commonType, resultType))
```
siegfriedpammer, line 4382: This rejects every `commonType` that is a proper subtype of `resultType`, including when the conversion between them is identity/reference/boxing. Where `resultType` is the stack-type fallback `object` and the arms are tuples, the result is not equivalent to the input.

Repro (net10.0, Release):

```csharp
public string JsonPath(int continuationCount)
{
    StringBuilder sb = new StringBuilder("$");
    (int frameCount, bool includeCurrent) = continuationCount switch {
        0 => (_count - 1, true),
        1 => (0, true),
        _ => (continuationCount, false)
    };
    for (int i = 0; i < frameCount; i++) sb.Append(i);
    if (includeCurrent) sb.Append('c');
    return sb.ToString();
}
```

master `2d9f852cd`:

```csharp
(int, bool) obj = continuationCount switch { ... };
int item = obj.Item1;
bool item2 = obj.Item2;
```

this branch `5d9cdce94`:

```csharp
public unsafe string JsonPath(int continuationCount)
{
    object obj = continuationCount switch { 0 => (object)(_count - 1, true), ... };
    int item = ((ValueTuple<int, bool>*)(&obj))->Item1;
    bool item2 = ((ValueTuple<int, bool>*)(&obj))->Item2;
```

`&obj` is the address of the reference, not of the boxed payload. Since C# 11 that is warning CS8500 rather than CS0208, so it still compiles. Running the original, master's output and the branch's output on the same input (`_count = 5`, `JsonPath(0)`):

```
original: $0123c
master:   $0123c
branch:   THREW OutOfMemoryException
```

Decompiling `Microsoft.NETCore.App/10.0.11` with both builds, the same shape hits `System.Text.Json.JsonPath()` and `PropertyPath()` and `System.Private.CoreLib.LCIDToLocaleName` (all three also gain `unsafe`). Redundant-cast-only hits: `System.Net.Http` (`EventCounter obj = ...` becomes `object obj = ...` plus `((EventCounter)obj)?.WriteMetric(...)`), `System.Linq.Expressions`, `System.Private.Xml`, and `Newtonsoft.Json 13.0.4`.

`TranslateSectionBody` has already run `ConvertTo(resultType, allowImplicitConversion: true)` on every arm, so a `commonType` that differs from `resultType` can only differ by a conversion the arms never had to spell out.

What I verified locally: keeping the old stack-type guard and adding a conversion check, instead of exact equality -

```csharp
if (NormalizeTypeVisitor.TypeErasure.EquivalentTypes(commonType, resultType))
    return true;
if (commonType.GetStackType() != inst.ResultType)
    return false;
var c = resolver.conversions.ImplicitConversion(commonType, resultType);
return c.IsIdentityConversion || c.IsReferenceConversion || c.IsBoxingConversion;
```

removes every regression above, keeps this branch's intended improvement in `System.Private.Xml` (`m_bLen = (byte)(x switch { 7 => 1, ... })` becomes `m_bLen = x switch { 7 => (byte)1, ... }`), and leaves the suite green. The `GetStackType()` guard is load-bearing - without it `System.Private.Xml` trips `Debug.Assert(cexpr.Type.GetStackType() == inst.ResultType)` at ExpressionBuilder.cs:205.

ICSharpCode.Decompiler/CSharp/Resolver/TypeInference.cs
```diff
@@ -1095,7 +1104,7 @@ static IType MergeTupleNames(IType a, IType b)
 				bool absorbed = false;
 				for (int i = 0; i < mergedBounds.Count && !absorbed; i++)
 				{
-					IType merged = MergeTupleNames(mergedBounds[i], bound);
+					IType merged = MergeSimilarTypes(mergedBounds[i], bound);
```
siegfriedpammer, line 1107: Five lines below this, `anyNamesMerged |= !ReferenceEquals(merged, mergedBounds[i])` is a proxy that only holds while merging always allocates. `MergeSimilarTypes(dynamic, object)` returns `a`, i.e. the entry already in the list, so when `dynamic` is the first bound and `object` a later one the flag stays `false` and the method returns the unmerged bounds - the merge is dropped.

`BestCommonTypeObjectAndDynamic` covers only the `(object, dynamic)` order. The reverse fails on this branch:

```csharp
[Test]
public void BestCommonTypeDynamicAndObject_Reversed()
{
    Assert.That(
        ti.GetBestCommonType(new[] {
            new ResolveResult(SpecialType.Dynamic),
            new ResolveResult(compilation.FindType(KnownTypeCode.Object))
        }, out bool success),
        Is.EqualTo(SpecialType.Dynamic));
    Assert.That(success);   // Expected: True  But was: False
}
```

Visible in the output as well:

```csharp
static void MixedConsumed(int i, object o, dynamic d)
{ dynamic x = i switch { 0 => o, 1 => d, _ => o }; x.Foo(); }

static void MixedConsumedDynFirst(int i, object o, dynamic d)
{ dynamic x = i switch { 0 => d, 1 => o, _ => o }; x.Foo(); }
```

The first loses its `(dynamic)o` casts on this branch; the second is unchanged from master. Only top-level `dynamic` bounds are affected - `List<dynamic>` vs `List<object>` goes through the `ParameterizedType` branch, which allocates, so the flag is set and the merge sticks.

`anyNamesMerged |= !ReferenceEquals(merged, mergedBounds[i]) || !bound.Equals(mergedBounds[i]);` makes the reversed test pass and drops the cast in the dynamic-first case, with the eight `BestCommonType*` tests still green.

siegfriedpammer (commented):
Two findings from reviewing this against master (`2d9f852cd`), both reproduced with locally built decompilers on Linux. `ICSharpCode.Decompiler.Tests` is green on the branch (3505 total, 0 failed, 45 skipped).

1. The new `EquivalentTypes(commonType, resultType)` check changes behaviour for tuple-typed switch expressions - `System.Text.Json.JsonPath()` and `System.Private.CoreLib.LCIDToLocaleName` among the hits.
2. The object/dynamic merge is silently skipped when `dynamic` is the first bound.

Details inline.

PR #4072, opened 2026-08-29
Fix #3704: keep the static type of a dynamic call target

### LINKED ISSUE #3704: dynamic issue
[dynamic.zip](https://github.com/user-attachments/files/26419078/dynamic.zip)

```C#
public static bool TestSingleKeyPairComplete(object viewModel, string validationKeyHex, string decryptionKeyHex, string viewStateB64, string modifier)
{
  ......................
			CompleteIntegration.WriteToRichTextBox((dynamic)viewModel, "[+] COMPLETE SUCCESS!", Colors.Green);
			CompleteIntegration.WriteToRichTextBox((dynamic)viewModel, "    Validation Algorithm: " + tuple.Value.Item1, Colors.Green);
			CompleteIntegration.WriteToRichTextBox((dynamic)viewModel, "    Decryption Algorithm: " + tuple.Value.Item2, Colors.Green);
			CompleteIntegration.WriteToRichTextBox((dynamic)viewModel, "    .NET Mode: " + tuple.Value.Item3, Colors.Green);
			Type typeFromHandle = typeof(CompleteIntegration);
			byte[] item = tuple.Value.Item4;
			typeFromHandle.WriteToRichTextBox((dynamic)viewModel, $"    Decrypted Data Length: {((item != null) ? item.Length : 0)} bytes", Colors.Green);
```

See last line, should be `CompleteIntegration.WriteToRichTextBox((dynamic)viewModel, $"    Decrypted Data Length: {((item != null) ? item.Length : 0)} bytes", Colors.Green)` ?

siegfriedpammer:
Fixes #3704.

A dynamic call site that names a type passes `typeof(T)` as its target. When a later argument contains control flow, the compiler stores that `typeof` in a temporary, and `ILInlining` does not undo it: it inlines a store only into the immediately following instruction, so an intervening statement leaves the `typeof` behind. `ExpressionBuilder` then printed the temporary instead of the type, producing output that does not compile:

```csharp
Type typeFromHandle = typeof(CompleteIntegration);
byte[] item = tuple.Value.Item4;
typeFromHandle.WriteToRichTextBox((dynamic)viewModel, $"...{((item != null) ? item.Length : 0)}...", Colors.Green);
```

Substituting the `typeof` back into the target slot is not an option: a call there has side effects, so the inliner can no longer move the remaining arguments across it and results of dynamic operations stop collapsing into their consumer (18 `DynamicTests` configurations regress that way).

The type is stored on the instruction instead. `DynamicInvokeConstructorInstruction` already resolved it that way but kept the field private, so `ExpressionBuilder` re-derived it from the target argument and degraded to an error expression on the spilled form. `DynamicInvokeMemberInstruction` gets the same field. Both drop the target from `Arguments`, which lets the existing dead-argument bookkeeping retire the store; `ArgumentInfo` keeps its entry for the target, because the invocation symbol is built from it.

Those two are the only binder method kinds that ever see `CSharpArgumentInfoFlags.IsStaticType` - every other way of naming a type as the receiver binds statically and leaves only a conversion of the dynamic operand behind.

Tests: `DynamicTests` covers a discarded result, a used result, explicit type arguments and a dynamic object creation, each with the `typeof` stranded behind a surviving temporary; `DynamicAwait` covers the state-machine spill across an await.

Full `ICSharpCode.Decompiler.Tests` run: 3460 passed, 0 failed, 45 skipped.

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
@@ -4573,9 +4573,8 @@ namespace ICSharpCode.Decompiler.CSharp
 
 		protected internal override TranslatedExpression VisitDynamicInvokeConstructorInstruction(DynamicInvokeConstructorInstruction inst, TranslationContext context)
 		{
-			if (!(inst.ArgumentInfo[0].HasFlag(CSharpArgumentInfoFlags.IsStaticType) && IL.Transforms.TransformExpressionTrees.MatchGetTypeFromHandle(inst.Arguments[0], out var constructorType)))
-				return ErrorExpression("Could not detect static type for DynamicInvokeConstructorInstruction");
-			var arguments = TranslateDynamicArguments(inst.Arguments.Skip(1), inst.ArgumentInfo.Skip(1)).ToList();
+			var constructorType = inst.Type;
+			var arguments = TranslateDynamicArguments(inst.Arguments, inst.ArgumentInfo.Skip(1)).ToList();
 			var constructor = CreateDynamicConstructorSymbol(constructorType, inst.ArgumentInfo.Skip(1).ToArray());
 			return new ObjectCreateExpression(ConvertType(constructorType), arguments.Select(a => a.Expression))
 				.WithILInstruction(inst)
@@ -4585,7 +4584,11 @@ namespace ICSharpCode.Decompiler.CSharp
 		protected internal override TranslatedExpression VisitDynamicInvokeMemberInstruction(DynamicInvokeMemberInstruction inst, TranslationContext context)
 		{
 			Expression targetExpr;
-			var target = TranslateDynamicTarget(inst.Arguments[0], inst.ArgumentInfo[0]);
+			var target = inst.StaticTargetType is IType staticTargetType
+				? new TypeReferenceExpression(ConvertType(staticTargetType))
+					.WithoutILInstruction()
+					.WithRR(new TypeResolveResult(staticTargetType))
+				: TranslateDynamicTarget(inst.Arguments[0], inst.ArgumentInfo[0]);
 			if (inst.BinderFlags.HasFlag(CSharpBinderFlags.InvokeSimpleName) && target.Expression is ThisReferenceExpression)
 			{
 				targetExpr = new IdentifierExpression(inst.Name);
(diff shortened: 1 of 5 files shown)


ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -4569,9 +4569,9 @@ protected internal override TranslatedExpression VisitDynamicGetMemberInstructio
 
 		protected internal override TranslatedExpression VisitDynamicInvokeConstructorInstruction(DynamicInvokeConstructorInstruction inst, TranslationContext context)
 		{
-			if (!(inst.ArgumentInfo[0].HasFlag(CSharpArgumentInfoFlags.IsStaticType) && IL.Transforms.TransformExpressionTrees.MatchGetTypeFromHandle(inst.Arguments[0], out var constructorType)))
+			if (inst.Type is not IType constructorType)
```
dgrunwald: Weird way of writing a null check

ICSharpCode.Decompiler/IL/Instructions/DynamicInstructions.cs
```diff
@@ -356,17 +373,21 @@ public override CSharpArgumentInfo GetArgumentInfoOfChild(int index)
...
 	{
-		readonly IType? resultType;
-
 		public IReadOnlyList<CSharpArgumentInfo> ArgumentInfo { get; }
 
+		/// <summary>
+		/// The type being constructed. The target is never in <see cref="Arguments"/>, but ArgumentInfo[0] still
+		/// describes it.
+		/// </summary>
+		public readonly IType? Type;
```
dgrunwald: Should this be nullable? I think the transform now always provides a type.

dgrunwald (approved):

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -4585,7 +4584,11 @@ protected internal override TranslatedExpression VisitDynamicInvokeConstructorIn
 		protected internal override TranslatedExpression VisitDynamicInvokeMemberInstruction(DynamicInvokeMemberInstruction inst, TranslationContext context)
 		{
 			Expression targetExpr;
-			var target = TranslateDynamicTarget(inst.Arguments[0], inst.ArgumentInfo[0]);
+			var target = inst.StaticTargetType is IType staticTargetType
```
dgrunwald, line 4587: This is another weird null check.

dgrunwald (approved):

PR #4077, opened 2026-08-29
Improvements for type inference

dgrunwald:
The deduplication of equivalent types is changed to an algorithm much closer to what Roslyn is doing:
 * uses a dictionary instead of quadratic merging step
 * no need to split the merged candidates back into lower/upper lists
 * supports merging types with different nullability

### DIFF
--- a/ICSharpCode.Decompiler.Tests/Semantics/TypeInferenceTests.cs
+++ b/ICSharpCode.Decompiler.Tests/Semantics/TypeInferenceTests.cs
@@ -1258,13 +1258,25 @@ namespace ICSharpCode.Decompiler.Tests.Semantics
 		}
 
 		[Test]
+		public void BestCommonTypeObjectAndNullableString()
+		{
+			Assert.That(
+				ti.GetBestCommonType([
+					new ResolveResult(compilation.FindType(KnownTypeCode.Object).ChangeNullability(Nullability.NotNullable)),
+					new ResolveResult(compilation.FindType(KnownTypeCode.String).ChangeNullability(Nullability.Nullable))
+				], out bool success),
+				Is.EqualTo(compilation.FindType(KnownTypeCode.Object).ChangeNullability(Nullability.Nullable)));
+			Assert.That(success);
+		}
+
+		[Test]
 		public void BestCommonTypeObjectAndNullableObject()
 		{
 			Assert.That(
-				ti.GetBestCommonType(new[] {
+				ti.GetBestCommonType([
 					new ResolveResult(compilation.FindType(KnownTypeCode.Object).ChangeNullability(Nullability.NotNullable)),
 					new ResolveResult(compilation.FindType(KnownTypeCode.Object).ChangeNullability(Nullability.Nullable))
-				}, out bool success),
+				], out bool success),
 				Is.EqualTo(compilation.FindType(KnownTypeCode.Object).ChangeNullability(Nullability.Nullable)));
 			Assert.That(success);
 		}
--- a/ICSharpCode.Decompiler.Tests/Semantics/TypeInferenceTests.cs
+++ b/ICSharpCode.Decompiler.Tests/Semantics/TypeInferenceTests.cs
@@ -23,6 +23,7 @@ using System.Collections.Immutable;
 using System.Collections.ObjectModel;
 using System.IO;
 using System.Linq;
+using System.Reflection.Metadata;
 
 using ICSharpCode.Decompiler.CSharp.Resolver;
 using ICSharpCode.Decompiler.Metadata;
@@ -713,6 +714,15 @@ namespace ICSharpCode.Decompiler.Tests.Semantics
 				ImmutableArray.CreateRange(elementNames));
 		}
 
+		FunctionPointerType MakeFunctionPointerType(ICompilation comp, IType returnType)
+		{
+			return new FunctionPointerType(
+				(MetadataModule)comp.MainModule,
+				SignatureCallingConvention.Default, ImmutableArray<IType>.Empty,
+				returnType, returnIsRefReadOnly: false,
+				ImmutableArray<IType>.Empty, ImmutableArray<ReferenceKind>.Empty);
+		}
+
 		[Test]
 		public void BestCommonTypeMergesTupleElementNames()
 		{
@@ -727,6 +737,21 @@ namespace ICSharpCode.Decompiler.Tests.Semantics
 					new ResolveResult(MakeTupleType(comp, "a", "c"))
 				}, out success),
 				Is.EqualTo(MakeTupleType(comp, "a", null)));
+			Assert.That(success);
+		}
+
+		[Test]
+		public void BestCommonTypeMergesFunctionPointerTupleElementNames()
+		{
+			var comp = RefAssemblyCompilation.Instance;
+			var inference = new TypeInference(comp);
+
+			Assert.That(
+				inference.GetBestCommonType(new[] {
+					new ResolveResult(MakeFunctionPointerType(comp, MakeTupleType(comp, "a", "b"))),
+					new ResolveResult(MakeFunctionPointerType(comp, MakeTupleType(comp, "a", "c")))
+				}, out bool success),
+				Is.EqualTo(MakeFunctionPointerType(comp, MakeTupleType(comp, "a", null))));
 			Assert.That(success);
 		}
 
--- a/ICSharpCode.Decompiler.Tests/Semantics/TypeInferenceTests.cs
+++ b/ICSharpCode.Decompiler.Tests/Semantics/TypeInferenceTests.cs
@@ -854,25 +854,24 @@ namespace ICSharpCode.Decompiler.Tests.Semantics
 		}
 
 		[Test]
-		public void FixingDoesNotMergeBoundsThatDifferInNullability()
+		public void FixingMergesBoundsThatDifferInNullability()
 		{
 			// Signature:  M<T>(T x, T y)
 			// Invocation: M(nullableArrayOfString, arrayOfString);
-			// Merging nullability requires the variance of the position, which this
-			// implementation does not track, so such bounds stay distinct and fixing fails
-			// (csc infers string[]?).
+			// Merging nullability in this covariant position should result in T=string[]? (the nullable array type).
 			var comp = RefAssemblyCompilation.Instance;
 			var T = new DefaultTypeParameter(comp, SymbolKind.Method, 0, "T");
 			IType stringType = comp.FindType(KnownTypeCode.String);
 
-			new TypeInference(comp).InferTypeArguments(new ITypeParameter[] { T },
-				new[] {
+			var result = new TypeInference(comp).InferTypeArguments([T],
+				[
 					new ResolveResult(new ArrayType(comp, stringType, 1, Nullability.Nullable)),
 					new ResolveResult(new ArrayType(comp, stringType))
-				},
-				new IType[] { T, T },
+				],
+				[T, T],
 				out bool success);
-			Assert.That(success, Is.False);
+			Assert.That(success, Is.True);
+			Assert.That(result, Is.EqualTo([new ArrayType(comp, stringType, 1, Nullability.Nullable)]));
 		}
 
 		[Test]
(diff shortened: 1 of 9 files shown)


ICSharpCode.Decompiler/CSharp/Resolver/TypeInference.cs
```diff
@@ -1210,8 +1221,57 @@ IReadOnlyList<IType> FindTypesInBounds(IReadOnlyList<IType> lowerBounds, IReadOn
...
+					{
+						var merged = MergeSimilarTypes(existing, bound, variance);
+						Log.WriteLine(" Merged similar types " + existing + " and " + bound + " into " + merged);
+						if (merged != null)
+						{
+							candidateMergeDict[key] = merged;
+						}
+						else
+						{
+							Debug.Fail("MergeSimilarTypes should always be able to merge;"
```
siegfriedpammer, line 1308: **Blocking, reproduced.** This assert is reachable, and a Debug build dies on it.

`KeyForTypeMerging` erases object/dynamic and nullability at *every* nesting level (`VisitChildren` recurses through all type constructors), but `MergeSimilarTypes` only rebuilds through `NullabilityAnnotatedType`, `TupleType`, `ParameterizedType`, `ArrayType` and `PointerType`. `ByReferenceType`, `ModifiedType` and `FunctionPointerType` collide on one key and cannot merge.

Repro:

```csharp
public class C
{
    public object objField;
    public dynamic dynField;
    public static void Use(ref object o) { }
    public void N(bool c) { Use(ref (c ? ref objField : ref dynField)); }
}
```

| build | `ilspycmd` on that assembly |
| --- | --- |
| this PR, Debug | `Process terminated. Assertion failed. MergeSimilarTypes should always be able to merge...` |
| `5d9cdce94`, Debug | decompiles fine |
| this PR, Release | decompiles fine, output identical to base |

Path: `ExpressionBuilder.VisitIfInstruction:4153` -> `GetBestCommonType` -> `Fix:1009` -> `FindTypesInBounds:1268` -> `AddCandidates:1258`.

Because `Debug.Fail` is `[Conditional("DEBUG")]`, Release drops the `else` branch and silently keeps only the first bound, so Debug and Release disagree on the candidate set.

Also reproduced directly through `FindTypeInBounds`: `{ref object, ref dynamic}` (both keys `dynamic&`) and `{modopt(IDisposable) object, modopt(IDisposable) dynamic}` (both keys `dynamic modopt(System.IDisposable)`). `FunctionPointerType` is the same gap by inspection - I did not construct one.

Widening `MergeSimilarTypes` case by case leaves the same trap open for the next type constructor; keeping the key visitor and the merger in lockstep, or falling back to a key comparison instead of asserting, closes it.

*Written by Claude Code (Opus 5) on my behalf; I have checked the reproductions.*

siegfriedpammer (commented):
Reviewed at `3564e61d4` vs merge base `5d9cdce94`. Full `ICSharpCode.Decompiler.Tests` on the branch is green (3463 passed, 0 failed, 45 skipped; `ILSpy-tests` absent so round-trip did not run).

One blocking finding: a Debug build aborts on a `ref object` / `ref dynamic` conditional that the base decompiles fine. The rest are latent or cosmetic; each comment says which.

I checked and cleared these, so they don't need a second look: the `MergeNullability` table is correct in all 27 cells against Roslyn's `Join`/`Meet`/`EnsureCompatible`, and `Combine` faithfully reimplements `GetTypeArgumentVariance`; `Debug.Assert(topLevelNullability.HasValue)` always holds; dropping `pa.GenericType.Equals(pb.GenericType)` is safe (the `ParameterizedType` ctors enforce arity); `MoveToImmutable()` is filled to capacity; no value-type candidate can gain a `?` (`ChangeNullability` returns `this` when `IsReferenceType == false`).

*This review was produced by Claude Code (Opus 5) on my behalf; I have NOT checked the reproductions.*

ICSharpCode.Decompiler/CSharp/Resolver/TypeInference.cs
```diff
@@ -1224,6 +1284,15 @@ IReadOnlyList<IType> FindTypesInBounds(IReadOnlyList<IType> lowerBounds, IReadOn
 				c => candidateTypes.All(o => conversions.ImplicitConversion(o, c).IsValid)
 			).ToList();
 
+			// Apply the merged top-level nullability:
+			Debug.Assert(topLevelNullability.HasValue);
+			// (Roslyn has a different approach in MergeOrRemoveCandidates, but to me that just looked
+			//  like an overly complicated way of achieving the same thing.)
+			for (int i = 0; i < candidateTypes.Count; i++)
+			{
+				candidateTypes[i] = candidateTypes[i].ChangeNullability(topLevelNullability.Value);
```
siegfriedpammer, line 1346: **Latent, measured.** This stamp is discarded on the path the decompiler actually takes.

`candidateTypes.Clear()` a few lines below throws the stamped list away whenever the spec-fixing step did not leave exactly one candidate and the algorithm is `Improved` - which is what `ExpressionBuilder.cs:112` uses. The improved algorithm then rebuilds candidates from `GetAllBaseTypeDefinitions()`/`GetAllTypeDefinitions()` and never re-applies it.

Measured through `FindTypeInBounds`:

| lower bounds | result |
| --- | --- |
| `{string?, Exception}` | `object`, `Oblivious` - accumulator computed `Nullable`, `Clear()` dropped it |
| `{string?, object}` | `object?` - spec path, kept |
| `{string?, string}` | `string?` - spec path, kept |

Output impact is annotation-only today (it surfaces as the printed cast in the conditional-expression path), so this is not urgent - but the same bounds yield an annotated or unannotated result depending purely on which exit produced the winner. Moving the loop to just before each `return candidateTypes;` would settle it.

*Written by Claude Code (Opus 5) on my behalf; I have checked the reproductions.*

ICSharpCode.Decompiler/CSharp/Resolver/TypeInference.cs
```diff
@@ -1210,8 +1221,57 @@ IReadOnlyList<IType> FindTypesInBounds(IReadOnlyList<IType> lowerBounds, IReadOn
...
+			{
+				// This helper function works like Roslyn's MethodTypeInference.AddAllCandidates().
+				// It deduplicates similar types and merges them into a single candidate type,
+				// handling differences in nullability, tuple element names, and object/dynamic,
+				// but only if the type is otherwise completely identical.
+				foreach (var bound in bounds)
+				{
+					if (topLevelNullability.HasValue)
+					{
+						topLevelNullability = MergeNullability(topLevelNullability.Value, bound.Nullability, variance);
```
siegfriedpammer, line 1291: **Latent, but the comment below it overstates the equivalence.**

One accumulator is folded left over the lower bounds (covariant) and then the upper bounds (contravariant), then stamped over the winner's own annotation. Roslyn merges per candidate: `AddOrMergeCandidate` only merges annotations between candidates its `EqualsIgnoringDynamicTupleNamesAndNullabilityComparer` calls equal, and `MergeOrRemoveCandidates` folds a removed candidate's annotation only into the shape-equivalent survivor.

Measured, with csc ground truth from a `SemanticModel` probe on `MLU<T>(T lower, Action<T> upper)`:

| bounds | this PR | csc |
| --- | --- | --- |
| lower `{string!}`, upper `{object?}` | `object!` | `object?` |
| lower `{string^oblivious}`, upper `{object?}` | `object` (`?` dropped) | - |

Reachable as `static void M<T>(T x, Action<T> a)` called with `M("s", (object? o) => {})`; in practice `IEnumerable<out T>` plus `Action<in T>`. Lower-bounds-only inference (`GetBestCommonType`) matches Roslyn, so only a type parameter with both a lower and an upper bound diverges.

It cannot flip the explicit-type-argument decision today, since `IsAppropriateCallTarget` compares under `TypeErasure`. The concrete cost is the comment calling `MergeOrRemoveCandidates` "an overly complicated way of achieving the same thing" - a two-bound case disproves that, and the next reader will trust it.

*Written by Claude Code (Opus 5) on my behalf; I have checked the reproductions.*

ICSharpCode.Decompiler/CSharp/Transforms/IntroduceQueryExpressions.cs
```diff
@@ -142,6 +142,8 @@ bool CanUseDiscardAssignment()
 			MemberReferenceExpression? mre = invocation.Target as MemberReferenceExpression;
 			if (mre == null || IsNullConditional(mre.Target))
 				return null;
+			if (mre.TypeArguments.Count > 0)
```
siegfriedpammer, line 145: **Guess - I could not reproduce this, flagging the mechanism only.**

The guard sees only the outermost call. The OrderBy/ThenBy case collapses the chain in its own `while` loop below, reassigning `mre = (MemberReferenceExpression)tmp.Target` with no `TypeArguments` re-check, and `ValidateThenByChain` - which already re-checks `IsNullConditional` per level - was not given the matching check. So for `src.OrderBy<TSource, TKey>(x => k1).ThenBy(x => k2)` the outer `ThenBy` passes the guard and the inner `OrderBy` is consumed without `DecompileQuery` ever seeing its type arguments.

I tried to make it bite: adding `from s in strings where s != null orderby s.Length, s select s.ToUpper()` over `IEnumerable<string?>` to the `NullableRefTypes` fixture round-trips with query syntax preserved, because inference succeeds there and no type arguments are emitted. I have no input that forces type arguments onto an inner `OrderBy` while leaving the outer `ThenBy` inferable, so this does not meet the "output is wrong" bar on its own.

Raising it only because the guard and `ValidateThenByChain` now disagree about which per-level checks the chain needs, and the guard has no fixture of its own - no Pretty expectation carries explicit type arguments on a LINQ method, so nothing in the suite would catch the reachable case if it appears.

*Written by Claude Code (Opus 5) on my behalf; I have checked the reproductions.*

ICSharpCode.Decompiler/Semantics/LambdaResolveResult.cs
```diff
@@ -165,7 +165,7 @@ public override Conversion IsValid(IType[] parameterTypes, IType returnType, CSh
 					return Conversion.None;
 				for (int i = 0; i < parameterTypes.Length; ++i)
 				{
-					if (parameterTypes[i].Equals(this.Parameters[i].Type))
+					if (NormalizeTypeVisitor.IgnoreNullability.EquivalentTypes(parameterTypes[i], this.Parameters[i].Type))
```
siegfriedpammer, line 168: **Not a bug - a question about which layer the fix belongs on.**

This does not loosen conversion validity at all. `IgnoreNullability` is strictly finer than `TypeErasure`, and `IdentityConversion` (`CSharpConversions.cs:347`) compares under `TypeErasure`, so every pair this branch short-circuits was already accepted two lines below. The only behavioural change is that `AttemptedConversionWithTypeMismatch` stops latching, and its sole consumer is `IntroduceQueryExpressions.MatchSimpleLambda`.

So this is a second net for the bug the type-inference commits already fix at the root, applied one layer under where the signal is read. Both `T09_Linq` fixtures pass with either change alone, so nothing in the suite distinguishes them.

The cost appears when inference and csc diverge for some other reason - the `M<T>(T, Action<T>)` case above is one: instead of falling back to method-call syntax, the transform emits query syntax and the range variable re-binds to the source's element type, so `from s in strings select s` over `IEnumerable<string?>` yields `IEnumerable<string?>` against a declared `IEnumerable<string>` (CS8619). `IgnoreNullability` also erases modopt/modreq, which the name does not say and which nothing here needed.

If the root fix is trusted, this line can go; if it is meant as a backstop, it would be worth a fixture that fails without it.

*Written by Claude Code (Opus 5) on my behalf; I have checked the reproductions.*

ICSharpCode.Decompiler.Tests/Semantics/TypeInferenceTests.cs
```diff
@@ -1387,8 +1422,12 @@ public void CommonSubTypeIEnumerableClonableIEnumerableComparableList()
 			// ReadOnlyCollectionBuilder<T> appears because the test compilation includes
 			// System.Core, which declares it as another public implementation of both
 			// IList and IList<T>.
+			var typesInBounds = FindAllTypesInBounds(Resolve(), Resolve(typeof(IEnumerable<ICloneable>), typeof(IEnumerable<IComparable>), typeof(IList)));
+			// As this finds all derived types, the result set contains compiler-generated types like <>z__ReadOnlyArray`1.
+			// We filter those out to make the test more robust against changes.
+			typesInBounds = typesInBounds.Where(t => !t.GetDefinition().IsCompilerGenerated()).ToArray();
```
siegfriedpammer, line 1453: Small thing: the comment reads as a guard against future corelib churn, but the churn is self-inflicted by this same file.

`<>z__ReadOnlyArray` occurs 0 times in the base test assembly (`5d9cdce94`) and 1 time in this branch's. The test compilation includes the test assembly itself, so the new candidate is Roslyn output from rewriting the other tests here to collection-expression syntax (`InferTypeArguments([T], ...)`, `Is.EqualTo([new ArrayType(...)])`) - production code cannot emit a type into the test assembly.

Not a product concern: the `GetAllTypeDefinitions()` branch only runs when `lowerBounds.Count == 0`, and every decompiler call site supplies lower bounds. Just worth either fixing the comment or keeping array-initialiser syntax in this one file, which removes the need for the filter.

*Written by Claude Code (Opus 5) on my behalf; I have checked the reproductions.*

PR #4091, opened 2026-08-30
Split StackType.O into StackType.Obj+StackType.VT

dgrunwald:
Object references and value types are quite different; so let's try keeping them in separate StackTypes throughout the whole compiler.
We have many assertions based on stack type comparisons, so the main effect of this change is to strengthen our invariants.

### DIFF
--- a/ICSharpCode.Decompiler/CSharp/CallBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/CallBuilder.cs
@@ -1187,6 +1187,18 @@ namespace ICSharpCode.Decompiler.CSharp
 								arguments.Add(new TranslatedExpression(expressionBuilder.GetDefaultValueExpression(type3).WithoutILInstruction()));
 						}
 						return true;
+					case ConversionResolveResult { Conversion.IsImplicitSpanConversion: true, Input: ArrayCreateResolveResult { Type: ArrayType { ElementType: var type3 }, SizeArguments: [{ ConstantValue: int arrayLength }] } }:
+						elementType = type3;
+						var expr = paramsArgument.Expression is CastExpression cast ? cast.Expression : paramsArgument.Expression;
+						arguments = new(((ArrayCreateExpression)expr).Initializer?.Elements.Select(e => new TranslatedExpression(e)) ?? []);
+						parameters = new List<IParameter>(arrayLength);
+						for (int i = 0; i < arrayLength; i++)
+						{
+							parameters.Add(new DefaultParameter(type3, string.Empty));
+							if (arguments.Count <= i)
+								arguments.Add(new TranslatedExpression(expressionBuilder.GetDefaultValueExpression(type3).WithoutILInstruction()));
+						}
+						return true;
 					default:
 						return false;
 				}
--- a/ICSharpCode.Decompiler/CSharp/CallBuilder.cs
+++ b/ICSharpCode.Decompiler/CSharp/CallBuilder.cs
@@ -1128,7 +1128,7 @@ namespace ICSharpCode.Decompiler.CSharp
 			ref List<TranslatedExpression> arguments)
 		{
 			var expressionBuilder = this.expressionBuilder;
-			if (ExtractArguments(out var elementType, out var expandedParameters, out var expandedArguments))
+			if (ExtractArguments(out var expandedParameters, out var expandedArguments))
 			{
 				expandedParameters.InsertRange(0, expectedParameters);
 				expandedArguments.InsertRange(0, arguments);
@@ -1144,18 +1144,30 @@ namespace ICSharpCode.Decompiler.CSharp
 			}
 			return false;
 
-			bool ExtractArguments([NotNullWhen(true)] out IType? elementType, [NotNullWhen(true)] out List<IParameter>? parameters, [NotNullWhen(true)] out List<TranslatedExpression>? arguments)
+			bool ExtractArguments([NotNullWhen(true)] out List<IParameter>? parameters, [NotNullWhen(true)] out List<TranslatedExpression>? arguments)
 			{
-				elementType = null;
 				parameters = null;
 				arguments = null;
+				// Every expanded argument binds to the element type of the params collection, so
+				// that is the type the parameters standing in for them carry. The argument itself
+				// may be an array of a more derived element type, because the collection is
+				// covariant in it. Unpack it the same way overload resolution does, and give up
+				// where overload resolution would give up on the expanded form as well.
+				IType paramsElementType;
+				if (parameter.Type is ArrayType { Dimensions: 1 } paramsArray)
+					paramsElementType = paramsArray.ElementType;
+				else if (parameter.Type.IsKnownType(KnownTypeCode.ReadOnlySpanOfT)
+					|| parameter.Type.IsKnownType(KnownTypeCode.SpanOfT)
+					|| parameter.Type.IsArrayInterfaceType())
+					paramsElementType = parameter.Type.TypeArguments[0];
+				else
+					return false;
 				switch (paramsArgument.ResolveResult)
 				{
 					case CSharpInvocationResolveResult { Member: IMethod method, Arguments: var args }:
 						// match System.Array.Empty<T>()
-						if (args is [] && method is { IsStatic: true, FullName: "System.Array.Empty", TypeArguments: [var type] })
+						if (args is [] && method is { IsStatic: true, FullName: "System.Array.Empty", TypeArguments: [_] })
 						{
-							elementType = type;
 							arguments = new();
 							parameters = new();
 							return true;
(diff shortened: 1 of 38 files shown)


ICSharpCode.Decompiler/CSharp/CallBuilder.cs
```diff
@@ -1187,6 +1187,18 @@ bool ExtractArguments([NotNullWhen(true)] out IType? elementType, [NotNullWhen(t
...
 						}
 						return true;
+					case ConversionResolveResult { Conversion.IsImplicitSpanConversion: true, Input: ArrayCreateResolveResult { Type: ArrayType { ElementType: var type3 }, SizeArguments: [{ ConstantValue: int arrayLength }] } }:
+						elementType = type3;
+						var expr = paramsArgument.Expression is CastExpression cast ? cast.Expression : paramsArgument.Expression;
+						arguments = new(((ArrayCreateExpression)expr).Initializer?.Elements.Select(e => new TranslatedExpression(e)) ?? []);
+						parameters = new List<IParameter>(arrayLength);
+						for (int i = 0; i < arrayLength; i++)
+						{
+							parameters.Add(new DefaultParameter(type3, string.Empty));
```
dgrunwald: Preexisting, but weird how the expected parameters are constructed based on the *argument* types.
Surely they should be based on the unexpanded parameter type instead?

ICSharpCode.Decompiler/CSharp/ExpressionBuilder.cs
```diff
@@ -3968,14 +3961,31 @@ TranslatedExpression TranslateArrayInitializer(Block block)
...
 				expr.Arguments.AddRange(newArr.Indices.Select(i => Translate(i).Expression));
-			return expr.WithILInstruction(block)
-				.WithRR(new ArrayCreateResolveResult(new ArrayType(compilation, type, dimensions), newArr.Indices.Select(i => Translate(i).ResolveResult).ToArray(), elementResolveResults));
+			ResolveResult rr = new ArrayCreateResolveResult(new ArrayType(compilation, type, dimensions),
+				newArr.Indices.Select(i => Translate(i).ResolveResult).ToArray(), elementResolveResults);
+			var initializer = expr.WithILInstruction(block).WithRR(rr);
+			if (arrayToSpan != null)
+			{
+				var arrayToSpanRR = new ConversionResolveResult(arrayToSpan.DeclaringType, rr, Conversion.ImplicitSpanConversion);
+				initializer = new CastExpression(ConvertType(arrayToSpan.DeclaringType), expr).WithoutILInstruction().WithRR(arrayToSpanRR);
```
dgrunwald: MatchArrayInitializerFinal checks the return type, but now this is using the declaring type instead?

ICSharpCode.Decompiler/IL/Transforms/ExpressionTransforms.cs
```diff
@@ -418,20 +416,12 @@ bool TransformSpanTCtorContainingStackAlloc(NewObj newObj, out ILInstruction loc
...
-				{
-					store.Variable = newVariable;
-				}
-				value.ReplaceWith(new LocAllocSpan(newObj.Arguments[1], type));
+				// The block addresses the allocation through the localloc pointer and only its
+				// result is the span, so the constructor becomes the block's final instruction
+				// instead of retyping the initializer variable to Span&lt;T&gt;.
+				initializer.FinalInstruction = new NewObj(newObj.Method) {
+					Arguments = { new LdLoc(initializerVariable), newObj.Arguments[1] }
+				}.WithILRange(newObj);
```
dgrunwald, line 426: Downside of this move: if the size of the allocation is not a compile-time constant, we can no longer inline the variable used in the size. But I'm not sure if this can happen in the case with initializer. (I only thought about this because I considered eliminating the LocAllocSpan instruction altogether, but that doesn't work for this reason)

If it is possible that we hit this case with a size variable that might need inlining, we'll have to use LocAllocSpan after all, and then rewrite all the initializing stores accordingly.

ICSharpCode.Decompiler/TypeSystem/MetadataModule.cs
```diff
@@ -602,6 +590,60 @@ IMethod ResolveMethodReference(MemberReferenceHandle memberRefHandle, GenericCon
...
+		/// than one it read from metadata.
+		/// </remarks>
+		public IMethod ResolveMethod(IType declaringType, string name, MethodSignature<IType> signature)
+		{
+			if (declaringType == null)
+				throw new ArgumentNullException(nameof(declaringType));
+			if (name == null)
+				throw new ArgumentNullException(nameof(name));
+			IEnumerable<IMethod> methods;
+			if (name == ".ctor")
```
dgrunwald: Any reason you don't use the same logic here as in the method above?
(I kinda expected that bit could also be extracted into  the common helper)

ICSharpCode.Decompiler/IL/Transforms/ExpressionTransforms.cs
siegfriedpammer: A stackalloc with an initializer requires a constant length, so the size can never be a variable that would need inlining:

```csharp
Span<int> s = stackalloc int[n] { 1, 2, 3 };
// error CS0150: A constant value is expected
```

`MatchesElementCount` accepts any instruction as the count, so hand-written IL could still get here with a non-constant one; no C# compiler can.

*Written by an AI agent (Claude) on Siegfried's behalf.*

siegfriedpammer: Completing the previous reply, since it only covered half of this. Both halves checked by compiling the matrix (SDK 11.0.100-preview.7, `net10.0`, `LangVersion=preview`):

Without an initializer, a runtime length is fully legal - `stackalloc T[n]` compiles with `n` a parameter or a local read from a method call, for a raw pointer, `Span<T>`, `ReadOnlySpan<T>`, and in direct argument position. So `LocAllocSpan` is still needed for that shape, exactly as you say.

With an initializer, the length must be a constant:

```csharp
Span<int> s = stackalloc int[n] { 1, 2, 3 };
// error CS0150: A constant value is expected
```

The error is identical for pointer, `Span<T>`, `ReadOnlySpan<T>`, argument-position and `scoped` targets, and for every unmanaged element type (`int`, `byte`, `double`, `char`, an enum, a user-defined struct, `T where T : unmanaged`). A `const` local, `const` field or constant expression is accepted; a `static readonly` field is not. The reason is visible in the neighbouring diagnostic: `stackalloc int[5] { 1, 2, 3 }` is `error CS0847: An array initializer of length '5' is expected`, so the compiler has to compare the element count against the length, which needs a constant.

So the initializer shape this changes can never carry a size that needs inlining. `MatchesElementCount` accepts any instruction as the count, so hand-written IL could still get there; no C# compiler can.

*Written by an AI agent (Claude) on Siegfried's behalf.*

siegfriedpammer:
`decompdiff` over the top-200 nuget.org corpus — `master` (dbf23c6e4) against this branch rebased onto that same commit, so the diff isolates the PR. 167 assemblies, 37,530 types.

Counters are flat and there are no new errors: 37,505 types byte-identical, 25 changed, 137 changed lines in total.

| | old | new | delta |
|---|---|---|---|
| lines | 4,055,141 | 4,055,142 | +1 |
| goto statements | 2,315 | 2,315 | 0 |
| //IL_ warnings | 24,191 | 24,191 | 0 |
| generated-name leaks | 2,544 | 2,544 | 0 |

One of the 25 changes is a behavioural difference.

## Invented `(byte)` cast truncates Regex character masks

7 of the changed types, all `System.Text.RegularExpressions.Generated.*`, in AWSSDK.Core, Dapper (x2), FluentValidation, OpenTelemetry and Renci.SshNet (x2).

```csharp
// master
if (((c = span[i]) < 0x80 ? (mask[c >> 4] & (1 << (c & 0xF)))
    : (CharInClass(c, "...") ? 1 : 0)) == 0)

// this branch
if (!((c = span[i]) < 0x80 ? ((byte)(mask[c >> 4] & (1 << (c & 0xF))) != 0)
    : CharInClass(c, "...")))
```

The bool-shaped conditional is an improvement and matches the shape the source generator actually emitted. The `(byte)` is not in the IL — there is no `conv.u1` anywhere in the method:

```
IL_0053: ldstr "\0\0..."   // the mask table
IL_0058: ldloc.2
IL_0059: ldc.i4.4
IL_005a: shr               // c >> 4
IL_005b: call instance char String::get_Chars(int32)
IL_0060: ldc.i4.1
IL_0061: ldloc.2
IL_0062: ldc.i4.s 15
IL_0064: and               // c & 15
IL_0065: ldc.i4.s 31
IL_0067: and               // & 31
IL_0068: shl               // 1 << ((c & 15) & 31)  -> up to 32768
IL_0069: and               // mask[c >> 4] & bit
IL_006a: ldc.i4.0
IL_006b: cgt.un            // != 0
```

The shift count is `(c & 15) & 31`, so the masked value reaches 32768 and `(byte)` discards every bit from 8 up. With AWSSDK's `mask[2] = U+7800`, character `'+'` (0x2B) tests bit 11: master computes `0x800 != 0` and matches, this branch computes `(byte)0x800 == 0` and does not. FluentValidation's mask holds U+FFFE and U+FFFF entries, so most of its character class is affected.

`git bisect run` over the branch points at 30552b01f, *Replace a couple of `inst.ResultType -> IType` conversions with `InferType()` calls* — specifically one line in `ExpressionBuilder.VisitIfInstruction`, on the path that picks a target type for a conditional whose branches have no common type:

```diff
- targetType = FindType(inst.ResultType, context.TypeHint.GetSign());
+ targetType = inst.InferType(compilation);
```

The old form derived the type from the stack type, so an I4 conditional became `int`. `InferType` narrows from the operands instead: `c & 15` has range 0–15 and fits a byte, but that inference must not survive `shl`, because `1 << n` is not byte-ranged. Reverting just this line at the branch tip removes the cast.

One thing to note for a fix: the guard removed from `VisitSwitchInstruction` in the same commit would not have caught this either, since it compares `GetStackType()` and both `byte` and `int` are `I4`.

Repro:

```
ilspycmd ~/.cache/nugetfuzz/awssdk.core/4.0.102.4/lib/net8.0/AWSSDK.Core.dll \
    -t "System.Text.RegularExpressions.Generated.<RegexGenerator_g>F23AE...__RoleSessionNameRegex_0"
```

## The other 18 changes

- **10 types** differ only in warning text: `//IL_xxxx: Expected O, but got Unknown` becomes `Expected Obj`, from the enum rename.
- **`BoolToTwoValuesConverter<TProvider>`** now declines instead of emitting `(TProvider v) => (object)v == (object)trueValue`. Master compared box identity where the tree compares values, so this is a fix.
- **`Logging.Logger`, `DebuggerDisplayFormatting`**: `LogLevel[] readOnlySpan` becomes `ReadOnlySpan<LogLevel>`.
- **`StackExchange.Redis.Condition`**: `RedisCommand num` becomes `redisCommand` — inference now yields the enum rather than its underlying integer.
- **SharpCompress ZStandard**: `(nint)(...)` becomes `(nint)((IntPtr)(...))`. Same type, so semantically identical, just slightly worse to read.

Corpus caveats: four assemblies are undecompilable on both sides for missing references (System.Web.Mvc, Microsoft.Azure.WebJobs.Host, MongoDB.Driver, MongoDB.Driver.Core) and twelve more decompile with unresolved references. Those are identical on both sides, so they are corpus gaps rather than PR effects.

PR #4114, opened 2026-09-06
Fix #4059: treat a Deconstruct out argument as a definition

### LINKED ISSUE #4059: Two back-to-back deconstructions in one method are not recognized
Two deconstructions of the same source type in one method are not recognized: the first one alone
decompiles correctly, but as soon as a second follows, both fall back to raw `Deconstruct` calls.

## Repro

```csharp
public class Src<T, T2>
{
    public void Deconstruct(out T a, out T2 b) { a = default; b = default; }
}

public class C
{
    private static Src<string, string> GetSource() => null;

    public static void One()
    {
        var (value, value2) = GetSource();
        Console.WriteLine(value);
        Console.WriteLine(value2);
    }

    public static void TwoBackToBack()
    {
        var (value, value2) = GetSource();
        Console.WriteLine(value);
        Console.WriteLine(value2);
        var (value3, value4) = GetSource();
        Console.WriteLine(value3);
        Console.WriteLine(value4);
    }
}
```

`One` round-trips. `TwoBackToBack` decompiles to:

```csharp
public static void TwoBackToBack()
{
    GetSource().Deconstruct(out var a, out var b);
    string value = a;
    string value2 = b;
    Console.WriteLine(value);
    Console.WriteLine(value2);
    GetSource().Deconstruct(out b, out a);   // note: same locals, swapped
    string value3 = b;
    string value4 = a;
    Console.WriteLine(value3);
    Console.WriteLine(value4);
}
```

## Why

csc reuses the same two out-slot temporaries for both calls, and the second call receives them in
the opposite order (`out b, out a`). `DeconstructionTransform` matches a deconstruction against
locals that the pattern owns; locals already written by an earlier deconstruction in the same
method do not satisfy that, so neither call is recognized.

Splitting the shared out-slot temporaries into per-deconstruction locals before pattern matching
should let both match independently.

## Scope

Reproduced with `-c Release` on the standalone case above, and in `DeconstructionTests` on all
eight compiler configurations the fixture runs (Roslyn 2.10, 3.11, 4.14 and latest, with and
without `Optimize`), so it is not compiler-version specific.

Found while porting shape coverage onto `DeconstructionTests`: 19 of 20 additional shapes pass,
this is the one that does not. The test case is checked in commented out, referencing this issue.

siegfriedpammer:
Fixes #4059.

Two deconstructions of the same source type in one method are not recognized: the first one alone
decompiles correctly, but as soon as a second follows, both fall back to raw `Deconstruct` calls,
because csc reuses the same two out-slot temporaries and hands them to the second call in the
opposite order.

## Why splitting the temporaries did not happen

`ReachingDefinitionsVisitor` does not track address loads, so a variable that is only ever written
through its address - which is every `out` slot - stays potentially uninitialized for its whole
life. `SplitVariables.GroupStores` merges all such loads of one variable into a single
representative, so the two deconstructions end up in one live range and neither can be matched.

The same holds for `out` arguments in general; it only becomes visible for `Deconstruct`, where
pattern matching depends on the split.

## The fix

An `ldloca` in argument position > 0 of a `Deconstruct` call is treated as a definition: it kills
the definitions reaching it and becomes the one that reaches the following loads. That reuses the
existing store machinery, so it stays flow-sensitive - where two such calls join, both reach the
load and are merged again, which a flow-insensitive rule would get wrong.

IL cannot express an address that is assigned without being read first; `out` is a C# convention
over `ref`. So the qualifying address loads have to be named one by one, and a `Deconstruct`
method written in IL could read its argument before assigning it. That possibility is knowingly
ignored - C#'s definite assignment rules rule it out for anything a compiler emits.

`SplitVariables` itself is unchanged: its `HandleLoad` on such an `ldloca` now runs against the
post-store state and merges the instruction with itself.

## Tests

The `TwoBackToBackDeconstructs_Custom` case that #4059 was filed from is checked in commented out;
this enables it. It fails on all eight compiler configurations before the change and passes after.

`ICSharpCode.Decompiler.Tests`: 3604 tests, no failures, 45 skipped (Windows-only configurations).

*Prepared by an AI agent (Claude, claude-opus-5, via Claude Code) and reviewed by @siegfriedpammer.*

### DIFF
--- a/ICSharpCode.Decompiler.Tests/TestCases/Pretty/DeconstructionTests.cs
+++ b/ICSharpCode.Decompiler.Tests/TestCases/Pretty/DeconstructionTests.cs
@@ -1240,17 +1240,15 @@ namespace ICSharpCode.Decompiler.Tests.TestCases.Pretty
 			Console.WriteLine(value2);
 		}
 
-		// #4059: csc reuses the same out-slot temporaries for both calls, and hands them to the
-		// second one in the opposite order, so neither deconstruction is recognized.
-		//public void TwoBackToBackDeconstructs_Custom()
-		//{
-		//	var (value, value2) = GetSource<string, string>();
-		//	Console.WriteLine(value);
-		//	Console.WriteLine(value2);
-		//	var (value3, value4) = GetSource<string, string>();
-		//	Console.WriteLine(value3);
-		//	Console.WriteLine(value4);
-		//}
+		public void TwoBackToBackDeconstructs_Custom()
+		{
+			var (value, value2) = GetSource<string, string>();
+			Console.WriteLine(value);
+			Console.WriteLine(value2);
+			var (value3, value4) = GetSource<string, string>();
+			Console.WriteLine(value3);
+			Console.WriteLine(value4);
+		}
 
 	}
 }
--- a/ICSharpCode.Decompiler/FlowAnalysis/ReachingDefinitionsVisitor.cs
+++ b/ICSharpCode.Decompiler/FlowAnalysis/ReachingDefinitionsVisitor.cs
@@ -39,7 +39,9 @@ namespace ICSharpCode.Decompiler.FlowAnalysis
 	/// * <c>StLoc</c>
 	/// * <c>TryCatchHandler</c> (for the exception variable)
 	/// * <c>ReachingDefinitionsVisitor.UninitializedVariable</c> for uninitialized variables.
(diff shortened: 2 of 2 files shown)


dgrunwald:
This needs proper investigation of how it interacts with the derived class (SplitVariables.GroupStores) -- and especially the partial support for ref locals that has.

Issue #3921, opened 2026-07-27
Explicit delegate constructor not distinguished in C# decompilation.


PetSerAl:
C# allows wrapping method group into delegate constructor. That disable delegate caching and construct new delegate instance on each invocation. That distinction visible in IL but lost in C# decompilation.
### Input code
```csharp
using System;

static class Test
{
	static Action Test1() => M;
	static Action Test2() => new(M);
	static void M() {}
}
```

### Erroneous output
```csharp
using System;

internal static class Test
{
	private static Action Test1()
	{
		return M;
	}

	private static Action Test2()
	{
		return M;
	}

	private static void M()
	{
	}
}
```
```msil
.class private auto ansi abstract sealed beforefieldinit Test
	extends [System.Runtime]System.Object
{
	// Nested Types
	.class nested private auto ansi abstract sealed beforefieldinit '<>O'
		extends [System.Runtime]System.Object
	{
		.custom instance void [System.Runtime]System.Runtime.CompilerServices.CompilerGeneratedAttribute::.ctor() = (
			01 00 00 00
		)
		// Fields
		.field public static class [System.Runtime]System.Action '<0>__M'

	} // end of class <>O


	// Methods
	.method private hidebysig static 
		class [System.Runtime]System.Action Test1 () cil managed 
	{
		// Method begins at RVA 0x2050
		// Header size: 1
		// Code size: 28 (0x1c)
		.maxstack 8

		// sequence point: (line 5, col 27) to (line 5, col 28) in M:\Temp\Test\Test.cs
		IL_0000: ldsfld class [System.Runtime]System.Action Test/'<>O'::'<0>__M'
		IL_0005: dup
		IL_0006: brtrue.s IL_001b

		IL_0008: pop
		IL_0009: ldnull
		IL_000a: ldftn void Test::M()
		IL_0010: newobj instance void [System.Runtime]System.Action::.ctor(object, native int)
		IL_0015: dup
		IL_0016: stsfld class [System.Runtime]System.Action Test/'<>O'::'<0>__M'

		IL_001b: ret
	} // end of method Test::Test1

	.method private hidebysig static 
		class [System.Runtime]System.Action Test2 () cil managed 
	{
		// Method begins at RVA 0x206d
		// Header size: 1
		// Code size: 13 (0xd)
		.maxstack 8

		// sequence point: (line 6, col 27) to (line 6, col 33) in M:\Temp\Test\Test.cs
		IL_0000: ldnull
[...]

PR #4126, opened 2026-09-08
Issue #3921 : preserve delegate construction

marcusmalloc:
Link to issue(s) this covers https://github.com/icsharpcode/ILSpy/issues/3921

### Problem
ILSpy decompiles and returns 'return M' and 'return new Action(M)' as 'return M.'

### Solution
As @dgrunwald recommended, this PR adds an ILast instruction to distinguish cached / fresh delegates throughout decompilation.
 
Please let me know if any changes need to be made to my implementation!


ICSharpCode.Decompiler/CSharp/TranslatedExpression.cs
```diff
@@ -270,6 +271,16 @@ TranslatedExpression Unwrapped(TranslatedExpression operand)
 							else if (Expression is ObjectCreateExpression oce && conversion.Conversion.IsMethodGroupConversion
 								  && oce.Arguments.Count == 1 && expressionBuilder.settings.UseImplicitMethodGroupConversion)
 							{
+								// C# 11 caches static method groups. Keep explicit construction when the IL creates a fresh delegate.
+								if (conversion.Conversion.Method.IsStatic
+									&& conversion.Conversion.Method is not LocalFunctionMethod { IsStaticLocalFunction: false }
+									&& conversion.Conversion.Method.Parameters.Count == type.GetDelegateInvokeMethod()?.Parameters.Count
+									&& expressionBuilder.settings.GetMinimumRequiredVersion() >= LanguageVersion.CSharp11_0
```
siegfriedpammer, line 278: That's an interesting way of using `GetMinimumRequiredVersion`... I'd prefer to have a separate setting, like with all the other features. Or maybe just omit it.

also patterns would fit nicely here: `conversion.Conversion.Method is { IsStatic:  true, ... }` maybe?

siegfriedpammer:
Thank you for your contribution... what's the commit message `[Decompiler] gpt-slop copy` supposed to mean? To what degree did you use LLMs to implement this? Please give us some background info on your approach. Thanks!

PR #4133, opened 2026-09-13
Remove the delegate cache left by a discarded method-group conversion

### LINKED ISSUE #3965: Discarded method-group conversion leaks the <>O delegate cache into the output
Decompiling a **discarded** method-group conversion leaves the compiler's delegate cache in the output, spelled with the unspeakable field names, so the result does not compile. The discard itself is also lost.

### Repro

```csharp
using System;

public class Leak
{
	private static void M(int x)
	{
	}

	public static void Main()
	{
		_ = (Action<int>)M;
	}
}
```

Compiled for `net10.0`, no special options.

### Decompiled output

```csharp
using System;

public class Leak
{
	[CompilerGenerated]
	private static class <>O
	{
		public static Action<int> <0>__M;
	}

	private static void M(int x)
	{
	}

	public static void Main()
	{
		if (<>O.<0>__M == null)
		{
			<>O.<0>__M = M;
		}
	}
}
```

### Expected

`_ = (Action<int>)M;`

The same conversion decompiles correctly when its result is used - `Action<int> a = M;` or passing it as an argument both come back without any trace of `<>O`. Only the discarded form leaks the cache.

### Version

Reproduced with `ilspycmd` built from master at fdc4c7c16.

dualfroz:
`_ = (Action<int>)M;` never reads the cache field back, and the transform required exactly one read, so the output kept `<>O.<0>__M`, which does not compile.
With zero reads, the cache check is now replaced by the delegate construction it guarded.

Closes #3965


siegfriedpammer:
Did you use AI to implement this? Please read https://github.com/icsharpcode/ILSpy/blob/master/CONTRIBUTING.md#contributing especially bullet 2.

dualfroz:
Hi. I didn't use the AI to write the code itself. I did however use AI to make it easier to read. Same with the comment - i wrote it myself and then told ai to please make it look nicer. Will copy-paste this response in the other PR.

ICSharpCode.Decompiler/IL/Transforms/CachedDelegateInitialization.cs
```diff
@@ -95,13 +94,25 @@ bool CachedDelegateInitializationWithField(IfInstruction inst)
...
-				return false;
-			var usages = nextInstruction.Descendants.Where(i => i.MatchLdsFld(field)).ToArray();
+			var usages = nextInstruction?.Descendants.Where(i => i.MatchLdsFld(field)).ToArray() ?? [];
+			if (usages.Length == 0)
+			{
+				// A discarded method-group conversion ("_ = (Action)M;") caches the
+				// delegate without ever reading the cache back. Keep the conversion,
+				// which still allocates, and drop the caching around it -- but only
+				// once the rest of the method is known not to read the field either.
+				if (context.Function.Descendants.Any(i => i != left && i.MatchLdsFld(field)))
```
siegfriedpammer, line 104: I think this check is not necessary (or rather harmful)...

I asked Claude about it, and it produced the following table:

Image

Would you be able to add more test cases to the pretty test?

siegfriedpammer:
LGTM thanks!

